PR #48, opened 2015-01-31
Restores the ability to list local flags

fabianofranz:
Because of `mergePersistentFlags()` the `flags` flag set is not actually storing only flags that are specific to the given command, but the complete set of flags that applies to it (local + persistent declared locally or on parents). This PR adds the `LocalFlags()` method that restores the ability to fetch the flags declared locally which can be useful when writing usage or help templates for example. Also updates the usage default template to have two separate groups of flags: `Flags` and `Global Flags` which makes it much more user friendly when you start to have multiple local + persistent flags in a single command.

### DIFF
--- a/command.go
+++ b/command.go
@@ -40,9 +40,9 @@ type Command struct {
 	Short string
 	// The long message shown in the 'help <this-command>' output.
 	Long string
-	// Set of flags specific to this command.
+	// Full set of flags
 	flags *flag.FlagSet
-	// Set of flags children commands will inherit
+	// Set of flags childrens of this command will inherit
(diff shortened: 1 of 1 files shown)


fabianofranz:
@spf13 If you don't like the idea of changing the default usage template I can revert that part.

vmarmol:
:+1: we're looking to do something similar GoogleCloudPlatform/kubernetes#4142

spf13:
@fabianofranz This is a great contribution. Merged as efb045e. Thanks!

vmarmol:
Thanks! :D

PR #49, opened 2015-01-31
Minor English correction

k4rtik:
"more information about that command" felt odd so verified the usage with `git help` and `go help`


spf13:
@k4rtik Thanks for the help. Merged as bd75ea1

PR #51, opened 2015-02-04
Several 'help' fixes

fabianofranz:
Need to make sure `--help` is registered before calling `pflags` which tries to handle the help flag itself.

`Find()` must be aware of the special command `help` and consider the context it's being called in order to display the help for the proper command (instead of root).

### DIFF
--- a/command.go
+++ b/command.go
@@ -18,13 +18,14 @@ package cobra
 import (
 	"bytes"
 	"fmt"
-	"github.com/inconshreveable/mousetrap"
-	flag "github.com/spf13/pflag"
 	"io"
 	"os"
 	"runtime"
 	"strings"
 	"time"
+
+	"github.com/inconshreveable/mousetrap"
+	flag "github.com/spf13/pflag"
 )
 
 // Command is just that, a command for your application.
@@ -370,6 +371,9 @@ func (c *Command) Find(arrs []string) (*Command, []string, error) {
 				// only accept a single prefix match - multiple matches would be ambiguous
 				if len(matches) == 1 {
 					return innerfind(matches[0], argsMinusX(args, argsWOflags[0]))
+				} else if len(matches) == 0 && len(args) > 0 && args[0] == "help" {
+					// special case help command
+					return innerfind(c, argsMinusX(append(args, "--help"), argsWOflags[0]))
 				}
 			}
 		}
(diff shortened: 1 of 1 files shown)


spf13:
@fabianofranz This isn't passing the tests. Can you revisit it and make it pass.

fabianofranz:
@spf13 Good to go.

command.go
```diff
@@ -751,7 +755,12 @@ func (c *Command) Runnable() bool {
...
 // Determine if the command has children commands
 func (c *Command) HasSubCommands() bool {
-	return len(c.commands) > 0
+	for _, sub := range c.commands {
+		if sub.Runnable() {
+			return true
+		}
+	}
+	return false
 }
```
eparis: Isn't this HasRunnableSubCommands()  ?  Why the change?

fabianofranz: Sorry, I wasn't aware of `HasRunnableSubCommands()`. Might have happened when I first created this PR. Reverted to the original behavior.

command.go
```diff
@@ -370,6 +371,9 @@ func (c *Command) Find(arrs []string) (*Command, []string, error) {
 				// only accept a single prefix match - multiple matches would be ambiguous
 				if len(matches) == 1 {
 					return innerfind(matches[0], argsMinusX(args, argsWOflags[0]))
+				} else if len(matches) == 0 && len(args) > 0 && args[0] == "help" {
+					// special case help command
+					return innerfind(c, argsMinusX(append(args, "--help"), argsWOflags[0]))
```
eparis: So this is interesting.  Before this change.

command help subcommand - works, shows help for subcommand
command subcommand help - fails, because 'help' is an unknown subsubcommand 

Are we certain we want the later?  (I personally like it)

fabianofranz: I changed this PR to revert that piece of code (on a separate commit so we don't loose it), since it may result in unexpected behavior not valid for every scenario.

The "help" string could be a valid argument that does not mean help. Using `git add` as an example:

`git add help`

Means I want to add the file called "help", which is perfectly plausible. I *do not* want the equivalent of `git add --help` here.

eparis:
i still can't figure out how to write a test that is somehow 'wrong'   can you help me figure out and example fabio?

command.go
```diff
@@ -798,7 +806,7 @@ func (c *Command) Flags() *flag.FlagSet {
 			c.flagErrorBuf = new(bytes.Buffer)
 		}
 		c.flags.SetOutput(c.flagErrorBuf)
-		c.PersistentFlags().BoolVarP(&c.helpFlagVal, "help", "h", false, "help for "+c.Name())
```
eparis: So the change here is that you are calling c.mergePersistentFlags() here. (And then making sure mergePersistentFlags() won't bug later)

So is there a real bug that somehow we missed a meaningful and necessary c.mergePersistendFlags() call somewhere else? I really wish I knew how to trigger it...

PR #53, opened 2015-02-09
added pre and post run hooks.

AlexanderThaller:
added two new hooks to the command type.

PreRun will run after parsing the flags and args but before starting Run.
PostRun will run after Run has completed.

### DIFF
--- a/command.go
+++ b/command.go
@@ -47,6 +47,12 @@ type Command struct {
 	// Run runs the command.
 	// The args are the arguments after the command name.
 	Run func(cmd *Command, args []string)
+	// PreRun runs the command after the flags are parsed and before run.
+	// The args are the arguments after the command name.
+	PreRun func(cmd *Command, args []string)
+	// PostRun runs the command after run.
+	// The args are the arguments after the command name.
+	PostRun func(cmd *Command, args []string)
 	// Commands is the list of commands supported by this program.
 	commands []*Command
 	// Parent Command for this command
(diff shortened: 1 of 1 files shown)


spf13:
I like the idea, but I can't figure out a good use case. What's the use case for this? I don't see why this isn't just part of Run?

tdavis:
I'll just chime in and say that I'm writing my first cli with cobra and already need a ``prerun`` type thing. Each sub-command needs some global configuration stuff setup using the vars from persistent root flags. Since the root's ``Run`` method isn't called when sub-commands run, there's no place to do global setup *after* the persistent flags have been parsed but *before* the current command runs.

spf13:
@tdavis Why not just make it the first line of Run? 

I could possibly see the need for a Global PreRunner or a tree based PreRunner (like the persistent flags), but have a hard time seeing the need for it on a per command basis. 

Overall I'm in favor of this idea, I just want to understand the need better.

tdavis:
@spf13 I am; more specifically ``cmd.Parent().Run()``, but it feels strange to manually invoke the command tree. I see even ``hugo`` does this with ``InitializeConfig()`` so who am I to argue your intuition? ;)

AlexanderThaller:
I needed it so I can setup some logging based on a flag I have defined. I wanted to do this with every sub command as well so it felt better to use a separate command and not duplicate code in every command Run function. Inheriting such a Pre/Post command to the sub commands would be better this was just to solve my problem and I saw that others had the same problem https://github.com/spf13/cobra/issues/47 so I made a pull request.

I can try to implement the inheriting if you still think that would be a good idea and update the pull request.

spf13:
Alexander,

Thanks for this. It helps quite a lot. I think the best approach here would be a persistent or single approach like we use with the flags. This is a first step towards that.

eparis:
I've now stepped on enough code that this can not be merged.  It seems spf13 was happy with the idea.  If you still want to do it, please rebase and add some documentation (readme)

eparis:
Closing, I rebased into #100

PR #54, opened 2015-02-11
help displays command names instead of usage in Available Commands

kazegusuri:
When a command has a subcommand whose usage is very log, it's difficult to look the help. After this change, the help displays just command names instead of usage.

For example, there are 2 subcommands bar and baz, and bar has a very long usage. The command's help displays like following:
```
Usage of foo:

Usage: 
  foo --very-very-long-use [flags]
  foo [command]

Available Commands: 
  bar [--foo-blah-blah-blah] [--bar-blah-blah-blah]  [--baz-blah-blah-blah] [--qux-blah-blah-blah] [--quux-blah-blah-blah] [--foobarbaz-blah-blah-blah] Description of bar
  baz --baz                                                                                                                                             Description of baz
  help [command]                                                                                                                                        Help about any command

 Available Flags:
  -h, --help=false: help for foo

Use "foo help [command]" for more information about that command.
```
After this change, the help displays like following:
```
Usage of foo:

Usage: 
  foo --very-very-long-use [flags]
  foo [command]

Available Commands: 
  bar         Description of bar
  baz         Description of baz

 Available Flags:
  -h, --help=false: help for foo

Use "foo help [command]" for more information about that command.
```

`help bar` still displays detailed usage as before. 

Or as an alternative option, `ShortUse` variable to display Available Commands.

### DIFF
--- a/cobra_test.go
+++ b/cobra_test.go
@@ -467,7 +467,15 @@ func TestRootHelp(t *testing.T) {
 		t.Errorf("--help shouldn't trigger an error, Got: \n %s", x.Output)
 	}
 
+	if strings.Contains(x.Output, cmdEcho.Use) {
+		t.Errorf("--help shouldn't display subcommand's usage, Got: \n %s", x.Output)
+	}
+
 	x = fullSetupTest("echo --help")
+
+	if strings.Contains(x.Output, cmdTimes.Use) {
+		t.Errorf("--help shouldn't display subsubcommand's usage, Got: \n %s", x.Output)
+	}
 
 	checkResultContains(t, x, "Available Commands:")
 	checkResultContains(t, x, "for more information about that command")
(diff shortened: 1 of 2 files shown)


spf13:
I'd like to get other's feedback on this before merging. It's a pretty big change I think. How do similar tools (go, git, etc) work?

spf13:
@brendandburns @smarterclayton @jlowdermilk 

I'd love feedback on this. I think the approach makes sense, but wanted another opinion.

kazegusuri:
Sorry for not touching this for a while. I think it is better to make displaying available commands or not  to option (displaying available commands in default for backward compatibilty).

eparis:
It appears this was already merged.  (and that's a good thing, being consistent with both git and go)

PR #55, opened 2015-02-14
New "example" section of commands

eparis:
We have a long and short description.  This adds an "Example" section.
Which can used to create better docs than putting it all in Long.


brendandburns:
Closing this PR as it has been merged into the tree via other means.

PR #58, opened 2015-02-28
Explicitly support local flags overwriting persistent/inherited flags

jlowdermilk:
The current (and arguably, desired) behavior when a Command specifies a flag that has the same name as a persistent/inherited flag, is that the local definition takes precedence. This change makes that behavior explicitly supported and updates the various `*Flags` functions accordingly:

* `LocalFlags`: now returns only the set of flags and persistent flags attached to the Command itself.
* `InheritedFlags`: now returns only the set of persistent flags inherited from the Command's parent(s), excluding any that are overwritten by a local flag.
* `NonInheritedFlags`: changed to an alias of `LocalFlags`.
* `AllPersistentFlags`: removed; it returned the set of all persistent flags attached to either the Command or its parent(s), which isn't very useful.

Default `UsageTemplate` updated to use `LocalFlags` and `InheritedFlags` for `Flags: ...`, `Global Flags: ...` output respectively.

### DIFF
--- a/cobra_test.go
+++ b/cobra_test.go
@@ -11,10 +11,13 @@ var _ = fmt.Println
 
 var tp, te, tt, t1 []string
 var flagb1, flagb2, flagb3, flagbr bool
-var flags1, flags2, flags3 string
+var flags1, flags2a, flags2b, flags3 string
 var flagi1, flagi2, flagi3, flagir int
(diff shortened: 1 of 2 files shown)


brendandburns:
Closing as this is merged (not via github merge)

PR #60, opened 2015-03-03
Add Command's RemoveCommand method

dohzya:
This method removes children commands of an existing command.

This allows to build CLI clients that can be extended by 3rd party tools, by adding some commands and removing/replacing some others. For instance to replace the “version” command.

Note: For now the 1st defined command is executed, so it is not possible to override an existing command.
But anyway, deleting an old command then adding a new one is the ultimate way to be certain there is no confusion.


spf13:
Can you follow up with some documentation (in the read me) on how and why this is useful?

PR #61, opened 2015-03-03
added a Files field to Command

baruchlubinsky:
For programs that process files provided as arguments after the flags. The Files member of the Command should describe the files that are input.

Add `[files ...]` to the usage line, and section to display the explanation.

### DIFF
--- a/command.go
+++ b/command.go
@@ -42,6 +42,8 @@ type Command struct {
 	Long string
 	// Examples of how to use the command
 	Example string
+	// Files that are passed to the program after flags
+	Files string
 	// Full set of flags
 	flags *flag.FlagSet
 	// Set of flags childrens of this command will inherit
@@ -204,7 +206,7 @@ func (c *Command) UsageTemplate() string {
 	} else {
 		return `{{ $cmd := . }}
 Usage: {{if .Runnable}}
-  {{.UseLine}}{{if .HasFlags}} [flags]{{end}}{{end}}{{if .HasSubCommands}}
+  {{.UseLine}}{{if .HasFlags}} [flags]{{end}}{{if .HasFiles}} [files ...]{{end}}{{end}}{{if .HasSubCommands}}
   {{ .CommandPath}} [command]{{end}}{{if gt .Aliases 0}}
 
 Aliases:
@@ -212,6 +214,10 @@ Aliases:
 {{end}}{{if .HasExample}}
 Examples:
 {{ .Example }}
+{{end}}
+{{if .HasFiles}}
+Files:
+  {{ .Files }}
 {{end}}{{ if .HasSubCommands}}
 Available Commands: {{range .Commands}}{{if .Runnable}}
   {{rpad .Name .NamePadding }} {{.Short}}{{end}}{{end}}
@@ -708,6 +714,11 @@ func (c *Command) HasParent() bool {
 	return c.parent != nil
 }
 
+// Does the command take file inputs
+func (c *Command) HasFiles() bool {
+	return c.Files != ""
+}
+
 // Get the complete FlagSet that applies to this command (local and persistent declared here and by all parents)
 func (c *Command) Flags() *flag.FlagSet {
 	if c.flags == nil {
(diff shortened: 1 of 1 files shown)


spf13:
@baruchlubinsky Thanks for the contribution. I'm not entirely sure what the value here is. Is it just to display [files ...] in the help? Can you add some documentation around it so it's clear what the new Field is for?

baruchlubinsky:
The CLI's that I've been writing are for file processing, and in some cases take one or more file names as arguments after the flags. The usage is the same as many linux programs such as grep. I was thinking that this is such a common use case that it does not merit writing a custom help template.

I'll put some working into improving the change and commit again.

spf13:
I just had an interesting conversation with another developer who's working on something very similar, but to use it not only for help but also to generate shell autocomplete. Important for his purposes is that you can specify a list of possible extensions as well as files. I'll point him this way and see if you can benefit from each other.

eparis:
@baruchlubinsky I'm the guy he's talking about.  I did work over in https://github.com/spf13/cobra/pull/69/ which still needs review.  I took a rather different approach.  I added generic 'annotations' to both pflags and commands.  One of my annotations, I called cobra.BashCompFilenameExt, is an annotation you can put on a pflag which, is is a list of acceptable filename extensions.  My particular program does not take filenames, instead certain flags do.

Following my model (which hasn't gotten any review) we could add a new annotation which you would put on the command (instead of on a specific flag).

But it does seem to me that you want to list specific acceptable files.  Which is actually fine.  We'd need a new annotation "cobra.BashCompFilename".

I'd say it's up to @spf13 if he likes the idea of more and more fields directly in pflags/command or if he likes my generic annotations stuff....

spf13:
I want to give this more thought. My gut is saying fields are preferable,
but I need to play with it a bit more.

--
Steve Francia
http://stevefrancia.com
http://spf13.com
http://twitter.com/spf13

eparis:
I preferred the annotations concept, because it let me pass any information I wanted.  They could be used for whatever my final application wanted.  In my case, I could pass annotations useful for my custom bash completions which cobra wasn't able to handle. It's generic and extensible and doesn't put cobra blocking anyone from making progress.  Of course, fields make things more discoverable and obvious.  I would probably really like to see both.  Some sort of generic extensibility option (my annotations) but at the same time, I can move the annotations I liked directly to fields to make it easier for others to find...

eparis:
https://github.com/GoogleCloudPlatform/kubernetes/compare/master...eparis:auto-generate-bash-completions

Should be the changes I've made to kubernetes to use my generator with annotations.

spf13:
I think that makes a lot of sense. Both sound right to me. 

Steve Francia
spf13.com
@spf13

eparis:
So I've not gone and done something else which is related.  #86 

I'm trying to make root command and regular commands 'less different.'  One way they are different is that root commands can not take args (other than flags) and complain about invalid subcommands.  subcommands always take args and never complain about invalid sub-sub-commands.  That PR adds a bool stating that a command (be it root or non-root) should accept arbitrary arguments or only look for subcommands. It defaults to 'false' which makes subcommand act like root commands do today...

eparis:
Obviously this ".Files" field implies ".TakesArgs"

I'm trying to decide my thoughts on the best way to address these things.  And figure out what fields we would want other than files....

baruchlubinsky:
I would agree that a generic solution is best. I've just written a command for an API client that takes one or more record ID's as inputs, following its flags. 

I've never implemented bash auto-completion before but I think its a really nice idea to include that here.

spf13:
@eparis Since you've made progress around this since this PR, what are your thoughts about it now?

spf13:
@eparis Ping

eparis:
I think this could be implemented on top of #120 very easily and cleanly....

ghodss:
+1 for the annotations on pflags and commands method.

eparis:
After 3 years I think I'm going to close this PR. I apologize that I was so slow in commenting all along, but I believe this should all be possible via annotations on commands and flags at this point.

PR #66, opened 2015-03-11
Fix up stripFlags to be more robust to different flag combinations

brendandburns:


command.go
```diff
@@ -264,8 +265,15 @@ func stripFlags(args []string) []string {
 				inQuote = true
 			case strings.Contains(y, "=\""):
 				inQuote = true
+			case strings.HasPrefix(y, "--") && !strings.Contains(y, "="):
```
filbranden: I think this will not work if I use `--boolflag`, in which case the second word should actually be the command, right?

command.go
```diff
@@ -264,8 +265,15 @@ func stripFlags(args []string) []string {
...
+				inFlag = true
+			case strings.HasPrefix(y, "-") && len(y) == 2:
```
filbranden: What about if I mix the flags together? For instance, boolean flag `-a` and boolean flag `-b`, I can write `-ab` for it...

filbranden:
I think the solution is to be able to tell spf13/pflag to stop processing flags once an argument is found, that way you can *know* that the next word is actually the command. Then you can add the new flags for the command and call spf13/pflag again to do a second round of parsing of flags including the ones for the specific command.

Does that make sense?

Sorry I haven't gotten to it... been busy with other stuff.

Cheers,
Filipe

command.go
brendandburns: ok, fixed this to include a check for bool flags.

brendandburns:
What you suggest makes sense, but I think that my solution is simpler and accomplishes the same thing.

Please take another look, I've added a bunch of test cases that I think validate that my code is working.

spf13:
LGTM

filbranden:
Ok looks good. I still think solving it in the pflag layer would have been more appropriate, but this works so that's fine. Thanks for the very comprehensive test coverage!

filbranden:
@spf13 Please go ahead and merge, I don't really have write permissions here, only in spf13/pflag...

spf13:
Merged as bbdea35. @filbranden You now have access here as well.

PR #68, opened 2015-03-14
fix missing newline in help output 

jmoiron:
The current help output for a simple program with commands looks like this:

```bash
Usage:
  iq [command]
Available Commands:
  command1
  command2
  ...
```

In the README and past versions there is a newline between the `Usage` and `Available Commands` section;  this behavior was changed (I suspect accidentally) in bf480fe628.  It also seems like there would be a missing newline for `Examples` sections, which this also adds back.


bep:
Looks good to me. Merged in e784a59

PR #69, opened 2015-03-16
Bash Autocompletion Generator

eparis:
That's right.  A golang program the writes a bash program!

### DIFF
--- a/README.md
+++ b/README.md
@@ -348,6 +348,9 @@ Like help the function and template are over ridable through public methods.
 
     command.SetUsageTemplate(s string)
 
+## Generating bash completions for your command
+
+Cobra can generate a bash completions file. If you add more information to your command these completions can be amazingly powerful and flexible.  Read more about [Bash Completions](bash_completions.md)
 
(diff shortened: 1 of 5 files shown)


eparis:
Requires https://github.com/spf13/pflag/pull/8 before this will build

eparis:
I pulled out annotations and hard coded a new "ValidArgs" field.  We don't really validate them, but we certainly could...

eparis:
An example of what this can do:

https://github.com/eparis/kubernetes/blob/931044be011e0bd910dcaa7ccbcceb2363195582/contrib/completions/bash/kubectl

fabianofranz:
@eparis Add README docs about how to use it.

eparis:
Now with 55% more README and 32% more tests!

fabianofranz:
Very nice! I could think of a better integration of `ValidArgs` (auto validation of the number of args, receive pointers the same way flags do, etc); but those could be enhancements. Looks good for now.

PR #70, opened 2015-03-16
Add mergePersistentFlags in strip flags since we now look at the flag set.

brendandburns:
@jlowdermilk @spf13 @smarterclayton ptal.

Thanks!
--brendan

### DIFF
--- a/command_test.go
+++ b/command_test.go
@@ -5,6 +5,8 @@ import (
 	"testing"
 )
 
+// TODO Test persistent flags on subcommands.  This requires testing the full execute function.
+
 func TestStripFlags(t *testing.T) {
 	tests := []struct {
 		input  []string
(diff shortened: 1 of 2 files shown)


smarterclayton:
Test case? :)

brendandburns:
Test cases added.

ptal.

Thanks!
--brendan

command_test.go
```diff
@@ -68,6 +76,7 @@ func TestStripFlags(t *testing.T) {
 	var flagi int
 	var flagstr string
 	var flagbool bool
+	cmdPrint.PersistentFlags().BoolVarP(&flagbool, "persist", "p", false, "help for persistent one")
```
jlowdermilk: Can you add a subcommand of `cmdPrint` with a persistent flag, and verify that stripFlags works?

jlowdermilk: TODO is fine, thanks.

jlowdermilk:
lgtm

PR #71, opened 2015-03-18
Default usage output to stdout

eparis:
If the command has not set an output explicitly everything will go to
stderr.  This makes a lot of sense, but is a huge PITA for people who
want to be able to grep the help output.  It is very common for users to
want to do

command --help | grep flag

This patch fixes that by default help output (but not error output like
an invalid command) to stdout instead of defaulting to stderr.


bep:
Merged in 6de96b8

PR #74, opened 2015-03-19
Help uses Short message if Long is not available

edgard:
I'm a little green on templates but I think that should do it.


eparis:
LGTM

bep:
Merged in 384c059

Issue #44, opened 2015-01-14
Global flags aren't used/parsed


jbeda:
If someone creates a flag into the `pflag.CommandLine` flag set, it isn't imported/used/parsed.  It should instead be merged with the rest of the persistent flags.

eparis:
This is now working!

PR #75, opened 2015-03-19
Make flags declared outside spf13.pflag persistent flags on the parent

eparis:
Some projects pick up flags from other projects they include.  A great
example would be projects that use glog and thus get all of those flags.
Kubernetes, for example, merges those flags manually into its commands.
    
This was reported in https://github.com/spf13/cobra/issues/44
    
What this patch does is merge those flags into the PersistentFlags on
the highest parent.  This allows kubernetes to stop having to merge
things themselves...

### DIFF
--- a/command.go
+++ b/command.go
@@ -942,6 +942,13 @@ func (c *Command) mergePersistentFlags() {
 		c.PersistentFlags().VisitAll(addtolocal)
 	}
 	rmerge = func(x *Command) {
+		if ! x.HasParent() {
+			flag.CommandLine.VisitAll(func(f *flag.Flag) {
+				if x.PersistentFlags().Lookup(f.Name) == nil {
+					x.PersistentFlags().AddFlag(f)
+				}
+			})
+		}
 		if x.HasPersistentFlags() {
 			x.PersistentFlags().VisitAll(func(f *flag.Flag) {
 				if c.Flags().Lookup(f.Name) == nil {


eparis:
Solves #44

ApsOps:
@eparis Hi. I'm trying to use glog with cobra, but unable to get `-v` flag working. Do I need to add this flag in cobra as well?

Can you please share an example snippet explaining how to do this?

eparis:
try using `--v` ?

ApsOps:
@eparis I tried that. It says `unknown flag: --v`

eparis:
https://github.com/kubernetes/contrib/blob/master/mungegithub/github/github.go#L359

Is an example that I know works. Just this morning I used `--logtostderr --v=10` to get glog flags...

ApsOps:
@eparis Ah, thanks a lot!
I was missing `RootCmd.PersistentFlags().AddGoFlagSet(flag.CommandLine)`. Probably missed this in docs.

Okay. But now I'm also getting `ERROR: logging before flag.Parse` before every log line. I tried adding `flag.CommandLine.Parse([]string{})` but then logs aren't shown at all. And I'm not entirely sure what this means.

Looks like I'm missing a couple of things. Here is my `cmd/root.go`. Can you please suggest what all is required?

```go
package cmd

import (
	"flag"
	"fmt"
	"os"

	"github.com/golang/glog"
	"github.com/spf13/cobra"
)

var cfgFile string

// RootCmd represents the base command when called without any subcommands
var RootCmd = &cobra.Command{
	Use:   "cobra_glog",
	Short: "A brief description of your application",
	Long:  `A longer description.`,
		glog.Info("Info log")
		glog.V(3).Info("level 3 log")
	},
}

// Execute adds all child commands to the root command sets flags appropriately.
// This is called by main.main(). It only needs to happen once to the rootCmd.
func Execute() {
		fmt.Println(err)
		os.Exit(-1)
	}
}

func init() {
	RootCmd.PersistentFlags().StringVar(&cfgFile, "config", "", "config file (default is $HOME/.cobra_glog.yaml)")
	RootCmd.Flags().BoolP("toggle", "t", false, "Help message for toggle")
	RootCmd.PersistentFlags().AddGoFlagSet(flag.CommandLine)
}
```

eparis:
Ugggh, I forgot glog had the bit of PITA. This example seems to work for me.
```
package main

import (
        goflag "flag"
        "fmt"
        "os"

        "github.com/golang/glog"
        "github.com/spf13/cobra"
)

var cfgFile string

// RootCmd represents the base command when called without any subcommands
var RootCmd = &cobra.Command{
        Use:   "cobra_glog",
        Short: "A brief description of your application",
        Long:  `A longer description.`,
                goflag.CommandLine.Parse([]string{})
                glog.Info("Info log")
                glog.V(3).Info("level 3 log")
        },
}

// Execute adds all child commands to the root command sets flags appropriately.
// This is called by main.main(). It only needs to happen once to the rootCmd.
func Execute() {
                fmt.Println(err)
                os.Exit(-1)
        }
}

func init() {
        RootCmd.PersistentFlags().StringVar(&cfgFile, "config", "", "config file (default is $HOME/.cobra_glog.yaml)")
        RootCmd.Flags().BoolP("toggle", "t", false, "Help message for toggle")
        RootCmd.PersistentFlags().AddGoFlagSet(goflag.CommandLine)
}

func main() {
        glog.Errorf("glog message before parse\n")
        Execute()
}
```

iwinux:
For anyone who has Google'd "cobra glog" and found this issue:

```go
root := cobra.Command{
	PersistentPreRun: func(cmd *cobra.Command, args []string) {
        // to convince glog that we have called `flag.Parse()`,
        // without actually consuming CLI flags
		stdFlag.CommandLine.Parse([]string{})
	},
}
	
// bonus: log to stderr by default,
// override with `--logtostderr=false`
stdFlag.Set("logtostderr", "true")
	
// make glog flags available in Cobra's CLI
root.PersistentFlags().AddGoFlagSet(stdFlag.CommandLine)
```

PR #79, opened 2015-03-26
Prevent incorrect flag message on bad subcommand parse

deads2k:
If an invalid argument is passed to a subcommand, but the root command is runnable, the flags are reparsed and the wrong error message appears to a user.

For instance 
```
openshift start node --images=foo --master=bar
```
If `openshift start node` has a `--images` flag, but does not have a `--master` flag, the parse will fail (good).  But after that failure, if openshift is runnable and doesn't have any flags, then the message displayed to the user `unknown flag: --images` which is incorrect.

I've added a test to demonstrate.

### DIFF
--- a/cobra_test.go
+++ b/cobra_test.go
@@ -417,6 +417,33 @@ func TestTrailingCommandFlags(t *testing.T) {
 	}
 }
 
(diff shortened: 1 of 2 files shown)


eparis:
Closed via #84

PR #84, opened 2015-04-03
Stop special casing runnable root commands

eparis:
The special case code to handle a runnable root command had some
problems.  It was noticed that if you created a runnable root and a
subcommand.  And the subcommand was then executed with both a valid and
invalid flag, the error message was about the valid flag being invalid.
For example

./command subcommand --goodflag=10 --badflag=10

Would fail and tell you that --goodflag was an invalid flag. Instead if
we just do away with the special Command.execute() for the root command
the parser for subcommand is what prints the error and it gets it
right...

### DIFF
--- a/command.go
+++ b/command.go
@@ -305,6 +305,8 @@ func stripFlags(args []string, c *Command) []string {
 				inFlag = true
 			case inFlag:
 				inFlag = false
+			case y == "":
+				// strip empty commands, as the go tests expect this to be ok....
 			case !strings.HasPrefix(y, "-"):
 				commands = append(commands, y)
 				inFlag = false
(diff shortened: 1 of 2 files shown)


eparis:
@deads2k

https://github.com/spf13/cobra/pull/79

What do you think about solving the above like this?

deads2k:
No objection.  I was just trying for minimal change.

command.go
```diff
@@ -341,6 +341,10 @@ func (c *Command) Find(arrs []string) (*Command, []string, error) {
 		return c.Root(), arrs, nil
 	}
 
+	for len(arrs) > 0 && len(arrs[0]) == 0 {
```
deads2k: Stripping leading empty arguments?  Is this related to the bad problem or just something else you found on the way?

deads2k:
It's difficult to read the diff and determine complete fidelity of error reporting with the existing code, but the new version is more readable.

lgtm.

command.go
eparis: It related to the fact that root commands have special code to ignore it later. So I'm just ripping them out for all commands.

Issue #42, opened 2014-12-27
If no subcommand, how to get args?


kk44:
example:

``` go
package main

import (
    "fmt"
    "github.com/spf13/cobra"
)

func main() {
    var echo = &cobra.Command{
        Use:    "echo [string...]",
        Short:  "echo all strings",
        Long:   "echo string string string string...",
        Run:    func(cmd *cobra.Command, args []string) {
            fmt.Printf("%v\n", args)
        },
    }

    echo.Execute()
}
```

```
[bash]# go run echo.go aa bb cc dd

Usage:
  echo [string...] [flags]

 Available Flags:
  -h, --help=false: help for echo
Error: unknown command "aa"
Run 'help' for usage.

echo: invalid command [`aa` `bb` `cc` `dd`]
Run 'echo help' for usage
```

```
[bash]# go run echo.go echo aa bb cc dd
[echo aa bb cc dd]
```

```
[bash]# go run echo.go
[]
```

Thanks!

ghodss:
I'm also having this issue. Is having zero subcommands supported? It doesn't seem to be working.

eparis:
It isn't possible.  The code treats the root command "special".  I don't really love it.

Root commands my only take flags and subcommands. Validation for subcommands is only done for root commands.

Subcommands will be called with an argument instead of failing if an invalid sub-sub-command was intended, but typod.

I'm thinking about what to do here in https://github.com/spf13/cobra/pull/86

But any kind of consistency is going to break previous behavior. Personally I'm leaning towards a 'takesNoArgs' flag which defaults to false. Meaning the root command, and all subcommands, would get unknown args.

You would only get subcommand validation and complain if you set takesNoArgs=true

viktorbenei:
We recently started to migrate from codegangsta / urfave CLI, and we love Cobra so far! But we have a couple of tools which are primarily designed to run other programs (think about something `/usr/bin/env` or `/usr/bin/ssh`, but with a couple of utility commands like `init` built into the tool), and we can't figure out how we could do that with Cobra.

IMO what @ghodss described (and what's essentially how this works in codegangsta / urfave CLI) at https://github.com/spf13/cobra/pull/120#issuecomment-116910479 / https://github.com/spf13/cobra/pull/86#issuecomment-102215537 makes sense, or be able to specify a "default" command in any other way.

Is there a workaround for this use case for now?

viktorbenei:
[...]

PR #86, opened 2015-04-06
Require commands to explicitly state they take arbitrary arguments

eparis:
The code is riddled with inconsistencies between the root command and
subcommands.  One of those is that the root command cannot take
arbitrary arguments and will complain about invalid subcommands.  But
subcommands will always take arbitrary arguments and will never complain
about invalid sub-sub-commands. I make arbitrary arguments explicitly
listed. You now need to set cmd.TakesArgs = true if you need to be able
to accept arguments tha aren't flags.

This is a pretty major change.

### DIFF
--- a/cobra_test.go
+++ b/cobra_test.go
@@ -23,30 +23,41 @@ const strtwoParentHelp = "help message for parent flag strtwo"
 const strtwoChildHelp = "help message for child flag strtwo"
 
 var cmdPrint = &Command{
-	Use:   "print [string to print]",
-	Short: "Print anything to the screen",
-	Long:  `an absolutely utterly useless command for testing.`,
+	Use:       "print [string to print]",
+	Short:     "Print anything to the screen",
+	Long:      `an absolutely utterly useless command for testing.`,
+	TakesArgs: true,
 	Run: func(cmd *Command, args []string) {
 		tp = args
 	},
 }
 
 var cmdEcho = &Command{
-	Use:     "echo [string to echo]",
-	Aliases: []string{"say"},
-	Short:   "Echo anything to the screen",
-	Long:    `an utterly useless command for testing.`,
+	Use:       "echo [string to echo]",
+	Aliases:   []string{"say"},
+	Short:     "Echo anything to the screen",
+	Long:      `an utterly useless command for testing.`,
+	TakesArgs: true,
 	Run: func(cmd *Command, args []string) {
 		te = args
 	},
 }
 
 var cmdTimes = &Command{
-	Use:   "times [# times] [string to echo]",
-	Short: "Echo anything to the screen more times",
-	Long:  `a slightly useless command for testing.`,
+	Use:       "times [# times] [string to echo]",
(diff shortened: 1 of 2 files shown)


eparis:
@spf13 this might be pretty controversial....   It will require updates to a lot of code...

spf13:
@eparis I think I see the value of this, but I also see it as an extra step. I'm hesistant to merge this in. I'd definitely want to hear what others think first. 

@smarterclayton @brendandburns @bep  care to weigh in?

My take is, I don't dislike this, but I think perhaps it should be the other way around. Instead of TakeArgs, what if it's "Final" or  "No Args" or something. The default is that args are allowed, but you can be explicit to make it not accept args. This would allow us to add this type of feature and not break all existing code.

ghodss:
How about, if the given command has no further subcommands, pass the args directly. But if it does have subcommands and you can't match against one, return an error. That would satisfy 99% of use cases without needing to add another parameter/complexity of setup. Right now it's nearly a straight up bug that you can't use Cobra without subcommands but with arguments. See #42.

I implemented the above approach for the root command only in #108. Should be feasible to improve it to work generically.

spf13:
@ghodss I must admit I don't think I've fully thought through the implications of this change, but I like the logic you are proposing. I think it drives a nice compromise that keeps the interface simple and resolves the issue you mention. Thank you for giving this the extra thought to come up with this more elegant solution. 

@eparis what are your thoughts here?

eparis:
I like the logic. It also means we should be able to write cleaner code and continue to eliminate the 'root command is a special snowflake' stuff still floating in there. We may need an eventual 'CmdTakesArgs' flag to support people who want to pass arbitrary arguments to subcommands but I'm happy to break those people to make this work   :)

eparis:
If that CmdTakesArgs was a string used in the help text, that would solve #61 as well...

ghodss:
Actually, I changed my mind on this. I think returning an error anytime there are subcommands is bad behavior. Here's what really should happen:

Start at the root command and start matching arguments to the subcommand tree. When you can't match anymore, see if the command that you've matched against has a Run function. If it does, run it, passing in all unmatched args. If it doesn't, print unknown command + help.

Incidentally this is how codegangsta/cli works, and I think it's the right behavior. Because if a developer has specified a run function AND subcommands, they are obviously expecting the run function to be run if a subcommand isn't matched - why else would they have specified a run function? If they want to print a custom error when that happens, they can do that. And more importantly, if they want to actually take the arguments and do something with them, they can do that too. We shouldn't be taking away their ability to specify behavior when a command's args don't match the given subcommands, as we are today with the root command.

This design would also allow the cmd.TakesArgs parameter in this PR to be implicitly derived from the above logic.

Ideally #108 would still be merged since it gets us closer to the above behavior, but to make it consistent in every case there likely needs to be some more extensive refactoring.

pdf:
@ghodss this logic appears sound to me.

eparis:
I'm going to close this as I like the behavior in #120 more....

PR #90, opened 2015-04-07
Deprecated subcommands

eparis:
They should still work, but shouldn't show up in help or usage output

### DIFF
--- a/cobra_test.go
+++ b/cobra_test.go
@@ -51,6 +51,15 @@ var cmdEchoSub = &Command{
 	},
 }
 
+var cmdDeprecated = &Command{
+	Use:        "deprecated [can't do anything here]",
+	Short:      "A command which is deprecated",
+	Long:       `an absolutely utterly useless command for testing deprecation!.`,
+	Deprecated: "Please use echo instead",
+	Run: func(cmd *Command, args []string) {
+	},
+}
+
 var cmdTimes = &Command{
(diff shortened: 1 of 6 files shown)


eparis:
@thockin and deprecation of subcommands

cobra_test.go
```diff
@@ -42,9 +42,10 @@ var cmdEcho = &Command{
 }
 
 var cmdTimes = &Command{
-	Use:   "times [# times] [string to echo]",
-	Short: "Echo anything to the screen more times",
-	Long:  `a slightly useless command for testing.`,
+	Use:        "times [# times] [string to echo]",
+	Short:      "Echo anything to the screen more times",
+	Long:       `a slightly useless command for testing.`,
+	Deprecated: true,
```
thockin: rather than a bool, maybe this should a be a redirect string?

Deprecated: "see foobar instead"

eparis: And use that in place of Use (or short or long)?  Should projects just change "Use" and we just not support 'deprecation' ?

thockin: The difference between deprecation and use is that the deprecation warning
should actually print whenever you use a deprecated feature.

eparis: That's a good thought. Do you think we should hide them from --help text?

thockin: yes, I think so.

eparis:
@thockin rewritten to use a string!

PR #94, opened 2015-04-22
Fix a bug where errors were masked.

brendandburns:
@eparis @jlowdermilk @spf13

### DIFF
--- a/command.go
+++ b/command.go
@@ -27,6 +27,14 @@ import (
 	"time"
 )
 
+type UnknownCommandError struct {
+  commandStr string
+}
+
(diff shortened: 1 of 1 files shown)


eparis:
@brendandburns Looks like your push isn't quite right.  Is this trying to fix the same thing as #79 which I hopefully fixed in #84

brendandburns:
I will update our vendored cobra to pick up your edits and see...

(and I'll fix the bad merge...)

--brendan

brendandburns:
*sigh* yes it appears you already fixed this bug...  That will teach me to not sync to head.

PR #95, opened 2015-04-24
prevent removal of valid arguments

deads2k:
running a command like: 
```
openshift admin policy add-role-to-user admin my-user
```
where `admin` and `my-user` are arguments to the `openshift admin policy add-role-to-user` command, removes the `admin` argument (`args[4]`).

This pull stops removing more than one token from the `args` for command matches.

@smarterclayton Related to https://github.com/openshift/origin/issues/1894

### DIFF
--- a/command.go
+++ b/command.go
@@ -328,15 +328,18 @@ func stripFlags(args []string, c *Command) []string {
 	return commands
 }
 
-func argsMinusX(args []string, x string) []string {
-	newargs := []string{}
-
-	for _, y := range args {
-		if x != y {
-			newargs = append(newargs, y)
+// argsMinusFirstX removes only the first x from args.  Otherwise, commands that look like
+// openshift admin policy add-role-to-user admin my-user, lose the admin argument (arg[4]).
+func argsMinusFirstX(args []string, x string) []string {
+	for i, y := range args {
+		if x == y {
+			ret := []string{}
+			ret = append(ret, args[:i]...)
+			ret = append(ret, args[i+1:]...)
+			return ret
 		}
 	}
-	return newargs
+	return args
 }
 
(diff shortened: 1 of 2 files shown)


smarterclayton:
Add a test case please.

deads2k:
test added.

smarterclayton:
LGTM.  @brendandburns can you review as well?  Impacts kube in a few places.

command.go
```diff
@@ -328,12 +328,17 @@ func stripFlags(args []string, c *Command) []string {
...
+func argsMinusFirstX(args []string, x string) []string {
+	removedFirstX := false
 	newargs := []string{}
 
 	for _, y := range args {
-		if x != y {
+		if removedFirstX || (x != y) {
 			newargs = append(newargs, y)
+		} else {
+			removedFirstX = true
```
eparis: Being always the bikeshedder, I like this better (and think it's faster, more readable, more efficient...)  What do you think?

https://gist.github.com/eparis/458542ddda01fa9850f9

deads2k: > What do you think?

I'm not against it, except for it not working: 
```
[deads@deads-dev-01 cobra]$ go test
Error: unknown command "one"
Run 'print help' for usage.
--- FAIL: TestChildSameName (0.00 seconds)
	cobra_test.go:345: Wrong command called
	cobra_test.go:348: Command didn't parse correctly
```

I checked in a similar version that passes.

smarterclayton:
LGTM

command.go
eparis: @deads2k yeah, so I guess my version was actually mutating the original slice, rather than creating a new slice with just the correct data.  Thanks for the new (working) version!

PR #102, opened 2015-05-04
Fixed Persistent-Run function propagation

andronat:
This is a fix for the following use case:

If I create a command A that is going to be a subcommand on B and then I add B as a subcommand to C (with this particular order on AddCommand), then A is not going to inherit PersistentPostRun and PersistenPreRun functions.


eparis:
can you add a test to cobra_test.go otherwise looks good

andronat:
sure

andronat:
Done

PR #108, opened 2015-05-13
Fix root command without subcommands but with arguments

### LINKED ISSUE #42: If no subcommand, how to get args?

ghodss:
Fixes #42.

### DIFF
--- a/cobra_test.go
+++ b/cobra_test.go
@@ -4,6 +4,7 @@ import (
 	"bytes"
 	"fmt"
 	"os"
+	"reflect"
 	"runtime"
 	"strings"
 	"testing"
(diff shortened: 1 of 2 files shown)


ghodss:
See https://github.com/spf13/cobra/pull/86#issuecomment-101480961 for discussion.

PR #109, opened 2015-05-13
genmarkdown: add optional frontmatter- and linkadjustment-funcs

bep:
The automatic Markdown generator works great!

But to use it to render the documentation in Hugo, we need front matter and slightly different links.

This commit adds optional callback funcs to add that.


bep:
@eparis some great tooling you add to Cobra. What about this adjustment?

eparis:
I'd love to see some documentation on how to use these new function, otherwise LGTM

bep:
Sure - I'll add some lines to the README.

PR #110, opened 2015-05-14
Add Global Normalization Function

andronat:
Added a global function to add normalization functions to all pFlag sets. That helps in cases that someone has lots of subcommands and want to handle all flag names in a consistent and global way.


spf13:
Thanks andronat. This looks really helpful.

PR #116, opened 2015-06-22
Correctly print the unknown command name

sgotti:
by now, if someone calls: `program --validflag unknowncommand` the
output will be:

```
Error: unknown command "--validflag"
Run 'program help' for usage.
```

This patch strips out flags so the unknown command is printed:

```
Error: unknown command "unknowncommand"
Run 'program help' for usage.
```

### DIFF
--- a/command.go
+++ b/command.go
@@ -424,7 +424,7 @@ func (c *Command) Find(args []string) (*Command, []string, error) {
 
 	// If we matched on the root, but we asked for a subcommand, return an error
 	if commandFound.Name() == c.Name() && len(stripFlags(args, c)) > 0 && commandFound.Name() != args[0] {
-		return nil, a, fmt.Errorf("unknown command %q", a[0])
+		return nil, a, fmt.Errorf("unknown command %q", stripFlags(args, c)[0])
 	}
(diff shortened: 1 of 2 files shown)


jonboulle:
lgtm, fwiw

command.go
```diff
@@ -424,7 +424,7 @@ func (c *Command) Find(args []string) (*Command, []string, error) {
 
 	// If we matched on the root, but we asked for a subcommand, return an error
 	if commandFound.Name() == c.Name() && len(stripFlags(args, c)) > 0 && commandFound.Name() != args[0] {
-		return nil, a, fmt.Errorf("unknown command %q", a[0])
+		return nil, a, fmt.Errorf("unknown command %q", stripFlags(args, c)[0])
```
eparis: Should it be args or a, I guess they are the same at this point...

eparis:
@jonboulle @sgotti What do you think of this fix? Should handle your fix, as well as other stuoopid...

eparis:
whoops, means to say #119 NOT this fix   :)

obviously you were good with this fix!

PR #117, opened 2015-06-22
fall back to filedir completion

liggitt:
If no known completions exist, it seems better to fall back to filedir completion than to do nothing.
@eparis, opinions on this?

### DIFF
--- a/bash_completions.go
+++ b/bash_completions.go
@@ -88,6 +88,11 @@ __handle_reply()
     fi
     COMPREPLY=( $(compgen -W "${completions[*]}" -- "$cur") )
 
+    if [[ ${#completions[@]} -eq 0 ]]; then
+    	__debug "${FUNCNAME}: no known completions, falling back to _filedir"
+    	_filedir
+    fi
+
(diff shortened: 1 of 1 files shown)


bash_completions.go
```diff
@@ -88,6 +88,11 @@ __handle_reply()
     fi
     COMPREPLY=( $(compgen -W "${completions[*]}" -- "$cur") )
 
+    if [[ ${#completions[@]} -eq 0 ]]; then
+    	__debug "${FUNCNAME}: no known completions, falling back to _filedir"
+    	_filedir
+    fi
+
```
eparis: I'd rather a commend had to say it would take an arbitrary file as an argument instead of just always doing it. Also, if we do just sorta randomly default to _filedir should it go after the __custom_func handler? I don't want people to start randomly completely filenames and never running the __custom_func...

liggitt: ah, ok... I misread how that worked

eparis: also, easy way to get that in your code would be just add _filedir to the end of your __custom_func....

liggitt: agreed, closing

PR #119, opened 2015-06-22
Correctly print the unknown command name

eparis:
by now, if someone calls: program --validflag unknowncommand the
output will be:

```
Error: unknown command "--validflag"
Run 'program help' for usage.
```
This patch strips out flags so the unknown command is printed:
```
Error: unknown command "unknowncommand"
Run 'program help' for usage.
```

Reported-by: Simone Gotti in https://github.com/spf13/cobra/pull/116

This also fixes a problem with the original code where if you had a
root command and a grand child with the same name, the parser would
break and would not run the grandchild. The code was special casing if
the immediate child had the same name, but didn't handle grand-children

### DIFF
--- a/cobra_test.go
+++ b/cobra_test.go
@@ -378,6 +378,24 @@ func TestChildSameName(t *testing.T) {
 	}
 }
 
+func TestGrandChildSameName(t *testing.T) {
+	c := initializeWithSameName()
+	cmdTimes.AddCommand(cmdPrint)
+	c.AddCommand(cmdTimes)
+	c.SetArgs(strings.Split("times print one two", " "))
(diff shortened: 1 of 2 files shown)


sgotti:
@eparis LGTM. I was doing something similar in another patch as I noticed the same problem (but wanted to keep the two patches separated for clearness).

eparis:
@sgotti that's fair. I'll merge your patch and rebase mine on top. I'm going to push a new PR with a whole slew of changes you might want to check out. Really cleans up a bunch.

sgotti:
@eparis sure! Thanks!

Issue #63, opened 2015-03-04
Unknown commands '-h' and '--help'


kizbitz:
Using the default example with 'Short' & 'Long' added to rootCmd:
```
package main

import (
	"fmt"
	"github.com/spf13/cobra"
	"strings"
)

func main() {

	var echoTimes int

	var cmdPrint = &cobra.Command{
		Use:   "print [string to print]",
		Short: "Print anything to the screen",
		Long: `print is for printing anything back to the screen.
        For many years people have printed back to the screen.
        `,
		Run: func(cmd *cobra.Command, args []string) {
			fmt.Println("Print: " + strings.Join(args, " "))
		},
	}

	var cmdEcho = &cobra.Command{
		Use:   "echo [string to echo]",
		Short: "Echo anything to the screen",
		Long: `echo is for echoing anything back.
        Echo works a lot like print, except it has a child command.
        `,
		Run: func(cmd *cobra.Command, args []string) {
			fmt.Println("Print: " + strings.Join(args, " "))
		},
	}

	var cmdTimes = &cobra.Command{
		Use:   "times [# times] [string to echo]",
		Short: "Echo anything to the screen more times",
		Long: `echo things multiple times back to the user by providing
        a count and a string.`,
		Run: func(cmd *cobra.Command, args []string) {
			for i := 0; i < echoTimes; i++ {
				fmt.Println("Echo: " + strings.Join(args, " "))
			}
		},
	}

	cmdTimes.Flags().IntVarP(&echoTimes, "times", "t", 1, "times to echo the input")

	var rootCmd = &cobra.Command{
		Use:   "usage string",
		Short: "short string",
		Long:  "long string",
	}

	rootCmd.AddCommand(cmdPrint, cmdEcho)
	cmdEcho.AddCommand(cmdTimes)
	rootCmd.Execute()
}
```
Output of `app` and `app help` is correct and lists '-h' and '--help' as global flags:
```
jbaker: )➤ app                                                                                                                                                                 
long string

Usage:
  usage [command]

Available Commands:
  print [string to print]   Print anything to the screen
  echo [string to echo]     Echo anything to the screen
  help [command]            Help about any command


Global Flags:
  -h, --help=false: help for usage

Use "usage help [command]" for more information about a command.
```
```
jbaker: )➤ app help                                                                                                                                                                 
long string

Usage:
  usage [command]

Available Commands:
  print [string to print]   Print anything to the screen
  echo [string to echo]     Echo anything to the screen
[...]

PR #120, opened 2015-06-22
Allow commands to specify if they take arbitrary arguments

eparis:
This does a lot of cleanup, largely to Find() and Execute(). I'm trying to make the code easy to read and comprehend. After the cleanup we add the ability to specify if your command takes arguments or not. (Although defaulting to the nutty legacy behavior)

The legacy behavior is
- Any command with no subcommands takes arbitrary arguments
- A subcommand takes arbitrary arguments
- A root command with subcommands can never take arbitrary arguments

The new behavior is explicit per command and root v subcommand is irrelevant.

### DIFF
--- a/cobra_test.go
+++ b/cobra_test.go
@@ -68,6 +68,7 @@ var cmdDeprecated = &Command{
 	Deprecated: "Please use echo instead",
 	Run: func(cmd *Command, args []string) {
 	},
+	TakesArgs: None,
 }
 
 var cmdTimes = &Command{
@@ -80,6 +81,7 @@ var cmdTimes = &Command{
 	Run: func(cmd *Command, args []string) {
 		tt = args
 	},
+	TakesArgs: Arbitrary,
 }
 
 var cmdRootNoRun = &Command{
@@ -92,9 +94,20 @@ var cmdRootNoRun = &Command{
 }
 
 var cmdRootSameName = &Command{
-	Use:   "print",
-	Short: "Root with the same name as a subcommand",
-	Long:  "The root description for help",
+	Use:       "print",
+	Short:     "Root with the same name as a subcommand",
+	Long:      "The root description for help",
+	TakesArgs: None,
+}
+
+var cmdRootTakesArgs = &Command{
+	Use:   "root-with-args [random args]",
+	Short: "The root can run it's own function and takes args!",
+	Long:  "The root description for help, and some args",
+	Run: func(cmd *Command, args []string) {
+		tr = args
+	},
+	TakesArgs: Arbitrary,
 }
 
 var cmdRootWithRun = &Command{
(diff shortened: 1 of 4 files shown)


cobra_test.go
```diff
@@ -80,6 +81,8 @@ var cmdTimes = &Command{
 	Run: func(cmd *Command, args []string) {
 		tt = args
 	},
+	TakesArgs: Arbitrary,
+	ValidArgs: []string{"one", "two", "three", "four"},
```
eparis: Maybe this should be `TakesArgs: ValidOnly`   or something like that?

eparis:
This should fix https://github.com/spf13/cobra/issues/63

eparis:
@sgotti Thoughts?

ghodss:
I don't believe this is the right approach. It makes the framework more complicated than necessary. As I mentioned in https://github.com/spf13/cobra/pull/86#issuecomment-102215537, the simplest behavior is:

> Start at the root command and start matching arguments to the subcommand tree. When you can't match anymore, see if the command that you've matched against has a Run function. If it does, run it, passing in all unmatched args. If it doesn't, print unknown command + help.

What this gives you is full flexibility to enable any outcome you might want without requiring people to learn and add new fields. You can enable "ValidArgs" with a simple if statement and most of the TakesArgs options are implementable with the default behavior above. If you really want something special, then you probably should be dispatching to the right subcommands yourself in your Run statement.

eparis:
Logic today when we descend the tree and have left over args.
- root commands with no subcommands can take arbitrary arguments
- root commands with subcommands will do subcommand validity checking
- subcommands will always accept arbitrary arguments and do no subsubcommand validity checking
- if you descend to a non-runnable command you get help, but no mention of the extra 'args'

@ghodss wants something very close.
- if command runnable, run it and pass args (this is how subcommands, but not root) work today.
- if command not runnalbe - complain about subcommand if args != 0

So the only real change is around the root command. Which I am fine with. I think normalizing, even if it breaks someone, would be a good idea!

But that doesn't solve the #1 think I what I really want to do here. I want to make it easy to say "does not take args". Some of the other stuff is just nice to have. But I've got too many Run functions that start with if `len(args) != 0 { printf; return }`

I'd like if the library was able to do that checking for me.  That is my TakesArgs: None. Do you think we should support this?
You suggest that my TakesArgs: Arbitrary should be irrelevant/default. Ok.
But you think that ValidArgs should be pushed into the application. I don't love it, but I'm ok with it.

kofalt:
FWIW - as some looking at using this library, your described change above sounds like a great idea. I've used libraries before that are not very smart with arbitrary args + nested commands. 

I also like not considering the root command as a special case; ideally it shouldn't be. :+1:

dnephin:
I'd like to be able to use this as well.  The default behaviour of allowing arbitrary args on every command requires a user who wants strict validation to check the lengths of `args` everywhere. 

It would be a lot nicer if this could be handled with a declaration (as it is in this PR).

I've rebased this PR and fixed some typos/working in the README in https://github.com/spf13/cobra/compare/master...dnephin:define_args

@eparis are you still interested in this PR, or should I open a new one with that branch?

bogem:
Closed in #284

PR #126, opened 2015-08-09
Add new BashCompSubdirsInDir annotation

anthonyfok:
This first `cd` to a specified directory, then
lists the subdirectories therein with `_filedir -d`.

This can be used by e.g. `hugo --theme=[Tab][Tab]`, which would
give a list of subdirectories under the `themes` directory.
See Pull Request spf13/hugo#1343 for more information.

### DIFF
--- a/bash_completions.go
+++ b/bash_completions.go
@@ -13,6 +13,7 @@ import (
 const (
 	BashCompFilenameExt     = "cobra_annotation_bash_completion_filename_extentions"
 	BashCompOneRequiredFlag = "cobra_annotation_bash_completion_one_required_flag"
+	BashCompSubdirsInDir    = "cobra_annotation_bash_completion_subdirs_in_dir"
 )
 
 func preamble(out *bytes.Buffer) {
@@ -98,6 +99,12 @@ __handle_filename_extension_flag()
 {
     local ext="$1"
     _filedir "@(${ext})"
+}
+
+__handle_subdirs_in_dir_flag()
+{
+    local dir="$1"
+    pushd "${dir}" >/dev/null 2>&1 && _filedir -d && popd >/dev/null 2>&1
 }
 
 __handle_flag()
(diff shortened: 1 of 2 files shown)


bash_completions.go
```diff
@@ -100,6 +101,12 @@ __handle_filename_extension_flag()
     _filedir "@(${ext})"
 }
 
+__handle_subdirs_in_dir_flag()
+{
+    local dir="$1"
+    cd "${dir}" >/dev/null 2>&1 && _filedir -d && cd - >/dev/null
```
eparis: I prefer `pushd` and `popd` instead of `cd` `cd -` as push/pop don't get confused if cwd changes are made by the intervening operations. In this case it doesn't look like _filedir() bash completions code uses cd/push either, so it would be fine. But since half of the improvements to open source are small iterations of the work which has come before, I'd rather your code be harder for someone else to get wrong in the future.

anthonyfok:
Thank you very much for the code review and feedback!
Indeed the use of `pushd` and `popd` is much more robust solution!
Thanks for the great tip!
I will make the changes and push it again right away.

eparis:
should have said this already.  so sorry.  But can you please write something in bash_completions_test.go which at least trys to make sure the correct bash is getting written out?

eparis:
(yes, I realize most of the tests are poor, but at least get a minimum of testing there)

anthonyfok:
Oops, I totally forgot about the tests too!  I am too forgetful and lazy.  &lt;grin, duck, run&gt;
And thanks for the reminder!

Here is the latest `git push` with a minimal test.
I don't know what to name the variable/flag for the `subdirs_in_dir` test,
so, being very uncreative, I just used `theme`.  Hope that is okay.  :-)

If there are any other issues, please do not hesitate to let me know.
For example, I am not sure whether "subdirs_in_dir" is a good name,
and I am open to suggestion to other names.  :-)

Thanks again!

Issue #129, opened 2015-08-23
Enable the specification of the Help flag's shorthand character and usage value


akutz:
I would like to be able to specify the value used for the help flag's shorthand character so that when there's a global "host" flag, it can use the "h" for its shorthand and the help flag can use "?". Additionally, the ability to format the help flag's usage value would also be nice for applications with different casing standards for help output.

It appears that the help flag is automatically generated as part of the core cobra package (when a flag set is nil), so I am submitting a patch that should make it easy for this routine to take into account user preference.

eparis:
A work around would be to define the `help` flag yourself with either a different shorthand or no shorthand. Cobra's help will only get added if you didn't do it...

akutz:
Well crap on a stick, how did I miss that? I saw you can do that for the command, but not the flag. I tried visiting all flags for all commands, setting each help flag's shorthand to what I wanted when I matched the name, but still the panic.

FWIW, here's the commit with which I was going to open a pull request. It doesn't seem necessary now. I'll test your suggestion and if it works, close this issue.

eparis:
(I could be wrong, wouldn't be the first time)

akutz:
Are you sure you're not thinking of the command? The source for how the help flag is created doesn't seem to allow for the previous existence of said flag:

```
// Get the complete FlagSet that applies to this command (local and persistent declared here and by all parents)
func (c *Command) Flags() *flag.FlagSet {
	if c.flags == nil {
		c.flags = flag.NewFlagSet(c.Name(), flag.ContinueOnError)
		if c.flagErrorBuf == nil {
			c.flagErrorBuf = new(bytes.Buffer)
		}
		c.flags.SetOutput(c.flagErrorBuf)
		c.PersistentFlags().BoolVarP(&c.helpFlagVal, "help", "h", false, "help for "+c.Name())
	}
	return c.flags
}
```

You cannot access Flags without forcing the creation of the help flag from what I'm able to tell.

akutz:
It doesn't look like you can do that...

```
[0]akutz@poppy:rexray$ rexray
panic: rexray flag redefined: help

goroutine 1 [running]:
github.com/spf13/pflag.(*FlagSet).AddFlag(0xc8200f0580, 0xc82011c240)
	/home/akutz/go/src/github.com/spf13/pflag/flag.go:522 +0x3f6
[...]

PR #130, opened 2015-08-23
Help Flag Shorthand & Usage Customization

akutz:
This patch relates to Issue #129.

This patch allows users to the value used for the help flag's
shorthand character so that when there's a global "host" flag, it can
use the "h" for its shorthand and the help flag can use "?".
Additionally, the ability to format the help flag's usage value would
also be nice for applications with different casing standards for help
output.

Users should simply set cobra.HelpFlagShorthand (defaults to "h") to the
desired shorthand value while using cobra.HelpFlagUsageFormatString to
specify a format string for the flag's usage value. The format string
accept a single substitution, the command's name.

### DIFF
--- a/cobra.go
+++ b/cobra.go
@@ -34,6 +34,15 @@ var EnablePrefixMatching bool = false
 // enables an information splash screen on Windows if the CLI is started from explorer.exe.
 var EnableWindowsMouseTrap bool = true
 
+// HelpFlagShorthand is the character used to for the help flag's shorthand notation. The
+// default value is "h".
+var HelpFlagShorthand string = "h"
+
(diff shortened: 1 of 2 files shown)


eparis:
What do you think of #133? It lets you do it all on your own and seems a bit less 'magic' to me...

akutz:
Well I don't like my patch referred to as 'magic,' and it *does* let you do it on its own, but dammit sir, I like your patch and the cut of your jib! Well played! Yeah, #133 is a better solution.

akutz:
I'm closing this PR in favor of PR #133 which provides an answer to Issue #129 in a cleaner fashion.

PR #132, opened 2015-08-28
turn generated man pages non k8s specific.

AntonioMeireles:
/cc @eparis


AntonioMeireles:
thank you!

eparis:
So sorry that got in there at all!

PR #133, opened 2015-08-30
Initialize the --help flag as late as possible

eparis:
This should let users overwrite it with something custom. Like the help
sub-command.

### DIFF
--- a/command.go
+++ b/command.go
@@ -561,6 +561,9 @@ func (c *Command) Execute() (err error) {
 }
 
 func (c *Command) initHelp() {
+	if c.Flags().Lookup("help") == nil {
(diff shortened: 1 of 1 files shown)


eparis:
This does cause `--help` to typically not show up in auto generated docs, bash completions, etc. Much the same was that the help subcommand doesn't show up in those. Unless the user creates their own...

eparis:
I feel like this is a cleaner way to get what is wanted in #130

akutz:
Well I don't like my patch (#130) referred to as 'magic,' and it *does* let you do it on its own, but dammit sir, I like your patch and the cut of your jib! Well played! Yeah, #133 is a better solution.

PR #135, opened 2015-09-01
Template Function Injection

akutz:
This patch enables developers to add one to many template functions that
can be used by custom Usage and Help templates. Here is an example that
is included in the file cobra_test.go as the test function named
TestAddTemplateFunctions:

    AddTemplateFunc("t", func() bool { return true })
    AddTemplateFuncs(template.FuncMap{
        "f": func() bool { return false },
        "h": func() string { return "Hello," },
        "w": func() string { return "world." }})

    const usage = "Hello, world."

    c := &Command{}
    c.SetUsageTemplate(`{{if t}}{{h}}{{end}}{{if f}}{{h}}{{end}} {{w}}`)

    if us := c.UsageString(); us != usage {
        t.Errorf("c.UsageString() != \"%s\", is \"%s\"", usage, us)
    }

In the above example four functions are added to the template function
map used when the Usage and Help text is generated from the templates
that enable custom logic as well as data injection during template
execution.

### DIFF
--- a/cobra.go
+++ b/cobra.go
@@ -25,6 +25,13 @@ import (
 	"text/template"
 )
 
+var templateFuncs template.FuncMap = template.FuncMap{
+	"trim": strings.TrimSpace,
+	"rpad": rpad,
+	"gt":   Gt,
+	"eq":   Eq,
+}
+
 var initializers []func()
 
 // automatic prefix matching can be a dangerous thing to automatically enable in CLI tools.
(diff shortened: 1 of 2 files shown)


cobra.go
```diff
@@ -39,6 +46,20 @@ var MousetrapHelpText string = `This is a command line tool
 You need to open cmd.exe and run it from there.
 `
 
+//AddTemplateFunc adds a template function that's available to Usage and Help
+//template generation.
+func AddTemplateFunc(name string, tmplFunc interface{}) {
```
eparis: Any time I see `interface{}` I gotta ask, is that the best we can do?

eparis: http://golang.org/pkg/text/template/#FuncMap  say yes  :-(

akutz: I was going to say, it's not *my* choice. It's golang's choice to persist a void pointer instead of implementing generics!

PR #143, opened 2015-09-04
Feature/hidden commands

sdomino:
It would be nice if there was a quick way to designate a command as "for internal use only" (hidden/private) so that consumers would never see the command but cobra could still run it as any other command.

This implementation is very simple and similar in design to the "deprecated" field.

### DIFF
--- a/command.go
+++ b/command.go
@@ -51,6 +51,8 @@ type Command struct {
 	BashCompletionFunction string
 	// Is this command deprecated and should print this string when used?
 	Deprecated string
+	// Is this command hidden and should NOT show up in the list of available commands?
+	Hidden bool
 	// Full set of flags
 	flags *flag.FlagSet
 	// Set of flags childrens of this command will inherit
@@ -256,9 +258,9 @@ Aliases:
 {{end}}{{if .HasExample}}
 
 Examples:
-{{ .Example }}{{end}}{{ if .HasNonHelpSubCommands}}
+{{ .Example }}{{end}}{{ if .HasAvailableSubCommands}}
 
-Available Commands: {{range .Commands}}{{if (not .IsHelpCommand)}}
+Available Commands: {{range .Commands}}{{if .IsAvailableCommand}}
   {{rpad .Name .NamePadding }} {{.Short}}{{end}}{{end}}{{end}}{{ if .HasLocalFlags}}
 
 Flags:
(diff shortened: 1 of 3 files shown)


eparis:
Personally I like `hidden` rather than `private`, but that's bikeshedding, I know.

You need to update `HasNonHelpSubCommands` (maybe rename that func?) So that it won't print `Available Commands` if you only have a single subcommand which is a private command? (Basically everywhere we check `deprecated` you want to do the same thing)

Also add some _test.go so we know that private/hidden commands work, but aren't showing up in any of the doc types?

command.go
```diff
@@ -258,8 +260,8 @@ Aliases:
 Examples:
 {{ .Example }}{{end}}{{ if .HasNonHelpSubCommands}}
 
-Available Commands: {{range .Commands}}{{if (not .IsHelpCommand)}}
-  {{rpad .Name .NamePadding }} {{.Short}}{{end}}{{end}}{{end}}{{ if .HasLocalFlags}}
+Available Commands: {{range .Commands}}{{if (not .IsHelpCommand)}}{{if (not .Hidden)}}
+  {{rpad .Name .NamePadding }} {{.Short}}{{end}}{{end}}{{end}}{{end}}{{ if .HasLocalFlags}}
```
eparis: Do we still need the template changes after the Has/IsHelp changes?

sdomino: As it was, that was still necessary; however I just pushed some updates that has that removed from the template. To do so, I had to update/add some functions to better determine how commands should appear in the default usage/help text.

sdomino:
Note that one of the tests is commented out. When run, that particular test passes, however it has some associated output that shows up as part of the execution, and I wasn't sure if that wanted to be seen when running the test suite.

command_test.go
```diff
@@ -5,6 +5,21 @@ import (
 	"testing"
 )
 
+// test to ensure hidden commands run as intended
+func TestHiddenCommandExecutes(t *testing.T) {
+	cmdHidden.Execute()
```
eparis: check that is is != "hidden" before the Execute() ?

sdomino: yea, that would probably be good to check. I'll add that right away.

command.go
```diff
@@ -850,42 +852,64 @@ func (c *Command) HasSubCommands() bool {
 	return len(c.commands) > 0
 }
 
+// Determine if a command is available as a non-help command (this includes all
+// non deprecated/hidden commands)
+func (c *Command) IsAvailableCommand() bool {
+
+	// a command is 'available' if it is runnable and is not deprecated/hidden
+	return c.Runnable() && (len(c.Deprecated) == 0) && !c.Hidden
```
eparis: unnecessary `()` around all of your `len() !=` tests through here. Otherwise squash and I'll merge.

command.go
```diff
@@ -850,42 +852,64 @@ func (c *Command) HasSubCommands() bool {
 	return len(c.commands) > 0
 }
 
+// Determine if a command is available as a non-help command (this includes all
```
eparis: I said I was done commenting, but maybe I lied? Your comments are great. If you start them with the name of the following function they will get turned into godocs...   So this would start with `// IsAvailableCommand`

eparis: I'll do it if you aren't sick of this.

sdomino: I'm fine doing it, I appreciate the feedback. Normally I do start with the name of the function for godocs, however recently I heard somewhere that was no longer a requirement (but I'm actually fond of it, so I don't mind doing it anyway).

eparis: I'm HORRIBLE at doing it   :)

PR #144, opened 2015-09-04
hidden flags

sdomino:
It would be nice if there was a quick way to designate a flag as "for internal use only" (hidden/private) so that consumers would never see the flag but cobra could make it available to commands.

Initially I thought that simply having a blank usage could be a way of indicating a private command, but decided that this was more of just a 'quick-fix' rather than a true solution
if f.Usage != "" {
  local.AddFlag(f)
}

Then I considered just using the already built-in Flag.Deprecated, however it seemed it was designed for something slightly different, and while it would do the job, it didn't feel like the right solution.
if f.Deprecated == "" {
  local.AddFlag(f)
}

Ultimately I feel that this, coupled with a pull request from spf13/pflag (https://github.com/spf13/pflag/pull/53), is the best approach; adding a Flag.Private field with the corresponding method to set a flag as private is the cleanest and most responsible solution to the problem.

### DIFF
--- a/cobra_test.go
+++ b/cobra_test.go
@@ -18,8 +18,8 @@ var _ = os.Stderr
 
 var tp, te, tt, t1, tr []string
 var rootPersPre, echoPre, echoPersPre, timesPersPre []string
-var flagb1, flagb2, flagb3, flagbr, flagbp bool
-var flags1, flags2a, flags2b, flags3 string
+var flagb1, flagb2, flagb3, flagbr, flagbp, flagbh, flagbph bool
+var flags1, flags2a, flags2b, flags3, outs string
 var flagi1, flagi2, flagi3, flagir int
 var globalFlag1 bool
 var flagEcho, rootcalled bool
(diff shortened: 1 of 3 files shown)


eparis:
I'd love to see some _test.go to really make is clear all of places you expect these flags to work and where you don't expect them to work...

eparis:
I expected them to always work but we'd just never show the user about them when I was thinking about them in pflag, but I think you want them to explicitly not function/not be usable from the command line. In either case, trying to make it clear in _test.go will make sure no one breaks your expectations in the future.

sdomino:
The expectation here is the same as in pflag; The command should function as normal, just never show up in any help/usage text. I'll work on some tests that make that expectation clear.

sdomino:
@eparis; Upon further exploration I believe these changes are no longer necessary to achieve what I needed. With the update to pflag (specifically https://github.com/spf13/pflag/blob/master/flag.go#L409), that seems to accomplish this w/o having to modify cobra at all.

I think this didn't occur to me since I was submitting both changes simultaneously; since the changes in pflag have been merged the functionality does now exist in cobra, and this can probably just be closed, would you agree?

eparis:
I would agree, sounds good.

sdomino:
I suppose the changes aren't completely unnecessary, and would just provide an extra layer of ensuring that hidden flags do not appear in any usage/help text. Also, removing the flags from the LocalFlags() or InheritedFlags() does ensure that if there are only hidden flags, then the 'Flags:' heading doesn't appear (with nothing underneath) in the default usage/help text.

So there may still be merit in merging this in, but I'm good with w/e you decide.

sdomino:
@eparis; I thought pull requests could be re-opened, but I can't seem to find it. Let me know if you'd like me to resubmit a pull request for this (based off my last comment).

PR #146, opened 2015-09-09
Update IsAvailableCommand checks

eparis:
This slightly changes IsAvailableCommand in that a non-runnable command
with a runnable subcommand is now 'Available'

We also use IsAvailableCommand in the rest of the codebase instead of
half kinda sorta doing it incorrectly other places.


eparis:
@sdomino

sdomino:
@eparis; Looks good to me

PR #147, opened 2015-09-10
always return error on Execute and do no print it by default downstream.

AntonioMeireles:
@eparis, @spf13 

i've found existing behaviour on command.Execute regarding error handling a bit incoherent, as it either prints the error by default while also returning it or doesn't even return it, printing it just. 

attached patch attempts to fixe that. fwiw i've left previous behaviour of on error displaying (inline on Execute) usage info not because i think it slso makes _too much_ sense but simply because lefting it (a minor issue) made simpler to get tests keep passing. 

best regards

### DIFF
--- a/cobra_test.go
+++ b/cobra_test.go
@@ -481,8 +481,8 @@ func TestChildCommandFlags(t *testing.T) {
 		t.Errorf("invalid flag should generate error")
 	}
 
-	if !strings.Contains(r.Output, "unknown shorthand") {
-		t.Errorf("Wrong error message displayed, \n %s", r.Output)
+	if !strings.Contains(r.Error.Error(), "unknown shorthand") {
+		t.Errorf("Wrong error message displayed, \n %s", r.Error)
 	}
 
 	if flagi2 != 99 {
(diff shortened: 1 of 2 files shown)


eparis:
I am not loving this. You made no actual change to the return of errors. Although you did make the code a touch more readable/obvious. Where you added a return would actual have fallen through and just returned the error.  /me dislikes functions which use named return values...

So really all you did was remove the printing of errors in Execute(). Personally I think that is the right thin, but programs already have this printing, so they likely aren't doing it themselves. I'm scared to remove it....

AntonioMeireles:
eparis - fair points. trying to explaining myself better - *all* i'd like to have is the ability on 'my' side (i.e. above Execute()) of having  full control on how to handle / display output in case of error. (right now, even if i want to tweak it, i always have to swallow whatever errored as is). actually, one could perhaps preserve exact same behaviour as today but with the error printing that is happening inside a default PostRun func/'template' that one could override at will.  would that be ok to you ?

eparis:
I'm totally ok with some way for you to suppress the error output.

spf13:
Where does this stand @AntonioMeireles ?

spf13:
I believe this PR has been superseded by #169.. Closing. Open with comment if this is incorrect.

PR #148, opened 2015-09-11
trim trailing whitespace

tamird:
This trims some whitespace from the built-in templates.

In https://github.com/cockroachdb/cockroach, we have a test that asserts on the output of our CLI commands. This change makes it easier to write that test case when using a whitespace-trimming editor.

### DIFF
--- a/command.go
+++ b/command.go
@@ -248,8 +248,7 @@ func (c *Command) UsageTemplate() string {
 	if c.HasParent() {
 		return c.parent.UsageTemplate()
 	} else {
-		return `{{ $cmd := . }}
-Usage: {{if .Runnable}}
+		return `{{ $cmd := . }}Usage:{{if .Runnable}}
   {{.UseLine}}{{if .HasFlags}} [flags]{{end}}{{end}}{{if .HasSubCommands}}
   {{ .CommandPath}} [command]{{end}}{{if gt .Aliases 0}}
 
@@ -260,7 +259,7 @@ Aliases:
 Examples:
 {{ .Example }}{{end}}{{ if .HasAvailableSubCommands}}
 
-Available Commands: {{range .Commands}}{{if .IsAvailableCommand}}
+Available Commands:{{range .Commands}}{{if .IsAvailableCommand}}
   {{rpad .Name .NamePadding }} {{.Short}}{{end}}{{end}}{{end}}{{ if .HasLocalFlags}}
 
 Flags:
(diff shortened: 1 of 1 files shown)


command.go
```diff
@@ -286,8 +285,7 @@ func (c *Command) HelpTemplate() string {
 		return c.parent.HelpTemplate()
 	} else {
 		return `{{with or .Long .Short }}{{. | trim}}{{end}}
-{{if or .Runnable .HasSubCommands}}{{.UsageString}}{{end}}
-`
```
eparis: This one seems required, you need the trailing \n...

tamird: You get one anyway, though I'm not totally sure why.

tamird: My output looks like this:

```
$ ./cockroach help

Usage:
  cockroach [command]

Available Commands:
  init        init new Cockroach cluster
  start       start a node by joining the gossip network
  cert        create ca, node, and client certs
  exterminate destroy all data held by the node
  quit        drain and shutdown node

  log         make log files human-readable

  sql         open a sql shell
  kv          get, put, conditional put, increment, delete, scan, and reverse scan key/value pairs
  user        get, set, list and remove users
  range       list, split and merge ranges
  zone        get, set, list and remove zones

  version     output version information

Flags:
      --alsologtostderr=false: log to standard error as well as files
      --color="auto": colorize standard error output according to severity
      --log-backtrace-at=:0: when logging hits line file:N, emit a stack trace
      --log-dir="": if non-empty, write log files in this directory
      --logtostderr=true: log to standard error instead of files
      --verbosity=0: log level for V logs
      --vmodule=: comma-separated list of file=N settings for file-filtered logging


Use "cockroach [command] --help" for more information about a command.
```
Notice that there's also a leading newline, which I would also like to get rid of, but not sure how.

tamird: There are also two newlines between the usage template and the help template (one is the spurious leading newline, the other is the spurious trailing newline). Would be nice to get this down to one as well (by removing the spurious leading newline).

eparis: The "leading" line is probably because you don't have a Short or a Long.

eparis: (I have a fix for that, looking at the extra line after Flags:)

eparis: The extra line after the "Flags:" section I don't think I can reasonably fix until https://github.com/golang/go/issues/9969  which is golang 1.6   :-(

chriseaton:
Any reason why this couldn't or should also be implemented in html/template?

PR #149, opened 2015-09-11
Allows command suggestions along with "unknown command" errors

fabianofranz:
Adds *optional* support to suggestions when "unknown command" errors happen. This allows Cobra to mimic the behaviour of `git`, like in:

```
$ git fethc
git: 'fethc' is not a git command. See 'git --help'.

Did you mean this?
	fetch
```

### DIFF
--- a/command.go
+++ b/command.go
@@ -106,6 +106,11 @@ type Command struct {
 	helpCommand   *Command                 // The help command
 	// The global normalization function that we can use on every pFlag set and children commands
 	globNormFunc func(f *flag.FlagSet, name string) flag.NormalizedName
+
+	// Disable the suggestions based on Levenshtein distance that go along with 'unknown command' messages
+	DisableSuggestions bool
+	// If displaying suggestions, allows to set the minimum levenshtein distance to display, must be > 0
+	SuggestionsMinimumDistance int
 }
 
 // os.Args[1:] by default, if desired, can be overridden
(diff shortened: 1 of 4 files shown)


fabianofranz:
@eparis some friday-evening awesomeness.  :)

eparis:
I sort of feel like "on by default"....  (really cool)

fabianofranz:
@eparis Yeah, I kept this off by default so we don't break the current behaviour. But it's going to be on in OpenShift.  :)

eparis:
@spf13 I see no reason not to just turn it on and do away with the option altogether...

spf13:
I can't think of a reason to not want this either. I think it's just a always on feature. Fewer knobs, more goodness all around.

apriendeau:
I agree. Its a good default way of handling the misspelled. A little more class then just failing or showing help.

fabianofranz:
I'm fine with making this on by default but I'd suggest we keep the knobs
to disable if needed. I can think of some scenarios, like potentially
dangerous administrative commands, with little (or huge) number of
subcommands, where users are expected to know exactly what they are doing
and giving suggestions that may not be completely accurate (since they are
automatic) could be dangerous.

If you are fine with this proposal I'll adjust the PR.
Em 12/09/2015 3:46 PM, "Austin Riendeau" <notifications@github.com>
escreveu:

fabianofranz:
Actually I went ahead and adjusted it already. :) So now we are *on* by default, and have the know to disable it.

If you are good with it this should be good to go.

apriendeau:
I think the only suggestion/comments that I have, is that it should skip deprecated and hidden commands.

fabianofranz:
@apriendeau good catch, fixed.

fabianofranz:
@apriendeau definitely, I didn't notice that method. Fixed to use `IsAvailableCommand()`.

apriendeau:
after I deleted my comment because I didnt want to knit pick. Nice job. This is super useful!

fabianofranz:
:+1:

command.go
```diff
@@ -419,9 +424,31 @@ func (c *Command) Find(args []string) (*Command, []string, error) {
...
+				}
+			}
+			if len(similar) > 0 {
+				suggestions += "\n\nDid you mean this?\n"
+				for _, s := range similar {
+					suggestions += fmt.Sprintf("\t%v\n", s)
+				}
+			}
+		}
+		return commandFound, a, fmt.Errorf("unknown command %q for %q%s", argsWOflags[0], commandFound.CommandPath(), suggestions)
```
eparis: So if I disable suggestions, and type a wrong command, do I have a trailing `\n`?

eparis: We are being so nit picky for something so awesome. Sorry, that's how I get when I show things to users   :)

apriendeau: Its how we show love :heart:

fabianofranz: With suggestions disabled it behaves exactly like before, there's no new `\n`. The [test for unknown command](https://github.com/fabianofranz/cobra/blob/b4087da7eb1bf3822d8413008dbd4c3c6c3c5f2d/cobra_test.go#L788) didn't change.

eparis: If you say so, I'll believe you   :)

fabianofranz: Never believe in me (but you can trust in tests).  :)

fabianofranz:
:+1:

PR #151, opened 2015-09-11
No leading newline if Short and Long both unset

eparis:
Before --help would start
```

Usage:
  command [flags]
  [...]
```

After --help will show
```
Usage:
  command [flags]
  [...]
```


eparis:
@tamird

tamird:
LGTM, thanks.

PR #152, opened 2015-09-11
No newline after Flags in usage

eparis:
The flags usage template from pflags has a trailing \n. We need to
include a newline in case there are no flags in our template. This will
trim the newline from the end of the flags from pflag and we can do it
right outselves.


eparis:
@tamird

PR #153, opened 2015-09-11
UsageTemple: remove useless variable assignment

eparis:


eparis:
@tamird  this is your punishment for touching the template. now you get to see me touching it.

tamird:
Ha. Could be worse.

PR #154, opened 2015-09-12
If nothing after flags: sting print a newline

eparis:


shurcooL:
What does "sting" mean in this context?

eparis:
s/sting/string/

PR #155, opened 2015-09-14
adds check to see if command name and os.Args[0] is the same before...

apriendeau:
applying os.Args as it was leading appending false arguments from the test suite. This was causing the test suite to fail. All tests should be passing now.

### DIFF
--- a/command.go
+++ b/command.go
@@ -566,7 +566,7 @@ func (c *Command) Execute() (err error) {
 
 	var args []string
 
-	if len(c.args) == 0 {
+	if len(c.args) == 0 && os.Args[0] == c.Name() {
 		args = os.Args[1:]
 	} else {
 		args = c.args


eparis:
what test was failing?

apriendeau:
The RunRoot with no args and the hidden command tests because it was trying
to use the go test args. You should see it on master currently.

apriendeau:
--- FAIL: TestRunnableRootCommandNilInput (0.00s)
        cobra_test.go:656: Root Function was not called

Usage:
  hide [secret string to print] [flags]

Flags:
      --flagOnCommandLine="": about my flag

Error: unknown shorthand flag: 't' in -test.v=true
--- FAIL: TestHiddenCommandExecutes (0.00s)
        command_test.go:21: Hidden command failed to run!

eparis:
so you ran `go test -v ./...` and the tests failed.  But `go test ./...` works fine. Which is why I'm not seeing a problem and master seems happy?  You are saying this patch will let me use `go test -v ./...` ?

apriendeau:
Yes because -v is just the verbose output of what you are running. The fact
that the suite was passing baffles me.

eparis:
The suite passes because it is correct and passing. Adding `-v` is actually what broke it!

What's happening is that `go test` is (behinds the scene) importing "testing" into our code.  "testing" will then create new (go flags) called "--test.*" in our program. In the "testing" package `init()` (or some similar thing) it will change the `-v` you added to be `-test.v` in the command line arguments to our program. "Testing" will then call `flag.Parse()` so the "testing" package will will have access to its results. Ok all great.

BUT cobra doesn't use the golang "flag" package, it uses "pflag" which does its own command line argument parsing. The pflag command line has no idea what `-test.v` is. In fact, since the `go test` command uses a single `-` instead of a `--` pflag believes the flag name is just `t` with an argument of `est.v`.

Whew. Ok, maybe that makes sense why `go test` works and travis was correctly passing, but `go test -v` fails. Given all that, I'll try to understand how your patch solves the problem that `go test -v` does not work.

Thanks for sticking with me!

apriendeau:
Course. The patch just checks if the command is actually the cobra command
before applying the os.Args. So it's just another layer of validation.
That's all.

apriendeau:
@eparis is there anything I need to do with this PR?

anthonyfok:
Hi @apriendeau, @spf13 and @eparis,

Now that `go test -v ./...` is used as per commit 5c40aa85cac15d3e5fe0d9f47cb6d688d2ea62fa for #157, Travis CI test is failing, as expected.

Should we revisit this Pull Request #155?  :-)

apriendeau:
Sure. I can put a PR up for it today again.

anthonyfok:
Cool, thanks!  :-)

PR #159, opened 2015-09-15
Update travis to use more versions of go

eparis:


apriendeau:
Sorry. I guess I could have done that rather than create an issue.

Issue #78, opened 2015-03-26
Suggestion: "did you mean" for misspelled commands


synful:
In git, if you execute `git commt`, you'll get the following message:

```
git: 'commt' is not a git command. See 'git --help'.

Did you mean this?
	commit
```

It would be a very useful (and simple to implement) feature for Cobra to have. Thoughts?

pravj:
Hi All,

This is a nice feature to have.

Although, I'm not sure it's right to have it here or not, because *git* has all the *commands* all the time, but in our case users may or may not have that much commands.

If we're going ahead, I'll propose some *similarity index* that returns N most similar commands to what users have entered. *Git* seems to be the doing same https://github.com/git/git/blob/master/help.c

synful:
I think that's a good way to go, although I think it'd be best to add a utility feature which would automatically pick the most similar command(s) which matched some similarity threshold and print them. Presumably we'd want this feature to be off by default.

eparis:
#149

fabianofranz:
This should be fixed by https://github.com/spf13/cobra/pull/149 and https://github.com/spf13/cobra/pull/163.

marians:
Just recognized this got implemented. Thanks!

PR #163, opened 2015-09-25
Improve suggestions

fabianofranz:
Multiple improvements to suggestions when "unknown command" errors happen. Basically a `HasPrefix` will now also match, and adds the `SuggestFor` attribute to allow explicit suggestions. By setting that attribute you can for example suggest the valid command `remove` when users try `delete`, if for some reason you don't want actual aliases.

### DIFF
--- a/README.md
+++ b/README.md
@@ -442,6 +442,18 @@ or
 
     command.SuggestionsMinimumDistance = 1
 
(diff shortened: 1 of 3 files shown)


fabianofranz:
@eparis Goodness and this should also fix https://github.com/spf13/cobra/issues/78.

fabianofranz:
CC @spf13

fabianofranz:
UP

eparis:
I looked through it, and it looked ok, but i'd like to look at it again. I'm traveling monday, but if I didn't commit, can you bother me again tuesday?

PR #170, opened 2015-10-27
adds alternative error handling to readme

apriendeau:
I just added a section to README covering the new error handling. I agree with @bep here, what is with CircleCI failing?


spf13:
Merged as bc980b8

PR #174, opened 2015-10-31
Ensure the usage command is printed only once

### LINKED ISSUE #171: Usage is printed twice
Using the full example in readme as-is would result in printing usage information twice.

executing ```./myapp```  which only contains the code from the example, without any arguments or subcommands, would result in the following output:
```
Usage:
  app [command]

Available Commands:
  print       Print anything to the screen
  echo        Echo anything to the screen

Flags:
  -h, --help[=false]: help for app

Use "app [command] --help" for more information about a command.

Usage:
  app [command]

Available Commands:
  print       Print anything to the screen
  echo        Echo anything to the screen

Flags:
  -h, --help[=false]: help for app

Use "app [command] --help" for more information about a command.
```

Note how the usage is printed twice here.

PS: Many thanks for this great library.

jbowes:
Reverse a swap in logic introduced in #169 that would cause the usage
output to be printed twice.

Fixes #171

### DIFF
--- a/cobra_test.go
+++ b/cobra_test.go
@@ -1075,3 +1075,14 @@ func TestAddTemplateFunctions(t *testing.T) {
 		t.Errorf("c.UsageString() != \"%s\", is \"%s\"", usage, us)
 	}
 }
+
+func TestUsageIsNotPrintedTwice(t *testing.T) {
+	var cmd = &Command{Use: "root"}
+	var sub = &Command{Use: "sub"}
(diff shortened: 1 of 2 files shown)


apriendeau:
I can't believe I overlooked that `err == flag.ErrHelp`. Nice catch. though that would change output which I don't know if its okay. @spf13 and @eparis  thoughts?

gerred:
:+1: Can't wait for this to be in.

PR #178, opened 2015-11-03
Modifying bash completions so they can run on OSX

mattnenterprise:
OSX has an older version of bash. So it doesn't have compopt.

### DIFF
--- a/bash_completions.go
+++ b/bash_completions.go
@@ -60,7 +60,9 @@ __handle_reply()
     __debug "${FUNCNAME}"
     case $cur in
         -*)
-            compopt -o nospace
+            if builtin compopt > /dev/null 2>&1; then
+              compopt -o nospace
+            fi
             local allflags
             if [ ${#must_have_one_flag[@]} -ne 0 ]; then
                 allflags=("${must_have_one_flag[@]}")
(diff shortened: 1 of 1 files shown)


bash_completions.go
```diff
@@ -60,7 +60,9 @@ __handle_reply()
     __debug "${FUNCNAME}"
     case $cur in
         -*)
-            compopt -o nospace
+						if builtin compopt > /dev/null 2>&1; then
```
eparis: Whitespace damage needs fixed.

bash_completions.go
```diff
@@ -216,7 +220,13 @@ func postscript(out *bytes.Buffer, name string) {
 }
 
 `, name)
-	fmt.Fprintf(out, "complete -F __start_%s %s\n", name, name)
+	fmt.Fprintf(out, `
+		if builtin compopt > /dev/null 2>&1; then
+			complete -F __start_%s %s
+		else
+			complete -o nospace -F __start_%s %s
```
eparis: This means that on OS X things will just never have a space after the tab'd item...   is that preferable?

eparis: If you think that's best, I'm ok with it. Squash and I'll merge.

mattnenterprise: I have squashed it.

PR #179, opened 2015-11-03
Check for _init_completion not_init_completionS

eparis:


eparis:
@mattnenterprise and you can review this too, since I added the bug for OS X support. (I don't have OS X)

mattnenterprise:
LGTM

PR #180, opened 2015-11-04
Persistent flags completions

fabianofranz:
Inherited flags should also be part of completions for every command that inherits it. Also allows marking persistent flags as required or filename for completions.

### DIFF
--- a/bash_completions.go
+++ b/bash_completions.go
@@ -396,6 +396,11 @@ func (cmd *Command) MarkFlagRequired(name string) error {
 	return MarkFlagRequired(cmd.Flags(), name)
 }
 
+// MarkPersistentFlagRequired adds the BashCompOneRequiredFlag annotation to the named persistent flag, if it exists.
+func (cmd *Command) MarkPersistentFlagRequired(name string) error {
+	return MarkFlagRequired(cmd.PersistentFlags(), name)
+}
+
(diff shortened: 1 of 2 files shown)


fabianofranz:
@eparis PTAL

eparis:
gimmie some _test.go.  But it seems ok/good/right

fabianofranz:
@eparis added some.

fabianofranz:
@eparis circleci failure seems unrelated.

spf13:
merged as 7290a94.

Yes, circle is unrelated and I've already contacted them about it.

anthonyfok:
The circleci failure seems to be the same as Travis CI build failure:

```
=== RUN TestRunnableRootCommandNilInput-2
--- FAIL: TestRunnableRootCommandNilInput-2 (0.00s)
	cobra_test.go:667: Execute() failed with unknown shorthand flag: 't' in -test.v=true
	cobra_test.go:671: Root Function was not called
```

It is triggered by the use of `-v` flag for `go test`.  Please see #155 for a potential solution.

anthonyfok:
Forgot to say: @fabianofranz, this is awesome!  :-)  Now bash completion for Hugo works quite a bit better indeed, e.g. the `--watch` in `hugo server --watch` can now be tab-completed.  :-)
Many thanks!

fabianofranz:
:+1:

PR #182, opened 2015-11-07
fixes verbose tests by checking if the first arg belongs to cobra

apriendeau:
This should fix the tests by checking if first arg is the same as the name as the root command in cobra. Replaces PR #155 @eparis @spf13 @anthonyfok

### DIFF
--- a/command.go
+++ b/command.go
@@ -618,7 +618,7 @@ func (c *Command) Execute() (err error) {
 
 	var args []string
 
-	if len(c.args) == 0 {
+	if len(c.args) == 0 && os.Args[0] == c.Name() {
 		args = os.Args[1:]
 	} else {
 		args = c.args


spf13:
pushed as 62e859a

anthonyfok:
Hi @spf13,

Sorry for my earlier premature recommendation for the fix.  It turns out that this patch has undesirable side-effects:  Specifying full path to the executable, or renaming the executable, would make it ignore any command-line arguments.  Using `hugo` as example:

        $ ~/go/bin/hugo --help
        
        $ mv hugo hhuuggoo
        $ hhuuggoo server

All subcommands and flags are ignored...

I discovered this problem when trying to `sudo ~/go/bin/hugo genautocomplete`.  At first, I thought it was the `sudo`, but it turns out `sudo hugo` works well; the problem came from the full path to the hugo executable.

@apriendeau, any ideas of an alternative solution?  :-)

PR #185, opened 2015-11-08
adds a flag to disable autogen tag at the bottom of man and markdown …

### LINKED ISSUE #142: Remove timestamp in generated documentation
Hello,

First of all, thanks for the time you put in this library, I find it really useful.

I'd like to setup a commit hook for my team, to automatically update the documentation (using both markdown and man). 
This works, however there is a timestamp in the generated docs which prevents me to roll it out (namely, every commit now has a diff in the docs).
`###### Auto generated by spf13/cobra at 2015-09-03 23:32:45.933601388 +0000 UTC`

Would it be okay for you guys to remove it? 
I was going to suggest a configuration option but it would seem overly complicated here :)

Best

apriendeau:
…generators. Inherits from parent commands all the way to the root command. Fixes #142


spf13:
merged as 5144a3aa19b64be9931d984ef359ccb8f7c39f60

PR #187, opened 2015-11-08
changes to contains rather than equality

apriendeau:
@anthonyfok Something like this instead of an equality check should fix the issue.

### DIFF
--- a/command.go
+++ b/command.go
@@ -618,7 +618,7 @@ func (c *Command) Execute() (err error) {
 
 	var args []string
 
-	if len(c.args) == 0 && os.Args[0] == c.Name() {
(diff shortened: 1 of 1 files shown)


anthonyfok:
Hello @apriendeau,

Thank you for your improve fix, though a strings.Contains() would fail if the user were to rename e.g. `hugo` to `Hugo` or even `hhuuggoo`.

I finally decided that perhaps making the exception only when the base name of `os.Args[0]` is `cobra.test` would minimize the chance of the end user running into problem, so I went ahead and committed that change as e5762054c9a6429fa7c234897707d53740d6ad5d.

Nevertheless, the credit goes to you for noticing the `go test -v` FAIL and providing a working initial solution.  Thank you!  :-)

Cheers,
Anthony

PR #192, opened 2015-11-17
Creates the ExecuteC function while maintaining Execute signature

### LINKED ISSUE #191: Command.Execute should return the Command
See #168

The command is always run from the root and the caller will never know which command that is actually executed. So, even if the errors are propagated up the chain, it is hard to do proper centralized error handling without the command as context (usage logging etc).

An API change is maybe a hard sell, but what about a `Command.ExecuteC` or something?

apriendeau:
This should address and fix #191

### DIFF
--- a/cobra_test.go
+++ b/cobra_test.go
@@ -248,6 +248,18 @@ func simpleTester(c *Command, input string) resulter {
 	return resulter{err, output, c}
 }
 
+func simpleTesterC(c *Command, input string) resulter {
+	buf := new(bytes.Buffer)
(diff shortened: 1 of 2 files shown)


eparis:
our API is getting uglier and uglier and uglier.  But I don't see a lot of a better way...

apriendeau:
Hence my comment on the issue. It gets uglier and uglier till we decide... 2.0! ;)

apriendeau:
We could start coming up with a plan to deprecate functions.

PR #196, opened 2015-11-23
add GNU GPL3 as a license

goern:
this will give the user GNU GPL3 as a select-able license on `cobra add`


anthonyfok:
Good stuff, @goern!  Thank you!

Merged as b167d9b.

(Some minor details may need to be fixed, but I will look at those and tweak it later today.)

Issue #195, opened 2015-11-22
Factor out external dependencies


magiconair:
I was trying to use cobra in one of my projects and I've noticed that cobra depends on a couple of libraries don't really belong to the core. The reason I'm bringing this up is because I'm managing my deps manually via vendoring into a sub dir and separate commits per lib. It feels a bit odd to copy dependencies which I don't need but cobra won't compile without them.

The libraries in question are:

`github.com/inconshreveable/mousetrap` 
`github.com/cpuguy83/go-md2man/md2man`

The first one is only needed on Windows and the second one only for the man page generation. The man and markdown generation also seems needlessly tied to the `cobra.Command` object. 

Both of them can be factored out either with conditional compilation (for mousetrap) or move the code into a sub directory (man and md generation).

I have PRs ready for both refactorings but would like to know whether this is in line with the project.

magiconair:
I forgot to mention that 

`github.com/cpuguy83/go-md2man/md2man`

itself has two additional dependencies

```
github.com/shurcooL/sanitized_anchor_name
github.com/russross/blackfriday
```

bep:
This makes totally sense to me. @eparis and @spf13 might have something to say about it.

apriendeau:
Well I know that even hugo uses `github.com/russross/blackfriday`. I could see removing `mousetrap` but blackfriday is pretty amazing all around. So would this remove the all the dependencies or just the two that you mentioned?

spf13:
I'm a bit confused about why people have issues with "unneeded dependencies". If you are using a tool to manage your dependencies it doesn't matter if there are 2 or 20. The binary itself isn't slower or heavier because of unused dependencies. 

While we could optimize for fewer included dependencies we would do so at a high cost of integrating cobra significantly harder. 

Please help me understand why this is an issue.

bep:
I'm totally in line with the thread starter here.

>If you are using a tool to manage your dependencies it doesn't matter if there are 2 or 20. 

There no perfect tool for that. They may handle the versioning etc., but if you drag a dependency into your project you would still have to do a QA about its license and quality etc. Normally not an issue in a personal toy project, but certainly a issue in commercial software.

>The binary itself isn't slower or heavier because of unused dependencies. 

It will be bigger?

>... a high cost of integrating cobra significantly harder. 

[...]

PR #197, opened 2015-11-23
Issue #195: Compile mousetrap only on Windows

magiconair:
* Create command_win.go and command_notwin.go for windows only code
* Move call to mousetrap hook into separate preExecHook() function

### DIFF
--- a/cobra.go
+++ b/cobra.go
@@ -40,14 +40,6 @@ var initializers []func()
 // Set this to true to enable it
 var EnablePrefixMatching bool = false
 
-// enables an information splash screen on Windows if the CLI is started from explorer.exe.
-var EnableWindowsMouseTrap bool = true
-
-var MousetrapHelpText string = `This is a command line tool
-
-You need to open cmd.exe and run it from there.
-`
-
 //AddTemplateFunc adds a template function that's available to Usage and Help
 //template generation.
 func AddTemplateFunc(name string, tmplFunc interface{}) {
--- a/command.go
+++ b/command.go
@@ -21,11 +21,8 @@ import (
 	"io"
 	"os"
 	"path/filepath"
-	"runtime"
 	"strings"
-	"time"
 
-	"github.com/inconshreveable/mousetrap"
 	flag "github.com/spf13/pflag"
 )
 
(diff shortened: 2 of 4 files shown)


cobra.go
```diff
@@ -40,14 +40,6 @@ var initializers []func()
 // Set this to true to enable it
 var EnablePrefixMatching bool = false
 
-// enables an information splash screen on Windows if the CLI is started from explorer.exe.
-var EnableWindowsMouseTrap bool = true
```
eparis: I have no idea if anyone was using it, but this means you can't turn off the mousetrap/exit, right?

command.go
```diff
@@ -626,13 +623,8 @@ func (c *Command) ExecuteC() (cmd *Command, err error) {
...
 
-	if EnableWindowsMouseTrap && runtime.GOOS == "windows" {
-		if mousetrap.StartedByExplorer() {
-			c.Print(MousetrapHelpText)
-			time.Sleep(5 * time.Second)
-			os.Exit(1)
-		}
-	}
+	// windows hook
+	preExecHook(c)
```
eparis: It's bikeshedding, but we already have user exposed things called PreExec stuff. I'm trying to think of a better name, but I don't love this one.

cobra.go
magiconair: You can replace the code with an empty function. Now that I think about it it should be more like this (sans better name. Don't like it either):

```
var preExecHook = preExecHookWin

func preExecHookWin() {...}
```

This way a user can override the function with a custom one - or replace it with nil to disable it.

command.go
magiconair: yeah, don't like it either. Feel free to be more creative :)

magiconair:
@eparis pushed a version which has uses the function var approach. Did a force push on the previous version (used to gerrit). Not sure whether I messed something up.

cobra.go
eparis: 1. preExecHook is a private function, so I can't overwrite it from outside
2. This is an API change vs the variable. So while I think this is a better API, I'd rather not redo it unless we need to.

spf13:
@eparis . I think this latest push looks good and the approach is sound. Do you want to review also since you provided feedback earlier?

eparis:
I like the approach, but people can't turn it off. (Which is fine if @spf13 is fine with it)

I'm not allowed to call   `cobra.preExecHook = myFunc` but I was allowed to do `cobra.EnableWindowsMouseTrap = false`

magiconair:
The reason I made the preExecHook private was that I felt that this isn't something people should hook into. So it is more an implementation detail than a public entry point. Hence also the somewhat generic name. I agree that disabling the functionality on Windows is no longer possible but that I could fix by reintroducing the flag.

spf13:
I can't think of why someone would turn it off. I suppose we can ship
without the flag and add it if someone asks for it? Less configuration is a
good thing.

PR #198, opened 2015-11-23
Issue #195: Move doc generation into separate pkg

magiconair:
* Move man_docs and md_docs into new doc pkg
* Replace *bytes.Buffer with io.Writer
* Replace c == cmd.helpCommand with c.IsHelpCommand()
* Remove redundant len(children) == 0 check in HasSeeAlso
* Duplicate test setup for doc generation
* Move byName sorter into doc_utils since it is used by both man and md


spf13:
merged as b3f29e9

PR #199, opened 2015-11-24
Fixed import path issue on Windows

jbydeley:
I sort of feel crazy here but trying out Cobra, I'm finding that it isn't working on Windows. It's fairly easy to reproduce. 

Test Case 1 (This one works)

```
C:\go\src\cobra.exe init test
C:\go\src\test\cat .\main.go
```

Which shows an import path of `import test/cmd`

Test Case 2 (This one is broken and fixed with this PR)

```
C:\go\src\directory\cobra.exe init test
C:\go\src\directory\cat .\main.go
```

Which shows an import path of `import directory\test/cmd`

Along with running into this issue, I ran the tests inside cobra/cmd and found that they weren't working. 

It's entirely possible I'm crazy and this is working for everyone else. I'm just not sure I see how it could be though based on the windows slashes being different from import slashes.

### DIFF
--- a/cobra/cmd/helpers.go
+++ b/cobra/cmd/helpers.go
@@ -104,7 +104,7 @@ func guessImportPath() string {
 		er("Cobra only supports project within $GOPATH")
 	}
 
-	return filepath.Clean(strings.TrimPrefix(projectPath, getSrcPath()))
+	return filepath.ToSlash(filepath.Clean(strings.TrimPrefix(projectPath, getSrcPath())))
 }
 
 func getSrcPath() string {
--- a/cobra/cmd/helpers.go
+++ b/cobra/cmd/helpers.go
@@ -143,7 +143,7 @@ func guessProjectPath() {
 	srcPath := getSrcPath()
 	// if provided, inspect for logical locations
 	if strings.ContainsRune(inputPath, os.PathSeparator) {
-		if filepath.IsAbs(inputPath) {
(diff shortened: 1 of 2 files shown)


cobra/cmd/helpers.go
```diff
@@ -104,7 +104,7 @@ func guessImportPath() string {
 		er("Cobra only supports project within $GOPATH")
 	}
 
-	return filepath.Clean(strings.TrimPrefix(projectPath, getSrcPath()))
+	return filepath.ToSlash(filepath.Clean(strings.TrimPrefix(projectPath, getSrcPath())))
```
jbydeley: Fix the import to use forward slashes on Windows.

cobra/cmd/helpers.go
```diff
@@ -143,7 +143,7 @@ func guessProjectPath() {
 	srcPath := getSrcPath()
 	// if provided, inspect for logical locations
 	if strings.ContainsRune(inputPath, os.PathSeparator) {
-		if filepath.IsAbs(inputPath) {
+		if filepath.IsAbs(inputPath) || filepath.HasPrefix(inputPath, string(os.PathSeparator)) {
```
jbydeley: `filepath.IsAbs` doesn't play nice on Windows. 

Linux: `filepath.IsAbs("/foo/bar") == true`
Windows: `filepath.IsAbs("/foo/bar") == false`

spf13:
Merged as ee6224d

PR #203, opened 2015-12-03
Always show help if requested, even if SilenceErrors is enabled

aperum:
Enabling SilenceErrors also stops --help from working. One could add SilenceHelp to the command struct if this should be configurable but i think the help system should always work.

I'm aware that "help <command>" still works but users will almost certainly try -h, --help first.

### DIFF
--- a/command.go
+++ b/command.go
@@ -661,13 +661,16 @@ func (c *Command) ExecuteC() (cmd *Command, err error) {
 	}
 	err = cmd.execute(flags)
 	if err != nil {
+		// Always show help if requested, even if SilenceErrors is in
+		// effect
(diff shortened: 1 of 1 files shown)


eparis:
@bep @apriendeau I feel like this is the behavior some wanted. They wanted complete control.  But I do tend to agree, --help should just give you the --help.

apriendeau:
I agree. That was an unwanted side effect which says we need some tests. :+1: That would be my only suggestion to this PR. Test both scenarios.

bep:
Help should give help, so this is a welcomed fix.

spf13:
merged as 2bd8a73

PR #205, opened 2015-12-05
Add flaghash to bash_completion to keep flag value

nak3:


nak3:
For example in Kubernetes, we will get wrong completion below

~~~
$ kubectl --kubeconfig=/point/to/another_server describe pod <TAB>
~~~

I hope that it keeps the flag values to make them available.

bash_completions.go
```diff
@@ -134,6 +134,9 @@ __handle_flag()
         must_have_one_flag=()
     fi
 
+    # keep flag value with flagname as flaghash
+    flaghash[${flagname}]=${words[ $((c+1)) ]}
```
nak3: Sorry, I noticed that it is not good. `--flagname foo` will work, but `--flagname=foo` will not work.

spf13:
merged as 60b8213

anthonyfok:
For reference:

It turns out that `declare -A` is a new feature in Bash 4.x, which, because of its superior GPL v3 license, OS X will likely never ship, hence Issue #240 and the workaround in 801364a.

Cc: @nak3

PR #210, opened 2015-12-10
Bash completion for names with ':' character.

boz:
This makes completions work with commands that have a colon in the name.  The issue came up when using [cmdns](https://github.com/gosuri/cmdns), but probably isn't limited to cases where this is used.

### DIFF
--- a/bash_completions.go
+++ b/bash_completions.go
@@ -103,6 +103,8 @@ __handle_reply()
     if [[ ${#COMPREPLY[@]} -eq 0 ]]; then
         declare -F __custom_func >/dev/null && __custom_func
     fi
+
+    __ltrim_colon_completions "$cur"
 }
 
 # The arguments should be in the form "ext1|ext2|extn"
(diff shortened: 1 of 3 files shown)


nak3:
`COMP_WORDBREAKS=${COMP_WORDBREAKS//:}` doesn't help this issue? 

ref: http://tiswww.case.edu/php/chet/bash/FAQ - E13

boz:
Modifying `COMP_WORDBREAKS` seems to be frowned upon as it affects all other completion functions.  `__ltrim_colon_completions` was created for a lighter-touch solution.

bash_completions.go
```diff
@@ -103,6 +103,8 @@ __handle_reply()
     if [[ ${#COMPREPLY[@]} -eq 0 ]]; then
         declare -F __custom_func >/dev/null && __custom_func
     fi
+
+    __ltrim_colon_completions "$cur"
```
eparis: Anyone have OS X and able to prove that __ltrim_colon_completions exists?

boz: I'm working on OSX.  `__ltrim_colon_completions` was added to bash-completion in [2009](http://anonscm.debian.org/cgit/bash-completion/bash-completion.git/commit/bash_completion?id=42302658705e3545471f2f4e53c2f758355a2bcc)

PR #215, opened 2016-01-03
Replace *bytes.Buffer with io.Writer

mvdan:
Also adds support for generating bash completions on writers other than
just buffers.

Found via github.com/mvdan/interfacer.

### DIFF
--- a/bash_completions.go
+++ b/bash_completions.go
@@ -3,6 +3,7 @@ package cobra
 import (
 	"bytes"
 	"fmt"
+	"io"
 	"os"
 	"sort"
 	"strings"
@@ -16,7 +17,7 @@ const (
 	BashCompSubdirsInDir    = "cobra_annotation_bash_completion_subdirs_in_dir"
 )
 
-func preamble(out *bytes.Buffer) {
+func preamble(out io.Writer) {
 	fmt.Fprintf(out, `#!/bin/bash
 
 __debug()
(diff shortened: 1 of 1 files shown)


mvdan:
I just realised that all the buffer functions were ignoring errors since buffers will never error on writes. I need to fix this.

mvdan:
Done. The patch is larger now, and the writing code is a bit uglier, but I don't think of a better way to treat write errors.

eparis:
looks like the patch needs to be rebased...

eparis:
should we do something similar in the markdown docs generation?

mvdan:
@eparis yeah, but the rebase should be trivial. Will do.

Regarding markdown doc generation, I didn't notice since it wasn't using `*bytes.Buffer`. Limiting to a buffer and not checking errors is fine, as is using a writer and checking errors. But mixing the two really doesn't work.

I can also fix that code if you confirm that adding error checking verbosity everywhere is fine. If you have suggestions on how to avoid `err != nil` on every write, I'm all ears. Otherwise, I'll update this PR.

eparis:
That's just golang, we're going to have to do the `if err != nil`...   You can save a line using 
```
if _, err := fmt.Fprintf(w, "__start_%s()\n", name); err != nil {
```
But basically, I think you did the best you can....

mvdan:
Indeed, I can save a line. But if you look at the diff, I avoided doing that in cases where the write is multi-line. Mainly to avoid awkwardness such as:

```go
if _, err := fmt.Fprintf(w, `some
multiline
string`); err != nil {
    // ...
}
```

So should I include the markdown generation error stuff in this PR as well, or should I do that in a separate PR to keep this one smaller and easier to merge?

eparis:
You tell me what makes you happy. I'll go with that   :)

mvdan:
Cool! I've just rebased this PR and will do the other in a separate one then. This one is big enough as it is.

PR #217, opened 2016-01-06
README: add etcd

gyuho:
We have been migrating our CLI to cobra starting with:
- https://github.com/coreos/etcd/tree/master/etcdctlv3
- https://github.com/coreos/etcd/tree/master/tools

Thanks for the great project!


spf13:
Merged as 2a426b5

gyuho:
Thanks!

Issue #219, opened 2016-01-08
Pre Run and Post Run chain functions 


algrebe:
W.r.t https://github.com/spf13/cobra/issues/216

Persitent\* functions execute only if the child has not overridden the parent.

Consider the following scenario:
Root command has a PersistentPreRun in which it defines a logger, sets the log level and starts a profiler.
Root command has a PersistentPostRun function in which it stops the profiler safely, closes file descriptors etc, whatever is needed for a clean and safe exit.

Child command also has a PersistentPreRun and PersistentPostRun for its own reasons - for its own child commands.
Since the child command has defined its own Persistent_, the root command Persistent_ are not called. 

To solve this, there need to be hooks that get fired irrespective of the child having its own hook - these are chain hooks.

PreRunChain gets called in the order of parent to child

PostRunChain gets called in the order of child to parent
i.e. root PreRunChain -> child PreRunChain

And PostRunChain gets called in the order of child to parent
i.e. child PostRunChain -> root PostRunChain

So for the previous scenario

root PreRunChain sets logger and profiler
child PreRunChain does its own initialization etc

```
/* the usual stuff that existed before - the overriding logic
PersistentPreRun
PreRun
Run
PostRun
PersistentPostRun
*/
```

child PostRunChain frees up its own things, for its safe exit
root PostRunChain stops the profiler

henvic:
When you have no Run for a given command and you end up printing the help for that given command: pre/post are silently skipped. This is the related [and not a feature], correct?

bogem:
Duplicate of #252

Issue #216, opened 2016-01-05
child command is not calling the parents PersistentPreRun function


algrebe:
Hi, I've been reading the documentation on the readme that says that the  `When the subcommand is executed, it will run the root command's PersistentPreRun but not the root command's PersistentPostRun` , but the PersistentPreRun isn't getting executed.

```go
package main

import (
    "fmt"
    "github.com/spf13/cobra"
)

func main() {
    var rootCmd = &cobra.Command{
        Use:   "root [sub]",
        Short: "My root command",
        PersistentPreRun: func(cmd *cobra.Command, args []string) {
            fmt.Printf("Inside rootCmd PersistentPreRun with args: %v\n", args)
        },
        PreRun: func(cmd *cobra.Command, args []string) {
            fmt.Printf("Inside rootCmd PreRun with args: %v\n", args)
        },
        Run: func(cmd *cobra.Command, args []string) {
            fmt.Printf("Inside rootCmd Run with args: %v\n", args)
        },
        PostRun: func(cmd *cobra.Command, args []string) {
            fmt.Printf("Inside rootCmd PostRun with args: %v\n", args)
        },
        PersistentPostRun: func(cmd *cobra.Command, args []string) {
            fmt.Printf("Inside rootCmd PersistentPostRun with args: %v\n", args)
        },
    }

    var subCmd = &cobra.Command{
        Use:   "sub [no options!]",
        Short: "My subcommand",
        PersistentPreRun: func(cmd *cobra.Command, args []string) {
                fmt.Printf("Inside subCmd PersistentPreRun with args: %v\n", args)
        },
        PreRun: func(cmd *cobra.Command, args []string) {
            fmt.Printf("Inside subCmd PreRun with args: %v\n", args)
        },
        Run: func(cmd *cobra.Command, args []string) {
            fmt.Printf("Inside subCmd Run with args: %v\n", args)
        },
        PostRun: func(cmd *cobra.Command, args []string) {
            fmt.Printf("Inside subCmd PostRun with args: %v\n", args)
        },
        PersistentPostRun: func(cmd *cobra.Command, args []string) {
            fmt.Printf("Inside subCmd PersistentPostRun with args: %v\n", args)
        },
    }

    rootCmd.AddCommand(subCmd)
    _ = rootCmd.Execute()
}

```

```bash
root@vagrant-ubuntu-vivid-64:/tmp# go version
go version go1.5.2 linux/amd64

[...]

PR #220, opened 2016-01-08
Issue #219: added PreRunChain and PostRunChain hooks

algrebe:
PreRunChain functions run in the order of root to child.
PostRunChain functions run in the order of child to root.

The overall exection order for a command is
PreRunChain
PersistentPreRun
PreRun
Run
PostRun
PersistentPostRun
PostRunChain

The following gist describes the usage and functionality
https://gist.github.com/algrebe/23cc8bf4739a129a6d0d

### DIFF
--- a/README.md
+++ b/README.md
@@ -641,22 +641,15 @@ command.SetUsageTemplate(s string)
 
 ## PreRun or PostRun Hooks
 
-It is possible to run functions before or after the main `Run` function of your command. The `PreRunChain`, `PersistentPreRun` and `PreRun` functions will be executed before `Run`. `PersistentPostRun`, `PostRun` and `PostRunChain` will be executed after `Run`.  The `Persistent*Run` functions will be inherrited by children if they do not declare their own.  These function are run in the following order:
+It is possible to run functions before or after the main `Run` function of your command. The `PersistentPreRun` and `PreRun` functions will be executed before `Run`. `PersistentPostRun` and `PostRun` will be executed after `Run`.  The `Persistent*Run` functions will be inherrited by children if they do not declare their own.  These function are run in the following order:
 
-- `PreRunChain`
 - `PersistentPreRun`
 - `PreRun`
 - `Run`
 - `PostRun`
 - `PersistentPostRun`
-- `PostRunChain`
 
-
-`PreRunChain` executes all PreRunChain functions from the root to the child. `PostRunChain` executes all PostRunChain functions from the child to the root. These
-are useful in cases such as setting the logger and profiling at the root level in `PreRunChain`, and stopping the profiler once the proram finishes in the `PostRunChain`, while still allowing subcommand `PreRunChain` and `PostRunChain` commands to execute for their children.
-
-An example of two commands which use all of these features is below.  When the subcommand is executed, it will run the root command's `PersistentPreRun` but not the root command's `PersistentPostRun` - because the subcommand has overridden the root command's `PersistentPostRun`. However, it will still run the root command's `PreRunChain` and `PostRunChain` in addition to the subcommand's `PreRunChain` and `PostRunChain`, because chains are supposed to execute all defined
-chain functions, and not just the overridden one.
+An example of two commands which use all of these features is below.  When the subcommand is executed, it will run the root command's `PersistentPreRun` but not the root command's `PersistentPostRun`:
 
 ```go
 package main
@@ -672,9 +665,6 @@ func main() {
 	var rootCmd = &cobra.Command{
 		Use:   "root [sub]",
 		Short: "My root command",
-        PreRunChain: func(cmd *cobra.Command, args []string) {
-            fmt.Printf("Inside rootCmd PreRunChain with args: %v\n", args)
-        },
 		PersistentPreRun: func(cmd *cobra.Command, args []string) {
 			fmt.Printf("Inside rootCmd PersistentPreRun with args: %v\n", args)
 		},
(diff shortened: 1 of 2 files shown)


mitake:
Hi @algrebe , @spf13 , nice to meet you.

I'm using cobra and I really want this feature! @spf13 do you have a plan to merge this PR?

algrebe:
@mitake Hi there. Yes, I'm waiting for it to be pulled too after a review - I think they might want a better name for it.
@eparis was actively following up with me regarding https://github.com/spf13/cobra/issues/216
so I guess its just a matter of time.

mitake:
@algrebe thanks for your reply, I'm looking forward to seeing next news!

spf13:
I'm fine with the functionality in general. I'm a bit worried this is a bit of an overcomplicated solution, but I must admit I don't really understand the use cases that would require all of this. This totals up to 7 different Run actions per command. It seems like overkill. 

Not a criticism, but more of a request to help me understand why we need all of these. It seems like a simpler approach could exist that would give us all the functionality we need.

algrebe:
@spf13 thanks for your reply
@spf13 good point. Before taking any new functionality, you should decide whether this is required, and whether we can make do with what is there.

1. Run function is needed to run the command.
2. Pre and Post are needed as hooks just before the command.
3. Persistent functions (existing logic ) that are inherited and can be overridden.

For me, I just need something that always runs - irrespective of whether the child has its own hooks or not. I had always assumed `Peristent*` did this, but turns out they can be overridden. Hence, I would need
4. Functions that cannot be overridden and run whenever any of its children / self is executed.

As for the usecase - this was the issue I was facing https://github.com/spf13/cobra/issues/219 .
Also, I feel a parent should have the option of running something irrespective of the child. 

I have a suggestion, which may / may not be acceptable but just throwing it out here - 
Can the `*Chain` functions take the place of `Persistent*` functions ? Since Persistent actually means always run.

spf13:
Chain without persistent makes sense to me.

Should the default behavior of pre and post just be behavior described as
chain here? I feel like we can probably cover all these use cases with just
that.

algrebe:
@spf13 Yes - makes sense to me. 
Shall I modify this pull request to remove the chains and put that behavior in Pre and Post ?

This is a change in the existing functionality though, so previous users will have to be aware that Pre and Post have changed right ?

spf13:
Let's have @eparis weigh in first.

eparis:
I'll start with that fact that I think chaining makes more sense than what we have now. What we have now is copying the Flags()/PersistentFlags() where 'chaining' just doesn't make sense.

We have what we have so I'm all over the place on what to do. Imagine I have a subcommand pre/post that is calling the parent command pre/post. With a change to make 'persistent' 'chained' my parent command is going to get called twice.

Imagine the same case where the parent pre/post isn't called today and shouldn't be, now it will be and the user of cobra won't have a reasonable way to know of this change in behavior.

2.0 API should ONLY have chained....

spf13:
I agree. There's a good solution here thought that doesn't break compatibility and doesn't introduce a more complicated mechanism.

We add a flag to Cobra that defaults to the current behavior, but can be flipped to the chain behavior. Cobra 2.0 we keep the flag, but change the default. Cobra 3.0 we remove the flag. 

That way new implementations can immediately take advantage of this cleaner approach while existing implementations can have time to migrate. 

Does that make sense? 

I recognize it's a big change, but I'm super weary of adding complexity to the interface and would much prefer to refine and simplify what we have now instead. If we can do that without introducing breaking changes that I think it's a win/win.

eparis:
@spf13 +1

bep:
+1

spf13:
@algrebe can you update the PR accordingly? I'd gladly merge.

algrebe:
@spf13 yes, this sounds good.

Just to confirm
1. Add a flag to cobra ( --use-chain-hooks ) that is false by default
2. PreRun and PostRun examine the --use-chained-hooks to decide whether to call it or not.

I'm confused about the order:
Currently PersistentPreRun is called before PreRun, and PersistentPostRun after PostRun.
If there is chaining, where would you want it ? Before PersistentPreRun / After PostRun or like how it is right now? Or depending on the flag ?

spf13:
We don't need persistent anymore. Just chaining.

With chaining it should cascade down from the root to the run of the
executed command (pre) then back the other direction with the post.

eparis:
Also I don't know about `--use-chain-hooks`. The end user shouldn't (need) to know about chaining, only the programmer. Which makes that that an external variable `UseChainHooks`.

However I'm not sure what behavior we should get if the var is set on the root command but not on the subcommand being executed...

spf13:
It should only be able to be set (or checked) on the root. It's effectively
global. This can be accomplished with private fields and functions. See how
Exec() is crafted.

algrebe:
@spf13 Hey, I think I've got a neat solution based on all the feedback - here it is.
You can also check out this gist https://gist.github.com/algrebe/cae234e77dae809bfe18
to test out its functionality.

spf13:
@algrebe I think we can do this without the persistent funcs at all... or at least I can't think of why we would need them. Just Pre & Post with a toggle option to make them universally chain-able or not. 

Am I oversimplifying it?

algrebe:
@spf13 I cant find a reason where someone would want a hybrid of both chaining and running a hook just for that command in the same setup, so simplifying it makes sense to me.
Although, just in case we've missed a valid scenario - this means loss of a feature right ? Which can easily be introduced later if necessary I guess. Shall we wait for others to pitch on or should I go ahead with the changes mentioned ?

spf13:
We should talk to @eparis. I'm not sure how much this existing feature is
in use. I don't use it, but I have far from the most complex usage of Cobra
out there. I'm ok with a breaking change if it doesn't affect many people.
If it does affect people then we should hold off on taking it out until we
make a breaking release.

algrebe:
@spf13 @eparis Hi there ! Long time. I was looking at the issues on the page and #252  seems related to this as well. Please let me know what else I can do to take this feature to completion.

LukasMa:
@spf13  @eparis  Hi. Don't want to bother you but I see a lot of open PRs and issues. Some are months old. Is this project still maintained?

PR #222, opened 2016-01-12
Allow functions within BashCompletionFunction to run unconditionally, limit their scope to a particular command

cpfair:
It's a pain to include all custom completion functions in a single bash function, and constantly check to make sure you're generating suggestions for the right command/s. So, this PR adds new `__custom_func` variants to handle both points. `__custom_func`'s behaviour is unchanged, so this should be backwards-compatible with existing handlers (excepting behaviour intended to run once on script load, which will now run on every completion).

The following is copy-pasted from the docs:

The `BashCompletionFunction` member can be used to define bash functions which generate completion suggestions. Within `BashCompletionFunction` you can define some of the following four bash functions, which should modify `COMPREPLY` as appropriate.

| Function | Runs always | Runs when no earlier handler produced completion results |
| --- | --- | --- |
| **Runs for Command** | `__custom_unconditional_command_func` | `__custom_command_func` |
| **Runs for Command & children** | `__custom_unconditional_func` | `__custom_func` |

If you define more than one `__custom*_func`, they will execute in the following order (latest takes precedence). If there is more than one instance of a certain function defined in a command hiearchy during completion, the most specific function will supplant less specific functions (which will _not_ execute at all).
- `__custom_unconditional_func`
- `__custom_unconditional_command_func`
- `__custom_func`
- `__custom_command_func`

### DIFF
--- a/bash_completions.go
+++ b/bash_completions.go
@@ -103,9 +103,17 @@ __handle_reply()
     fi
     COMPREPLY=( $(compgen -W "${completions[*]}" -- "$cur") )
 
+    declare -F __custom_unconditional_func >/dev/null && __custom_unconditional_func
+    declare -F __custom_unconditional_command_func >/dev/null && __custom_unconditional_command_func
+
     if [[ ${#COMPREPLY[@]} -eq 0 ]]; then
         declare -F __custom_func >/dev/null && __custom_func
+        declare -F __custom_command_func >/dev/null && __custom_command_func
     fi
+
+     # These are declared as globals within the specific commands' handler functions
+     # We must undefine them here, otherwise they will persist outside of their intended lifetime
+     unset -f __custom_func __custom_unconditional_func
 
     __ltrim_colon_completions "$cur"
 }
@@ -177,6 +185,8 @@ __handle_noun()
 
 __handle_command()
 {
+	# Clear handlers meant to be command-specific
+	unset -f __custom_command_func __custom_unconditional_command_func
     __debug "${FUNCNAME}: c is $c words[c] is ${words[c]}"
 
     local next_command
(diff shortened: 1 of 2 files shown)


spf13:
@eparis since you guys use this feature extensively, can you give it a review?

bash_completions.go
```diff
@@ -177,6 +185,8 @@ __handle_noun()
 
 __handle_command()
 {
+	# Clear handlers meant to be command-specific
+	unset -f __custom_command_func __custom_unconditional_command_func
```
eparis: whitespace nit.

eparis:
I'm trying to understand how to actually use these?  Should we replace the example with something that uses the new _command_ versions?

anthonyfok:
Hi @cpfair!  Thank you for your contribution!  Could you please follow up your pull request which enhances bash completion?
I guess you must have missed @eparis' code review and the subsequent request for more information from over a year ago, where he asked:

> I'm trying to understand how to actually use these? Should we replace the example with something that uses the new command versions?

Thank you!

cpfair:
If I remember correctly, the issue was that the normal method of providing custom autocomplete was only useful when used on the root function - as the docs noted (see diff, where they were changed). This meant that if you wanted different autocomplete behavior for different command stems, you had to manually add & maintain that switching logic yourself. This change adds a built-in slot in which one can insert command-specific autocompletion functions, which are automatically selected based on the stem which the user has already typed. I'm honestly not sure what the unconditional variants were for.

In any case, if it's been this long and nobody else has raised a similar issue, it may not be worth the trouble of figuring out exactly what Past Me was thinking here...

bogem:
@cpfair Will you solve all merge conflicts or should I close this PR?

Issue #227, opened 2016-01-17
Document generation documentation out-of-date


garthk:
* `GenManTree` and `GenMarkdownTree` live in `github.com/spf13/cobra/doc`, and take a command as their first argument
* `NewKubectlCommand` lives in `k8s.io/kubernetes/pkg/kubectl/cmd`, now, and takes the factory as an argument
* While we're at it, it'd be nice to have a simple Markdown example matching the simple man page example

Issue #226, opened 2016-01-17
GenManTree, GenMarkdownTree require trailing slash on dir argument


garthk:
… because they concatenate strings instead of using [`filepath.Join`](https://golang.org/pkg/path/filepath/#Join):

```
filename = dir + strings.Replace(filename, " ", "_", -1) + ".md"
```

garthk:
PR incoming.

PR #228, opened 2016-01-17
Fix doc generation and its documentation

garthk:
Fixes for #226 and #227.


garthk:
Thanks, @eparis! That was so fast, I didn't manage to merge my work-around into the project I'm working on. ツ

PR #230, opened 2016-01-17
Ensure SEE ALSO list has no leading comma.

### LINKED ISSUE #229: Hidden commands can cause leading comma in SEE-ALSO lists
`genMan` in `man_docs.go`:
```go
		children := cmd.Commands()
		sort.Sort(byName(children))
		for i, c := range children {
			if !c.IsAvailableCommand() || c.IsHelpCommand() {
				continue
			}
			if cmd.HasParent() || i > 0 {
				fmt.Fprintf(buf, ", ")
			}
			fmt.Fprintf(buf, "**%s-%s(%s)**", dashCommandName, c.Name(), header.Section)
		}
```
If the first member of `children` is not available, we `continue`. On the next command, `i > 0`, so we print a comma despite not having printed a command name.

garthk:
Fixes #229.


eparis:
willing to write a unit test?

garthk:
Yep. Not sure when I'll get to it, though. Might not be today.

garthk:
I've added a test for the SEE ALSO list, @eparis. Please let me know if you want `AssertLineFound` and `AssertNextLineEquals` moved to `doc/cmd_test.go` or some other file.

PR #242, opened 2016-02-07
Allow bash-completion to work with full-path executable

anthonyfok:
This allows, for example, bash-completion to work properly when the end user specifies the executable with a directory leading to it, e.g., `/usr/bin/hugo`, `~/go/bin/hugo`, instead of just `hugo`.


spf13:
Merged as 65a708c

PR #254, opened 2016-03-15
Must only add "[flags]" to the end of usage if not yet present

fabianofranz:
The commander must not infer that every command wants `[flags]` appended at the end of the usage string - some commands actually can't take flags there, for example something like:

```
rsh [flags] host -- REMOTE_COMMAND
```

This PR makes `[flags]` to only be added if it's not yet in the body of the usage string.

### DIFF
--- a/cobra.go
+++ b/cobra.go
@@ -27,11 +27,12 @@ import (
 )
 
 var templateFuncs template.FuncMap = template.FuncMap{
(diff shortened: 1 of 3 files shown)


fabianofranz:
@eparis ptal

eparis:
I hate golang templates.

PR #255, opened 2016-03-20
Advanced custom bash completion of flags

sttts:
kubectl-completion

- add completion after `--flag=` or `--flag=abc`
- add flag annotation for custom flag completion, like e.g. for `--namespace=` in Kubernetes kubectl

### DIFF
--- a/bash_completions.go
+++ b/bash_completions.go
@@ -12,6 +12,7 @@ import (
 
 const (
 	BashCompFilenameExt     = "cobra_annotation_bash_completion_filename_extentions"
+	BashCompCustom          = "cobra_annotation_bash_completion_custom"
 	BashCompOneRequiredFlag = "cobra_annotation_bash_completion_one_required_flag"
 	BashCompSubdirsInDir    = "cobra_annotation_bash_completion_subdirs_in_dir"
 )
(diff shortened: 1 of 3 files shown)


bash_completions.md
```diff
@@ -147,3 +147,35 @@ hello.yml                     test.json
...
 So while there are many other files in the CWD it only shows me subdirs and those with valid extensions.
+
+# Specifiy custom flag completion
+
+Similar to the filename completion and filtering usingn cobra.BashCompFilenameExt, you can specifiy
+a custom flag completion function with cobra.BashCompCustom:
+
+```go
+	annotation := make(map[string][]string)
+	annotation[cobra.BashCompFilenameExt] = "__handle_namespace_flag"
```
eparis: I this right?  annotation is map of strings to slices of strings, but you only set a string...

bash_completions.go
```diff
@@ -29,14 +30,6 @@ __debug()
     fi
 }
 
-# Homebrew on Macs have version 1.3 of bash-completion which doesn't include
-# _init_completion. This is a very minimal version of that function.
-__my_init_completion()
-{
-    COMPREPLY=()
-    _get_comp_words_by_ref cur prev words cword
-}
```
eparis: Why did you do this? _init_completion is the right thing to do. You just removed functionality from people who can use it and forced us all to be stuck with what OS X people have...

bash_completions.md
sttts: Oh, right. Outdated docs. Fixing...

bash_completions.go
sttts: Have to check what _init_completion does in addition.

In any case, we have to call `_get_comp_words_by_ref` with `-n =` to exclude `=` as a word separator. Otherwise, bash 4+ and base <4 will behave differently.

sttts: Fixed. Actually, we call _init_completion already with `-s` which implies `-n =`. Only on the Mac, I had to add `-n =` manually.

PR #256, opened 2016-03-22
Fix escaping of % in bash code strings

sttts:


sttts:
Ptal @eparis

eparis:
Seems more like: https://github.com/spf13/cobra/pull/256/files#diff-3db5cab63dbb32fff4264f5449fd2904L25 should be a `Fprint`, instead of `Fprintf`

sttts:
Either is fine. In fact, there is another line with `%%` which is slightly broken due to printf interpretation.

eparis:
I'd rather see less crazy escapism since it makes it easier for the next person to stumble along.

PR #257, opened 2016-03-25
Add noun aliases

sttts:
Example:
- "rc" for replicationcontroller
- "pods" for "pod".


sttts:
ptal @eparis

eparis:
sry.  How about a unit test of some sort?  otherwise it LGTM.

sttts:
Good point @ tests. Will add them of course.

sttts:
Here we go. Tests added. ptal.

PR #260, opened 2016-03-31
Fixing golint warnings

ianwalter:
* Moving final return outside of if-else
* Removing type declarations that Go can infer from values
* Cleaning up some existing comments
* Changing snake_case variables to camelCase
* Declaring slices using var instead of make
* Removing unused `err`
* Small gofmt change

### DIFF
--- a/cobra/cmd/helpers.go
+++ b/cobra/cmd/helpers.go
@@ -196,7 +196,7 @@ func isEmpty(path string) (bool, error) {
 		if err != nil {
 			return false, err
 		}
-		list, err := f.Readdir(-1)
+		list, _ := f.Readdir(-1)
 		// f.Close() - see bug fix above
 		return len(list) == 0, nil
 	}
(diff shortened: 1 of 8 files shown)


cobra/cmd/helpers.go
```diff
@@ -196,7 +196,7 @@ func isEmpty(path string) (bool, error) {
 		if err != nil {
 			return false, err
 		}
-		list, err := f.Readdir(-1)
```
eparis: how did this even compile before?

ianwalter: Good question. I guess the compiler allows it in that particular situation? Note: I think I actually found this through gometalinter and not golint.

kkirsche:
Overall I like this, but why remove the error instead of handling it? That seems like a bad practice to me.

ianwalter:
I agree it should be handled but I was just trying to take the path of least resistance and clean up without changing/adding too much logic. Sure enough it was merged really quickly!

PR #265, opened 2016-04-02
Run shellcheck on bash completion

sttts:


bash_completions.go
```diff
@@ -226,12 +228,12 @@ __handle_word()
...
         return
     fi
-    __debug "${FUNCNAME}: c is $c words[c] is ${words[c]}"
+    __debug "${FUNCNAME[0]}: c is $c words[c] is ${words[c]}"
     if [[ "${words[c]}" == -* ]]; then
         __handle_flag
     elif __contains_word "${words[c]}" "${commands[@]}"; then
         __handle_command
-    elif [[ $c -eq 0 ]] && __contains_word "$(basename ${words[c]})" "${commands[@]}"; then
+    elif [[ $c -eq 0 ]] && __contains_word "$(basename "${words[c]}")" "${commands[@]}"; then
```
eparis: This seems odd and I don't know the right answer.  But now ${words[c]} is not in quotes?

bash_completions.go
```diff
@@ -203,20 +205,20 @@ __handle_noun()
...
-    __debug "${FUNCNAME}: c is $c words[c] is ${words[c]}"
+    __debug "${FUNCNAME[0]}: c is $c words[c] is ${words[c]}"
 
     local next_command
     if [[ -n ${last_command} ]]; then
         next_command="_${last_command}_${words[c]//:/__}"
     else
         if [[ $c -eq 0 ]]; then
-            next_command="_$(basename ${words[c]//:/__})"
+            next_command="_$(basename "${words[c]//:/__}")"
```
eparis: This seems to unquote the variable....

bash_completions.go
```diff
@@ -88,11 +88,13 @@ __handle_reply()
...
                 if [[ ${index} -ge 0 ]]; then
-                    COMPREPLY=() PREFIX="" cur="${cur#*=}"
+                    COMPREPLY=()
+                    PREFIX=""
+                    cur="${cur#*=}"
                     ${flags_completion[${index}]}
                     if [ -n "${ZSH_VERSION}" ]; then
                         # zfs completion needs --flag= prefix
-                        eval COMPREPLY=( "\${COMPREPLY[@]/#/${flag}=}" )
+                        eval "COMPREPLY=( \"\${COMPREPLY[@]/#/${flag}=}\" )"
```
eparis: And this just melts my mind....

bash_completions.go
sttts: What do you mean? The `$(...)` creates a new "quote context".

bash_completions.go
sttts: single quotes might be more suitable.

bash_completions.go
sttts: As below, `"$(..."$X"...)"` is correct. It does not "unquote" the variable.

bash_completions.go
sttts: We need the `${flag}` interpolation. Single quotes don't make it easier.

Issue #267, opened 2016-04-11
Command with only hidden flags still shows "Flags:" section


johnSchnake:
Create a command with 1 flag marked as hidden, run `./app command -h` and you see the "Flags:" header but no flags.  The command should be checking that there are _available_ flags, not just `.HasFlags` which is what the template does.

Example program:
```
package main

import (
	"fmt"

	"github.com/spf13/cobra"
)

func main() {
	var s string
	RootCmd := &cobra.Command{
		Use: "mycmd",
		Run: func(cmd *cobra.Command, args []string) {
			fmt.Println("foo bar", s)
		},
	}
	RootCmd.Flags().StringVar(&s,"flagname","","a string flag",)
	RootCmd.Flags().MarkHidden("flagname")
	RootCmd.Execute()
}
```

Output of `go run onlyhidden.go -h`

```
Usage:
  mycmd [flags]

Flags:

```

apriendeau:
When you call `.Flags()`, it returns an instance of `*flag.FlagSet`. So you are actually marking `pflag` to hide flags. So you are actually calling [flagset.MarkHidden](https://godoc.org/github.com/spf13/pflag#FlagSet.MarkHidden). You will need to create a template for the that command to hide the flags since its not actually telling cobra to hide anything.

johnSchnake:
I agree that the hiding of a flag is handled on the pflags level by changing the hidden field on the flag and modifying the usage. However, the default usage template relies on the `HasFlags()` logic which I think is inappropriate in this user facing context.

Just like cobra checks if there are available subcommands before showing the subcommands header, we should be checking if there are available flags before showing the flags header.

apriendeau:
I agree with you here and I think your solution is nicely put. I was thinking on how to solve this issue, you nailed it, IMO.

johnSchnake:
OK, just waiting for pflag to get updated then will update cobra.

eparis:
pflag done.

PR #268, opened 2016-04-13
Issue #267: Ensure hidden flags/subcommands do not change help

johnSchnake:
If a command has one flag which is hidden, it should not, for
instance, show the `Flags: ` heading.  Likewise there are other
items in the help template which should respect hidden/deprecated
state.

### DIFF
--- a/command.go
+++ b/command.go
@@ -261,7 +261,7 @@ func (c *Command) UsageTemplate() string {
 		return c.parent.UsageTemplate()
 	}
 	return `Usage:{{if .Runnable}}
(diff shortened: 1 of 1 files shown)


johnSchnake:
@eparis Seems to have not pulled in the new changes to pflag; seems like it needs to be "Rebuild without cache"; I'm assuming I don't have the rights to that since I don't see that button in the UX anywhere.

spf13:
Merged as e14e47b

Thanks for this fix!

Issue #266, opened 2016-04-07
Ability to add nested commands through `cobra add`


ianwalter:
I would love to use `cobra add` to create nested commands, for example: 

```
cobra add db migrate
```

Cobra would then see that there are multiple arguments, create a directory called `db` if it doesn't exist and a command file inside of it called `migrate.go`. It would also create another file either inside of `db` or at the same directory level called `db.go` or `root.go` that will simply create a `dbCommand` so that nested commands can be added to it. I can submit a PR for this after my existing PR is resolved. Thanks.

spf13:
@ianwalter I love the idea, please contribute this. 

If I can, I'll provide my ideas on what this feature should do/work like.

I think that the syntax should be `cobra add db/migrate`.
This would create `db.go` (if needed), `db/` (if needed) and `db/migrate.go`.
It would also add the appropriate `AddCommand(addCmd)` calls to support the tree structure. 

Currently the add command does support the `--parent` flag which does operate in a similar way, but requires more steps and doesn't put it in a subdirectory.

ianwalter:
That sounds good to me. I will work on this. I must have missed `--parent`. Should I leave it intact? It seems to me that it's functionality would be completely covered by the proposed solution.

spf13:
Lets leave it for now. It works slightly differently and supports a few use
cases this doesn't.

waitingkuo:
Currently `cobra add db/migrate` does generate the subdirectory. But the behavior is strange. It generates a `var db/migrate = &cobra.Command{}`. Should we warn when the command contains "/" first?

spf13:
I don't think it should warn. I think it should just do the right thing.
Seems like the only bug is how the variable name is created. Should we just
strip the "/"?

kasisnu:
Looks like this bug still exists. I'm happy to work on this.

spf13:
Please do :)

Wulfheart:
@kasisnu what's the current status?

geniass:
For me this doesn't create the subdirectory automatically.

Once I create the subdir, it creates a .go file inside it, but as discussed above the file has invalid syntax.

The next problem is that the subdir is considered a separate package (I think?)
Since the parent command, e.g. `rootCmd` is not exported, it can't be directly referenced.

@spf13  Should we make the parent command public? Add a public function? Something else?

geniass:
I think this issue is a duplicate right?
https://github.com/spf13/cobra/issues/1059

[...]

PR #273, opened 2016-05-04
Adding ability to add nested commands through `cobra add`

ianwalter:
- Modified `cobra add` command to be able to parse argument into nested command structure (#266)
- Modified `createCmdFile` function to accept cmdPath
- Modified cmdFile template to specify package name, import it's child package if necessary, and to TitleCase command names so that they can be accessed by parent packages
- Added getChildNames helper function so that child commands can be added to their parent
- Simplified `writeTemplateToFile` function syntax.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,9 +1,14 @@
 language: go
 go:
-  - 1.3.3
-  - 1.4.2
-  - 1.5.1
+  - 1.4.3
+  - 1.5.4
+  - 1.6.2
   - tip
+
+matrix:
+  allow_failures:
+    - go: tip
+
 before_install:
   - mkdir -p bin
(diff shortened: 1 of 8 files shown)


ianwalter:
Not sure why this is failing with go 1.3.3 all of a sudden but it also failed in #271 so I don't think it is specific to my changes!

mattstratton:
Any idea when this will be merged? I think it's exactly what I need for a project right now...

jmdots:
I need this too :).   Looks dormant.

mattstratton:
Any thoughts on when this might be merged?

bogem:
@eparis

eparis:
I'll leave it for @spf13   I've never even looked at the code under `cmd/`

Issue #276, opened 2016-05-10
Cobra command doesn't work if GOPATH is multiple


ghost:
I have a GOPATH containing multiple paths like this: `/ssd/go:/home/username/go`. The cobra command wouldn't work because it treats it as a single path, resulting in stupid messages and errors.  
GOPATH is supposed to be a list of directories separated by a colon, as specified here: https://golang.org/cmd/go/#hdr-GOPATH_environment_variable

kbj:
There is a fix in my branch, but I can't figure out how to fix the tests. Does Travis CI supports a GOPATH with multiple entries?

dc0d:
My other projects are working fine (I have 3 paths) and I've decided to try cobra. Seems it is not fixed:

`Error: Cobra only supports project within $GOPATH`

jhernandezme:
I have the same issue with mulitple paths.

PR #280, opened 2016-05-20
Fixes issue #276.

kbj:
This commit fixes issue #276. If $GOPATH includes multiple paths (as
allowed by the spec), Cobra cannot guess the project path, so now init
bails out if a relative path is given. The original behaviour is not
changed, if $GOPATH only contains a single path (as in most cases).


kbj:
Gah. Tests, sorry. Will fix.

PR #283, opened 2016-05-25
Add support for setting a function to handle flag parsing errors

dnephin:
The default pflag error message only prints the bad flag. This enables an application to include a usage message or other details about the error.

I tried to match the design of the existing options (`UsageFunc` and `HelpFunc`).

### DIFF
--- a/command.go
+++ b/command.go
@@ -109,10 +109,11 @@ type Command struct {
 
 	flagErrorBuf *bytes.Buffer
 
-	args          []string                 // actual args parsed from flags
-	output        *io.Writer               // out writer if set in SetOutput(w)
-	usageFunc     func(*Command) error     // Usage can be defined by application
-	usageTemplate string                   // Can be defined by Application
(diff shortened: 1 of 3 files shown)


dnephin:
This PR is in use by [docker/docker](https://github.com/docker/docker/blob/master/cli/cobraadaptor/adaptor.go#L41) so that the error messages on unknown flags can include the  short usage of the command.

This PR includes a unit test for the new feature.

patrickod:
Have a project that'd benefit from this functionality. Are there specific blockers / changes that need to be resolved before it could be merged? An update to the README perhaps to list it as a new lifecycle function?

eparis:
I just totally missed this. If the original author will rebase I will merge it. So sorry.

dnephin:
rebased, and build is green

dnephin:
Thanks!

patrickod:
wow! thanks to both of you! ❤️

PR #284, opened 2016-05-27
Support validation of positional arguments

### LINKED ISSUE #42: If no subcommand, how to get args?

dnephin:
Fixes #42
Branched from #120

Adds a new field `Args` that allows a user to define the expected positional arguments that a command should receive. If a command doesn't receive the expected args it returns an error.

By accepting a validation function users can create their own custom validation for positional args. This PR includes validators for the most common cases.

### DIFF
--- a/README.md
+++ b/README.md
@@ -462,6 +462,38 @@ A flag can also be assigned locally which will only apply to that specific comma
 RootCmd.Flags().StringVarP(&Source, "source", "s", "", "Source directory to read from")
 ```
 
+### Specify if you command takes arguments
+
+There are multiple options for how a command can handle unknown arguments which can be set in `TakesArgs`
+- `Legacy`
+- `None`
+- `Arbitrary`
+- `ValidOnly`
+
+`Legacy` (or default) the rules are as follows:
+- root commands with no subcommands can take arbitrary arguments
+- root commands with subcommands will do subcommand validity checking
+- subcommands will always accept arbitrary arguments and do no subsubcommand validity checking
+
+`None` the command will be rejected if there are any left over arguments after parsing flags.
+
+`Arbitrary` any additional values left after parsing flags will be passed in to your `Run` function.
+
+`ValidOnly` you must define all valid (non-subcommand) arguments to your command. These are defined in a slice name ValidArgs. For example a command which only takes the argument "one" or "two" would be defined as:
+
+```go
+var HugoCmd = &cobra.Command{
+        Use:   "hugo",
+        Short: "Hugo is a very fast static site generator",
+	ValidArgs: []string{"one", "two", "three", "four"}
+	TakesArgs: cobra.ValidOnly
+        Run: func(cmd *cobra.Command, args []string) {
+            // args will only have the values one, two, three, four
+	    // or the cmd.Execute() will fail.
+        },
+    }
+```
+
 ### Bind Flags with Config
 
 You can also bind your flags with [viper](https://github.com/spf13/viper):
--- a/README.md
+++ b/README.md
@@ -462,36 +462,34 @@ A flag can also be assigned locally which will only apply to that specific comma
 RootCmd.Flags().StringVarP(&Source, "source", "s", "", "Source directory to read from")
 ```
 
-### Specify if you command takes arguments
+## Positional and Custom Arguments
 
-There are multiple options for how a command can handle unknown arguments which can be set in `TakesArgs`
-- `Legacy`
-- `None`
-- `Arbitrary`
-- `ValidOnly`
+Validation of positional arguments can be specified using the `Args` field.
 
-`Legacy` (or default) the rules are as follows:
-- root commands with no subcommands can take arbitrary arguments
-- root commands with subcommands will do subcommand validity checking
-- subcommands will always accept arbitrary arguments and do no subsubcommand validity checking
+The follow validators are built in:
 
-`None` the command will be rejected if there are any left over arguments after parsing flags.
+- `NoArgs` - the command will report an error if there are any positional args.
+- `ArbitraryArgs` - the command will accept any args.
+- `OnlyValidArgs` - the command will report an error if there are any positional args that are not in the ValidArgs list.
+- `MinimumNArgs(int)` - the command will report an error if there are not at least N positional args.
+- `MaximumNArgs(int)` - the command will report an error if there are more than N positional args.
+- `ExactArgs(int)` - the command will report an error if there are not exactly N positional args.
+- `RangeArgs(min, max)` - the command will report an error if the number of args is not between the minimum and maximum number of expected args.
 
-`Arbitrary` any additional values left after parsing flags will be passed in to your `Run` function.
-
-`ValidOnly` you must define all valid (non-subcommand) arguments to your command. These are defined in a slice name ValidArgs. For example a command which only takes the argument "one" or "two" would be defined as:
+A custom validator can be provided like this:
 
 ```go
-var HugoCmd = &cobra.Command{
-        Use:   "hugo",
-        Short: "Hugo is a very fast static site generator",
-	ValidArgs: []string{"one", "two", "three", "four"}
-	TakesArgs: cobra.ValidOnly
-        Run: func(cmd *cobra.Command, args []string) {
-            // args will only have the values one, two, three, four
-	    // or the cmd.Execute() will fail.
-        },
-    }
+
+Args: func validColorArgs(cmd *cobra.Command, args []string) error {
+  if err := cli.RequiresMinArgs(1)(cmd, args); err != nil {
+    return err
+  }
+  if myapp.IsValidColor(args[0]) {
+     return nil
+  }
+  return fmt.Errorf("Invalid color specified: %s", args[0])
+}
+
 ```
 
 ### Bind Flags with Config
@@ -511,6 +509,7 @@ In this example the persistent flag `author` is bound with `viper`.
 when the `--author` flag is not provided by user.
 
 More in [viper documentation](https://github.com/spf13/viper#working-with-flags).
+
 
 ## Example
 
--- a/args.go
+++ b/args.go
@@ -1,0 +1,98 @@
+package cobra
+
+import (
+	"fmt"
+)
+
+type PositionalArgs func(cmd *Command, args []string) error
+
+// Legacy arg validation has the following behaviour:
+// - root commands with no subcommands can take arbitrary arguments
+// - root commands with subcommands will do subcommand validity checking
+// - subcommands will always accept arbitrary arguments
+func legacyArgs(cmd *Command, args []string) error {
+	// no subcommand, always take args
+	if !cmd.HasSubCommands() {
+		return nil
+	}
+
+	// root command with subcommands, do subcommand checking
+	if !cmd.HasParent() && len(args) > 0 {
+		return fmt.Errorf("unknown command %q for %q%s", args[0], cmd.CommandPath(), cmd.findSuggestions(args[0]))
+	}
+	return nil
+}
+
+// NoArgs returns an error if any args are included
+func NoArgs(cmd *Command, args []string) error {
+	if len(args) > 0 {
+		return fmt.Errorf("unknown command %q for %q", args[0], cmd.CommandPath())
+	}
+	return nil
+}
+
+// OnlyValidArgs returns an error if any args are not in the list of ValidArgs
+func OnlyValidArgs(cmd *Command, args []string) error {
+	if len(cmd.ValidArgs) > 0 {
+		for _, v := range args {
+			if !stringInSlice(v, cmd.ValidArgs) {
+				return fmt.Errorf("invalid argument %q for %q%s", v, cmd.CommandPath(), cmd.findSuggestions(args[0]))
+			}
+		}
+	}
+	return nil
+}
+
+func stringInSlice(a string, list []string) bool {
+	for _, b := range list {
+		if b == a {
+			return true
+		}
+	}
+	return false
+}
+
+// ArbitraryArgs never returns an error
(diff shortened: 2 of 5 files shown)


args.go
```diff
@@ -0,0 +1,98 @@
...
+
+	// root command with subcommands, do subcommand checking
+	if !cmd.HasParent() && len(args) > 0 {
+		return fmt.Errorf("unknown command %q for %q%s", args[0], cmd.CommandPath(), cmd.findSuggestions(args[0]))
+	}
+	return nil
+}
+
+// NoArgs returns an error if any args are included
+func NoArgs(cmd *Command, args []string) error {
```
deads2k: Can a noarg command have subcommands?  This seems like the trigger I wanted in #285, right?

dnephin: Yes, subcommand lookup happens before this is called in `Find`, so these arg handling functions only take care of non-subcommand args.  I believe this `NoArgs` option is what you want in #285.

dnephin:
There's a problem with this PR. Arg validation happens before the `--help` flag is checked, so they can report errors incorrectly when `--help` is used.

I think the order of these operations needs to be adjusted.

dnephin:
Ok, that issue should be fixed. Now legacy arg validation happens at the same time (before any flags are parsed), but any new arg validation will happen after flag parsing, so that flags like `--help` will behave correctly.

I updated the tests to catch this case.

cobra_test.go
```diff
@@ -513,6 +523,11 @@ func TestSubCmdTakesArgs(t *testing.T) {
 func TestCmdOnlyValidArgs(t *testing.T) {
 	result := noRRSetupTest("echo times one two five")
 
+	if result.Error == nil {
+		t.Errorf("Expected an error")
+		t.FailNow()
```
stevekuznetsov: The `testing` package provides:
```go
t.Fatal("Expected an error")
```

as a convenience method for wrapping the two calls you're making here.

dnephin: Thanks. I've made that change.  It did seem strange that I had to call two functions, but somehow I missed this.

dnephin:
This PR is in use by https://github.com/docker/docker to validation positional args.

By using a function instead of an enum for the `Args` field, it allows us to do things like [validate the format of args](https://github.com/docker/docker/blob/1bbac428de015a24a0724a9d0990d500c445dcad/api/client/service/scale.go#L19-L42) before they are passed to any `Run` function.

rileytg:
Am I understanding correctly?

In addition to the documented features, you can now validate arguments in a custom `PositionalArgs` type func passed to `Args:`.

contrived example (imagine if instead of :
```go
func validColorArgs(cmd *cobra.Command, args []string) error {
  if err := cli.RequiresMinArgs(1)(cmd, args); err != nil {
    return err
  }
  if myapp.IsValidColor(args[0]) {
     return nil
  }
  return fmt.Errorf("Invalid color specified: %s", args[0])
}
```

dnephin:
@rileytg yup, that's correct. You can write custom validation functions for the positional args.

rileytg:
what do you think about changing the language to express something along the lines of: (im thinking maybe keep it like urs and add a `custom` to the end of the list with my example.) 


### Positional and Custom Arguments

Validation of positional arguments can be specified using the `Args` field. A custom validator can be provided like this:
```go
Args: func validColorArgs(cmd *cobra.Command, args []string) error {
  if err := cli.RequiresMinArgs(1)(cmd, args); err != nil {
    return err
  }
  if myapp.IsValidColor(args[0]) {
     return nil
  }
  return fmt.Errorf("Invalid color specified: %s", args[0])
}
```

The follow validators are built in:

 - `NoArgs` - the command will report an error if there are any positional args.
 - `ArbitraryArgs` - the command will accept any args.
 - `OnlyValidArgs` - the command will report an error if there are any positiona 
   args that are not in the `ValidArgs` list.
 - `MinimumNArgs(int)` - the command will report an error if there are not at 
   least N positional args.
 - `MaximumNArgs(int)` - the command will report an error if there are more than 
   N positional args.
 - `ExactArgs(int)` - the command will report an error if there are not 
   exactly N positional args.
 - `RangeArgs(min, max)` - the command will report an error if the number of args 
   is not between the minimum and maximum number of expected args.

dnephin:
That sounds good. I'll update the PR with that change

Julio-Guerra:
What about the help message of such commands? It would be great to automatically generate `command [flags] arg1 arg2 arg3`, etc. according to `Args` or whatever. Mostly because of `[flags]` which is already automatically concatenated to the `Use` field. So we can't simply put `arg1 arg2 arg3` in the `Use` field, otherwise we end up with `command arg1 arg2 arg3 [flags]` instead of `command [flags] arg1 arg2 arg3`.

dnephin:
I find the `[flags]` placement to be unreliable in general, and I disable it by setting a custom template. The problem you describe already exists today with a use line. You can fix it by setting the use line to `command [flags] arg1 arg2 arg3` explicitly.

dnephin:
Rebased and updated the README text as suggested

jwhitcraft:
I would really love this feature to be merged, so I don't have to use the fork and be able to get all other fixes.  

@dnephin @eparis what is the status of this?

dnephin:
Still waiting for feedback from the project maintainers. 

This code is being used in `docker/docker` (as of 1.13).

I pulled the latest from `spf13/master` into my fork, so you could use that for now if you want.

jwhitcraft:
@dnephin, i just pulled it down and it seems your master branch is missing the commits from this branch.  Could you rebase this branch with master so i can just use this branch from your fork?

Thanks!

dnephin:
The master branch has these changes, it just doesn't have the same commits. I think they were rebased from a different branch, so the commit ids are different.

cobra_test.go
```diff
@@ -109,6 +113,16 @@ var cmdRootSameName = &Command{
 	Long:  "The root description for help",
 }
 
+var cmdRootTakesArgs = &Command{
+	Use:   "root-with-args [random args]",
+	Short: "The root can run it's own function and takes args!",
```
kofalt: s/it's/its

harsimranmaan:
@spf13 @dnephin Is this still waiting on something?

deterok:
@dnephin what is the status of the PRs now?

dnephin:
Status hasn't changed. Waiting on someone to accept or reject, or provide more feedback about how to get it accepted.

README.md
```diff
@@ -423,6 +423,42 @@ A flag can also be assigned locally which will only apply to that specific comma
...
 ```
 
+### Positional Arguments
+
+Validation of positional arguments can be specified using the `Args` field, which accepts
+one of the following values:
+
+- `NoArgs` - the command will report an error if there are any positional args.
+- `ArbitraryArgs` - the command will accept any args.
+- `OnlyValidArgs` - the command will report an error if there are any positiona
```
bogem: Typo: if there are any **positional**

bogem:
You did really good job, but I have some questions:
1. I can't understand it, but what's the problem, to validate the arguments in `Run`?
2. What's the purpose for `ArbitraryArgs`? It does nothing, so instead of this user can just use `Args: nil` and command will take any argument.

It's very important PR, but we should consider it carefully, because it adds complexity to cobra.

bogem:
And could you please give an explanation or example when `ValidArgs` will be really useful?

dnephin:
> I can't understand it, but what's the problem, to validate the arguments in Run?

Sure, that's mostly possible, but that's also true for pre/post run hooks, and yet those exist. Validating positional arguments makes code using cobra more consistent. 

It also gives users a way to override the strange default args behaviour.

>What's the purpose for `ArbitraryArgs`? 

`ArbitraryArgs` is different from the old default behaviour. The old default (https://github.com/spf13/cobra/pull/284/files#diff-07ee57a49d1c0b692bd79676cc31dd6fR13) now called `legacyArgs()` is a bit complicated, and not the easiest thing to understand. If you want to override that you can use `ArbitraryArgs`.

> could you please give an explanation or example when ValidArgs will be really useful?

It would  be useful anytime your app uses positional args, or doesn't use positional args (so that you can raise an error). So basically always :)

bogem:
I understood first 2 points, but unfortunately I can't imagine an example, where `ValidArgs` could be useful.
For example, my app has command `echo times [number]`, that prints something `number` times.
`number` can be any positive number. So how in this case `ValidArgs` will be useful?
Maybe it's a bad example, but could you please show another concrete example, where it can be useful. I also didn't find the example of usage in docker.

dnephin:
I see, you'd like examples not use cases, no problem!

```
docker pull NAME
```
This command requires exactly one argument.

```
docker run [OPTIONS] IMAGE [COMMAND] [ARG...]
```
This command requires at least one argument.

```
docker service ls [OPTIONS]
```
This command accepts no arguments. Any argument should result in an error.

It is used in all the `docker/docker` packages under `cli/command/*`. Examples: https://github.com/docker/docker/blob/master/cli/command/image/history.go#L35
https://github.com/docker/docker/blob/master/cli/command/service/list.go#L30

bogem:
Yes, I understand this the concept of positional arguments. Maybe I said not so clear, but where **ValidFlags** could be useful. I didn't find the usage of this field in docker.

> For example, my app has command echo times [number], that prints something number times.
number can be any positive number. So how in this case ValidArgs will be useful?

dnephin:
What is ValidFlags? It's not in my PR. `ValidArgs` is an old thing that was in Cobra already. I don't think it's useful. The only time I could see it being used would be to validate an enum positional argument, but that must be a rare case.

bogem:
Because of my inattention I didn't it noticed it.
Actually, it looks good to me, but there is a typo in README (I already made a review 3 hours ago). And why did you insert "Positional Arguments" section in "Working with Flags"? I would insert docs about positional arguments right after "Working with Flags"  as h2 and not as h3.

dnephin:
Thanks, I've fixed the README

bogem:
Excellent!
But maybe `Args` field will be called as `CheckArgs` or somehow else so it will be clearer to understand the purpose of function.

bogem:
@dnephin what do you think of it:
> But maybe Args field will be called as CheckArgs or somehow else so it will be clearer to understand the purpose of function.
And I think built-in validators should be above custom.
Also IMHO it's more beautiful to conclude name of built in functions in backticks.

dnephin:
I've made those changes to the README.

I'm not really a fan of `CheckArgs`. Check them for what? `ValidateArgs` maybe, but there is already a `ValidArgs` which is a bit too similar.

bogem:
Ok, 'Args' is good too
Actually I like this PR. It adds really needed this feature and fix bug, to that there are many complains (#42).
We should only wait for @spf13 agreement

harsimranmaan:
Thanks @bogem and @dnephin for driving this forward. Hopefully, @spf13 would agree for this merge now. Much appreciated.

bogem:
Please rebase your branch

bogem:
[I solved the merge conflicts by myself](https://github.com/bogem/cobra/commit/50aacc2c543b724fcf1a1467b9637904a005ea04). Please copy code from my branch (`git pull https://github.com/bogem/cobra.git dnephin-define_args`) and paste to yours. After you push conflict solutions, I will write directly to @spf13 so he said his (final) decision.
I will not commit to master (except of bug fixes) till we merge it.

dnephin:
rebased

harsimranmaan:
@bogem @spf13 The ball in your court now. Thanks @dnephin .

bogem:
Waiting for @spf13 conclusion

kofalt:
Thanks @bogem and @dnephin! :tada:

notjames:
I know that this is coming from left field a little, however, I'm very new to Go and I'm trying to port an internal application and am currently very stuck primarily because the documentation simply does not make sense with regard to custom argument validation. I've been through this entire issue and the README.md, which briefly mentions it, but gives NO context about where it should, how it should look, etc, and everything I try ends up being an error. Would someone for the love of all that is good PLEASE explain how to use `Args:` and friends IE custom argument validation? @rileytg ? 

Tangentially, the documentation for this project tries to be good, but it is seriously lacking some important details methinks. It almost seems like nobody is keeping it updated or at the very least is keeping it updated only minimally. 

For instance, I thought someone hijacked the documentation when mentioning even the ability for custom arg validation because when I tried to use it, my IDE screamed that it wasn't even available in the struct. Turns out, after reading this issue and seeing the change was just merged a couple of days ago that the problem was simply an outdated library on my system. Kudos for adding something about it at all in the documentation, but the details about how to properly use it are completely wrong or just missing. 

How are developers supposed to rely on this library if we can't track the changes or use the library reliably? Sorry for the rant. It's almost 4 AM so...it's just that I've lost a lot of time trying to figure this out and it seems like I shouldn't have had to.

bogem:
@notjames Sorry for inconvenience and this feature is documented really bad. I will make a PR now to fix this issue and we can discuss it there.

PR #285, opened 2016-05-27
allow suggestions for non-root commands

deads2k:
Suggests were only appearing against the root command, but for nested command structures it makes sense to look for suggestions at the deepest command that was found.

@eparis

### DIFF
--- a/command.go
+++ b/command.go
@@ -430,13 +430,13 @@ func (c *Command) Find(args []string) (*Command, []string, error) {
 	}
 
 	// root command with subcommands, do subcommand checking
-	if commandFound == c && len(argsWOflags) > 0 {
+	if commandFound != nil && len(argsWOflags) > 0 {
 		suggestionsString := ""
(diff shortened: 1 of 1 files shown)


deads2k:
@eparis Is there a corporate signature for red hat?

deads2k:
Well clearly that had unintended consequences.

deads2k:
Good unit tests.  I want to be able to use suggestions for nested commands, so that `rootcommand subcommand misspelled-subsubcommand` can get a suggestion.  For that to work, I guess I'll need to annotate the `Command` somehow to indicate that it has no non-flag args.

@juanvallejo want to see what you can do?  See the linked pull for usage.

dnephin:
>  For that to work, I guess I'll need to annotate the Command somehow to indicate that it has no non-flag args.

Sounds related to the PR I just submitted in #284. I also factored out this `findSuggestions()` function which should make it easier to re-use. Maybe it should be public.

PR #286, opened 2016-05-31
Merge persistent flags before checking for a help flag

dnephin:
Small bug fix. Without this change calling `UsageFlags()` in the `Run()` of a command that doesn't have any other flags would use the wrong `help` flag, if  a custom help flag was set as a PersistentFlag() on a parent command.

### DIFF
--- a/command.go
+++ b/command.go
@@ -684,6 +684,7 @@ func (c *Command) ExecuteC() (cmd *Command, err error) {
 }
 
 func (c *Command) initHelpFlag() {
+	c.mergePersistentFlags()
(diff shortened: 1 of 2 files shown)


eparis:
Would you be willing to write a test case so we don't ever regress?

dnephin:
Sure, I'll do that

dnephin:
I've updated this PR with the test to cover the bug fix.

dnephin:
This PR was fixed a while ago, anything blocking a merge?

PR #287, opened 2016-06-03
Add Bash support for command-substitute flags

elyscape:
These are useful to mark a flag as taking the place of a command. For example, `kubectl create` can be called using either [a subcommand](http://kubernetes.io/docs/user-guide/kubectl/kubectl_create/#see-also) or by [providing a file as input](http://kubernetes.io/docs/user-guide/kubectl/kubectl_create/):

    kubectl create --filename=file-input.yml

    kubectl create secret generic [...]

This allows the generated Bash completions to reflect this and suggest both commands and command-substitute flags.


elyscape:
Updated with documentation added to `bash_completions.md`.

eparis:
What do you think about #288 instead.  Seems to do what you want and no changes required to other's code!

PR #288, opened 2016-06-03
Do not display bash commands if local flag specified

eparis:
If a user specified a non-persistent flag to a command they cannot then call a subcommand. Thus we should not display subcommands in the bash completions.

This also changes things such that we show both 'required flags' and 'subcommands' if a 'required flag' exists. Since you may be able to run a subcommand instead of use the 'required' flag.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -19,6 +19,18 @@ _cgo_export.*
 
 _testmain.go
 
+# Vim files https://github.com/github/gitignore/blob/master/Global/Vim.gitignore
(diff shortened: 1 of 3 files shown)


eparis:
@elyscape

elyscape:
Ooh, nice.

eparis:
as in merge it and close #287 or you think #287 is better, or you want to look more?

elyscape:
This one looks fine to me.

PR #290, opened 2016-06-07
Fixed typos in README and docstring

joe2far:


joe2far:
I have updated the user in the commit to joe2far and I have also now signed the CLA.

PR #293, opened 2016-06-14
Sort commands by their names

andreykurilin:
When the number of commands is big, help message become unreadable (It is hard to read "Available Commands" section and find proper command).
Example of output from Kubernetes project(`kubectl --help`):

```
Available Commands:
  get            Display one or many resources
  set            Set specific features on objects
  describe       Show details of a specific resource or group of resources
  create         Create a resource by filename or stdin
  replace        Replace a resource by filename or stdin.
  patch          Update field(s) of a resource using strategic merge patch.
  delete         Delete resources by filenames, stdin, resources and names, or by resources and label selector.
  edit           Edit a resource on the server
  apply          Apply a configuration to a resource by filename or stdin
  namespace      SUPERSEDED: Set and view the current Kubernetes namespace
  logs           Print the logs for a container in a pod.
  rolling-update Perform a rolling update of the given ReplicationController.
  scale          Set a new size for a Deployment, ReplicaSet, Replication Controller, or Job.
  cordon         Mark node as unschedulable
  drain          Drain node in preparation for maintenance
  uncordon       Mark node as schedulable
  attach         Attach to a running container.
  exec           Execute a command in a container.
  port-forward   Forward one or more local ports to a pod.
  proxy          Run a proxy to the Kubernetes API server
  run            Run a particular image on the cluster.
  expose         Take a replication controller, service, deployment or pod and expose it as a new Kubernetes Service
  autoscale      Auto-scale a Deployment, ReplicaSet, or ReplicationController
  rollout        rollout manages a deployment
  label          Update the labels on a resource
  annotate       Update the annotations on a resource
  taint          Update the taints on one or more nodes
  config         config modifies kubeconfig files
  cluster-info   Display cluster info
  api-versions   Print the supported API versions on the server, in the form of "group/version".
  version        Print the client and server version information.
  explain        Documentation of resources.
  convert        Convert config files between different API versions
  completion     Output shell completion code for the given shell (bash or zsh)
```

After this pr, it will look like:
```
Available Commands:
  annotate       Update the annotations on a resource
  api-versions   Print the supported API versions on the server, in the form of "group/version".
  apply          Apply a configuration to a resource by filename or stdin
  attach         Attach to a running container.
  autoscale      Auto-scale a Deployment, ReplicaSet, or ReplicationController
  cluster-info   Display cluster info
  completion     Output shell completion code for the given shell (bash or zsh)
  config         config modifies kubeconfig files
  convert        Convert config files between different API versions
  cordon         Mark node as unschedulable
  create         Create a resource by filename or stdin
  delete         Delete resources by filenames, stdin, resources and names, or by resources and label selector.
  describe       Show details of a specific resource or group of resources
  drain          Drain node in preparation for maintenance
  edit           Edit a resource on the server
  exec           Execute a command in a container.
  explain        Documentation of resources.
  expose         Take a replication controller, service, deployment or pod and expose it as a new Kubernetes Service
  get            Display one or many resources
  label          Update the labels on a resource
  logs           Print the logs for a container in a pod.
  namespace      SUPERSEDED: Set and view the current Kubernetes namespace
  patch          Update field(s) of a resource using strategic merge patch.
  port-forward   Forward one or more local ports to a pod.
  proxy          Run a proxy to the Kubernetes API server
  replace        Replace a resource by filename or stdin.
  rolling-update Perform a rolling update of the given ReplicationController.
  rollout        rollout manages a deployment
  run            Run a particular image on the cluster.
  scale          Set a new size for a Deployment, ReplicaSet, Replication Controller, or Job.
  set            Set specific features on objects
  taint          Update the taints on one or more nodes
  uncordon       Mark node as schedulable
  version        Print the client and server version information.
```

### DIFF
--- a/cobra.go
+++ b/cobra.go
@@ -41,6 +41,10 @@ var initializers []func()
 // Set this to true to enable it
 var EnablePrefixMatching = false
 
+//EnableCommandSorting controls sorting of the slice of commands, which is turned on by default.
+//To disable sorting, set it to false.
+var EnableCommandSorting = true
+
 //AddTemplateFunc adds a template function that's available to Usage and Help
 //template generation.
 func AddTemplateFunc(name string, tmplFunc interface{}) {
--- a/command.go
+++ b/command.go
@@ -21,6 +21,7 @@ import (
 	"io"
 	"os"
 	"path/filepath"
+	"sort"
 	"strings"
 
 	flag "github.com/spf13/pflag"
@@ -103,6 +104,8 @@ type Command struct {
 	commandsMaxUseLen         int
 	commandsMaxCommandPathLen int
 	commandsMaxNameLen        int
+	// is commands slice are sorted or not
+	commandsAreSorted bool
 
 	flagErrorBuf *bytes.Buffer
 
@@ -721,8 +724,20 @@ func (c *Command) ResetCommands() {
 	c.helpCommand = nil
 }
 
-//Commands returns a slice of child commands.
+// Sorts commands by their names
+type commandSorterByName []*Command
+
+func (c commandSorterByName) Len() int           { return len(c) }
+func (c commandSorterByName) Swap(i, j int)      { c[i], c[j] = c[j], c[i] }
+func (c commandSorterByName) Less(i, j int) bool { return c[i].Name() < c[j].Name() }
+
+// Commands returns a sorted slice of child commands.
 func (c *Command) Commands() []*Command {
+	// do not sort commands if it already sorted or sorting was disabled
+	if EnableCommandSorting && !c.commandsAreSorted{
+		sort.Sort(commandSorterByName(c.commands))
+		c.commandsAreSorted = true
+	}
 	return c.commands
 }
 
@@ -751,6 +766,7 @@ func (c *Command) AddCommand(cmds ...*Command) {
 			x.SetGlobalNormalizationFunc(c.globNormFunc)
 		}
 		c.commands = append(c.commands, x)
+		c.commandsAreSorted = false
 	}
 }
 
--- a/command_test.go
+++ b/command_test.go
@@ -133,3 +133,44 @@ func Test_DisableFlagParsing(t *testing.T) {
 		t.Errorf("expected: %v, got: %v", as, targs)
 	}
 }
+
+func TestCommandsAreSorted(t *testing.T) {
+	EnableCommandSorting = true
+
+	originalNames := []string{"middle", "zlast", "afirst"}
+	expectedNames := []string{"afirst", "middle", "zlast"}
+
+	var tmpCommand = &Command{Use: "tmp"}
+
+	for _, name := range(originalNames) {
+		tmpCommand.AddCommand(&Command{Use: name})
+	}
+
+	for i, c := range(tmpCommand.Commands()) {
+		if expectedNames[i] != c.Name() {
+			t.Errorf("expected: %s, got: %s", expectedNames[i], c.Name())
+		}
+	}
+
+	EnableCommandSorting = true
+}
+
+func TestEnableCommandSortingIsDisabled(t *testing.T) {
+	EnableCommandSorting = false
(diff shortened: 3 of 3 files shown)


command.go
```diff
@@ -721,8 +724,19 @@ func (c *Command) ResetCommands() {
...
+
+func (c commandSorterByName) Len() int           { return len(c) }
+func (c commandSorterByName) Swap(i, j int)      { c[i], c[j] = c[j], c[i] }
+func (c commandSorterByName) Less(i, j int) bool { return c[i].Name() < c[j].Name() }
+
+// Commands returns a sorted slice of child commands.
 func (c *Command) Commands() []*Command {
+	// do not sort commands if it doesn't needed
+	if !c.commandsAreSorted{
+		sort.Sort(commandSorterByName(c.commands))
```
eparis: wouldn't you need to set `commandsAreSorted` ?

eparis:
If you write an _test I'll very happily merge...

andreykurilin:
@eparis eh...forgot to set commandsAreSorted.. fixed :) will add some tests

dnephin:
I believe right now the order is based on the order commands are added using `AddCommand()`, which means that any application using cobra can choose to sort the commands before calling `AddCommand()`, and any application that wants a different ordering is able to keep the order they defined.

With this change it wont be possible to keep the custom ordering. Does that sound right?

andreykurilin:
@dnephin: yes, you are right about order. From my point of view, there are no reasons to support old approach, but I can be wrong.

To support both orders, Command struct can be extended by one more public field (SortCommands bool). What do you think about this?
cc @eparis

eparis:
What about a way for an outside program to set CommandsAreSorted ?  I know it is a 'breaking' change, but would rather we default to alphabetical sorting and give others and opt-out.

andreykurilin:
@eparis 

> What about a way for an outside program to set CommandsAreSorted ? 

hm... It should help to decrease number of new fields, but it can produce misunderstood in cases when some part of outside already called Commands() before setting CommandsAreSorted. 
Also CommandsAreSorted will change value to false each time when `addCommand` is called.

> I know it is a 'breaking' change, but would rather we default to alphabetical sorting and give others and opt-out.

agree

andreykurilin:
@eparis I added test. Since it is my first test in go, I'll appreciate for any comments.

command_test.go
```diff
@@ -133,3 +134,22 @@ func Test_DisableFlagParsing(t *testing.T) {
 		t.Errorf("expected: %v, got: %v", as, targs)
 	}
 }
+
+func TestCommandsAreSorted(t *testing.T) {
+
+	originalNames := []string{"middle", "zlast", "afirst"}
```
eparis: I would prefer you create a manual list of output names, so we aren't using the `sort` package to test the `sort` package...

andreykurilin: done

andreykurilin:
@eparis: What do you think about making transition from commandsAreSorted bool to commandsSorting int (negative value means that sorting is disabled; 0 means unsorted; positive value - commands are sorted)? DisableCommandsSorting() will set commandsSorting to -1.

eparis:
Maybe somewhat copy `EnablePrefixMatching`?  Create a var `EnableCommandSorting` which defaults to true. Only if that external variable is true do you do sort and keep the internal `commandsAreSorted` variable separate?

andreykurilin:
@eparis done

cobra.go
```diff
@@ -41,6 +41,9 @@ var initializers []func()
 // Set this to true to enable it
 var EnablePrefixMatching = false
 
+// the slice of commands is sorted by default. To disable sorting, set this ti false.
```
eparis: use Godoc style.  start it `// EnableCommandSorting ....`  also s/ti/to/

andreykurilin:
@eparis done

andreykurilin:
@eparis thanks:)

PR #294, opened 2016-06-21
Fix man page gen bugs

dnephin:
* use `Command.UseLine()` instead of `Command.Name()` to get the correct args and usages string in the `SYNOPSIS`
* fixes a bug where a flags shorthand was being displayed even when it was deprecated
* fixed a bug where the man page section from the header was not being used as part of the filename (section 1 was assumed)
* add a new `GenManTreeFromOpts()` which takes additional options while leaving `GenManTree()` backwards compatible. Currently only takes one additional option, but it can be expanded without breaking backwards compatibility.

### DIFF
--- a/doc/man_docs.go
+++ b/doc/man_docs.go
@@ -45,9 +45,12 @@ func GenManTree(cmd *cobra.Command, header *GenManHeader, dir string) error {
 			return err
 		}
 	}
-	needToResetTitle := header.Title == ""
+	section := "1"
+	if header.Section != "" {
+		section = header.Section
+	}
 
-	basename := strings.Replace(cmd.CommandPath(), " ", "_", -1) + ".1"
+	basename := strings.Replace(cmd.CommandPath(), " ", "_", -1) + "." + section
 	filename := filepath.Join(dir, basename)
 	f, err := os.Create(filename)
 	if err != nil {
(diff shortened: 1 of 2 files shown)


doc/man_docs.go
```diff
@@ -34,35 +34,51 @@ import (
...
+	section := "1"
+	if header.Section != "" {
+		section = header.Section
+	}
 
-	basename := strings.Replace(cmd.CommandPath(), " ", "_", -1) + ".1"
-	filename := filepath.Join(dir, basename)
+	basename := fmt.Sprintf("%s.%s",
+		strings.Replace(cmd.CommandPath(), " ", opts.CommandSeparator, -1),
+		section)
```
eparis: This is getting pretty unreadable. Can we do this in 2 steps?

dnephin: Done, also added a default for `CommandSeparator`

doc/man_docs.go
```diff
@@ -34,35 +34,53 @@ import (
...
+	})
+}
+
+// GenManTreeFromOpts generates a man page for the command and all descendants.
+// The pages are written to the opts.Path directory.
+func GenManTreeFromOpts(cmd *cobra.Command, opts GenManTreeOptions) error {
+	header := opts.Header
 	if header == nil {
 		header = &GenManHeader{}
 	}
```
eparis: Why wouldn't we use `fillHeader()`

dnephin: Not sure I understand the question.

`fillHeader()` populates the header struct with default values for the command, so it has to be called with the child command not the root command, otherwise the header is wrong. It's still being called by `GenMan()` which I believe is the correct place for it.

PR #296, opened 2016-06-23
Set a different stream for deprecation errors

dnephin:
If `Command.SetOutput()` is used, all messages will end up on the same stream. However some messages, like deprecation messages, should always be on stderr.

This adds a new `io.Writer` for writing deprecation messages.

### DIFF
--- a/command.go
+++ b/command.go
@@ -109,13 +109,14 @@ type Command struct {
 
 	flagErrorBuf *bytes.Buffer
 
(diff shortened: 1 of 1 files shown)


dnephin:
A better way to handle this would probably be to fix `Out()` to actually be the err stream, and support a way to set the "output" stream. 

I've found a way to work around this for now, so going to close this PR.

PR #299, opened 2016-06-30
Add a TraverseChildren option to allow for flags on each command in the hierarchy

### LINKED ISSUE #277: reusing flags for subcommand.
same Flags should be allowed to be reuse in sub-commands.

Problem:
- few flags like `-f` are synonymous to `--force` or `--file` flags. 
- Since usage of same flag is not allowed in one command tree (command, sub-commands), changing UX is required, which is not good as per UX design.

dnephin:
Fixes #277

Currently cobra only parses flags on the last command in the chain. If you use `PersistentFlags()` you force the parent flag onto the child commands, cluttering help, and reducing the short flags available to child commands.

`TraverseChildren` allows each command in the hierarchy to have its own flags. Flags are parsed as commands are found during the `Traversal()`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -344,6 +344,19 @@ A flag can also be assigned locally which will only apply to that specific comma
 RootCmd.Flags().StringVarP(&Source, "source", "s", "", "Source directory to read from")
 ```
 
+### Local Flag on Parent Commands
+
+By default Cobra only parses local flags on the target command, any local flags on 
+parent commands are ignored. By enabling `Command.TraverseChildren` Cobra will 
+parse local flags on each command before executing the target command.
+
+```go
+command := cobra.Command{
+  Use: "print [OPTIONS] [COMMANDS]",
+  TraverseChildren: true,
+}
+```
+
 ### Bind Flags with Config
 
 You can also bind your flags with [viper](https://github.com/spf13/viper):
--- a/command.go
+++ b/command.go
@@ -125,6 +125,9 @@ type Command struct {
 	// Must be > 0.
 	SuggestionsMinimumDistance int
 
+	// TraverseChildCommands parses flags on all parents before the child command
+	TraverseChildCommands bool
+
 	// name is the command name, usually the executable's name.
 	name string
 	// commands is the list of commands supported by this program.
@@ -475,13 +478,14 @@ func argsMinusFirstX(args []string, x string) []string {
 	return args
 }
 
+func isFlagArg(arg string) bool {
+	return ((len(arg) >= 3 && arg[0] == '-') ||
+	        (len(arg) >= 2 && arg[1] == '-' && arg[1] != '-'))
+}
+
 // Find the target command given the args and command tree
 // Meant to be run on the highest node. Only searches down.
 func (c *Command) Find(args []string) (*Command, []string, error) {
-	if c == nil {
-		return nil, nil, fmt.Errorf("Called find() on a nil Command")
-	}
-
 	var innerfind func(*Command, []string) (*Command, []string)
 
 	innerfind = func(c *Command, innerArgs []string) (*Command, []string) {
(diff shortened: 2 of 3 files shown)


bogem:
But does it show all flags in help?

dnephin:
It does. It shows the flags on the command where the flags are relevant.

bogem:
It's an API change, so we should wait for @spf13 
Please rebase your PR to master

dnephin:
rebased

spf13:
added replies on #277

t00f:
Having a similar issue here. I have a `rootCmd` that has multiple subcommands. 3 of these subcommands needs the same parameter `param` and are added like:

```golang
subcommand1.Flags().StringSliceP("param", "p", nil, "Allow to pass multiple parameters to the command1")
viper.BindPFlag("param", subcommand1.Flags().Lookup("param"))

subcommand2.Flags().StringSliceP("param", "p", nil, "Allow to pass multiple parameters to the command2")
viper.BindPFlag("param", subcommand2.Flags().Lookup("param"))

subcommand3.Flags().StringSliceP("param", "p", nil, "Allow to pass multiple parameters to the command3")
viper.BindPFlag("param", subcommand3.Flags().Lookup("param"))
```

When I try to retrieve the flag value using `viper.GetStringSlice("param")`, it is sometimes empty, depending on the command. 

I tried to use Persistant Flags instead - it would work for me as subcommands does not have subcommands - but without success. 

@dnephin how can I use your `TraverseChildren` option to solve this issue ?

bogem:
@t00f I can say you that it will not fix your issue. It's because you can not bind different flags with the same name. Viper will bind only params flag of last command, in your case it's `subcommand3`.
It's spread problem that is solved for example in [hugo](https://github.com/gohugoio/hugo/blob/master/commands/hugo.go#L118) and in [my CLI app](https://github.com/bogem/nehm/blob/master/commands/commands.go#L56). You should move the binding of a flag from init or main func to `Run` func of corresponding subcommand, so the `params` flag only of launched subcommand will be binded.

dnephin:
rebased

bogem:
Could you please gofmt your changes and write some documentation to README? I want to merge it.

dnephin:
`gofmt` done. Where should it be documented in the README? A new section before the docs generation section?

bogem:
I think, it‘s better to put it after positional arguments.

dnephin:
Test failure is from #522. I just tried it on master and it fails there as well, so not related to this PR.

bogem:
> Test failure is from #522. I just tried it on master and it fails there as well, so not related to this PR.

To fix it on your local machine just update pflag library: `go get -u github.com/spf13/pflag`.

dnephin:
I added some text to the README, let me know if it makes sense or if I should add more.

command.go
```diff
@@ -125,6 +125,9 @@ type Command struct {
 	// Must be > 0.
 	SuggestionsMinimumDistance int
 
+	// TraverseChildren parses flags on all parents before executing child command
```
bogem: Please add a dot at end of comment.

README.md
```diff
@@ -344,6 +344,12 @@ A flag can also be assigned locally which will only apply to that specific comma
 RootCmd.Flags().StringVarP(&Source, "source", "s", "", "Source directory to read from")
 ```
 
+### Local Flag on Parent Commands
+
+By default Cobra only parses local flags on the target command, any local flags on 
+parent commands are ignored. By enabling `Command.TraverseChildren` Cobra will 
+parse local flags on each command before executing the target command.
+
```
bogem: Please add an example of usage.

command.go
```diff
@@ -993,7 +1046,20 @@ func (c *Command) HasAlias(s string) bool {
 	return false
 }
 
-// NameAndAliases returns string containing name and all aliases
+// HasNameOrAliasPrefix returns true if the Name or any of aliases start
+// with prefix
+func (c *Command) HasNameOrAliasPrefix(prefix string) bool {
```
bogem: Why do we need it public? It's better to make it private for having in API only necessary methods.

dnephin: ok, makes sense. I don't remember why I exported this one

dnephin:
> in your tests you use cmd.Traverse method to make it work. 

Yes, the tests are testing the new functionality, which is in `cmd.Traverse()`, which is why they are called `TestTraverse*`. They are not testing the `TraverseChildren` option because that is a simple boolean if/else and there's not really any significant logic to test there.

I could add a test for `ExecuteC()` to show that it reads the `TraverseChildren` option if you think it's necessary.

> cmd.TraverseChildren option is only used in cmd.SuggestionsFor and nowhere else at all.

No, it is used in `ExecuteC()` not `SuggestionsFor`.

> The expected behaviour for user is to just enable cmd.TraverseChildren option, execute cmd.Execute and everything will work. Why can't we do like this?

Yes, that is correct, that's how it works. What do you mean by "Why can't we do like this?" ? Do you mean for the test cases?

README.md
dnephin: ok, sounds good.

command.go
dnephin: ok, makes sense. I don't remember why I exported this one

README.md
```diff
@@ -344,6 +344,22 @@ A flag can also be assigned locally which will only apply to that specific comma
...
+```go
+
```
bogem: Please delete the newlines at the beginning and the end of example.

dnephin: Cool, done!

Issue #15, opened 2014-06-15
Match on everything else


ghigt:
First, thank you for this package.

I wondered if there was a way of being redirected on commands or something else when the first parameter is not known as a command?
For example:

``` sh
$ todo add 'work on project#1'
```

> redirected on command `add`.

``` sh
$ todo list
```

> redirected on command `list`.

``` sh
$ todo 'work on project#1'
```

> redirected on a command matching with everything except the existing commands (`add` and `list`) instead of raising an error.

PR #301, opened 2016-07-04
use root command if no other command matches and root is runnable

viktorbenei:
A very simple solution for https://github.com/spf13/cobra/issues/15 and https://github.com/spf13/cobra/issues/42

TL;DR; pick the root command if it's runnable and no other command matched.

### DIFF
--- a/cobra_test.go
+++ b/cobra_test.go
@@ -934,6 +934,11 @@ func TestRootSuggestions(t *testing.T) {
 
 	cmd := initializeWithRootCmd()
 	cmd.AddCommand(cmdTimes)
+	origCmdRun := cmd.Run
+	defer func() {
+		cmd.Run = origCmdRun
+	}()
+	cmd.Run = nil
 
 	tests := map[string]string{
 		"time":     "times",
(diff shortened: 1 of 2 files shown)


viktorbenei:
I did change the related failing tests too. These test changes work with my modification in `command.go`, as well as with the original code.

Obviously I might have missed something, but so far it worked for us.

IMO this is a more logical handling of the case when the root command is runnable, as the current one which will invoke the run method of the root command, but only in case there's no argument specified at all.

klingtnet:
I wonder why this doesn't get merged.
I really like cobra for building CLI's but I stumble over this exact problem every time I am developing an interface with this library.

bogem:
Actually, it breaks the suggestions for command. For example:
Before this update:
```
$ hugo lst
Error: unknown command "lst" for "hugo"

Did you mean this?
	list

Run 'hugo --help' for usage.
```

After:
```
$ hugo lst
Error: Unable to locate Config file. Perhaps you need to create a new site.
       Run `hugo help new` for details. (Config File "config" Not Found in "[/Users/albert]")

```
Sorry, but I don't think it's right solution, although it's simple

PR #303, opened 2016-07-06
Help command must use Stdout instead of Stderr

fabianofranz:
Help when called from `-h` or the `help` command must print to Stdout. Right now it only happens in the root cmd, any subcommand (e.g. `cmd subcmd -h`) prints to Stderr.

Other minor code reorg.

### DIFF
--- a/command.go
+++ b/command.go
@@ -134,25 +134,6 @@ func (c *Command) SetArgs(a []string) {
 	c.args = a
 }
 
-func (c *Command) getOut(def io.Writer) io.Writer {
-	if c.output != nil {
-		return *c.output
-	}
-
-	if c.HasParent() {
-		return c.parent.Out()
-	}
-	return def
-}
-
-func (c *Command) Out() io.Writer {
-	return c.getOut(os.Stderr)
-}
-
-func (c *Command) getOutOrStdout() io.Writer {
-	return c.getOut(os.Stdout)
-}
-
 // SetOutput sets the destination for usage and error messages.
 // If output is nil, os.Stderr is used.
 func (c *Command) SetOutput(output io.Writer) {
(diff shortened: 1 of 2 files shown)


fabianofranz:
PTAL @eparis @spf13

command.go
```diff
@@ -230,6 +231,33 @@ func (c *Command) HelpFunc() func(*Command, []string) {
 	}
 }
 
+// Output the usage for the command
+// Used when a user provides invalid input
+// Can be defined by user by overriding UsageFunc
+func (c *Command) Usage() error {
+	c.mergePersistentFlags()
+	err := tmpl(c.getOutOrStderr(), c.UsageTemplate(), c)
```
dnephin: I think this is wrong. It's not using `UsageFunc()` so if someone has set `SetUsageFunc()` it wont print the correct usage.

fabianofranz: @dnephin  Unless I'm missing something, it actually works the other way - `UsageFunc()` [uses this one](https://github.com/fabianofranz/cobra/blob/77530f537eb200c0781aac768e568bf35191525c/command.go#L208), and if another func were provided through `SetUsageFunc`, [it's going to be used](https://github.com/fabianofranz/cobra/blob/77530f537eb200c0781aac768e568bf35191525c/command.go#L200-L202) and this one will be ignored. 

I mostly did this to improve code clarity and conciseness: `UsageFunc()` calls `Usage()` the same way `HelpFunc()` were already calling `Help()`.

fabianofranz:
UP

command.go
dnephin: The problem is that `Usage()` is public, so anyone calling `Usage()` will not get the same result.  If `Usage()` were `usage()` that would be fine.

fabianofranz: @dnephin like in https://github.com/spf13/cobra/pull/305/commits/d6bf4ef243b7f3b96d8a4249e63af711ef33d53e

PR #305, opened 2016-07-15
Fixes how we expose help and usage and expose output without making assumptions

fabianofranz:
@eparis PTAL

### DIFF
--- a/cobra_test.go
+++ b/cobra_test.go
@@ -646,7 +646,7 @@ func TestSubcommandArgEvaluation(t *testing.T) {
 	second := &Command{
 		Use: "second",
 		Run: func(cmd *Command, args []string) {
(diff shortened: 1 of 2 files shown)


fabianofranz:
@DavidVorick agree about the need of a deprecation period, sorry about that. 

Anyways, the concept and implementation here were very wrong - you would get different output on `Help()` - `HelpFunc()(args)`, `Usage()` - `UsageFunc()(args)` - `UsageString()` depending on how you set custom funcs for help and usage. We must not make blind assumptions, but be very explicit and concise instead. Another example is `Out()` which were printing to `Stderr` if not set until recently, but was refactored to `OutOrStdour()` and `OutOrStderr()`.

Issue #307, opened 2016-07-18
Backward Compatibility


nelsam:
My tests (in [hel](https://github.com/nelsam/hel)) recently failed because I was using `cmd.Help`, which was removed recently.  It's not a big issue, but it did cause my project to be in a failing state for a couple of days (which, really, is my fault for not publishing binaries yet).

As a result of the breaking change, I'm just questioning what sort of backward compatibility guarantees cobra has, and whether or not it would consider supporting version tags for anyone who wants to use something like `gopkg.in` to help prevent breaking changes.  I'm probably happy enough continuing to pull from master as long as breaking changes are small (like the removed `Help` method), but is there at least a policy that you follow related to more major breaking changes?  Some way that we can know ahead of time to update our code to stop using the deprecated methods?

I've got no interest in vendoring dependencies, so I'm hoping that you'll support something other than that.

awh:
The removal of `Usage` has broken not only my master branch but also historic tags, which I now have to go back and back and make point releases for. Was this really necessary?

Note also this comment https://github.com/spf13/cobra/commit/d6bf4ef243b7f3b96d8a4249e63af711ef33d53e#commitcomment-18273101 by @davidvorick

ultimateboy:
This issue broke our builds as well. We were using `cmd.Help` to print the command help whenever an invalid set of flags were passed.

zquestz:
+1 to breaking changes causing my project to fail. I realize I should use a versioned copy of the library, however I figured something as simple as Help() would be maintained. Is this repo available in gopkg.in or another similar service?

frodopwns:
Anyone experiencing this issue can replace cmd.Help() with

```
hfunc := cmd.HelpFunc()
hfunc(cmd, args)
```

for a quick fix.

zquestz:
Exactly what I did... was not crazy about the cmd.HelpFun()(cmd, args) pattern....

hectorj2f:
+1 ... why did you break our projects :**( ?

krak3n:
Use something like `govendor` to pin at specific commits.

The project really should have version branches / tags to help people pin their dependencies.  Or use something like `gopkg.in`. Master will always be considered bleeding edge and likely to contain breaking changes so pinning your dependencies is important.

hectorj2f:
Sure, there is also something called `backwards compatibility` :D. Anyway, the usage of a specific version should avoid these situations.

eparis:
[...]

PR #308, opened 2016-07-20
Restore Help() and Usage() for backwards compatibility

fabianofranz:
Fixes https://github.com/spf13/cobra/issues/307

@eparis PTAL


shurcooL:
Should these be marked as Deprecated?

PR #313, opened 2016-08-02
Display neither hidden nor deprecated flags in bash completion

euank:
As per the title.

I'm not certain if there's a reason not to hide both of these, but I'm open to opinions or discussion otherwise. Subcommands that are deprecated/hidden are already hidden it looks like.


eparis:
LGTM let me know when the CLA and CI pass

euank:
All green

lucab:
Sweet (and branch naming :heart:)!

I know this is off-topic here, but any chance we can get a tagged release to sync upon, @eparis?

PR #314, opened 2016-08-09
Go tool path style

nikandfor:
Command `cobra init ./` gets different result than `cobra init .`. that's a bit strange.
Also you can't place your command a bit deeper than project root using relative syntax. e.g. `cobra init github.com/user/project/many_cmds/prog1`

I propose to use syntax in go tool style:
- absolute path -> use as is
- relative path with prefix `./` -> `$PWD/given/path/of/any/depth`
- other relative path -> `$GOPATH/src/repo/name/path/to/command/dir`
- relative path without `./` prefix with only one depth level (`dirname` or either `dirname/`) -> `$GOPATH/src/github.com/dirname/`

### DIFF
--- a/cobra/cmd/init.go
+++ b/cobra/cmd/init.go
@@ -72,7 +72,7 @@ func initializePath(path string) {
 			er(err)
 		}
 	} else { // If path exists and is not empty don't use it
-		empty, err := exists(path)
+		empty, err := isEmpty(path)
 		if err != nil {
 			er(err)
 		}
--- a/cobra/cmd/helpers.go
+++ b/cobra/cmd/helpers.go
@@ -141,42 +141,31 @@ func guessProjectPath() {
 	}
 
 	srcPath := getSrcPath()
-	// if provided, inspect for logical locations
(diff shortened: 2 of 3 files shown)


cobra/cmd/init.go
```diff
@@ -72,7 +72,7 @@ func initializePath(path string) {
 			er(err)
 		}
 	} else { // If path exists and is not empty don't use it
-		empty, err := exists(path)
```
nikandfor: this is misspell I suppose

nikandfor:
I tried to use it in format `cobra init` and it still do wrong thing:
It creates LISENCE on right directory, but main.go and cmd/root.go on $GOPATH/src/github.com/

bogem:
Thank you for your contribution! It is very useful.
Sorry for closing your PR, but the code of cmd was not so good, so I rewrote it in #428 and fixed described situations in https://github.com/spf13/cobra/commit/db6b9a8b3f3f400c8ecb4a4d7d02245b8facad66. 
I'm glad to see new contributions from you!

PR #317, opened 2016-08-15
Add the ability to tell which of the aliases was the Command actually…

jltallon:
… called with

  Introduces a CalledName() method on Command (plus a private member in Command)
which stores the alias that the command was originally called as (or Name() if
## it was the "Use" name)

This is a somewhat naïve approach, but it does solve the issue for my use case: two commands that are _almost_ exactly the same (to be precise, one is the opposite of the other), and so implementing the command twice makes little sense. With CalledName() one can determine which was the actual command invoked in order to set the proper variable.

(Thank you for Cobra, btw, Steve. Nice commander you created :)

### DIFF
--- a/command.go
+++ b/command.go
@@ -96,6 +96,8 @@ type Command struct {
 	PersistentPostRunE func(cmd *Command, args []string) error
 	// DisableAutoGenTag remove
 	DisableAutoGenTag bool
+	// Contains the 'calledName' rather than the alias that matched
+	calledName string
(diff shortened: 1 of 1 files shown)


jltallon:
Bump. Had to sign the CLA thrice already :O

bogem:
@spf13

eparis:
I just merged #567 which should accomplish this. Thank you and so sorry to take so long to ever look!

PR #327, opened 2016-08-20
Some small fixes

bogem:


eparis:
I hope we don't break anyone fixing the spelling of the constant...   please please please...

PR #329, opened 2016-08-22
Bugfix/consistent error reporting when called with unknown subcommand

improbablejan:
We recently noticed that error reporting and auto-suggestion when a non-existent sub-command name is specified via command line only happens for the root command, not when specifying a sub-command of a sub-command. This behaviour was very unexpected, and I don't think there is any fundamental reason for it. Moreover, when looking into this issue, I found that the root command cannot take positional arguments if it has any sub-commands whereas sub-commands can do so, even if they have sub-commands.

In general, the problem of whether something is specifying a sub-command or a positional argument is unsolvable, but if a command has no `Run` or `RunE` specified, it's safe to assume we are referencing a sub-command.

This PR changes the behaviour such that if a command has sub-commands, an argument is provided, and it does not match any of the sub-command names:
- an error is printed if the command doesn't specify `Run` or `RunE`.
- `Run` or `RunE` is called with the remaining args, if specified.

Furthermore, this PR unifies this behaviour across commands, such that it's the same for the root and sub-commands or lower.

Tests have been added but due to them not being completely hermetic, I had to change some details elsewhere. I think the tests should be rewritten to always get fresh instances to prevent this in the future; however, I didn't want to make this PR too large and contentious.

I think running command with no `Run`, `RunE`, and no sub-commands with an argument should also be an error. However, I can see how that's arguable so didn't want to add it to this PR - please do let me know if you think the behaviour should change there so it's unified.

I hope you find this contribution useful,
Jan

### DIFF
--- a/bash_completions_test.go
+++ b/bash_completions_test.go
@@ -53,7 +53,7 @@ COMPREPLY=( "hello" )
 )
 
 func TestBashCompletions(t *testing.T) {
-	c := initializeWithRootCmd()
+	c := initializeWithRootCmdWithRun()
 	cmdEcho.AddCommand(cmdTimes)
 	c.AddCommand(cmdEcho, cmdPrint, cmdDeprecated, cmdColon)
 
--- a/cobra_test.go
+++ b/cobra_test.go
@@ -27,6 +27,7 @@ var versionUsed int
 
 const strtwoParentHelp = "help message for parent flag strtwo"
 const strtwoChildHelp = "help message for child flag strtwo"
+const nonExistentCommand = "non_existent_command"
 
 var cmdHidden = &Command{
 	Use:   "hide [secret string to print]",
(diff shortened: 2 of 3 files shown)


Helcaraxan:
Friendly ping? Is there anybody who could have look at this and give some feedback.

This PR has been outstanding for quite a while.

eparis:
I think the processing of root vs subcmd being different is wrong and stoopid. But, it's what we have. Any change to it is an API change. It might break a working application. I believe there was work to make this more consistent along with the `PositionalArgs` addition, but my head is really fuzzy. I think if you set PositionArgs in any way, you will get sanity/consistency.

I know this is really old and we ignored you for 2 years, but I'm going to close this PR, because even though the API here today stinks I'm not willing to break any users.

dnephin:
> I think if you set PositionArgs in any way, you will get sanity/consistency.

This is correct. The `PositionalArgs` field allowed you to override the default behaviour, so you can make it consistent.

PR #331, opened 2016-08-23
gofmt, govet, run those in Travis

broady:
Also small cleanup to cobra_test.

### DIFF
--- a/cobra_test.go
+++ b/cobra_test.go
@@ -916,15 +916,14 @@ func TestRootUnknownCommand(t *testing.T) {
 
 func TestRootUnknownCommandSilenced(t *testing.T) {
 	r := noRRSetupTestSilenced("bogus")
-	s := "Run 'cobra-test --help' for usage.\n"
 
 	if r.Output != "" {
-		t.Errorf("Unexpected response.\nExpecting to be:\n %q\nGot:\n %q\n", s, r.Output)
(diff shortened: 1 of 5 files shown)


cobra_test.go
```diff
@@ -916,15 +916,14 @@ func TestRootUnknownCommand(t *testing.T) {
...
 
 	if r.Output != "" {
-		t.Errorf("Unexpected response.\nExpecting to be:\n %q\nGot:\n %q\n", s, r.Output)
+		t.Errorf("Unexpected response.\nExpecting to be: \n\"\"\n Got:\n %q\n", r.Output)
 	}
 
 	r = noRRSetupTestSilenced("--strtwo=a bogus")
 	if r.Output != "" {
-		t.Errorf("Unexpected response.\nExpecting to be:\n %q\nGot:\n %q\n", s, r.Output)
+		t.Errorf("Unexpected response.\nExpecting to be:\n\"\"\nGot:\n %q\n", r.Output)
```
eparis: This seems to be wrong...  What are you trying to do here?

broady: Why is it wrong?

The test is checking that `r.Ouput` is `""`.

The message was incorrect before.

eparis: No, I'm just an idiot.

broady:
Hmmm, @eparis - not sure if the build matrix stuff I just did is correct. I guess we'll see when it hits master.

The PR build seemed to only build for 1.6.3:
https://travis-ci.org/spf13/cobra/builds/154566855

Issue #330, opened 2016-08-22
cobra init not working


danesparza:
I'm using Go 1.7 on Windows.  I have used cobra on other projects.  I'm trying to start a new project and getting the following error:

`Error: Cobra only supports project within $GOPATH`.  

My GOPATH is a single path and is set correctly.  Several other projects build just fine.  Help!

elliotmr:
I have exactly the same problem.  It has to do with windows case insensitivity.  I have my GOPATH set to "c:\Users\Elliot\go", but the comparison checks against "C:\Users\Elliot\go\..." with a straight string comparison.

Unfortunately filepath.Clean() does not normalize the drive letter and cobra uses a straight string comparison.

tomdavidson:
I am on Debian Testing:

```
$ echo $GOPATH 
/home/tom/Projects/s4-go

$ echo $PWD
/home/tom/Projects/s4-go

$ cobra init
Error: Cobra only supports project within $GOPATH
```

eparis:
I believe you would need $PWD == $GOPATH/src/$REPO/$PROJECT

tomdavidson:
@eparis thanks for the reply. I'm pretty sure $PWD is suppose to be the current working directory. It was included in my comment to show that I am running `cobra init` from the the directory that is $GOPATH.

elliotmr:
The point is that you wouldn't want to actually run the command directly in the $GOPATH dir, it needs to be a subdirectory where you actually want the application to be created.

kytrinyx:
I ran into this on a couple of windows machines. In both cases:

- `GOPATH` was set
- `$GOPATH/src/github.com` existed
- The directory matching the user's username *did not exist* (`$GOPATH/src/github.com/<username>`)

To work around it we manually created their username directory, and then successfully ran the `cobra init github.com/<username>/<appname>` command.

/cc @spf13 @ashleymcnamara

kytrinyx:
> The point is that you wouldn't want to actually run the command directly in the $GOPATH dir

@elliotmr I believe the desired behavior is that if you pass it an import path (not just a single directory name) then it creates it in the correct location, even if you're typing the command from somewhere else (e.g. the $GOPATH).

kytrinyx:
@tomdavidson `cobra init` takes an argument which is the import path to your new application.

E.g.

```
$ echo $GOPATH
/Users/kytrinyx/code/go
[...]

PR #335, opened 2016-08-25
Changed string comparison for child path to use filepath.Rel

elliotmr:
Fix for #330


elliotmr:
Alright, I need to think about this.  The tests pass on windows, but giving a windows path with different drive case on linux makes the test fail.

PR #349, opened 2016-09-27
Set the defaults BEFORE considering a --config CLI argument

dldinternet:
viper.setConfigName will also set configFile == "" therefor we must set all defaults prior to considering command line arguments.


bogem:
Thank you for your contribution! It fixed in #428.

PR #351, opened 2016-10-02
Compatibility for multiple GOPATH workspaces

### LINKED ISSUE #276: Cobra command doesn't work if GOPATH is multiple

irfansharif:
This commit adds the necessary changes to support users with multiple
`GOPATH` workspaces. Previously the `getSrcPath()` function in
`cobra/cmd/helpers.go` would not take into account the fact the the
system `GOPATH` returns a colon-separated list if system was configured
with multiple GOPATH workspaces. If the GOPATH was set to be
`/foo/a:/bar/b`, it would simply append `src/` to the end of the string
literal giving `/foo/a:/bar/b/src` which resulted in undesired behavior.

Following the specs the changeset here adds support for multiple GOPATH
workspaces and if a given package directory is not to be found in any of
the workspaces the first one is chosen, i.e. for `cobra init` it will
first check all workspaces for the specified project and settling on the
first (`/foo/a`) if nothing found.
This preserves backwards compatibility as single workspace environments
are provably unaffected by these changes.

This fixes #276, cc @eparis.

---
This change is [Reviewable](https://reviewable.io/reviews/spf13/cobra/351)

### DIFF
--- a/cobra/cmd/helpers.go
+++ b/cobra/cmd/helpers.go
@@ -26,10 +26,6 @@ import (
 	"github.com/spf13/viper"
 )
 
-// var BaseDir = ""
-// var AppName = ""
-// var CommandDir = ""
-
 var funcMap template.FuncMap
 var projectPath = ""
 var inputPath = ""
@@ -99,16 +95,24 @@ func guessCmdDir() string {
 
 func guessImportPath() string {
 	guessProjectPath()
-
-	if !strings.HasPrefix(projectPath, getSrcPath()) {
-		er("Cobra only supports project within $GOPATH")
+	srcPaths := getSrcPaths()
+	for _, srcPath := range srcPaths {
+		if strings.HasPrefix(projectPath, srcPath) {
+			return filepath.ToSlash(filepath.Clean(strings.TrimPrefix(projectPath, srcPath)))
+		}
 	}
 
-	return filepath.ToSlash(filepath.Clean(strings.TrimPrefix(projectPath, getSrcPath())))
+	er("Cobra only supports project within $GOPATH")
+	return ""
 }
 
-func getSrcPath() string {
-	return filepath.Join(os.Getenv("GOPATH"), "src") + string(os.PathSeparator)
+func getSrcPaths() []string {
+	paths := strings.Split(os.Getenv("GOPATH"), string(os.PathListSeparator))
+	for i, gopath := range paths {
+		srcPath := filepath.Join(gopath, "src") + string(os.PathSeparator)
+		paths[i] = srcPath
+	}
+	return paths
 }
 
 func projectName() string {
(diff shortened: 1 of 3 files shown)


eparis:
@spf13 because I've never really looked at this part of the code at all..

cobra/cmd/helpers_test.go
```diff
@@ -29,8 +29,8 @@ func reset() {
 }
 
 func TestProjectPath(t *testing.T) {
-	checkGuess(t, "", filepath.Join("github.com", "spf13", "hugo"), filepath.Join(getSrcPath(), "github.com", "spf13", "hugo"))
-	checkGuess(t, "", filepath.Join("spf13", "hugo"), filepath.Join(getSrcPath(), "github.com", "spf13", "hugo"))
+	checkGuess(t, "", filepath.Join("github.com", "spf13", "hugo"), filepath.Join(getSrcPaths()[0], "github.com", "spf13", "hugo"))
+	checkGuess(t, "", filepath.Join("spf13", "hugo"), filepath.Join(getSrcPaths()[0], "github.com", "spf13", "hugo"))
```
eparis: no new test for the new code?

cobra/cmd/helpers.go
```diff
@@ -167,13 +185,20 @@ func guessProjectPath() {
...
 		} else {
-			projectPath = filepath.Join(srcPath, projectBase, inputPath)
+			for _, srcPath := range srcPaths {
+				fpath := filepath.Join(srcPath, projectBase, inputPath)
+				if b, _ := exists(fpath); b {
+					projectPath = fpath
+					return
+				}
+			}
+			projectPath = filepath.Join(srcPaths[0], projectBase, inputPath)
```
eparis: I think after typing the same code 3 times it'd be best to make it its own function.

irfansharif: agreed, I was thinking `findProjectPath` but the existence of `guessProjectPath` makes it non-intuitive. 
just as an aside the existence of global variables `projectPath`, `inputPath` and `projectBase` make it harder to refactor things around. would mean a bigger change set but changing the structure around those variables would make things much cleaner imo. thoughts?

cobra/cmd/helpers_test.go
irfansharif: any suggestions on how to mock out the `os.Getenv()` call in `getSrcPaths()`?

cobra/cmd/helpers.go
dixudx: A new function that wraps these three variables `srcPath`,`inputPath` and one of `""`, `"github.com"`and `projectBase`. Any comments?

sagikazarmark:
Any news?

PR #353, opened 2016-10-14
Fix condition in HelpFunc

bogem:
I think, it's more obvious now, what does this method

### DIFF
--- a/command.go
+++ b/command.go
@@ -231,12 +231,8 @@ func (c *Command) Usage() error {
 // HelpFunc returns either the function set by SetHelpFunc for this command
 // or a parent, or it returns a function with default help behavior.
 func (c *Command) HelpFunc() func(*Command, []string) {
-	cmd := c
-	for cmd != nil {
-		if cmd.helpFunc != nil {
-			return cmd.helpFunc
-		}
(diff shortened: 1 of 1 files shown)


eparis:
I think you broke it. but I might not understand. You are only looking at parent. The previous code checked any ancestor.

bogem:
@eparis you're right. I fixed it

command.go
```diff
@@ -247,6 +243,19 @@ func (c *Command) HelpFunc() func(*Command, []string) {
 	}
 }
 
+// checkHelpFunc checks if there is helpFunc in ancestors of c.
+func (c *Command) checkHelpFunc() func(*Command, []string) {
+	if c != nil {
```
eparis: Preference is for
```
if c == nil {
  return nil
}
```

command.go
```diff
@@ -247,6 +243,19 @@ func (c *Command) HelpFunc() func(*Command, []string) {
...
+		if c.helpFunc != nil {
+			return c.helpFunc
+		}
+		if c.parent != nil {
```
eparis: I believe this should be `if c.HasParent()`

PR #356, opened 2016-10-26
Fix the GPL license headers and texts

pombredanne:
- the current GPL 2 and 3 templates are incorrect on two counts:
  1. the headers make a reference to the Lesser GPL instead of the GPL
  2. the texts are incomplete and truncated: including a partial GPL
     is as good as not including it at all from a GPL standpoint.
- this patch corrects the header templates and also removes nice but
  not needed templating of appName.
- it also ensures that the full text of the GPL licenses are included. Both headers and
  texts are copied verbatim from these URLs (with the exceptions of backticks, replaced with single quote as it is not possible to escape backticks in raw strings):
  - https://www.gnu.org/licenses/gpl-3.0.txt
  - https://www.gnu.org/licenses/old-licenses/gpl-2.0.txt  

This is follow up on this discussion https://github.com/nexB/scancode-toolkit/issues/270#issuecomment-256116802 with @spf13 

Signed-off-by: Philippe Ombredanne pombredanne@nexb.com

### DIFF
--- a/cobra/cmd/licenses.go
+++ b/cobra/cmd/licenses.go
@@ -436,18 +436,21 @@ OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 		PossibleMatches: []string{"gpl2", "gnu gpl2"},
 		Header: `{{ .copyright }}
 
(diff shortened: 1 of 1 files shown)


pombredanne:
Let me fix the string syntax. This is likely an issue of quoting.

PR #359, opened 2016-11-03
Add a tags field for user defined tags on a Command

dnephin:
This `Tags` field would be used by an application to identify commands with specific properties. For example:
* experimental
* added-in version x

They could be used in custom help/usage text to group commands in different ways, or to filter the commands to hide based on tags.

See https://github.com/docker/docker/pull/28010 for a concrete use case.

cc @bep

### DIFF
--- a/command.go
+++ b/command.go
@@ -57,6 +57,9 @@ type Command struct {
 	Deprecated string
 	// Is this command hidden and should NOT show up in the list of available commands?
 	Hidden bool
+	// Annotations are key/value pairs that can be used by applications to identify or
+	// group commands
+	Annotations map[string]string
 	// Full set of flags
 	flags *flag.FlagSet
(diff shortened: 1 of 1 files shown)


kraman:
@spf13 Anything I can do to help this get merged?

dnephin:
Rebased.

@bep This is a small change, any feedback?

eparis:
do we want to copy https://github.com/spf13/pflag/blob/master/flag.go#L163 ? For some consistency?

dnephin:
I don't know if there is any need for the value to be `[]string{}`. I guess Annotations had some use case for them (bash completion).

eparis:
I guess I'd be find with the `map[string]string` if that's what you want. But following the naming at least makes sense.

dnephin:
Cool, done

PR #369, opened 2016-12-13
Added an UnknownCommandFunc to allow for custom handling of unknown commands

markbates:


markbates:
@spf13 ping

vjeantet:
I like this feature, can we see it merged ?

spf13:
@markbates This is a great idea. Sorry for sitting on it so long. I'm just catching up now on many of my projects and haven't reviewed this thoroughly. I would expect that the unknownCommandFunction would cascade up the tree to the root. Meaning I can write custom handling for my entire application on the root or I can define specific handling for subcommands. It doesn't appear that this code does this. What do you think?

markbates:
Yeah, if I remember this correctly, it just handles the use case of the calling an unknown command on another command. It doesn't walk up the tree. It would be great if it did.

spf13:
Do you want to extend it to do that? If you do (and document it), I'll gladly merge it.

bogem:
I also support this idea, but please don't forget to make tests for it!

vjeantet:
Hello,

This is my workaround to handle unknow command.

     root.go (RootCmd)  

```go
func Execute() {
	// Workaround to "forward" to a know command when no know command found
	_, _, err := RootCmd.Find(os.Args[1:])
	if err != nil && strings.HasPrefix(err.Error(), "unknown command") {
		// insert my command "run" in args
		os.Args = append(os.Args[:1], append([]string{"run"}, os.Args[1:]...)...)
	}

		fmt.Println("cmd execute error: ", err)
		os.Exit(-1)
	}
}
```

eparis:
I think this was a great idea, but it's been over a year. I'm going to close this PR, but if you ever do want to continue the project would be very interested in something which addresses both https://github.com/spf13/cobra/pull/369#issuecomment-296311314 and https://github.com/spf13/cobra/pull/369#issuecomment-343626299

PR #370, opened 2016-12-14
fmt package should be imported

pengqinglan:
fmt package should be imported in this code block
Signed-off-by: Qinglan Peng <qinglanpeng@zju.edu.cn>

### DIFF
--- a/README.md
+++ b/README.md
@@ -337,6 +337,7 @@ package cmd
 
 import (
 	"github.com/spf13/cobra"
+	"fmt"
 )
 
 func init() {


README.md
```diff
@@ -337,6 +337,7 @@ package cmd
 
 import (
 	"github.com/spf13/cobra"
+	"fmt"
 )
```
shurcooL: Why not do this similar to https://github.com/spf13/cobra#prerun-or-postrun-hooks, where `fmt` comes first, blank line, then `cobra`?

pengqinglan: you are right, I will fix it

PR #372, opened 2016-12-15
Add a hook function called when searching a command to be executed

ishidawataru:
This hook function is useful when we want to dynamically add commands.

An example of the usage:
https://github.com/ishidawataru/gobgp/commit/03f478c8fa81a7b3b3a239d01eb3d51ee1ea749d

GoBGP CLI supports commands like

`gobgp neighbor 1.1.1.1 shutdown`

In this case, `1.1.1.1` is a dynamic command, which can be any valid IP address.
Currently, GoBGP is using `Execute()` and `SetArgs()` to implement dynamic commands.
But current implementation can't display the help message of a dynamic command correctly.

With use of the hook ( `FindHookFn` ), we can implement these kinds of dynamic commands more easily and use the default help message mechanism.

### DIFF
--- a/command.go
+++ b/command.go
@@ -54,6 +54,9 @@ type Command struct {
 	// ValidArgs is list of all valid non-flag arguments that are accepted in bash completions
 	ValidArgs []string
 
+	// Dynamic
+	Dynamic func(name string) (bool, error)
+
 	// Expected arguments
 	Args PositionalArgs
 
@@ -191,7 +194,6 @@ type Command struct {
 	versionTemplate string
 }
 
-// SetArgs sets arguments for the command. It is set to os.Args[1:] by default, if desired, can be overridden
 // particularly useful when testing.
 func (c *Command) SetArgs(a []string) {
 	c.args = a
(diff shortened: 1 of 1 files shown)


eparis:
This feels a little bit like https://godoc.org/github.com/spf13/pflag#SetInterspersed  You don't really need a dynamic command, you could just accept `shutdown` as a child of `neighbor` if you got 1.1.1.1 in the args slice. I don't know if I like the idea of having one way to set interspersed `flags` and the way to get interspersed `nouns` is via dynamic command creation. Looking at your example I'm not sure how 'shutdown' is used. Is shutdown not a subcommand? It is just an arg? If it is just an arg I'm not sure what the custom command lookup/dynamic creation function is really getting you.

ishidawataru:
@edgard Thanks for your comment.

>  Is shutdown not a subcommand? It is just an arg?

It is a subcommand. The example above was not good.

Here is the simplified example of what I want to realize.

https://gist.github.com/ishidawataru/71dd20b3f0e602cadd132944040fd071

```shell
$ ./cobra
parent cmd called
$ ./cobra 1.1.1.1
impl cmd called: 1.1.1.1
$ ./cobra 1.1.1.2
impl cmd called: 1.1.1.2
$ ./cobra 1.1.1.2 sub
sub cmd called
$ ./cobra 1.1.1.1 sub
sub cmd called
```

I'd like to create a cli command which works like `ifconfig`

`ifconfig` accepts commands like

```shell
$ ifconfig eth0 up
$ ifconfig eth1 up
```

Can we do this using `SetInterspersed ` ?

PR #381, opened 2017-01-12
fixes config flag behaviour. related to #374 #375

ofpiyush:
**Expected behaviour:**

`--config` flag should override the default config path ($HOME/.\<project\>.yaml)

**Observed behaviour**
`--config` flag is overriden by the default config path

**Fix**
Run `SetConfigFile` after setting defaults.

Related to #374 and #375


calvn:
Will this be merged in the near future?

PR #383, opened 2017-01-17
Add EnableWindowsMouseTrap variable again 

sp-joseluis-ledesma:
In commit 193b182195dad17ec904d4851f33e3ee3696bf07 EnableWindowsMouseTrap variable was removed. This variable allowed to control whether mousetrap was enabled or not for Windows

### DIFF
--- a/cobra.go
+++ b/cobra.go
@@ -45,6 +45,9 @@ var EnablePrefixMatching = false
 // To disable sorting, set it to false.
 var EnableCommandSorting = true
 
+// EnableWindowsMouseTrap enables an information splash screen on Windows if the CLI is started from explorer.exe.
+var EnableWindowsMouseTrap = true
+
 // AddTemplateFunc adds a template function that's available to Usage and Help
 // template generation.
 func AddTemplateFunc(name string, tmplFunc interface{}) {
--- a/command_win.go
+++ b/command_win.go
@@ -18,9 +18,11 @@ You need to open cmd.exe and run it from there.
 `
 
 func preExecHook(c *Command) {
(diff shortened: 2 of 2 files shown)


bogem:
Maybe, we will not add global variable, but just make it this way: if user sets `MousetrapHelpText`  to "", mousetrap will not enable.

sp-joseluis-ledesma:
As a user of this library I would rather prefer having an specific variable to disable/enable this behaviour than setting MousetrapHelpText to an special value to handle it, but I can change the pull request if you feel is better this way.

bogem:
> As a user of this library I would rather prefer having an specific variable to disable/enable this behaviour than setting MousetrapHelpText to an special value to handle it

Can you please explain why?

It's better to use `MousetrapHelpText`, so we don't make any overhead to library. Also we should document it.

bogem:
Ping

sp-joseluis-ledesma:
Sorry, I've been busy. I found it makes the code more clear ( EnableWindowsMouseTrap=false is obvious what it does, meanwhile MousetrapHelpText="" it's not). Moreover this variable already existed by was silently wiped out.

Anyway I'll try to update the pullrequest with this change + documentation.

bogem:
> EnableWindowsMouseTrap=false is obvious what it does, meanwhile MousetrapHelpText="" it's not

Yes, it's not so obvious, but I don't want to add components to library, that are could be replaced with existing methods. So let's just document, what blank MousetrapHelpText disables mousetrap.

bogem:
Ping

PR #390, opened 2017-02-07
cache command name

dixudx:


eparis:
Historically this is set from the `Use` argument. Why is the insufficient? Also changing `c.name` after `c.Name()` has ever been called is a problem. So while I'm not sure we want this patch at all (maybe you can explain why it is required) it definitely should return an error if `c.name != ""`.

dixudx:
Currently `c.name` is always an empty string. When calling `Name()`, the first word (split by `" "`) in `Use` is returned and no value is assigned back to `c.name`. So the check `if c.name != ""` will always be false.

Also I think it may be redundant to split the string `Use` again and again when calling `Name()`. Well, it works when `c.Use` is changed.

eparis:
I never noticed that code was wrong. I always thought that `c.name` was assigned the first time that `c.Name()` was called. I'm so sorry! Personally I think that `c.Name()` should be updated to cache the result. We already have that unused `c.name`.

I'll also just looked through the code and maybe I was wrong. Maybe you can "safely" change `c.Name()`. I think all of the places the result of `c.Name()` is used in the code would be ok if it changed later.

So I guess I'm fine with this change. I'm just still not convinced why getting the `name` from `Use` doesn't work. If this is just about fixing the string parsing mess that's easier and doesn't create a new API.

dixudx:
Yes, you're right. All the codes call `c.Name()` instead of `c.name`. So making changes to `c.Name()` will not bring any conflicts.

And I will just modify `c.Name()` to cache the result. No needs to add a new API or function.

Will this be okay?

eparis:
Lets just go ahead and revert this, and add a test in the revert PR which tests the problem it caused for kubernetes?

shiywang:
This pr definitely break kubectl code base [here](https://github.com/kubernetes/kubernetes/blob/81631eceb80aace00dfb221b13440c7f658a18c7/pkg/kubectl/cmd/cmd.go#L359), I was wondering could we have a ci which could run integration test with k8s ?

dixudx:
@shiywang `name` field is removed from `Command`. xref #547.

PR #391, opened 2017-02-09
Fix subcommand locations with NoOptDefValue flags

eparis:
If one ran a command like

./root --boolFlag subcmd1 subcmd2

Thing worked fine. The code recognized that --boolFlag followed by a
space meant the next word was not the argument to --boolFlag. But other
flag types with a NoOptDefValue (like a Count flag) would not ignore the
"argument".  On a command like:

./root --countflag subcmd1 subcmd2

The processor, when looking for a subcommand, would first throw out the
`--countflag subcmd1` and then look for subcmd2 under root.

The fix is to ignore the next word after any NoOptDefVal flag, not just
boolean flags.

### DIFF
--- a/command.go
+++ b/command.go
@@ -384,20 +384,18 @@ func (c *Command) resetChildrensParents() {
 	}
 }
 
-// Test if the named flag is a boolean flag.
-func isBooleanFlag(name string, f *flag.FlagSet) bool {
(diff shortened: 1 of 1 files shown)


eparis:
I believe this fixes https://github.com/spf13/pflag/issues/112

ncw:
This patch works well for me - I've added it to the vendor branch for rclone.

PR #394, opened 2017-02-27
Some fixes related to help command

### LINKED ISSUE #366: Include 'help', if we have other commands at the top level
If other top-level commands are present, the displayed list of Available Commands should include the `help` command—however, it doesn't.

Since this repository seems the correct place for this issue, I mention it here (see Hugo issue [2349](https://github.com/spf13/hugo/issues/2349)).

### LINKED ISSUE #393: Misbehaviour of IsHelpCommand()
I notice, what IsHelpCommand always returns false even on built-in help command and I found out why:
Look at the docs of IsHelpCommand:
```
IsHelpCommand determines if a command is a 'help' command; a help command is
determined by the fact that it is NOT runnable/hidden/deprecated, and has no
sub commands that are runnable/hidden/deprecated.
```
In the docs it says, what help is not runnable, but if you look in `initHelpCmd` definition
```go

func (c *Command) initHelpCmd() {
	if c.helpCommand == nil {
		if !c.HasSubCommands() {
			return
		}

		c.helpCommand = &Command{
			Use:   "help [command]",
			Short: "Help about any command",
			Long: `Help provides help for any command in the application.
    Simply type ` + c.Name() + ` help [path to command] for full details.`,
			PersistentPreRun:  func(cmd *Command, args []string) {},
			PersistentPostRun: func(cmd *Command, args []string) {},

			Run: func(c *Command, args []string) {
				cmd, _, e := c.Root().Find(args)
				if cmd == nil || e != nil {
					c.Printf("Unknown help topic %#q.", args)
					c.Root().Usage()
				} else {
					cmd.Help()
				}
			},
		}
	}
	c.AddCommand(c.helpCommand)
}
```
, you can see, what built-in help command has Run function.
So the actual logic of IsHelpCommand is wrong. I think, the possible solution could be:
```go
func (c *Command) IsHelpCommand() bool {
    helpCmd := c.findHelpCommand()
    if c == helpCmd{
        return true
    }
    return false
}

func (c Command) findHelpCommand() *Command {
    if c.helpCommand == nil {
        if c.HasParent() {
             return c.Parent().findHelpCommand()
        }
    }
    return c.helpCommand
}
```

bogem:
Fixes #366

### DIFF
--- a/command.go
+++ b/command.go
@@ -155,12 +155,12 @@ func (c *Command) SetUsageTemplate(s string) {
 }
 
 // SetFlagErrorFunc sets a function to generate an error when flag parsing
-// fails
+// fails.
 func (c *Command) SetFlagErrorFunc(f func(*Command, error) error) {
 	c.flagErrorFunc = f
 }
 
-// SetHelpFunc sets help function. Can be defined by Application
+// SetHelpFunc sets help function. Can be defined by Application.
 func (c *Command) SetHelpFunc(f func(*Command, []string)) {
 	c.helpFunc = f
 }
@@ -187,7 +187,7 @@ func (c *Command) SetGlobalNormalizationFunc(n func(f *flag.FlagSet, name string
 	}
 }
 
-// OutOrStdout returns output to stdout
+// OutOrStdout returns output to stdout.
 func (c *Command) OutOrStdout() io.Writer {
 	return c.getOut(os.Stdout)
 }
--- a/command.go
+++ b/command.go
@@ -698,7 +698,6 @@ func (c *Command) Execute() error {
 
 // ExecuteC executes the command.
 func (c *Command) ExecuteC() (cmd *Command, err error) {
-
 	// Regardless of what command execute is called on, run on Root only
 	if c.HasParent() {
 		return c.Root().ExecuteC()
@@ -783,7 +782,7 @@ func (c *Command) initHelpCmd() {
 			Run: func(c *Command, args []string) {
 				cmd, _, e := c.Root().Find(args)
 				if cmd == nil || e != nil {
-					c.Printf("Unknown help topic %#q.", args)
+					c.Printf("Unknown help topic %#q\n", args)
 					c.Root().Usage()
 				} else {
 					cmd.Help()
--- a/command.go
+++ b/command.go
@@ -345,19 +345,19 @@ Aliases:
 {{end}}{{if .HasExample}}
 
 Examples:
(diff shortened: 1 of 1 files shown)


eparis:
lets talk in the #393 issue...  I don't think this is right, but I agree it is terribly confusing.

bogem:
Changed PR so it fixes only #366.
Fix of #393 is discussed ibid

command.go
```diff
@@ -783,7 +782,8 @@ func (c *Command) initHelpCmd() {
 			Run: func(c *Command, args []string) {
 				cmd, _, e := c.Root().Find(args)
 				if cmd == nil || e != nil {
-					c.Printf("Unknown help topic %#q.", args)
+					c.Printf("Unknown help topic %#q", args)
```
eparis: Can't we just use `"Unknown help topic %%q\n"`?

bogem: Yes, fixed

command.go
```diff
@@ -783,7 +782,7 @@ func (c *Command) initHelpCmd() {
 			Run: func(c *Command, args []string) {
 				cmd, _, e := c.Root().Find(args)
 				if cmd == nil || e != nil {
-					c.Printf("Unknown help topic %#q.", args)
+					c.Printf("Unknown help topic %%q\n", args)
```
eparis: you just went from `%#q` to `%%q` which doesn't seem to be right.

bogem: Now everything seems work

command.go
```diff
@@ -783,7 +782,7 @@ func (c *Command) initHelpCmd() {
 			Run: func(c *Command, args []string) {
 				cmd, _, e := c.Root().Find(args)
 				if cmd == nil || e != nil {
-					c.Printf("Unknown help topic %#q.", args)
+					c.Printf("Unknown help topic %q\n", args)
```
eparis: Why the change from `%#q` to   `%q` ? Not that I have a strong opion about `"` quoting vs `` ` `` quoting. But it seems like a needless change...

bogem: But you said in the first review about it.

bogem: 😅

eparis: Blech, I made a mistype in the first review. I wanted you to use `\n` and I mistyped `%%q\n`. Which now I realize is why you used `%%`.

How about `%#q\n`?  That work?

bogem: Yes, it works. 
How it looks like in my CLI:

PR #398, opened 2017-03-09
Rename IsHelpCommand to IsAdditionalHelpTopicCommand

### LINKED ISSUE #393: Misbehaviour of IsHelpCommand()

bogem:
Fixes #393

Created this PR just to end the discussion of #393

### DIFF
--- a/command.go
+++ b/command.go
@@ -356,7 +356,7 @@ Flags:
 Global Flags:
 {{.InheritedFlags.FlagUsages | trimRightSpace}}{{end}}{{if .HasHelpSubCommands}}
 
-Additional help topics:{{range .Commands}}{{if .IsHelpCommand}}
+Additional help topics:{{range .Commands}}{{if .IsAdditionalHelpTopicCommand}}
   {{rpad .CommandPath .CommandPathPadding}} {{.Short}}{{end}}{{end}}{{end}}{{if .HasAvailableSubCommands}}
 
 Use "{{.CommandPath}} [command] --help" for more information about a command.{{end}}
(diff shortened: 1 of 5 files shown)


eparis:
Its an API break but I'm going to merge (once CI passes). The only reason the function is public at all is because it has to be in order to be used in the template. If anyone complains we can re-add the old bad name.

PR #399, opened 2017-03-14
remove help command before adding it

tamird:
This fixes an issue where each Execute call grows the number of times
`help` appears in the help command by 1.

### DIFF
--- a/command.go
+++ b/command.go
@@ -788,6 +788,7 @@ func (c *Command) initHelpCmd() {
 			},
 		}
 	}
+	c.RemoveCommand(c.helpCommand)
 	c.AddCommand(c.helpCommand)
 }
 


tamird:
This was uncovered by https://github.com/spf13/cobra/pull/394/commits/f8b7358055ade2b35aa247066449c74215c05340 - the help command was not previously listed.

command.go
```diff
@@ -787,8 +787,8 @@ func (c *Command) initHelpCmd() {
 				}
 			},
 		}
+		c.AddCommand(c.helpCommand)
```
eparis: If I set a custom help command via `SetHelpCommand()` then `helpCommand` would be non nil but `AddCommand` would never have been called. Right?

tamird: Yeah, I think that's right. Pushed an update.

PR #402, opened 2017-03-24
Fix #401

### LINKED ISSUE #401: GenManTree and GenMarkdownTree require the output directory to exists. No error shown.
I was trying `GenManTree` and `GenMarkdownTree` unsuccessfully but without any error, until I realized that the directory in which the results are to be saved must exist prior to executing the binary. I think it would be helpful to make cobra either:

- Display an error message if the target directory does not exist.
- Create the target directory.

bogem:
* doc: Add and edit docs
* doc: Add tests

Fixes #401

### DIFF
--- a/doc/man_docs.go
+++ b/doc/man_docs.go
@@ -30,8 +30,8 @@ import (
 
 // GenManTree will generate a man page for this command and all descendants
 // in the directory given. The header may be nil. This function may not work
-// correctly if your command names have - in them. If you have `cmd` with two
-// subcmds, `sub` and `sub-third`. And `sub` has a subcommand called `third`
+// correctly if your command names have `-` in them. If you have `cmd` with two
+// subcmds, `sub` and `sub-third`, and `sub` has a subcommand called `third`
 // it is undefined which help output will be in the file `cmd-sub-third.1`.
 func GenManTree(cmd *cobra.Command, header *GenManHeader, dir string) error {
 	return GenManTreeFromOpts(cmd, GenManTreeOptions{
--- a/doc/man_docs_test.go
+++ b/doc/man_docs_test.go
@@ -13,9 +13,6 @@ import (
 	"github.com/spf13/cobra"
 )
 
-var _ = fmt.Println
-var _ = os.Stderr
-
 func translate(in string) string {
 	return strings.Replace(in, "-", "\\-", -1)
 }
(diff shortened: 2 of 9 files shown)


doc/man_docs.go
```diff
@@ -65,6 +65,12 @@ func GenManTreeFromOpts(cmd *cobra.Command, opts GenManTreeOptions) error {
 	if opts.CommandSeparator != "" {
 		separator = opts.CommandSeparator
 	}
+
+	err := os.Mkdir(opts.Path, 0766)
```
eparis: I'm not entirely convinced I like the idea at all. I'm trying to think of over unix commands which make a non-existing directory. This means that a typo will go unnoticed and generate docs where the user won't look for them. Before it would have said that the dir didn't exist. The use could either fix their typo or create the dir....

doc/man_docs_test.go
```diff
@@ -151,10 +150,7 @@ func TestGenManTree(t *testing.T) {
 		Use: "do [OPTIONS] arg1 arg2",
 	}
 	header := &GenManHeader{Section: "2"}
-	tmpdir, err := ioutil.TempDir("", "test-gen-man-tree")
-	if err != nil {
-		t.Fatalf("Failed to create tempdir: %s", err.Error())
-	}
+	tmpdir := filepath.Join(os.TempDir(), "test-gen-man-tree")
```
eparis: was there anything wrong here, or you could just do it "better"

doc/md_docs.md
```diff
@@ -15,7 +17,10 @@ func main() {
 		Use:   "test",
 		Short: "my test program",
 	}
-	doc.GenMarkdownTree(cmd, "/tmp")
+	err := doc.GenMarkdownTree(cmd, "/tmp")
+	if err != nil {
+		log.Fatal(err)
```
eparis: Why `Fatalln()`above but `Fatal()` here?

doc/md_docs.md
```diff
@@ -40,7 +46,10 @@ import (
 
 func main() {
 	kubectl := cmd.NewKubectlCommand(cmdutil.NewFactory(nil), os.Stdin, ioutil.Discard, ioutil.Discard)
-	doc.GenMarkdownTree(kubectl, "./")
+	err := doc.GenMarkdownTree(kubectl, "./")
+	if err != nil {
+		log.Fatal(err)
```
eparis: same `Fatal` v `Fatalln`

eparis:
If this patch didn't create dirs but did everything else (including Fatal v Fatalln reconciliation) I'm merge it immediately.

With the dir creation I'm unconvinced.

bogem:
I thought a lot about the Mkdir here too. And you're right, none unix command creates dir, at least I can't name one.
Also which permissions should be right in this case?

PR #406, opened 2017-03-28
Use go get so missing dependencies are fetched

campoy:
If the developer is missing any dependencies, for instance `github.com/spf13/cobra/cobra`, `go install` will fail:

```bash
../../spf13/viper/util.go:29:2: cannot find package "gopkg.in/yaml.v2" in any of:
        /Users/campoy/src/golang.org/x/go/src/gopkg.in/yaml.v2 (from $GOROOT)
        /Users/campoy/src/gopkg.in/yaml.v2 (from $GOPATH)
```

While `go get github.com/spf13/cobra/cobra` will build the tool successfully.

### DIFF
--- a/README.md
+++ b/README.md
@@ -180,7 +180,7 @@ commands you want. It's the easiest way to incorporate Cobra into your applicati
 
 In order to use the cobra command, compile it using the following command:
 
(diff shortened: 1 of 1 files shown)


bogem:
Please sign a CLA and I will merge it

PR #422, opened 2017-04-23
Improve template mechanisms

bogem:
* Delete `Eq`, `Gt`, `appendIfNotPresent` and `trim` functions
* Add `[flags]` in UseLine
* Simplify other functions
* Simplify templates

⚠️   **Public API changes:** Delete public `Eq`, `Gt` functions.
*`Eq` is used nowhere, `Gt` was used only in template.* I looked in some projects that use cobra, and I found out, what (almost) no one uses them.

---
Minor performance improvement.
Benchmark for command with 4 flags and one child command:
```
benchmark                   old ns/op     new ns/op     delta
BenchmarkCmdUsageFunc-4     335860        319290        -4.93%

benchmark                   old allocs     new allocs     delta
BenchmarkCmdUsageFunc-4     562            543            -3.38%

benchmark                   old bytes     new bytes     delta
BenchmarkCmdUsageFunc-4     21623         21037         -2.71%
```

/cc @spf13 @eparis

### DIFF
--- a/command.go
+++ b/command.go
@@ -327,23 +327,23 @@ func (c *Command) UsageTemplate() string {
 		return c.parent.UsageTemplate()
 	}
 	return `Usage:{{if .Runnable}}
-  {{if .HasAvailableFlags}}{{appendIfNotPresent .UseLine "[flags]"}}{{else}}{{.UseLine}}{{end}}{{end}}{{if .HasAvailableSubCommands}}
-  {{ .CommandPath}} [command]{{end}}{{if gt .Aliases 0}}
+  {{.UseLine}}{{end}}{{if .HasAvailableSubCommands}}
+  {{.CommandPath}} [command]{{end}}{{if gt (len .Aliases) 0}}
 
 Aliases:
   {{.NameAndAliases}}{{end}}{{if .HasExample}}
 
 Examples:
-{{ .Example }}{{end}}{{if .HasAvailableSubCommands}}
+{{.Example}}{{end}}{{if .HasAvailableSubCommands}}
 
 Available Commands:{{range .Commands}}{{if (or .IsAvailableCommand (eq .Name "help"))}}
   {{rpad .Name .NamePadding }} {{.Short}}{{end}}{{end}}{{end}}{{if .HasAvailableLocalFlags}}
 
 Flags:
-{{.LocalFlags.FlagUsages | trimRightSpace}}{{end}}{{if .HasAvailableInheritedFlags}}
+{{.LocalFlags.FlagUsages | trimTrailingWhitespaces}}{{end}}{{if .HasAvailableInheritedFlags}}
 
 Global Flags:
-{{.InheritedFlags.FlagUsages | trimRightSpace}}{{end}}{{if .HasHelpSubCommands}}
+{{.InheritedFlags.FlagUsages | trimTrailingWhitespaces}}{{end}}{{if .HasHelpSubCommands}}
 
 Additional help topics:{{range .Commands}}{{if .IsAdditionalHelpTopicCommand}}
   {{rpad .CommandPath .CommandPathPadding}} {{.Short}}{{end}}{{end}}{{end}}{{if .HasAvailableSubCommands}}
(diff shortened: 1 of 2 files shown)


command.go
```diff
@@ -922,15 +922,14 @@ func (c *Command) DebugFlags() {
 
 // Name returns the command's name: the first word in the use line.
 func (c *Command) Name() string {
-	if c.name != "" {
```
eparis: I'm good with the whole PR until here. Why do the entire work of the function inside a conditional? I much prefer early returns to blocks of code inside conditionals. But you are more active than me these days, if you strongly prefer this style fix I won't stand in your way.

bogem: I always do this way: if the field of struct is nil or blank, initialize it at first and then return it.

fabianofranz:
@bogem @eparis We should take better care in removing publicly exported functions (`Eq` and `Gt` in this case). `kubectl` in [Kubernetes](https://github.com/kubernetes/kubernetes/) uses it.

eparis:
sorry, will add them back.

eparis:
who knows who else we broke....

bogem:
@fabianofranz can you please show me the code, where these functions are used

fabianofranz:
@bogem we write our own templates, examples:

https://github.com/kubernetes/kubernetes/blob/master/pkg/kubectl/cmd/templates/templater.go#L102-L103
https://github.com/kubernetes/kubernetes/blob/master/pkg/kubectl/cmd/templates/templates.go#L34

bogem:
You can easily delete `eq` from `templateFuncs`, because it's used [nowhere](https://github.com/kubernetes/kubernetes/search?l=Go&q=eq&type=&utf8=%E2%9C%93) in kubernetes at all.

bogem:
`gt` is used only [here](https://github.com/kubernetes/kubernetes/blob/c19569f03f09eb4e8638d8f6a20b0a7a751abcc2/pkg/kubectl/cmd/templates/templates.go#L34), but you can rewrite it with standard function from text/template [in this way](https://github.com/spf13/cobra/blob/master/command.go#L368): `{{if gt (len .Aliases) 0}}Aliases:`.

fabianofranz:
Sure, fixing would have been simple. But we rely on [semver](http://semver.org/) and we also have users (e.g. OpenShift) which may be using them and we can't break anyone. API changes should only happen in major release.

bogem:
`(templater *templater) templateFuncs` is a private function, so changing that will not harm to API AFAIK. 
And the behaviour of `SectionAliases` will be fully same, if you will replace `gt .Aliases 0` with `gt (len .Aliases) 0`

fabianofranz:
Thanks but no. Kube cmd users could be calling `SetHelpTemplate` and `SetUsageTemplate`.

bogem:
Okay

command.go
dixudx: @eparis This change is breaking [Kubernetes](https://k8s-gubernator.appspot.com/build/kubernetes-jenkins/pr-logs/pull/53631/pull-kubernetes-bazel-test/9403/).

The `c.name` will not be override in [deprecatedAlias](https://github.com/kubernetes/kubernetes/blob/e57accbd03b1265340fdd266d0afe2b7a8e1c171/pkg/kubectl/cmd/cmd.go#L401-L414), since `c.name` will be assigned a value at the first calling and keeps unchanged. You can have a test using [Test_deprecatedAlias](https://github.com/kubernetes/kubernetes/blob/e57accbd03b1265340fdd266d0afe2b7a8e1c171/pkg/kubectl/cmd/cmd_test.go#L672-L720)

bogem: But I don't see any difference between the functionality of these two changes. In old version this function works the same: `c.name will be assigned a value at the first calling and keeps unchanged`.

bogem: I propose to delete `name` property of command and generate name on each call of Name() func.

dixudx: > But I don't see any difference between the functionality of these two changes. In old version this function works the same: `c.name will be assigned a value at the first calling and keeps unchanged`.

@bogem Right. I found the root cause is #390, another PR that made by me 😂.

> I propose to delete name property of command and generate name on each call of Name() func.

+1, since `name` is a private attribute and only used in this method. Seems useless. @eparis WDYT?

PR #418, opened 2017-04-20
cmd: Fix license issues

bogem:
Fixes #311
Fixes #318 
Fixes #322
Closes #356

/cc @MarkDBlackwell @eparis @spf13

PR #427, opened 2017-04-27
cmd: Licenses refactoring

bogem:
Notes from https://github.com/spf13/cobra/pull/418#discussion_r113724661.
Huge thanks @pombredanne for help and remarks.

/cc @pombredanne

### DIFF
--- a/cobra/cmd/license_agpl.go
+++ b/cobra/cmd/license_agpl.go
@@ -678,6 +678,7 @@ specific requirements.
   You should also get your employer (if you work as a programmer) or school,
 if any, to sign a "copyright disclaimer" for the program, if necessary.
 For more information on this, and how to apply and follow the GNU AGPL, see
-<http://www.gnu.org/licenses/>.`,
+<http://www.gnu.org/licenses/>.
+`,
 	}
 }
--- a/cobra/cmd/license_agpl.go
+++ b/cobra/cmd/license_agpl.go
@@ -4,21 +4,20 @@ func initAgpl() {
 	Licenses["agpl"] = License{
 		Name:            "GNU Affero General Public License",
 		PossibleMatches: []string{"agpl", "affero gpl", "gnu agpl"},
-		Header: `
-This file is part of {{ .appName }}.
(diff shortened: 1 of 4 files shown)


pombredanne:
awesome! :bowing_man:

bogem:
Is it ok to have copyright at the beginning of header?

bogem:
Yes, it's ok to have it:
https://www.gnu.org/licenses/gpl-3.0.html
```
 <one line to give the program's name and a brief idea of what it does.>
    Copyright (C) <year>  <name of author>

    This program is free software: you can redistribute it and/or modify
    it under the terms of the GNU General Public License as published by
    the Free Software Foundation, either version 3 of the License, or
    (at your option) any later version.

    This program is distributed in the hope that it will be useful,
    but WITHOUT ANY WARRANTY; without even the implied warranty of
    MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
    GNU General Public License for more details.

    You should have received a copy of the GNU General Public License
    along with this program.  If not, see <http://www.gnu.org/licenses/>.
```

bogem:
But!
```
If the program does terminal interaction, make it output a short notice like this when it starts in an interactive mode:

    <program>  Copyright (C) <year>  <name of author>
    This program comes with ABSOLUTELY NO WARRANTY; for details type `show w'.
    This is free software, and you are welcome to redistribute it
    under certain conditions; type `show c' for details.
The hypothetical commands `show w' and `show c' should show the appropriate parts of the General Public License. Of course, your program's commands might be different; for a GUI interface, you would use an “about box”.
```

@pombredanne should we really write it?? I never saw that.

pombredanne:
@bogem it is always OK to put a copyright and actually it is better. Now for a CLI notice the GPL-suggested are pretty weak... and not informative. It would be best to craft something that spells the license explicitly rather than just stating this is "free software" and IMHO using a single notice in all cases is probably a better choice.

bogem:
Ok
So I merge it

PR #437, opened 2017-05-09
Added a configuration file section to the README

deltaskelta:
Added a section to explain the relationship between viper and cobra, with a small example and links to viper docs


bogem:
Sorry, but I don't think you write clean for newcomers. The explanation should be more shorter and the title (`Reading Flags From a Config File`) is incorrect, because you can't read flags from config file.

bogem:
I will make myself PR to fix it.

PR #439, opened 2017-05-09
Partial revert of 4d2c4af 'Improve template mechanism'

eparis:
There were template functions which we defined and others started using.
Although we no longer want those functions, since others use them,
deleting them breaks our API.  Putting those (unused) functions back.

Fixes the report in https://github.com/spf13/cobra/pull/422#issuecomment-300263377

### DIFF
--- a/cobra.go
+++ b/cobra.go
@@ -19,14 +19,21 @@ package cobra
 import (
 	"fmt"
 	"io"
+	"reflect"
+	"strconv"
 	"strings"
 	"text/template"
 	"unicode"
 )
 
 var templateFuncs = template.FuncMap{
-	"trimTrailingWhitespaces": trimTrailingWhitespaces,
-	"rpad": rpad,
+	"trim":                    strings.TrimSpace,
+	"trimRightSpace":          trimRightSpace,
+	"trimTrailingWhitespaces": trimRightSpace,
+	"appendIfNotPresent":      appendIfNotPresent,
+	"rpad":                    rpad,
+	"gt":                      Gt,
+	"eq":                      Eq,
 }
 
 var initializers []func()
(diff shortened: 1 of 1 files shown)


eparis:
@bogem @fabianofranz

fabianofranz:
LGTM, also thanks for the FIXME notes.

eparis:
I'm going to self merge to unblock kubernetes.  @bogem we can talk about next steps if you disagree with what I'm doing here...

bogem:
We only should return Gt and Eq functions, if we want to return backward compatibility.
We don't need to restore all templateFuncs. `appendIfNotPresent` is even a private function, so other users can't use it.

eparis:
ok, will fix, but @fabianofranz you are free!!

fabianofranz:
Private funcs are also important since they can be used by people providing their templates through `SetHelpTemplate` and `SetUsageTemplate`.

bogem:
If they provide custom templates, they can provide their own functions by [AddTemplateFunc](https://godoc.org/github.com/spf13/cobra#AddTemplateFunc).

eparis:
But we don't break existing users, if someone could be using those functions via `Set*Template` we should not break them. (Until we decide to go to v2).

eparis:
Mind you I'm all for going to something we call v2. The inconsistency in handling of finding subcmds vs taking unnamed arguments has always bothered me a lot   :)

bogem:
If they want to use them, they can copy these functions and them by themselves. It's no problem.

eparis:
agreed, technically it is easy, fabiano would have created his own functions. But the contract we make with users is not to break their software unless we move to v2. If it worked yesterday, it must work tomorrow. Even if that means we are stuck with something subpar, as long as 1 person relied on it, we must keep it until we change the major version. (given we don't even have a 1.0 I know that 'major version' thing is hard to visualize)

bogem:
So what about a writing a roadmap to release v1.0?

bogem:
And we should consider #259 issue.

eparis:
Personally I think cobra works well enough that we could just call it 1.0 right now and start working on removing all the silly things like this PR...

bogem:
Then let's write a list of silly things, what we should remove.

eparis:
I can think of 2   :)   What's a good place? in #259?

bogem:
I think, we should make new issue for it.

PR #444, opened 2017-05-12
Update pflag link in README.

dpetersen:
Cobra is not using that version of pflag, but a fork of it.

It's a minor issue, but I ran into a bunch of dependency hell issues because I accidentally updated that one instead of the fork while trying to update cobra in a project.


bogem:
Thank you for this fix!

PR #445, opened 2017-05-13
Fix tests so they give correct args

bogem:
Consider these program:
```go
package main

import (
	"fmt"
	"os"
)

func main() {
	for _, a := range os.Args {
		fmt.Printf("%q ", a)
	}
}
```
While executing it as `go run main.go -i="qwe" "qwe qwe" -i=qwe` we become `"/path/to/temp/binary" "-i=qwe" "qwe qwe" "-i=qwe"`. Args are not split with space, as it is in tests. So we don't need to manually quotes. It makes shell.
`spf13/pflag` and standard `flag` libs are processing flags this way too.

### DIFF
--- a/cobra_test.go
+++ b/cobra_test.go
@@ -227,36 +227,36 @@ type resulter struct {
 	Command *Command
 }
 
-func fullSetupTest(input string) resulter {
+func fullSetupTest(args ...string) resulter {
 	c := initializeWithRootCmd()
 
-	return fullTester(c, input)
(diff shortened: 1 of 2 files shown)


eparis:
I've been trying to think of any reason that `"` handling code could be needed other than to support what we are doing in _test.go. You know I'm always reluctant to change anything that might break someone, but I really don't think anyone  could be relying on us doing these terrible contortions with `"`s...

PR #447, opened 2017-05-14
Simplify bash_completions.go

bogem:
Use buf instead of writing directly to w.
Improve test coverage from 80% to 85%.

### DIFF
--- a/bash_completions.go
+++ b/bash_completions.go
@@ -1,6 +1,7 @@
 package cobra
 
 import (
+	"bytes"
 	"fmt"
 	"io"
 	"os"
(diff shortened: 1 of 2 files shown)


eparis:
I don't love it. It definitely fixes some dumb things like: `b := (len(flag.NoOptDefVal) > 0; if !b {` (which I probably wrote and feel badly about looking back on it.

But most of this patch is just removing error handling code by storing everything in RAM instead of using the writer we were given. And eventually you have to have the data in RAM twice while you copy it from the buffer to the writer. Why is that an improvement? The testing % went up, but now, in a memory constrained environment we will panic in all the places we would have had an error.

It's not a nak but it is also not a LGTM from me. There are definitely a couple of things which are better in here. I'm neither going to block the merge nor click the button myself.

bogem:
> Why is that an improvement?

```
$ benchcmp before after
benchmark                     old ns/op     new ns/op     delta
BenchmarkBashCompletion-4     450778        120105        -73.36%

benchmark                     old allocs     new allocs     delta
BenchmarkBashCompletion-4     350            472            +34.86%

benchmark                     old bytes     new bytes     delta
BenchmarkBashCompletion-4     10877         36761         +237.97%
```
Yes, the RAM usage increased, but only from 11KB to 37KB. I don't think it's a great deal for PCs in 2017. But keep in mind, what performance increased in 3.7 times ...from 460μs to 120μs, not so much too.

> in  a memory constrained environment we will panic in all the places we would have had an error
I don't think, that someone's PC will not have ~20KB for this. 

The main thing is not in performance, but in this code style. I hate code, where there are `if _, err := fmt.Fprint(...); err != nil` in almost every line. With my changes code looks much better and more understandable. You don't have to consider these `if _, err := fmt.Fprint(...); err != nil` lines and we can focus more on more important aspects of bash completion generation.

bogem:
Ping @eparis

eparis:
I stick with my 'I don't love it'.  But I understand its appeal. Feel free to merge.

PR #459, opened 2017-06-05
__ltrim_colon_completions is not always available on macOS

lucymhdavies:
So bash-completion should check first

### DIFF
--- a/bash_completions.go
+++ b/bash_completions.go
@@ -133,7 +133,7 @@ __handle_reply()
     fi
 
     # available in bash-completion >= 2, not always present on macOS
-    if command -v __ltrim_colon_completions >/dev/null; then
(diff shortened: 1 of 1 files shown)


bash_completions.go
```diff
@@ -132,7 +132,10 @@ __handle_reply()
         declare -F __custom_func >/dev/null && __custom_func
     fi
 
-    __ltrim_colon_completions "$cur"
+    # available in bash-completion >= 2, not always present on macOS
+    if command -v __ltrim_colon_completions >/dev/null; then
```
eparis: Doesn't command only return true is there is a binary named `__ltrim_colon_completions`? This would fail on every system everywhere, wouldn't it? Do you mean `declare -F` ?

lucymhdavies: You would appear to be correct. :)

PR #461, opened 2017-06-08
README.md  #### Manually create rootCmd demo error

zh794390558:
func main(){} must be func initConfig(){}

PR #476, opened 2017-06-24
Rename main func to initConfig

johananl:
The function called `main` in the manual creation example should be called
`initConfig`.


bogem:
Duplicate of #461, but I will merge it because that PR has no progress.

PR #477, opened 2017-06-24
Print err instead of home

johananl:
When `home` can't be determined it doesn't make sense to print it.
The value of `err` should be printed instead.

### DIFF
--- a/README.md
+++ b/README.md
@@ -337,7 +337,7 @@ func main() {
 		// Find home directory.
 		home, err := homedir.Dir()
 		if err != nil {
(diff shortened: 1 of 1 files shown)


bogem:
I was so careless when I wrote this...
Thank you for your fixes!

johananl:
No problem @bogem :-)
I'm using the library heavily at the moment so these problems just pop up in front of me... I'll keep opening PRs if I find more issues if you don't mind.

bogem:
@johananl New PRs are only welcome!

Issue #412, opened 2017-04-04
Bug: Flags doesn't contain persistent flags


bogem:
As in the docs said:
```go
// Flags returns the complete FlagSet that applies
// to this command (local and persistent declared here and by all parents).
func (c *Command) Flags() *flag.FlagSet
```
But consider this script:
```go
package main

import (
	"fmt"

	"github.com/spf13/cobra"
	"github.com/spf13/pflag"
)

var parent = &cobra.Command{
	Run: func(*cobra.Command, []string) {
	},
}
var child = &cobra.Command{
	Run: func(*cobra.Command, []string) {
	},
}

func init() {
	parent.PersistentFlags().Bool("1st-persistent-parent-flag", true, "")
	parent.PersistentFlags().Bool("2st-persistent-parent-flag", true, "")
	child.Flags().Bool("1st-child-flag", true, "")
	child.Flags().Bool("2st-child-flag", true, "")
	parent.AddCommand(child)

	fmt.Println("child.Flags().VisitAll() before child.LocalFlags():")
	child.Flags().VisitAll(func(f *pflag.Flag) {
		fmt.Println(f.Name)
	})

	child.LocalFlags()

	fmt.Println()
	fmt.Println("child.Flags().VisitAll() after child.LocalFlags():")
	child.Flags().VisitAll(func(f *pflag.Flag) {
		fmt.Println(f.Name)
	})
}

func main() {
	fmt.Println()
	fmt.Printf("parent.Execute(): %v", parent.Execute())
}
```

Output:
```
child.Flags().VisitAll() before child.LocalFlags():
1st-child-flag
2st-child-flag

child.Flags().VisitAll() after child.LocalFlags():
1st-child-flag
1st-persistent-parent-flag
2st-child-flag
2st-persistent-parent-flag

parent.Execute(): <nil>
```
Only after the calling `child.LocalFlags()` user can see persistent flags in `child.Flags()`. I think, it's not obvious behaviour.

P.S. Instead of `child.LocalFlags()` you can use `child.InheritedFlags()` or another function that calls `c.mergePersistentFlags()` and you will see the same effect.

ghostsquad:
ya, these seems like very strange behavior.

bogem:
Maybe just call `mergePersistentFlags` in `VisitAll` 🤔

andig:
It seems the same (if I'm not mistaken) applies for `PersistentFlags`. A persistent flag should be available on the child commands too, but isn't in my tests.

PR #481, opened 2017-07-02
Issue #412: When AddCommands, update childs' parentsPflags

aimof:
I solve issue #412.

Before my PR, when AddCommands(), the children's parentsPFlags were not updated.
So there are some problems such as issue #412.

In my PR, update parentsPFlags when run AddCommands().

### DIFF
--- a/command.go
+++ b/command.go
@@ -844,6 +844,13 @@ func (c *Command) AddCommand(cmds ...*Command) {
 		}
 		c.commands = append(c.commands, x)
 		c.commandsAreSorted = false
+		if cmds[i].parentsPflags == nil {
+			cmds[i].parentsPflags = flag.NewFlagSet(cmds[i].Name(), flag.ContinueOnError)
+			cmds[i].parentsPflags.SetOutput(c.flagErrorBuf)
+			cmds[i].parentsPflags.SortFlags = false
+		}
+		cmds[i].parentsPflags.AddFlagSet(c.PersistentFlags())
+		cmds[i].Flags().AddFlagSet(cmds[i].parentsPflags)
 	}
 }
 


bogem:
Yes, it solves this issue, but creates one more and breaks the behaviour. Consider this code:
```go
package main

import (
	"github.com/spf13/cobra"
)

var c1 = &cobra.Command{Use: "c1"}
var c2 = &cobra.Command{Use: "c2"}

func init() {
	c1.PersistentFlags().Bool("boolFlag", true, "flag")
	c1.AddCommand(c2)

	c2.Flags().Bool("boolFlag", true, "flag")
}

func main() {}
```
Running this code we see a panic: `panic: c2 flag redefined: boolFlag`.

aimof:
I understood.

```go
var child = &cobra.Command{Use: "child"}
var parent = &cobra.Command{Use: "parent"}
```

1. We must never define child.flags in a same name before child.Flags().Bool()
2. We can run child.AddCommand(parent) before define child.flags
3. If we run child.AddCommand(parent), we must run child.mergePersistentFlags() or sth like this after child.AddCommand(parent).
4. We must run child.mergePersistentFlags() after both defining child.flags and running child.AddCommand(parent).

Simple Solution is define a function like this
```go
func (c *Command) MergePersistentFlags() {
    c.mergePersistentFlags()
}
```
and run this after definition of all child.flags and running child.AddCommand(parent) but it seems to be more difficult to use this library.

fastreles:
The persistence of "persistent" flags seems pretty integral to the library. The issues surrounding them, between them sometimes not being inherited, and sometimes being hijacked by siblings, risk making the whole cobra-pflag-viper ecosystem a bit of a liability for people looking to integrate them into major projects.

Once multiple people are reporting related bugs, and the lib doesn't reliably work as documented, I believe that satisfied the "until someone really need[s] it" condition.

chancez:
This is super annoying. Is this fixed or not, the bug was closed so it seems not?

Issue #479, opened 2017-06-29
Default config path contradicts the system default from windows and the xdg-definitions for linux


Fjolnir-Dvorak:
The configruation path of cobra generated files defaults to "$HOME" which is no place the configruation belongs to. For windows there is the environment variable %APPDATA% which is always set and for linux systems it should be best practive to use the $XDG_CONFIG_HOME environment variable which defaults to $HOME/.config and must not be set.

Reference with explanation: [StackOverflow](https://stackoverflow.com/questions/11113974/what-is-the-cross-platform-way-of-obtaining-the-path-to-the-local-application-da/39868933#39868933)
Reference to xda: [freedesktop.org env vars](https://specifications.freedesktop.org/basedir-spec/latest/ar01s03.html)

Fjolnir-Dvorak:
Found project to easily solve system-dependend config path's: [github.com/shibukawa/configdir](https://github.com/shibukawa/configdir). Licensed under MIT. Will create a merge-request tomorrow containing the fix.

Fjolnir-Dvorak:
I am not finding any System or Environment variables for Darwin/MacOS. I do not want to hard-code those strings. I will look a few days longer for a sollution.

bogem:
What do you think about backward compatibility?
Currently users have config in their home path, but after this change a lot of people will encounter a problem that their configs are not to be found.
I don't think this issue is a real problem for users.

Fjolnir-Dvorak:
It would not change anything for them. My change only changes the behaviour of new projects. The config path is defined when generating a new project using Cobra.
But every new project would have the correct config settings. And I think that the temptation is quite low to change the config path from $xdg_config_home to $home.

PR #485, opened 2017-07-07
Changed the default config location from  to the system-directory for…

Fjolnir-Dvorak:
This is my Request for the ticket [479](https://github.com/spf13/cobra/issues/479)

I changed the default config location for new generated projects to the system default configruation directory. In linux-systems this is $xdg_config_home and in Windows %APPDATA%. To help with that I added a library managing the system environment defined directories.


Fjolnir-Dvorak:
Is there a beautiful way to fulfill the build-server test? I could generate the new test file and test it on my own and replace the old test-documents. That does not seems like the way to do. It does not provide any information on any tested functionality and if that functionality is working, just plain-editing the test file...

bogem:
Sorry, but I can't understand which problems this PR solves.
You enforce the support of GNU standards, but actually not so many people really care about them. You is the only one, who made the issue about it, but without the description of problem do you have and what do you want from this PR.
If you want to have standard config paths, use it, but the cobra command generates just minimum boilerplate code to easily start the CLI developing. I don't think it's a good idea to make it more complex because of this issue.

Fjolnir-Dvorak:
I thought the whole point of cobra is to provide a GNU standard conform easy to use framework? And I do not understand the argument: Because other people are doing it wrong and do not care about, everyone should do it wrong and learn it wrong because there are libraries which I trust that they do it the way it is intended to be done.
And the pull request is not only intended for linux where people do not seem to care about standards of their system they are using (as it seems), but also for windows and mac.

bogem:
The whole point of cobra/cmd is to make a minimal boilerplate code for your future CLI project.
User should decide himself/herself, if he/she wants to support this XDG standard. And AFAIK XDG is not a part of GNU or POSIX standards:

>freedesktop.org (fd.o) is a project to work on interoperability and shared base technology for free software desktop environments for the X Window System (X11) on Linux and other Unix-like operating systems.
freedesktop.org was formerly known as the X Desktop Group, and the abbreviation "XDG" remains common in their work.

Correct me please, if I'm not right.


I don't think it's a good idea to make cobra more complex because of this issue.

Fjolnir-Dvorak:
And if it would be included via option-out flag like "cobra init --no-default-config-dir" or "--config-in-home"? Than it would be the decision for the user if he/she wants to support os independend standards and for linux XDG, or not. If only linux is the issue I could change that to opt-in and every other system will be the location where the config files belongs to.

anthonyfok:
I think @Fjolnir-Dvorak has a point here.

Some years back, I was part of the Linux Standardization Workgroup in China where we worked, among other things, with Linux Standard Base (LSB) and freedesktop.org standards, which are required for proper interoperability among various platforms.  freefesktop.org is important not only in Linux but also in much of the UNIX world including FreeBSD, NetBSD, OpenBSD, Solaris, etc.  So please don't think of it as a niche thing and dismiss it outright.

> The whole point of cobra/cmd is to make a minimal boilerplate code for your future CLI project.
> User should decide himself/herself, if he/she wants to support this XDG standard.

@Fjolnir-Dvorak's changes go beyond XDG.  It also covers Windows and macOS too.  It is a heck a lot of work for the user to implement all that from scratch.

As for the two opposite opinions:

> "I thought the whole point of cobra is to provide a GNU standard conform easy to use framework?"  
vs
> "The whole point of cobra/cmd is to make a minimal boilerplate code for your future CLI project."

The three of us are not the original author, so I think it is best to let @spf13 answer that instead of us having preconceived idea of what "the whole point of cobra/cmd" is.

So, @Fjolnir-Dvorak's proposal is quite interesting, to me at least, and one that is worthy of consideration.

BUT!  I understand @bogem's concern because @Fjolnir-Dvorak's current implementation throws out the current behaviour without giving the user a choice.  No choice?  No good.

> And if it would be included via option-out flag like "cobra init --no-default-config-dir" or "--config-in-home"?  ...  If only linux and XDG is the issue I could change that to opt-in and every other system will be the location where the config files belongs to.

That's much better!  Though I am initially thinking of an "opt-in" for _all_ systems, keeping the current behaviour as the default.  :wink:

But actually, for behavioural or user-facing changes like this, we should really let @spf13 to make the decision.

bogem:
It would be great to make it optional and actually agree with @anthonyfok that @spf13 should make a final decision.

Fjolnir-Dvorak:
What is the status? Updating the branch without any information if those system-standard enhancements are wished or not does not seem like something worth of doing.

amyadzuki:
I just want to say that this is POSIX (or actually FHS) compliance, not GNU-specific behavior, and there are lots of us that would appreciate this change, such as myself.

Fjolnir-Dvorak:
well, I can update the pull request this week.
I had actually abandoned this request into Oblivion.

Fjolnir-Dvorak:
ok, I will create a flag for that on "cobra init", backed by viper, default false.
I understand your point, but it feels a little bit wrong to make a standard opt-in and not opt-out. When I generated my first cobra-project I did not look into the params for "init". Well, it is the default configuration and the default configuration will make everything perfect as it is. And I would assume that a lot of developers wouldn't care about flags in that case, too.

anthonyfok:
> I understand your point, but it feels a little bit wrong _to make a standard opt-in and not opt-out_

Just so that we are clear, it is not about making **the one only** standard as an opt-in and not opt-out.

The fact of the matter is, there is not **one single** standard, but rather, **multiple** standards.

The $xdg_config_home is **great** standard, but it is NOT universal yet.  As @spf13 mentioned, when he wrote Cobra, it was **already** following the GNU and POSIX standard.  It may be old, but still a standard nonetheless.

So, the real issue here is not of "opt-in/opt-out", but of "choosing WHICH standard".

That brings me to a suggestion: Since most people run `cobra init` only once, in the very beginning of their new project, I doubt people would try to run `cobra init` in an automated script.  If everyone agrees that is the case, how about making `cobra init` interactive?  Upon running `cobra init`, present to the end user a list of config file locations (i.e. the various different standards) to choose from, with explanation.  That way, the end user wouldn't miss it.

And, if so desired, a flag can be added (if the end user really wanted to skip the interactive prompts), and again, not to "opt-in" to a single standard, but to "choose from multiple standards".

Fjolnir-Dvorak:
I would like the interactive mode for cobra where the user can add every needed information. That would make additions added in the future more visible even for those who are not new to cobra. I think I will open a new issue for that today evening.

Another Idea:
What would be about to split the cofiguration even more apart, that a user can use APPDATA or LOCALAPPDATA for windows and one of the many standards in linux? Currently if the user does not want to use the linux environ path, he/she has to handle all the other OS configuration paths by him/herself. That would make the initialization a little bit more complex, but then it is totally up to the user what to use.

Fjolnir-Dvorak:
I searched a bit for linux / unix standards which are written down. I only references to XDG.
Also the [linux foundation](http://refspecs.linuxfoundation.org/FHS_3.0/fhs/index.html) only mentioned [XDG](http://standards.freedesktop.org/basedir-spec/basedir-spec-latest.html) and [GLIB](https://web.archive.org/web/20160817034659/https://developer.gnome.org/glib/unstable/glib-Miscellaneous-Utility-Functions.html#g-get-user-config-dir)  (which is using XDG) in the Filesystem Hierarchy Standard.

Supporting Plan-9 style for linux/unix/gnu based systems feels kind of wrong because it is another whole operation system which is based on unix but is different in a lot of ways as far as I have researched (I did not find anything for unix explicitly, so I am assuming they are using the FHS standard as well).
Also I am not finding any written down standards for the filesystem hierarchy in plan9 so I would be gratefull of some literature so I can implement it.

What I did not implement was the difference between a local installation or a global installation on linux, windows and macOS. Speaking about different operration systems: How many operation systems does cobra officially support and which operation system configuration style should be assumed for systems which are not yet supported?

Fjolnir-Dvorak:
I rewrote  my project environ with support for windows, macOS and Linux. If any other system should be supported I would be glad about some resources / definitions

Fjolnir-Dvorak:
I no longer intend to wait on this pull request or to maintain it. I will close it for now. If someone wants to claim it feel free to do so.

PR #494, opened 2017-07-20
Add ability to get available sub commands

0x04C2:
Hi @bogem, @spf13 

I'm using cobra in our project and reading cobra project. I found that we don't have ability to retrieve available sub commands, but we have ability to determine that a command has available sub commands. Does it make senses to add this functionality in cobra?

Thanks

### DIFF
--- a/command.go
+++ b/command.go
@@ -1073,6 +1073,18 @@ func (c *Command) HasAvailableSubCommands() bool {
 	return false
 }
 
+// AvailableSubCommands returns available sub commands.
+func (c *Command) AvailableSubCommands() []*Command {
+	subCmds := []*Command{}
+
+	for _, sub := range c.commands {
+		if sub.IsAvailableCommand() {
+			subCmds = append(subCmds, sub)
+		}
+	}
+	return subCmds
+}
+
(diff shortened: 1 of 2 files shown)


bogem:
> we don't have ability to retrieve available sub commands

But you have:
```go
availableSubCmds := []*Command{}

for _, sub := range c.Commands() {
    if sub.IsAvailableCommand() {
        subCmds = append(subCmds, sub)
    }
}
```

0x04C2:
@bogem  Yes, we have and can use the code snippet in the project. But it seems DRY if we use the snippet in many projects. That's why that I make this PR. :)

bogem:
I don't think that this function will be used really often by other users, and anyway if user wants to get available subcommands, he/she can use the method I described above.
I can't merge PR and add functionality that is really needed only for you. Command struct has a lot of methods now (77) and we should keep it as simple as possible and don't add overhead methods.

bogem:
Thank you for your PR anyway! We are glad to see new PRs, but at first it would be great to open issues with explanation of problem you have ;)

0x04C2:
@bogem Yes! Totally agree with you. 👍

PR #496, opened 2017-07-24
Fix TestGlobalNormFuncPropagation() on gccgo

anthonyfok:
According to golang/go#9504 and https://golang.org/pkg/reflect/#Value,
== should not be used on two reflect.Values, but comparing the results
of their Interface() method does not work in this case, so let's
compare the results of their Pointer() method instead.

See https://stackoverflow.com/questions/9643205/how-do-i-compare-two-functions-for-pointer-equality-in-the-latest-go-weekly

### DIFF
--- a/cobra_test.go
+++ b/cobra_test.go
@@ -1166,7 +1166,7 @@ func TestGlobalNormFuncPropagation(t *testing.T) {
 
 	rootCmd := initialize()
 	rootCmd.SetGlobalNormalizationFunc(normFunc)
-	if reflect.ValueOf(normFunc) != reflect.ValueOf(rootCmd.GlobalNormalizationFunc()) {
+	if reflect.ValueOf(normFunc).Pointer() != reflect.ValueOf(rootCmd.GlobalNormalizationFunc()).Pointer() {
 		t.Error("rootCmd seems to have a wrong normalization function")
 	}
 


anthonyfok:
Background Information:

The issue was discovered by Debian's recent buildd run on s390x where gccgo is explicitly used for compatibility with older IBM z10 microprocessor which gccgo-7 can still support, whereas the main Go 1.8.3 targets z10's successor z196 with newer CPU instructions.

The build failed at:

```
=== RUN   TestGlobalNormFuncPropagation
--- FAIL: TestGlobalNormFuncPropagation (0.00s)
	cobra_test.go:1099: rootCmd seems to have a wrong normalization function
```

Full build log at: https://buildd.debian.org/status/fetch.php?pkg=golang-github-spf13-cobra&arch=s390x&ver=0.0~git20170716.0.d994347-1&stamp=1500733885&raw=0


And then I could reproduce the same on my amd64 machine, so it is not a s390x-specific issue:


```
~/go/src/github.com/spf13/cobra$ go version
go version go1.8.3 linux/amd64
~/go/src/github.com/spf13/cobra$ go test -v -run TestGlobalNormFuncPropagation
=== RUN   TestGlobalNormFuncPropagation
--- PASS: TestGlobalNormFuncPropagation (0.00s)
PASS
ok  	github.com/spf13/cobra	0.006s

~/go/src/github.com/spf13/cobra$ go-7 version
go version go1.8.1 gccgo (Debian 7.1.0-10) 7.1.0 linux/amd64
~/go/src/github.com/spf13/cobra$ go-7 test -v -run TestGlobalNormFuncPropagation
=== RUN   TestGlobalNormFuncPropagation
--- FAIL: TestGlobalNormFuncPropagation (0.00s)
	cobra_test.go:1170: rootCmd seems to have a wrong normalization function
FAIL
exit status 1
FAIL	github.com/spf13/cobra	0.201s
```

Initially, I thought it were gccgo's fault, but it turns out that using == for comparing reflect.Values is not supposed to work, and there are even a proposal of making reflect.Value uncomparable in Go2, see golang/go#18871.

Thank you in advance for your review(s)!

bogem:
Thank you for your PR 👍

PR #497, opened 2017-07-24
Issue #107: add basic zsh completion (command hierarchy only)

bpicode:
Support for flags still missing though. Maybe we can go from here.

### DIFF
--- a/zsh_completions.go
+++ b/zsh_completions.go
@@ -2,8 +2,8 @@ package cobra
 
 import (
 	"bytes"
-	"io"
 	"fmt"
+	"io"
 	"strings"
 )
 
--- a/zsh_completions.go
+++ b/zsh_completions.go
@@ -1,0 +1,123 @@
+package cobra
+
+import (
+	"bytes"
+	"io"
+	"fmt"
+	"strings"
+)
+
+type fWriter struct {
+	io.Writer
+}
+
+func (fw *fWriter) fWriteLn(format string, a ...interface{}) (int, error) {
+	return io.WriteString(fw, fmt.Sprintf(format+"\n", a...))
+}
+
+// GenZshCompletion generates a zsh completion file and writes to the passed writer.
+func (cmd *Command) GenZshCompletion(w io.Writer) error {
+	buf := new(bytes.Buffer)
+	fw := &fWriter{buf}
+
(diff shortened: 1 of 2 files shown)


zsh_completions.go
```diff
@@ -0,0 +1,123 @@
...
+	"io"
+	"strings"
+)
+
+type fWriter struct {
+	io.Writer
+}
+
+func (fw *fWriter) fWriteLn(format string, a ...interface{}) (int, error) {
+	return io.WriteString(fw, fmt.Sprintf(format+"\n", a...))
```
bogem: I don't like this fWriter struct.
Why don't you use just `fmt.Fprintln(w, format, a1, a2)` like in bash_completions.go? It's simpler and nicer

zsh_completions.go
```diff
@@ -0,0 +1,123 @@
...
+
+func writeLevel(fw *fWriter, root *Command, i int) {
+	fw.fWriteLn(fmt.Sprintf("    case $words[%d] in", i))
+	defer fw.fWriteLn("    esac")
+
+	commands := filterByLevel(root, i)
+	byParent := groupByParent(commands)
+
+	for p, c := range byParent {
+		names := names(c)
```
bogem: variable `names` defined but not used.

bogem:
@eparis @anthonyfok

zsh_completions.go
bpicode: Yeah sure, I can change to fmt.Fprintln, I don't mind.

zsh_completions.go
bpicode: It is used two lines below?

bogem: Oh sorry, I didn't notice it.

bogem:
LGTM but I know nothing about zsh syntax and how it works.
So I will wait to merge till someone, who knows something about zsh, reviews it and says the conclusion.

bpicode:
I'm also far from being a zsh expert ;)

nono:
LGTM

I'm a zsh user that has written some zsh syntax files, but I'm far from being an expert. This PR adds basic completion (as stated in the title of this PR). I think it is a good start for having a complete zsh completion support in cobra.

Issue #206, opened 2015-12-06
Question: how to make flags required


tleyden:
I'm defining some flags like this:

```
    follow_sync_gwCmd.PersistentFlags().String("url", "", "Sync Gateway URL")
    follow_sync_gwCmd.MarkPersistentFlagRequired("url")
```

and even explicitly calling `ParseFlags` (out of desperation):

```
    Run: func(cmd *cobra.Command, args []string) {

        if err := cmd.ParseFlags(args); err != nil {
            log.Printf("err: %v", err)
        }
                ...
    },
```

and calling:

```
$ go run main.go follow_sync_gw
```

with no `--url` flag, but cobra isn't throwing errors.

I was expecting that if I set `MarkPersistentFlagRequired` but failed to pass the flag on the command line then the command would fail with an error and a usage example.

apriendeau:
I am not sure that is implemented as it uses, [pflag](https://godoc.org/github.com/spf13/pflag)

So there is no direct way unless you do it in `PreRun` function with an `if` or `switch` but I could be wrong but that is how I have handled it in the past.

eparis:
Really its just a bad name since it was never implemented.  It will cause bash completions and things to always suggest that flag, but we aren't enforcing it.  You doing it in a PreRunE is a good idea.

tleyden:
I didn't know about the PreRun function, I'll look into that.  Yeah, the name is misleading .. so does that mean calling `MarkPersistentFlagRequired` has no effect whatsoever?

apriendeau:
@tleyden use `PreRunE` :+1:

eparis:
At this time its only effect is on the bash completions   :-(

I will try to write a PreRunE which enforces it....

ApsOps:
`MarkFlagRequired` and `MarkPersistentFlagRequired` should enforce the flag IMO. Or the function name should be different at least. I just wasted an hour wondering why it's not working :/

crux:
just a reminder, if it is required it is an parameter, or argument. If it is and option or flag it is NOT required. That is the very nature of flags and options. Every command line library supporting "required" or mandatory options/flags I would considered broken by design.

davidcpell:
@crux like the `heroku` toolbelt which which requires the name of the application you're dealing with to be passed to the `-a` flag?

tsupertramp:
[...]

PR #502, opened 2017-07-27
enforce required flags

dixudx:
Currently when we set/mark a flag as required, `cobra` will not raise an error if unset.

This PR does enforce checking the required flags.

/cc @eparis

### DIFF
--- a/command.go
+++ b/command.go
@@ -693,6 +693,9 @@ func (c *Command) execute(a []string) (err error) {
 		c.PreRun(c, argWoFlags)
 	}
 
+	if err := c.validateRequiredFlags(); err != nil {
+		return err
+	}
 	if c.RunE != nil {
 		if err := c.RunE(c, argWoFlags); err != nil {
 			return err
@@ -808,6 +811,25 @@ func (c *Command) ValidateArgs(args []string) error {
 		return nil
 	}
 	return c.Args(c, args)
+}
+
+func (c *Command) validateRequiredFlags() error {
+	flags := c.Flags()
+	missingFlagNames := []string{}
+	flags.VisitAll(func(pflag *flag.Flag) {
+		requiredAnnotation, found := pflag.Annotations[BashCompOneRequiredFlag]
+		if !found {
+			return
+		}
+		if (requiredAnnotation[0] == "true") && !pflag.Changed {
+			missingFlagNames = append(missingFlagNames, pflag.Name)
+		}
+	})
+
+	if len(missingFlagNames) > 0 {
+		return fmt.Errorf(`Required flag(s) "%s" have/has not been set`, strings.Join(missingFlagNames, `", "`))
+	}
+	return nil
 }
 
 // InitDefaultHelpFlag adds default help flag to c.
--- a/command_test.go
+++ b/command_test.go
@@ -438,3 +438,51 @@ func TestTraverseWithBadChildFlag(t *testing.T) {
 		t.Fatalf("wrong command %q expected %q", c.Name(), sub.Name())
 	}
 }
+
+func TestRequiredFlags(t *testing.T) {
+	c := &Command{Use: "c", Run: func(*Command, []string) {}}
+	output := new(bytes.Buffer)
+	c.SetOutput(output)
+	c.Flags().String("foo1", "", "required foo1")
+	c.MarkFlagRequired("foo1")
+	c.Flags().String("foo2", "", "required foo2")
+	c.MarkFlagRequired("foo2")
+	c.Flags().String("bar", "", "optional bar")
+
+	expected := fmt.Sprintf("Required flag(s) %q, %q have/has not been set", "foo1", "foo2")
+
+	if err := c.Execute(); err != nil {
+		if err.Error() != expected {
+			t.Errorf("expected %v, got %v", expected, err.Error())
+		}
+	}
+}
+
+func TestPersistentRequiredFlags(t *testing.T) {
+	parent := &Command{Use: "parent", Run: func(*Command, []string) {}}
+	output := new(bytes.Buffer)
+	parent.SetOutput(output)
+	parent.PersistentFlags().String("foo1", "", "required foo1")
+	parent.MarkPersistentFlagRequired("foo1")
+	parent.PersistentFlags().String("foo2", "", "required foo2")
+	parent.MarkPersistentFlagRequired("foo2")
+	parent.Flags().String("foo3", "", "optional foo3")
+
+	child := &Command{Use: "child", Run: func(*Command, []string) {}}
+	child.Flags().String("bar1", "", "required bar1")
+	child.MarkFlagRequired("bar1")
+	child.Flags().String("bar2", "", "required bar2")
+	child.MarkFlagRequired("bar2")
+	child.Flags().String("bar3", "", "optional bar3")
+
+	parent.AddCommand(child)
+	parent.SetArgs([]string{"child"})
+
+	expected := fmt.Sprintf("Required flag(s) %q, %q, %q, %q have/has not been set", "bar1", "bar2", "foo1", "foo2")
+
+	if err := parent.Execute(); err != nil {
+		if err.Error() != expected {
+			t.Errorf("expected %v, got %v", expected, err.Error())
+		}
+	}
+}


command.go
```diff
@@ -756,6 +759,25 @@ func (c *Command) ValidateArgs(args []string) error {
...
+func (c *Command) ValidateRequiredFlags() error {
+	flags := c.Flags()
+	missingFlagNames := []string{}
+	flags.VisitAll(func(pflag *flag.Flag) {
+		requiredAnnotation, found := pflag.Annotations[BashCompOneRequiredFlag]
+		if !found {
+			return
+		}
+		if (requiredAnnotation[0] == "true") && !pflag.Changed {
+			missingFlagNames = append(missingFlagNames, pflag.Name)
```
bogem: Why not just return the single error immediately by finding the first not provided flag?

dixudx: @bogem I think explicitly returning all the missing required flags are convenient for users instead of one by one.

dixudx:
@bogem Add another test for persistent required flags. PTAL. Thanks.

command_test.go
```diff
@@ -347,3 +347,53 @@ func TestSetHelpCommand(t *testing.T) {
...
+	output := new(bytes.Buffer)
+	c.SetOutput(output)
+	c.Flags().String("foo1", "", "required foo1")
+	c.MarkFlagRequired("foo1")
+	c.Flags().String("foo2", "", "required foo2")
+	c.MarkFlagRequired("foo2")
+	c.Flags().String("bar", "", "optional bar")
+
+	expectedFmt := "Required flags %s have/has not been set"
+	expected := fmt.Sprintf(expectedFmt, fmt.Sprintf("%q, %q", "foo1", "foo2"))
```
bogem: I think, it would be simpler to write `expected := fmt.Sprintf(Required flags %q, %q have/has not been set, "foo1", "foo2")`

command_test.go
```diff
@@ -347,3 +347,53 @@ func TestSetHelpCommand(t *testing.T) {
...
+	child.MarkFlagRequired("bar1")
+	child.Flags().String("bar2", "", "required bar2")
+	child.MarkFlagRequired("bar2")
+	child.Flags().String("bar3", "", "optional bar3")
+
+	parent.AddCommand(child)
+	parent.SetArgs([]string{"child"})
+
+	expectedFmt := "Required flags %s have/has not been set"
+	expected := fmt.Sprintf(expectedFmt, fmt.Sprintf("%q, %q, %q, %q", "bar1", "bar2", "foo1", "foo2"))
```
bogem: Same as above

command.go
```diff
@@ -756,6 +759,25 @@ func (c *Command) ValidateArgs(args []string) error {
...
+		}
+	})
+
+	if len(missingFlagNames) > 0 {
+		return fmt.Errorf("Required flags \"%s\" have/has not been set", strings.Join(missingFlagNames, "\", \""))
```
bogem: Please change `\"%s\"` to `%q`.

bogem: Please change to `Required flag(s)`.

dixudx: @bogem If changed to `%q`, it's hard to generate desired string `"foo", "bar"`.

dixudx: The generated string will be `"foo\", \"bar"`

bogem: Ok, I'm not right

bogem: Or just write `return fmt.Errorf(`Required flags "%s" have/has not been set`, strings.Join(missingFlagNames, `", "`))`. Backslashes look ugly :(

dixudx:
@bogem Updated. PTAL. Thanks.

bogem:
Cool. LGTM
@eparis @anthonyfok

dixudx:
@eparis @anthonyfok PTAL. Need your approval to get it merged. Thanks.

agonzalezro:
I wonder if this is going to work with the flags that are set with an environment variable. In my case I had to add a small "hacky" way to check this:

```
// envVarIsSet return true if there is an  environment variable set for this flag.
//
// The variable name is uppercased to match with viper automatic environment variables,
// however, this is a little bit hacky since those vars can be prefixed. It is good for
// us since we don't prefix them, but I thought writting a not would be worthy.
func envVarIsSet(name string) bool {
	return os.Getenv(strings.ToUpper(name)) != ""
}

func checkRequiredFlags(cmd *cobra.Command, _ []string) error {
	var requiredFlagsNotFound []string
	cmd.Flags().VisitAll(func(flag *pflag.Flag) {
		required := len(flag.Annotations[cobra.BashCompOneRequiredFlag]) > 0 && flag.Annotations[cobra.BashCompOneRequiredFlag][0] == "true"
		if required && (!flag.Changed && !envVarIsSet(flag.Name)) {
			requiredFlagsNotFound = append(requiredFlagsNotFound, flag.Name)
		}
	})
	if len(requiredFlagsNotFound) > 0 {
		return fmt.Errorf("The following flags are required and were not present: %s", strings.Join(requiredFlagsNotFound, ", "))
	}
	return nil
}
```

dixudx:
@agonzalezro Not supporting environment variables.

Also `cobra` does not support environment variables as flags currently. There is no mappings between the required flags and the environment variables. You're just assuming they have the same name.

agonzalezro:
@dixudx the env vars are from `viper.AutomaticEnv()`. I know that I am doing that assumption, hence the comment I added, however, it's important for me not to fail if the flag has a value set with an environment variable.

Sadly there isn't a public way to get the environment variable name that would be generated: https://github.com/spf13/viper/blob/d8f2aa78d42b618f51e668d0b7e05fe203f889de/viper.go#L366

command.go
```diff
@@ -756,6 +759,25 @@ func (c *Command) ValidateArgs(args []string) error {
 	return c.Args(c, args)
 }
 
+func (c *Command) ValidateRequiredFlags() error {
```
eparis: Does this need to be a public function? Do we need to extend the API of Cobra for this?

I believe there are times when 1 of 2 flags might be required. But not both. I'd have to look at the bash that is generated. This would force both of them to be present to even run. Maybe we should we not re-use BashCompOneRequiredFlag. Should it be a first class property of some object, or should we just have another name to make 'no no, it really does need to be set' (which could also be used, in turn, by the bash completion code that only requires 1 of the set)?

dixudx: > Does this need to be a public function?

Not necessary. It can be private.

> Do we need to extend the API of Cobra for this?

@eparis  No. Since we have marked a param as required, we should enforce this validation as default, right? For the API, nothing need to be changed or updated.

> I believe there are times when 1 of 2 flags might be required. But not both. I'd have to look at the bash that is generated.

Well, but this will not help enforcing the required flag. The "required" only exists in the help message. If so, why not just adding this to the description message. They did the same thing.

When marking a flag as required, users/developers will assume that `cobra` will help validate such required flags. If unset, errors should be raised. And this is exactly what #206 wants. WDYT?

> Maybe we should we not re-use BashCompOneRequiredFlag. Should it be a first class property of some object, or should we just have another name to make

Keep open on this. I don't think we need to introduce a new const value. Maybe we can rename it. But this is not safe, since other project may reuse current `BashCompOneRequiredFlag`. Just keep it as it is. WDYT?

eparis: Let's just make this function private.

The open question is about the changing behavior. Consider a program that had:
```
cmd.Flags().Int("need-flag-1", ...)
cmd.Flags().MarkFlagRequired("need-flag-1")
cmd.Flags().Int("or-flag-2", ...)
cmd.Flags().MarkFlagRequired("or-flag-2")
```
Where the expectation is that you must set either `--need-flag-1` or you must set `--or-flag-2`. I believed kube had such a place, but I cannot find it. There appears to be one place in kube where it has `need-flag-1` AND `need-flag-2`.  `kubectl create secret docker-registry` requires both `--username` and `--password`. AND is the behavior this proposed code is correct for. Which means at least kube will be fine with this change.

However this is NOT the behavior that the bash completions implement today. They implement an OR relationship, not an AND relationship. See: https://github.com/spf13/cobra/blob/master/bash_completions.go#L169 This worries me. I don't really mind the bash being wrong, it won't break the application, but I'm concern anyone who expected the annotation to work like the bash completions do could be broken with your change. Maybe we could/should try to change the bash completions behavior to match this behavior. But that's a different issue.


Anyone who called a function called `Mark*FlagRequired()` and didn't actually want it to be required is a fool and I'm ok if we break them. Instead of worrying about those people I decided to search for anyone who used `BashCompOneRequiredFlag` directly and could have a problem with this change. That resulted in [8 hits](https://github.com/search?utf8=%E2%9C%93&q=BashCompOneRequiredFlag+-filename%3Abash_completions.go+in%3Acode+language%3Agolang&type=Code).

6 of those 8 implemented ~this function. 1 of those used the annotation for something which I don't quite understand, but this change will not break them.

[1 place](https://github.com/organicelement/ipvanish/blob/e736120447cf1013b0293d71b1fe43f13916860c/commands/ipvanish.go#L65) set the annotation on a flag directly and I'm not convinced this change would be safe/compatible for them.

@ccustine I believe you wrote/own ipvanish and since you are the only user of `BashCompOneRequiredFlag` I'm worried this change might break would you be willing to sign off here? Would this break you? Are you willing to update to accommodate this change?

dixudx: > Where the expectation is that you must set either --need-flag-1 or you must set --or-flag-2.

> AND is the behavior this proposed code is correct for.

Yes, all required flags that marked are needed. An `AND` operator.

> 6 of those 8 implemented ~this function. 1 of those used the annotation for something which I don't quite understand, but this change will not break them.

Thanks @eparis for the result. This helps us know this change is useful. It is convincible.

dixudx:
ping @eparis Updated. PTAL. Thanks.

eparis:
why not. lets go.

czerasz:
Below a usage example:

```go
ScanCmd.Flags().StringP("path", "p", "", "sd card path")
ScanCmd.MarkFlagRequired("path")
```

PR #513, opened 2017-08-07
First pass at implementing lolgopher for truecolor terminals

kris-nova:
Adding `--fab` to cobra

## Warning

Adding a dependency on `lolgopher` implicitly adds a dependency on the `ncurses` C package: https://github.com/kris-nova/lolgopher/blob/master/has_colors.go#L4

### DIFF
--- a/cobra/cmd/root.go
+++ b/cobra/cmd/root.go
@@ -15,7 +15,9 @@ package cmd
 
 import (
 	"fmt"
+	"os"
 
+	lol "github.com/kris-nova/lolgopher"
 	homedir "github.com/mitchellh/go-homedir"
 	"github.com/spf13/cobra"
 	"github.com/spf13/viper"
@@ -24,6 +26,7 @@ import (
 var (
 	// Used for flags.
 	cfgFile, userLicense string
+	fab                  bool
 
 	rootCmd = &cobra.Command{
 		Use:   "cobra",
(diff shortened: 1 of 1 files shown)


cobra/cmd/root.go
```diff
@@ -24,13 +26,20 @@ import (
...
 
 	rootCmd = &cobra.Command{
 		Use:   "cobra",
 		Short: "A generator for Cobra based Applications",
 		Long: `Cobra is a CLI library for Go that empowers applications.
 This application is a tool to generate the needed files
 to quickly create a Cobra application.`,
+		Run: func(command *cobra.Command, args []string) {
+			if fab {
+				command.SetOutput(&lol.Writer{Output: os.Stdout, ColorMode: lol.ColorModeTrueColor})
```
kris-nova: This defaults to `truecolor` which render well in *every* terminal.. the safest bet is to go with 256, but the coloring in that case isn't quite as smooth/rainbowy

Safe solution: `256`
Less safe but potentially prettier solution: `truecolor`

Thoughts?

bogem:
Hi @kris-nova!
Thanks for this PR, but I think it's a huge overhead adding `ncurses` dependency and new flag for such a funny but non-useful thing.
cobra/cmd should stay tiny and does only necessary things.

kris-nova:
@bogem we are working on getting rid of the `ncurses` dependency - would that change your mind?

kris-nova:
Dependency removed here: https://github.com/kris-nova/lolgopher/pull/14

bogem:
Anyway it adds a third-party dependency and this PR doesn't add some useful features. As I said,
cobra/cmd should stay simple and do only necessary things.

Issue #407, opened 2017-03-29
Bash error when hitting tab after a required command


dtmistry:
I've a root command without flags and a child command with some flags. Only one of the flag is required for the child command. 

The generated completion script is installed on my Mac using homebrew. 

Autocomplete works fine up to the required flag, but fails with the below error when I hit tab after the flag (for autocompleting a file location for a flag)

```
$ clutch add-spec [tab][tab]   
--ctxRoot=    -c
```

`
$ clutch add-spec -c /test -d ~/workspace[tab]
`

`
$ clutch add-spec -c /test -d ~/workspace/-bash: flaghash[${flagname}]: bad array subscript
`

`
$ clutch add-spec --ctxRoot=/test -d ~/workspace[tab]
`

`
$ clutch add-spec --ctxRoot=/test -d ~/work-bash: --ctxRoot=: attempted assignment to non-variable (error token is "=")
`

The error persist with whatever order of flags.

johnmccabe:
This seems to be related to the version of bash on OSX and its handling of hashes, this fails for me:
```
./my-cli --gateway=jh bash: --gateway=: attempted assignment to non-variable (error token is "=")
```
Which is falling over here in the `__handle_flag()` method:
```
flaghash[${flagname}]=${flagvalue}
```
This fails with the default bash `GNU bash, version 3.2.57(1)-release (x86_64-apple-darwin16)`, but works if I use brew to install a more recent version `GNU bash, version 4.4.12(1)-release (x86_64-apple-darwin16.3.0)`.

johnmccabe:
@dtmistry I've raised a PR with a potential fix if you want to try that out.

bacongobbler:
looks like this can be closed now that #520 has been merged. :)

PR #520, opened 2017-08-25
Fix generated bash completion for Bash 3 (OSX)

johnmccabe:
This PR address the issue reported in #407, the `flaghash` associative array introduced in #205 is only supported on Bash 4, for Bash 3 the attempted assignment to the array results in the error returned in #407. The fix introduced in 801364a only addresses the initial declare.

The fix in this case is to only attempt to assign to the `flaghash` on Bash versions greater than 3.

This PR also updates the names of the functions in the preamble, currently all bash completions generated by cobra have preamble functions with the same name, meaning that only one particular version of a function will be available at any given time - depending on the order in which the completions are sourced.

The fix here is to prefix these functions with the name of the command for which the completion is generated.

Note that the error observed in #407 is also currently thrown by kubectl, this fixes that error - the `flaghash` seems to have been originally added in #205 for use with `kubectl` but I couldn't actually see it being used?

### DIFF
--- a/bash_completions.go
+++ b/bash_completions.go
@@ -21,8 +21,8 @@ const (
 
 func writePreamble(buf *bytes.Buffer, name string) {
 	buf.WriteString(fmt.Sprintf("# bash completion for %-36s -*- shell-script -*-\n", name))
-	buf.WriteString(`
-__debug()
+	buf.WriteString(fmt.Sprintf(`
+__%[1]s_debug()
 {
     if [[ -n ${BASH_COMP_DEBUG_FILE} ]]; then
         echo "$*" >> "${BASH_COMP_DEBUG_FILE}"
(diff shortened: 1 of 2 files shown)


johnmccabe:
fyi @nak3 / @anthonyfok

eparis:
Wow, so I just merged #628 which fixes 50% of this problem. If you redo this patch with just the preamble changes (or make a new PR with those) I will merge them.

johnmccabe:
@eparis I've rebased against master and dropped the Bash 3 fix (682d488d424c20d713dd8f8832e503e51334b2cc) which ended up being merged in #628

johnmccabe:
ping @eparis just in case this falls off the radar again

superbrothers:
This PR causes the below bug because it does not take subcommands into consideration.

```
kubectl create -f [Tab] failed with the following message: kubectl create -f __create_handle_filename_extension_flag: command not found
```

https://github.com/kubernetes/kubernetes/issues/60517

@johnmccabe @eparis It is needed either reverting or fixing the bug.

johnmccabe:
Having a look at it now @superbrothers

Issue #521, opened 2017-08-29
Improving consistency of flags when setting name normalization functions


Itizir:
Because cobra commands keep multiple `FlagSets` that share the underlying flags, one can end up with consistency problems when setting a `normalizeNameFunc` (if setting it after defining flags): the names in the underlying flags will be updated, but not all `FlagSets` that have references to them will know how to properly lookup the new name.

Example of bug:
```
package main

import (
	"strings"

	"github.com/spf13/cobra"
	"github.com/spf13/pflag"
)

func main() {
	cmd := &cobra.Command{}
	flagVal := false
	n := func(f *pflag.FlagSet, name string) pflag.NormalizedName {
		return pflag.NormalizedName(strings.ToUpper(name))
	}

	cmd.Flags().BoolVar(&flagVal, "flagname", true, "this is a dummy flag")
	cmd.Usage()

	cmd.SetGlobalNormalizationFunc(n)
	cmd.Usage()
}
```

Submitting two pull requests for a proposed fix:
https://github.com/spf13/cobra/pull/522
https://github.com/spf13/pflag/pull/137

N.B. The consistency of the command flags is a bit tricky in general, and I'm not sure if I don't understand it quite correctly or if indeed there are other issues: as a symptom, some of the tests in `command_test.go` rely on side effects from other tests to pass (at least `TestChildCommandHelp` and `TestFlagAccess`). I can elaborate a bit on that if it's of any use...

PR #522, opened 2017-08-29
Proposed fix for issue #521

Itizir:
Related to issue #521

### DIFF
--- a/command.go
+++ b/command.go
@@ -1157,6 +1157,7 @@ func (c *Command) InheritedFlags() *flag.FlagSet {
 	}
 
 	local := c.LocalFlags()
+	c.iflags.SetNormalizeFunc(c.parentsPflags.GetNormalizeFunc())
 
 	c.parentsPflags.VisitAll(func(f *flag.Flag) {
 		if c.iflags.Lookup(f.Name) == nil && local.Lookup(f.Name) == nil {
@@ -1298,6 +1299,10 @@ func (c *Command) updateParentsPflags() {
 		c.parentsPflags = flag.NewFlagSet(c.Name(), flag.ContinueOnError)
 		c.parentsPflags.SetOutput(c.flagErrorBuf)
 		c.parentsPflags.SortFlags = false
+	}
+
+	if c.HasParent() {
+		c.parentsPflags.SetNormalizeFunc(c.Parent().PersistentFlags().GetNormalizeFunc())
 	}
 
 	c.Root().PersistentFlags().AddFlagSet(flag.CommandLine)
--- a/command.go
+++ b/command.go
@@ -1157,7 +1157,6 @@ func (c *Command) InheritedFlags() *flag.FlagSet {
 	}
 
 	local := c.LocalFlags()
-	c.iflags.SetNormalizeFunc(local.GetNormalizeFunc())
 
 	c.parentsPflags.VisitAll(func(f *flag.Flag) {
 		if c.iflags.Lookup(f.Name) == nil && local.Lookup(f.Name) == nil {
--- a/command.go
+++ b/command.go
@@ -806,6 +806,7 @@ Simply type ` + c.Name() + ` help [path to command] for full details.`,
 
 // ResetCommands used for testing.
 func (c *Command) ResetCommands() {
+	c.parent = nil
 	c.commands = nil
 	c.helpCommand = nil
 	c.parentsPflags = nil
@@ -1192,6 +1193,10 @@ func (c *Command) ResetFlags() {
 	c.flags.SetOutput(c.flagErrorBuf)
 	c.pflags = flag.NewFlagSet(c.Name(), flag.ContinueOnError)
 	c.pflags.SetOutput(c.flagErrorBuf)
+
+	c.lflags = nil
+	c.iflags = nil
+	c.parentsPflags = nil
 }
 
 // HasFlags checks if the command contains any flags (local plus persistent from the entire structure).
--- a/command.go
+++ b/command.go
@@ -1133,7 +1133,9 @@ func (c *Command) LocalFlags() *flag.FlagSet {
 		c.lflags.SetOutput(c.flagErrorBuf)
 	}
 	c.lflags.SortFlags = c.Flags().SortFlags
-	c.lflags.SetNormalizeFunc(c.Flags().GetNormalizeFunc())
+	if c.globNormFunc != nil {
+		c.lflags.SetNormalizeFunc(c.globNormFunc)
+	}
 
 	addToLocal := func(f *flag.Flag) {
 		if c.lflags.Lookup(f.Name) == nil && c.parentsPflags.Lookup(f.Name) == nil {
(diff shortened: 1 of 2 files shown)


Itizir:
Hm, sorry about the hesitation: got confused with what normalization function inherited flags should get... Probably the one of the parent, in fact.
Let me know if that makes sense and I'll change this, too.

bogem:
Thank you for this PR. LGTM.
I will merge it after https://github.com/spf13/pflag/pull/137/ will be merged.

bogem:
LGTM. I will merge it after https://github.com/spf13/pflag/pull/137/ will be merged.

Itizir:
Before that I should push the change updating inherited flag's normalisation function and try to come up with a couple of tests...

Itizir:
Here are the extra changes. Let me know if there's anything I should improve.

Bad news is it's not passing the tests; good news is that's precisely what we want... And it will be fixed with the improvement to pflag.
```--- FAIL: TestNormConsistent (0.00s)
	cobra_test.go:1250: Normalizing flag names should not result in duplicate flags```

**Comment by Itizir** [comment]
Also, InheritedFlags pulling the normalisation function out of the parent is pedantic, given that the function should be identical throughout the command tree, but... :P

**Comment by bogem** [comment]
I will take a look tomorrow. Now it's already night in my timezone and I want to sleep :(

**Comment by Itizir** [comment]
It's night in mine, too. ;)
Good night!

**Comment by bogem** [comment]
@ltizir good night too ☺️

**Comment by bogem** [comment]
> --- FAIL: TestNormConsistent (0.00s)
	cobra_test.go:1250: Normalizing flag names should not result in duplicate flags

It happens because of wrong test:
```go
if cmd.LocalFlags().Lookup("flagname") == cmd.LocalFlags().Lookup("FLAGNAME") {
    t.Error("Normalizing flag names should not result in duplicate flags")
} 
```
Here is the signature of Lookup func:
```go
// Lookup returns the Flag structure of the named flag, returning nil if none exists.
func (f *FlagSet) Lookup(name string) *Flag {
	return f.lookup(f.normalizeFlagName(name))
}
```
In this case `f.Lookup("flagname")` => `f.lookup(f.normalizeFlagName("flagname"))` => `f.lookup("FLAGNAME")` will be always equal to `f.Lookup("FLAGNAME")`. So I propose to change this test:
```
diff --git a/cobra_test.go b/cobra_test.go
index 30e1af7..1eae729 100644
--- a/cobra_test.go
+++ b/cobra_test.go
@@ -1244,10 +1244,14 @@ func TestNormConsistent(t *testing.T) {
 	cmd.LocalFlags()
 
 	cmd.SetGlobalNormalizationFunc(n)
-	cmd.SetGlobalNormalizationFunc(nil)
 
-	if cmd.LocalFlags().Lookup("flagname") == cmd.LocalFlags().Lookup("FLAGNAME") {
-		t.Error("Normalizing flag names should not result in duplicate flags")
+	var i int
+	cmd.LocalFlags().VisitAll(func(f *pflag.Flag) {
+		i++
+	})
+
+	if i > 1 {
+		t.Errorf("There should be only 1 flag, not %v", i)
 	}
 }
```
Anyway, this test fails:
```
$ go test -run=TestNormCons
--- FAIL: TestNormConsistent (0.00s)
	cobra_test.go:1254: There should be only 1 flag, not 2
FAIL
exit status 1
FAIL	github.com/spf13/cobra	0.018s
```
and we should figure out why.

bogem:
> Also, InheritedFlags pulling the normalisation function out of the parent is pedantic, given that the function should be identical throughout the command tree, but... :P

And that's why I think it's better to do `c.(lflags/iflags/parentsPflags).SetNormalizeFunc(c.GlobalNormalizationFunc())`, because SetGlobalNormalizationFunc says that `SetGlobalNormalizationFunc sets a normalization function to all flag sets and also to child commands.` and we don't want to have ambiguous behavior in library.

bogem:
> Bad news is it's not passing the tests; good news is that's precisely what we want... And it will be fixed with the improvement to pflag.

I restarted TravisCI and CircleCI builds. Now this test passed.

Itizir:
> I restarted TravisCI and CircleCI builds. Now this test passed.
Yup, now that pflag has been updated...

> And that's why I think it's better to do c.(lflags/iflags/parentsPflags).SetNormalizeFunc(c.GlobalNormalizationFunc()), because SetGlobalNormalizationFunc says that SetGlobalNormalizationFunc sets a normalization function to all flag sets and also to child commands and we don't want to have ambiguous behavior in library.
All right, will change that right away. Let me know if it's better.

Itizir:
Oops.

command.go
```diff
@@ -1298,6 +1310,10 @@ func (c *Command) updateParentsPflags() {
 		c.parentsPflags.SortFlags = false
 	}
 
+	if c.globNormFunc != nil {
```
bogem: I think, it's better not to check this. User may remove global normalization func with `cmd.SetGlobalNormalizeFunc(nil)` and it will not work.

Itizir: Ah yes you're right: for some reason I thought you could not pass `nil` to `SetGlobalNormalizeFunc`.
But then I think this check should also be removed from `Command.AddCommand`: otherwise you may end up in a situation where the parent command does not have a normalisation function set, but the child does, which contradicts the assumption that it is set consistently throughout the command tree.

bogem: I was wrong. It's better to check if c.globNormFunc is nil. If user wants to remove globNormFunc, it's better to use `func(f *FlagSet, name string) NormalizedName { return NormalizedName(name) }`, because using nil funcs can cause to nil-pointer errors.

bogem:
Everything is okay for me. Do you want to add something in this PR or can I merge it?

Itizir:
If it's ok with you -- following your suggestion above -- I'll remove all `if c.globNormFunc != nil` checks, and make sure in tests that a nil normalisation function in the root cmd resets the child's.

bogem:
> If it's ok with you -- following your suggestion above -- I'll remove all if c.globNormFunc != nil checks, and make sure in tests that a nil normalisation function in the root cmd resets the child's.

I wrote in review:
> I was wrong. It's better to check if c.globNormFunc is nil. If user wants to remove globNormFunc, it's better to use func(f *FlagSet, name string) NormalizedName { return NormalizedName(name) }, because using nil funcs can cause nil-pointer errors.

Itizir:
Oh no wait a second! When a child command's normalisation is set, it doesn't propagate up, does it?

So this thing about it being always consistent is not correct, right? So my original proposal where the parent's normalisation function is checked when getting inherited flags did make sense. Am I missing something?

Itizir:
Ah, you already merged... Well, I think it's still worth thinking whether the consistency provided by the current implementation makes sense or not. :P

> I was wrong. It's better to check if c.globNormFunc is nil. If user wants to remove globNormFunc, it's better to use func(f *FlagSet, name string) NormalizedName { return NormalizedName(name) }, because using nil funcs can cause nil-pointer errors.

Not sure it's an issue: in pflag, it is OK for the function to be nil, since the function is never used directly but always called through `GetNormalizeFunc`.

bogem:
> When a child command's normalisation is set, it doesn't propagate up, does it?

> SetGlobalNormalizationFunc sets a normalization function to **all flag sets** and also to child commands.

I think, it's also should be used for inherited flags.

Itizir:
Indeed, it propagates down, but not up...

bogem:
No problem. Thank you for your PR, you fixed really serious issue that nobody noticed earlier.

Itizir:
Very welcome. Let me know if there's anything else I can do!
Think about how you want to resolve this potential inconsistency between parents' and children's normalisation functions... ;)

bogem:
> Let me know if there's anything else I can do!

Any PR is welcome. Just take a look at issues list.

> Think about how you want to resolve this potential inconsistency between parents' and children's normalisation functions... ;)

Hopefully, we will drastically solve this issue in v2.

Itizir:
Nice. Good luck!

PR #523, opened 2017-08-29
dep init

zevdg:
This is not just a normal `dep init`.  This is 1 `dep init` in the project root *without* the vendor folder and then another `dep init` in the cobra subdirectory **with** the vendor folder.  This gives you reproducible builds on the tool without causing problems for clients of the library solving both problems outlined in #259 

I don't necessarily expect this to get merged as is.  It could be, but I initially created it just to further the dialog by showing an example of that this could look like.


Gopkg.lock
```diff
@@ -0,0 +1,123 @@
+# This file is autogenerated, do not edit; changes may be undone by the next 'dep ensure'.
+
+
+[[projects]]
+  name = "github.com/cpuguy83/go-md2man"
+  packages = ["md2man"]
+  revision = "a65d4d2de4d5f7c74868dfa9b202a3c8be315aaa"
+  version = "v1.0.6"
```
zevdg: This particular version is actually wrong and problematic.  It pulls in blackfriday v2 and breaks stuff.  Do not approve this PR until this dependency has been updated.

.travis.yml
```diff
@@ -2,20 +2,23 @@ language: go
...
   allow_failures:
-    - go: tip
+    - go: master
 
 before_install:
   - mkdir -p bin
   - curl -Lso bin/shellcheck https://github.com/caarlos0/shellcheck-docker/releases/download/v0.4.3/shellcheck
   - chmod +x bin/shellcheck
 script:
-  - PATH=$PATH:$PWD/bin go test -v ./...
```
zevdg: fyi, this line ignores the vendor folder as expected in go1.9, but not in previous versions of go.  The changes I've made work across all go versions.

zevdg:
AFAICT, this is safe to merge now.  Now it's just up to you all to decide if you're ready to adopt `dep`.

umarcor:
Now that modules are supported in go, should this PR be closed in favour of #817 (which will hopefully close #756, #720, #795...)?

zevdg:
Agreed on both counts.

PR #526, opened 2017-09-04
Make *Command receivers consistent

SamWhited:
Currently receivers for `*Command` use inconsistent naming in the docs:

consistentdocs

This fixes all receivers for that type to use the name `c`.


eparis:
Thank you for the patch. I agree it makes the docs look much better. Would you be wiling to sign the CLA/CDO? If you do I'd be glad to merge this!

SamWhited:
I can't imagine this is copyrightable, but no, I don't like to sign random things just to make a contribution, sorry about that.

PR #531, opened 2017-09-06
Fix handling of custom license, make license fields actual templates

nerdlem:
This pull request causes license fields to actually behave as templates where interpolation can occur. It also generates correct `LICENSE` files in the case of custom licenses specified via YAML config file.

Inconsistency in the values available for interpolation is fixed by introducing a single function to convert from the `Project` structure into a `map[string]...` that exposes information for template expansion.

No changes to the `README` are required as this pull request actually causes behavior to match documentation.

### DIFF
--- a/cobra/cmd/init.go
+++ b/cobra/cmd/init.go
@@ -16,7 +16,6 @@ package cmd
 import (
 	"fmt"
 	"os"
-	"path"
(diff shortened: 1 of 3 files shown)


bogem:
Please sign a CLA

PR #532, opened 2017-09-11
Support default value of $GOPATH

### LINKED ISSUE #527: Error: $GOPATH is not set
Hello,

As of go1.8 one does not need to explicitly set the GOPATH environment variable. Which is the way I have setup my local environment.

```
% go env
GOARCH="amd64"
GOBIN=""
GOEXE=""
GOHOSTARCH="amd64"
GOHOSTOS="linux"
GOOS="linux"
GOPATH="/home/marcus/go"
GORACE=""
GOROOT="/usr/lib/go"
GOTOOLDIR="/usr/lib/go/pkg/tool/linux_amd64"
GCCGO="gccgo"
CC="gcc"
GOGCCFLAGS="-fPIC -m64 -pthread -fmessage-length=0 -fdebug-prefix-map=/tmp/go-build932011383=/tmp/go-build -gno-record-gcc-switches"
CXX="g++"
CGO_ENABLED="1"
PKG_CONFIG="pkg-config"
CGO_CFLAGS="-g -O2"
CGO_CPPFLAGS=""
CGO_CXXFLAGS="-g -O2"
CGO_FFLAGS="-g -O2"
CGO_LDFLAGS="-g -O2"
```

The go tools use the default location ($HOME/go), but cobra complains about a missing GOPATH variable:
```
% cobra init
Error: $GOPATH is not set
```

Please make cobra aware of the default go path.

e-beach:
Fixes #527.

This handles the default `$GOPATH` by checking if the environment variable exists, otherwise using the output of `go env GOPATH`. It is the same method as the Glide project (https://github.com/Masterminds/glide/pull/798/files).

This `init` script is the only place where the `GOPATH` environment variable is accessed, so this change should be sufficient for the project.

A questionable feature is the `COBRA_GO_EXECUTABLE` environment variable - I copied it from Glide, but it might not be needed.

### DIFF
--- a/cobra/cmd/helpers.go
+++ b/cobra/cmd/helpers.go
@@ -18,10 +18,10 @@ import (
 	"fmt"
 	"io"
 	"os"
+	"os/exec"
 	"path/filepath"
 	"strings"
 	"text/template"
-	"os/exec"
 )
 
 var cmdDirs = [...]string{"cmd", "cmds", "command", "commands"}
--- a/cobra/cmd/helpers.go
+++ b/cobra/cmd/helpers.go
@@ -21,6 +21,7 @@ import (
 	"path/filepath"
 	"strings"
 	"text/template"
+	"os/exec"
 )
 
 var cmdDirs = [...]string{"cmd", "cmds", "command", "commands"}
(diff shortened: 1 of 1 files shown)


bogem:
Great PR, and it's better to leave `COBRA_GO_EXECUTABLE` because someone can have another alias to call go command.

e-beach:
Woo! Glad to help.

PR #544, opened 2017-10-09
Added CustomData

snwfdhmp:
Added CustomData : interface{}

In many situation, having a CustomData to store indexes or arbitrary informations about the Command.

Example: Storing Command index, to make it able to access himself through an array of Command after its declaration.
It helps automating Command generation from configuration files (see [here an example](https://github.com/snwfdhmp/duck2/blob/master/main.go) in function 'createCobraCommand' (in this case 'CustomData' is named 'DuckCmdIndex'))

### DIFF
--- a/command.go
+++ b/command.go
@@ -175,6 +175,9 @@ type Command struct {
 	// helpCommand is command with usage 'help'. If it's not defined by user,
 	// cobra uses default help command.
 	helpCommand *Command
+	
+	// CustomData permits you to store arbitrary data for specific needs
+	CustomData interface{}
 }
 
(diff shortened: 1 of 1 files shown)


bogem:
IMHO this option is redundant for cobra.Command. If you want to store some information about commands, make your own storage (map) in your code or use [viper](https://github.com/spf13/viper).

eparis:
You have `cmd.Annotations` if you can store your custom data as a string. I really hate having the same thing twice, when should one use annotations and when 'customData' ? I'll ponder some more. You can also consider

```
type myCmd struct {
  cobra.Command
  customData interface{}
```

myCmd will be able to be used as a cobra.Command....

PR #547, opened 2017-10-11
Delete name field from Command

bogem:
Discussion: https://github.com/spf13/cobra/pull/422#discussion_r143918343

/cc @dixudx @eparis


dixudx:
LGTM

eparis:
@dixudx I merged this change, but would you please write a test to capture the failure kube hit?

PR #548, opened 2017-10-12
add test for deprecate alias and c.Name()

dixudx:
Discussion: [#547 (comment)](https://github.com/spf13/cobra/pull/547#issuecomment-335825065)

/cc @bogem @eparis

### DIFF
--- a/command_test.go
+++ b/command_test.go
@@ -486,3 +486,15 @@ func TestPersistentRequiredFlags(t *testing.T) {
 		}
 	}
 }
+
+// TestUpdateName checks if c.Name() updates on changed c.Use.
+// Related to https://github.com/spf13/cobra/pull/422#discussion_r143918343.
+func TestUpdateName(t *testing.T) {
+	c := &Command{Use: "name xyz"}
+	originalName := c.Name()
+
+	c.Use = "changedName abc"
+	if originalName == c.Name() || c.Name() != "changedName" {
+		t.Error("c.Name() should be updated on changed c.Use")
+	}
+}


bogem:
Do we actually need to add a test of some functionality of Kubernetes?

dixudx:
@bogem Actually this is not specially for Kubernetes. The main purpose of this test is for `c.Name()`, which used to have a cached value. It is this cached value that breaking Kubernetes.

bogem:
@dixudx Then you can write simpler test:
```go
func TestUpdateName(t *testing.T) {
	c := &Command{Use: "name"}
	originalName := c.Name()

	c.Use = "changedName"
	if originalName == c.Name() {
		t.Error("c.Name() should be updated on changed c.Use")
	}
}
```

dixudx:
@bogem Yes, this is much simpler.

Originally I combined deprecated aliases testing with `c.Name()` together. Do we still need testing deprecated aliases?

bogem:
> Do we still need testing deprecated aliases?

I think, no. It's a functionality of Kubernetes, not cobra. We don't have to make tests more complex and simulate deprecatedAliases.

command_test.go
```diff
@@ -486,3 +486,15 @@ func TestPersistentRequiredFlags(t *testing.T) {
...
 }
+
+// TestUpdateName checks if c.Name() updates on changed c.Use.
+// Related to https://github.com/spf13/cobra/pull/422#discussion_r143918343.
+func TestUpdateName(t *testing.T) {
+	c := &Command{Use: "name xyz"}
+	originalName := c.Name()
+
+	c.Use = "changedName abc"
+	if originalName == c.Name() {
```
eparis: Is this the right test? Should it be:
```
if c.Name() != "changedName" {
```

bogem:
> If we made a change in cobra that broke any user we need to write a test to make sure we never ever break them again.

> But I want to be very adamant that if cobra changes, even subtly, and breaks any user, we need to fix the regression and always write a test mimicking the user's behavior to prevent us from breaking them again.

I totally agree with you, but I meant that we don't need to simulate the functionality of kubernetes and add a lot of code to just check a core problem. We should keep our codebase as simple as we can. The test I [proposed](https://github.com/spf13/cobra/pull/548#issuecomment-336052882) does the same thing but much simpler.

command_test.go
dixudx: @eparis The main purpose here is to see whether `c.Name()` gets changed after assigning a new value to `c.Use`.

We may add a extra statement here to see if `c.Name() != "changedName"`.

dixudx:
@eparis @bogem Updated. PTAL.

> The answer all cobra maintainers should have for that question is OF COURSE! If we made a change in cobra that broke any user we need to write a test to make sure we never ever break them again.

@eparis Right. We should be cautious on the changes that will affect users, such as flag name changing.

My original tests did include `c.Deprecated` and `c.Name()` together, which is exactly breaking Kubernetes. Since we already have `TestUseDeprecatedFlags`, there is no need to test deprecated flags again.

@bogem's [proposal](https://github.com/spf13/cobra/pull/548#issuecomment-336052882) does focus on `c.Name()` and be much simpler. Thanks.

PR #552, opened 2017-10-26
Add a field to disable [flags] in UseLine()

dnephin:
`Command.UseLine()` will append a `[flags]` string for any command that has flags. This behaviour is often wrong. For example, in the following cases:

* the use line documents flags using some other string (`[FLAGS]`, `[OPTIONS]`, etc)
* the use line includes positional arguments, and https://godoc.org/github.com/spf13/pflag#SetInterspersed is set to false. This requires that flags appear before any positional arguments, so appending the string to the end of the use line is incorrect.

To fix these two problems I've added a new field to disable adding `[flags]` to the `UseLine`. The application will be responsible for included the string in the appropriate position in the usage line.

### DIFF
--- a/command.go
+++ b/command.go
@@ -118,6 +118,10 @@ type Command struct {
 	// will be printed by generating docs for this command.
 	DisableAutoGenTag bool
 
+	// DisableFlagsInUseLine will disable the addition of [flags] to the usage
+	// line of a command when printing help or generating docs
+	DisableFlagsInUseLine bool
+
 	// DisableSuggestions disables the suggestions based on Levenshtein distance
 	// that go along with 'unknown command' messages.
 	DisableSuggestions bool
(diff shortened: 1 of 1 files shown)


dnephin:
cc @bogem

bogem:
I made a mistake, when I added the`[flags]` string to UseLine.
Actually I think, this bug doesn't worth to add more public fields and expand API. 
In version 2.0 I want to just revert the behaviour and delete `[flags]` from useline. I will start on working with 2.0 tomorrow.

dnephin:
Ok, do you think there is any way we can fix this in V1 without adding the new field? 

I was able to fix it for some cases by using a template for help, but that doesn't work for man page generation. Maybe the man page generation could accept a template as well?

bogem:
I thought a lot about a fix for this issue and can't come up with better solution than yours.
Anyway we should add `[flags]` at the end of UseLine in help if it's not provided. [Previous solution](https://github.com/spf13/cobra/blob/d83a1d7ccd00a9e1b5d234653837b498b9b27abd/command.go#L333) (adding [flags] directly in help) was wrong too.

bogem:
I would like to merge it now, but I want hear the opinion of @eparis too.

Issue #342, opened 2016-09-04
Man page generator generates inconsistent file names and SEE ALSO references


knz:
We encountered this in CockroachDB (xref https://github.com/cockroachdb/cockroach/issues/4917)
- `GenManTree` in `doc/man_docs.go` forces the command separator to be `"_"`, which causes `GenManTreeFromOpts` to generate file names for man pages to be `like_this.1`
- Meanwhile, `genMan` replaces spaces in names by `"-"` always.

As a result, when the reader reads the "SEE ALSO" section she might see a reference to `this-command(1)` whereas the command she has to type to read that man page is actually `man this_command`.

May I suggest to use dashes always, that is, set the options in `GenManTree` to use `"-"`.

dnephin:
I ran into this problem as well, and submitted a fix in #294 (e291587027b16ffd7fb0e5bdd6bf203e8ed2620f).

I think you can use `GenManTreeFromOpts()` and pass `CommandSeparator: "-",` to do what you want.

knz:
Please explain more. Why would you want to keep inconsistent default settings in the cobra code?

dnephin:
Hmm, I'm not sure. I guess some people might be expecting the old file names and don't use `SEE ALSO` ? I wanted to keep backwards compatibility, so I left the old default and added it as an option.

eparis:
owuld the right thing to do to fix 'see also'? as that would seem more backwards compat...

knz:
Look either are fine, I think this is subjective really. I'm not a cobra developer so I can't really make a call on this. Just inform me what the best solution will be; if my PR #343 ought to be written another way make a comment there and I'll update it accordingly.

anthonyfok:
I agree with @knz, especially this is the way `git` sub-command man pages work too, where `-` dashes are used to represent spaces, and sometimes, actual `-` dashes.

On the other hand, `_` underscores means actual underscores in the command name, e.g. `kbd_mode`, `xinput_calibrator`.  I couldn't really find an example (outside of cobra-generated man pages) in my /usr/share/man/man1/ directory where the `_` underscore is used to represent a space.

So, yes, I think I will simply go ahead and accept @knz's PR #343.

knz:
Thank you!

anthonyfok:
Thank you for bringing this up and providing the patch!  :-)

PR #565, opened 2017-11-03
enable changing the command separator when generating markdown

bacongobbler:
In certain cases, hyphens are preferred over underscores for command separators. For example, documentation hosted by markdown-generated sites like hugo use hyphens as the path separator.

The code here is inspired by #342 and is necessary to fix our documentation, so I figure I'd throw out a PR and see if this is something you'd be interested in supporting upstream.

Downstream bugs that are fixed through this enhancement are https://github.com/helm/helm-www/issues/60 and https://github.com/kubernetes/helm/issues/2840.


doc/md_docs.go
```diff
@@ -50,8 +66,8 @@ func GenMarkdown(cmd *cobra.Command, w io.Writer) error {
 	return GenMarkdownCustom(cmd, w, func(s string) string { return s })
 }
 
-// GenMarkdownCustom creates custom markdown output.
-func GenMarkdownCustom(cmd *cobra.Command, w io.Writer, linkHandler func(string) string) error {
```
eparis: I really like the idea. But this would break API compatibility. You may create a new function `GenMarkdownWithOpts` and call it from here. Please note this function as being deprecated in the godoc. Please make sure to also supply to end users (and use ourselves) default versions of the GenMarkdownOptions and GenMarkdownTreeOptions. So that an end user can easily use the new functions and get the default values.

bacongobbler: All good points. I’ll be happy to do so. :)

PR #567, opened 2017-11-05
Add `CalledAs` method to cobra.Command (w/ tests)

tep:
The `CalledAs` method returns the name of the command or alias that
invoked the command -- as long as the command was actually invoked.
Otherwise, it returns the empty string.

The opens up possibilies for commands to behave differently based on
which alias invoked the command (in the same vein as Linux programs
which adjust their behavior based on the value of argv[0]).


eparis:
Looks good to me, thank you!

gdey:
Any way to get this information in the PreRun routines?

PR #575, opened 2017-11-10
Improve error message for missing required flags

nmiyake:
Make the error message distinguish between singular and plural
to make it more human-readable.

### DIFF
--- a/command.go
+++ b/command.go
@@ -823,7 +823,11 @@ func (c *Command) validateRequiredFlags() error {
 	})
 
 	if len(missingFlagNames) > 0 {
-		return fmt.Errorf(`Required flag(s) "%s" have/has not been set`, strings.Join(missingFlagNames, `", "`))
+		errFmtStr := `Required flags "%s" have not been set`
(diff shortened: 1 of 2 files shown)


bogem:
Hello, @nmiyake! Thank you for your contribution, but sorry, we don't need to make logic more complex only for this. It's normal practice to use just `(s)` in the end of nouns.

nmiyake:
@bogem thanks for the feedback -- agree that extra logic is overkill and the `(s)` is consistent with the other messages in the program.

Opened https://github.com/spf13/cobra/pull/580 with a smaller suggested change. I think the `have/has` is the construction that seems most unnatural to me, so this PR suggests changing that. However, understand that this is a matter of taste and am fine if it is not made.

However, I do think that un-capitalizing the leading `R` is best for consistency -- all of the other Cobra errors follow this convention and it makes the default error output more readable.

Changes:

```
Error: Required flag(s) "project-dir" have/has not been set
```

to

```
Error: required flag(s) "project-dir" not set
```

PR #577, opened 2017-11-10
Added ValidArgs as object

themester:
to prevent memory wasting using new Commad as subsubcommands

### DIFF
--- a/args.go
+++ b/args.go
@@ -35,7 +35,14 @@ func NoArgs(cmd *Command, args []string) error {
 func OnlyValidArgs(cmd *Command, args []string) error {
 	if len(cmd.ValidArgs) > 0 {
 		for _, v := range args {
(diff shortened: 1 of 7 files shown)


args.go
```diff
@@ -35,7 +35,14 @@ func NoArgs(cmd *Command, args []string) error {
 func OnlyValidArgs(cmd *Command, args []string) error {
 	if len(cmd.ValidArgs) > 0 {
 		for _, v := range args {
-			if !stringInSlice(v, cmd.ValidArgs) {
+			if !func() bool {
```
dnephin: I think this would be great as a named function:

```go
func isValidArg(cmd *Command, arg string) bool {
   ..
}
```

bogem:
It actually breaks the backward compatibility. So we can‘t easily merge it.

PR #580, opened 2017-11-18
Update error message for missing required flags

nmiyake:
Make it so that first letter is not capitalized and rephrase
to remove "have/has".

### DIFF
--- a/command.go
+++ b/command.go
@@ -823,7 +823,7 @@ func (c *Command) validateRequiredFlags() error {
 	})
 
 	if len(missingFlagNames) > 0 {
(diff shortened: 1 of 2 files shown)


nmiyake:
PR suggests rephrasing error to remove the `have/has` construction, which doesn't read super fluently. However, understand that this is change is a matter of taste and am fine if it is not made.

However, I do think that un-capitalizing the leading `R` is best for consistency -- all of the other Cobra errors follow this convention and it makes the default error output more readable.

Changes default error output of the form:

```
Error: Required flag(s) "project-dir" have/has not been set
```

to

```
Error: required flag(s) "project-dir" not set
```

bogem:
It looks nice now :)

PR #581, opened 2017-11-19
remove hard-coded "[flags]" in UseLine

dixudx:
remove hard-coded `[flags]` introduced by #422, which is changing the usage template. This will break some downstream projects, like [kubernetes](https://github.com/kubernetes/kubernetes/pull/53631#issuecomment-344054069).

@bogem @eparis


bogem:
#552 offers more compromise way to disable it.

Issue #578, opened 2017-11-12
bash completion doesn't work correctly for alias of command


superbrothers:
bash completion generated by cobra does not work correctly for alias of command.

```
$ bash --version
GNU bash, version 4.4.12(1)-release (x86_64-unknown-linux-gnu)
Copyright (C) 2016 Free Software Foundation, Inc.
License GPLv3+: GNU GPL version 3 or later <http://gnu.org/licenses/gpl.html>

This is free software; you are free to change and redistribute it.
There is NO WARRANTY, to the extent permitted by law.
$ source $(brew --prefix)/share/bash-completion/bash_completion
$ source <(hello completion)
$ alias h="hello"
$ complete -o default -F __start_hello h
$ h <tab>
```

**Actual output:**
```
$ h <tab>
hello
```

**Expected output:**
```
$ h <tab>
completion  say
```

https://github.com/superbrothers-sandbox/hello-cli/tree/autocomplete-alias reproduces this issue.

/ref https://github.com/kubernetes/kubectl/issues/120

nikhita:
@superbrothers Looks like this can be closed and is fixed.

superbrothers:
@nikhita Yes, this has been already fixed by https://github.com/spf13/cobra/pull/638. Thank you for letting me know.

PR #588, opened 2017-11-29
#578: allow aliases to work with completions

rajatjindal:
This fixes the issue that auto completion does not work with aliases.

basically we are using a 'static' root_command name instead of getting it dynamically based on 'basename $0'

created a sample project using this and tested it to be working as expected.

### DIFF
--- a/bash_completions.go
+++ b/bash_completions.go
@@ -21,8 +21,8 @@ const (
 
 func writePreamble(buf *bytes.Buffer, name string) {
 	buf.WriteString(fmt.Sprintf("# bash completion for %-36s -*- shell-script -*-\n", name))
-	buf.WriteString(`
-__debug()
+	buf.WriteString(fmt.Sprintf(`
+__%[1]s_debug()
 {
     if [[ -n ${BASH_COMP_DEBUG_FILE} ]]; then
         echo "$*" >> "${BASH_COMP_DEBUG_FILE}"
@@ -31,13 +31,13 @@ __debug()
 
 # Homebrew on Macs have version 1.3 of bash-completion which doesn't include
 # _init_completion. This is a very minimal version of that function.
-__my_init_completion()
+__%[1]s_init_completion()
 {
     COMPREPLY=()
     _get_comp_words_by_ref "$@" cur prev words cword
 }
 
-__index_of_word()
+__%[1]s_index_of_word()
 {
     local w word=$1
     shift
@@ -49,7 +49,7 @@ __index_of_word()
     index=-1
 }
 
-__contains_word()
+__%[1]s_contains_word()
 {
     local w word=$1; shift
     for w in "$@"; do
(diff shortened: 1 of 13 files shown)


rajatjindal:
I am in process of getting approval from my work to sign the CLA

rajatjindal:
@bocan you please review this code.

Thanks
Rajat Jindal

bogem:
Actually, I don't understand bash syntax at all, especially for auto-completion. So please @eparis check the code. This change is small.

rajatjindal:
Thanks @bogem

@eparis can you please review this code

rajatjindal:
pinging @eparis

superbrothers:
I've verified that #578 will be fixed by this PR at https://github.com/superbrothers-sandbox/hello-cli/tree/autocomplete-alias. Thanks @rajatjindal.

rajatjindal:
@bogem can you come pls help get some attention here

eparis:
I'll get off my lazy backside today. If I don't, keep hitting me on slack over and over and over relentlessly. I'm so sorry.

rajatjindal:
@eparis I am pretty sure you are busy with more critical stuff. and i totally get that.

appreciate all the amazing open source work you guys are doing. 

I will continue to nag you though :)

rajatjindal:
@eparis another ping !

bash_completions.go
```diff
@@ -434,6 +438,16 @@ func writeRequiredNouns(buf *bytes.Buffer, cmd *Command) {
 	}
 }
 
+func writeCmdAliases(buf *bytes.Buffer, cmd *Command) {
+	sort.Sort(sort.StringSlice(cmd.Aliases))
+
+	for _, value := range cmd.Aliases {
+		buf.WriteString(fmt.Sprintf("    command_aliases+=(%q)\n", value))
+		buf.WriteString(fmt.Sprintf("    aliasHash[%q]=%q\n", value, cmd.Name()))
```
eparis: I FINALLY looked at this. Since the (ancient) version of bash on OS X does not support associative arrays, we have the HORRIBLE contortions with `flags_with_completion` and trying to find the same index in 2 different arrays.

Do you know how this script will react on bash versions that don't support arrays? I see that flagHash does exist, but we never actually use it...  I think it could/should be ripped out of the code.

Should we do something like grab the bash version or remember if the declare -A worked, and just skip the words[c] rewrite if the version is too old?

rajatjindal:
Hi @eparis 

do you think the alias support can be taken as a separate PR and get just the bash completion fix merged into it. bash completion fix is a much smaller change to review and is the big pain point for multiple projects.

If you agree to this, i will revert the changes done to ensure it only includes bash completion fix.

Thanks
Rajat Jindal

rajatjindal:
actually flaghash is being used in __handle_flag() function

eparis:
Oh my, I thought I was waiting for you here! I suck. In nay case, I just merged
https://github.com/spf13/cobra/pull/628
Which is because we didn't work on stock OS X bash on mac. Can you find a way to do something similar?

rajatjindal:
Hi @eparis

I will try to get this done this week

PR #590, opened 2017-12-02
Add documentation for '--version' flag to README

nmiyake:


nmiyake:
Adds documentation for https://github.com/spf13/cobra/pull/584

README.md
```diff
@@ -560,6 +560,43 @@ cmd.SetUsageFunc(f func(*Command) error)
...
 ```
 
+## Version Flag
+
+Cobra adds a top-level '--version' flag if the Version field is set on the root command.
+Running an application with the '--version' flag will print the version to stdout using
+the version template. The template can be customized using the
+`cmd.SetVersionTemplate(s string)` function.
+
+### Example
```
bogem: Actually, the docs are OK, but I find this example too obvious, so it would better to not leaving it here.

nmiyake: Sure, makes sense -- removed.

PR #615, opened 2018-01-16
test if flag is a bool before setting inFlag

wxio:
Bug fix to get the following to work 
```
cli --boolflag command
```
Currently `command` is attached to the boolean flag.
According to pflag 
https://github.com/spf13/pflag#command-line-flag-syntax
```
// boolean or flags where the 'no option default value' is set
-f
-f=true
-abc
but
-b true is INVALID
```

Therefore boolean flags not set inFlag.

There seems to be inconsistencies is c.Flags() earlier on, hence adding the call to merge.
Also I don't understand why `NoOptDefVal` is not set for booleans as is appeared to be automatic by pflag.

### DIFF
--- a/command.go
+++ b/command.go
@@ -436,10 +436,13 @@ func (c *Command) VersionTemplate() string {
 `
 }
 
-func hasNoOptDefVal(name string, fs *flag.FlagSet) bool {
-	flag := fs.Lookup(name)
+func hasDefault(name string, c *Command) bool {
+	flag := c.Flags().Lookup(name)
 	if flag == nil {
-		return false
+		return true
+	}
+	if flag.Value.Type() == "bool" {
+		return true
 	}
 	return flag.NoOptDefVal != ""
 }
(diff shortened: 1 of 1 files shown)


dnephin:
Could you write a test that shows the bug?

eparis:
If think you are saying that you have: `flag.Value.Type() == "bool"` but that `flag.NoOptDefVal == ""`  I believe that shouldn't happen...  Can you write a test how that happened.

wxio:
@dnephin @eparis  take a look at 
https://github.com/wxio/cobra/tree/pr615test/tests

The issue may involve the combination of
```
var rootCmd = &cobra.Command{
	TraverseChildren: true,
}
```
and
```
var subCmd = &cobra.Command{
	Use:                "sub",
	DisableFlagParsing: true,
		fmt.Printf("sub called with args %v", args)
	},
}
```

wxio:
see https://github.com/spf13/cobra/pull/616

PR #620, opened 2018-01-22
Add Set* functions to allow overriding of localised values after create

Naatan:
Currently you can only set the following properties at the time of creation:

 * Use
 * Short
 * Long
 * Example

These are all properties that would hold localised strings. This PR allows you to modify them after creation of the struct, which in turn opens the door for localisation.

Of course you could in theory localise these at the time of creation, but this falls apart when you use Cobra to specify the localisation.

Regardless of use case, it seems fairly harmless to allow users to override these after creation.

### DIFF
--- a/command.go
+++ b/command.go
@@ -186,6 +186,30 @@ type Command struct {
 	versionTemplate string
 }
 
+// SetUse allows you to set the Use property after the command has been created, primarily intended to
+// facilitate localisation
+func (c *Command) SetUse(s string) {
+	c.Use = s
+}
+
+// SetShort allows you to set the Short property after the command has been created, primarily intended to
+// facilitate localisation
+func (c *Command) SetShort(s string) {
+	c.Short = s
+}
+
+// SetLong allows you to set the Long property after the command has been created, primarily intended to
+// facilitate localisation
+func (c *Command) SetLong(s string) {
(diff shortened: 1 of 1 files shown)


command.go
```diff
@@ -186,6 +186,30 @@ type Command struct {
 	versionTemplate string
 }
 
+// SetUse allows you to set the Use property after the command has been created, primarily intended to
+// facilitate localisation
+func (c *Command) SetUse(s string) {
```
eparis: Don't we have a bunch of checks in AddCommand to make sure we never set the same command name twice? Would we need that here?

wfernandes:
+label:"needs investigation"
This seems like a reasonable ask to support these methods. This is also just adding to the API so I don't see it to be breaking anything.
@jharshman WDYT?

jharshman:
@wfernandes this sounds fine to me. I think we need to ensure that we don't have conflicts in command names as @eparis mentioned in his comment above.

marckhouzam:
All the mentioned properties are public so they can be modified after creation. Is this just for better design?

jharshman:
> All the mentioned properties are public so they can be modified after creation. Is this just for better design?

Yeah actually that's a good point. This PR really don't introduce any new functionality.

PR #627, opened 2018-02-03
Add usage example for required flags

jakekdodd:
Was looking for this feature and came across #502, thought an example in the readme would be helpful


eparis:
Please also add more/correct comments around the actual `func MarkFlagRequired` so we have good information in https://godoc.org/github.com/spf13/cobra#MarkFlagRequired

eparis:
That looks great, thank you so much!

jakekdodd:
👍 glad to help! Great library

PR #628, opened 2018-02-04
Fix attemptd assignment to non-variable in bash 3

superbrothers:
flaghash variable is an associative array which is only supported in bash > 3.

```
kubectl get service --namespace=kube-system k-bash: --namespace=: attempted assignment to non-variable (error token is "=")
```

fixes https://github.com/kubernetes/kubectl/issues/121, https://github.com/kubernetes/kubernetes/issues/29322, https://github.com/kubernetes/kubernetes/issues/32676

### DIFF
--- a/bash_completions.go
+++ b/bash_completions.go
@@ -176,7 +176,7 @@ __handle_flag()
 
     # keep flag value with flagname as flaghash
     # flaghash variable is an associative array which is only supported in bash > 3.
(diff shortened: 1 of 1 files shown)


superbrothers:
@eparis Do you know where is `flaghash` used? It seems not to be used in [kubernetes/kubernetes](https://github.com/kubernetes/kubernetes). If it is not used anywhere, we can revert #205. WDYT?

superbrothers:
I tested `kubectl completion bash/zsh` with https://github.com/superbrothers/kubernetes/tree/cobra-628.

PR #629, opened 2018-02-05
Add Execute func example in README.md

phillipj:
Hi,

I'm a golang rookie and had a hard time grasping where the `cmd.Execute()` function came from in the first Getting Started section of the `README.md`. Hopefully it helps other not spending unnecessary time connecting the dots 😄 

This trivial addition defines the `cmd.Execute()` function that is referenced further down in the `main.go` example.

Found it in another repo that contains a minimal Cobra example: https://github.com/dharmeshkakadia/cobra-example/blob/master/cmd/root.go#L41, seems reasonable and does the trick.

### DIFF
--- a/README.md
+++ b/README.md
@@ -192,6 +192,13 @@ var rootCmd = &cobra.Command{
     // Do Stuff Here
   },
 }
+
+func Execute() {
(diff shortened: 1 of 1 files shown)


eparis:
Thank you so much!

Issue #465, opened 2017-06-12
Obsolete bash completion example (kubernetes)... and use a better example anyway


rfay:
The current [bash_completions.md]
(https://github.com/spf13/cobra/blob/master/bash_completions.md) 

* uses a very complex example (kubernetes) and also 
* Has the wrong import ("github.com/GoogleCloudPlatform/kubernetes/pkg/kubectl/cmd"
) and code (it's shifted over time)

And it might be best to scale back the initial statement "is incredibly easy", since I'm a seasoned developer and can make no sense of the example at all, as it uses custom kube stuff that has moved around and doesn't seem relevant.

A basic example on a simpler go repo would work a lot better.

Thanks!

rfay:
Of course this *is* really easy. In our situation, we have a command RootCmd which has all the other commands added to it. Seems like an easy example like this could just be included. Works perfectly, thanks.

```
func main() {
	err := cmd.RootCmd.GenBashCompletionFile("bin/ddev_completion.sh")

	util.CheckErr(err)
}
```

eparis:
@rfay Ha. Hard to argue. I wrote the bash completions stuff so I could use it in kube. At the time it seemed so 'simple'. I'd love a PR that makes an actually simple example.

dfong:
it may be simple, but it is not self-explanatory.  would be helpful to have an explanation of how to generalize it to my own command.

menuka94:
It would be great if someone could provide an auto-completion example for a custom command :)

GodloveD:
+1  would love to see a generic example

PR #631, opened 2018-02-06
Include a basic example #465

Jason-Hendry:
#465 Obsolete bash completion example (kubernetes)... and use a better example anyway

### DIFF
--- a/bash_completions.md
+++ b/bash_completions.md
@@ -1,5 +1,40 @@
 # Generating Bash Completions For Your Own cobra.Command
 
+If you are using the generator you can create a completion command by running
+
+```bash
+cobra add completion
+```
+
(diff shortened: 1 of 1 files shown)


eparis:
This is great. Would you be willing to delete the kubernetes example entirely, and possibly make this an actual go example that shows up in godoc? I also often find that I use `Hidden: true,` in my definition of such a subcommand, so maybe show that, even if it is commented out?

itskingori:
This should be merged. It helped me figure out what the issue was.

eparis:
I'll merge this. But @itskingori any willingness to make a follow so this is an example which shows up in godoc instead of these impenetrable .md files?

eparis:
(and delete the kube example entirely now that we have a much better one?)

intel352:
@eparis what's the `<(bitbucket completion)` bit for?  It occurs twice in the long help command.

PR #638, opened 2018-02-19
Bash completion aliases

rajatjindal:
This is continuation of PR #588 ( i messed up branch during rebasing from master).

I've taken out the code to support "cobra command aliases" which needed to use flaghash. I will work on that as a separate PR.

This current changes are just to support bash aliases for the root command

so we can do like: 

alias aliasname=origname
complete -o default -F __start_origname aliasname

and then we can use aliasname with completions working.

### DIFF
--- a/bash_completions.go
+++ b/bash_completions.go
@@ -222,7 +222,7 @@ __%[1]s_handle_command()
         next_command="_${last_command}_${words[c]//:/__}"
     else
         if [[ $c -eq 0 ]]; then
-            next_command="_root_command"
+            next_command="_%[1]s_root_command"
         else
             next_command="_${words[c]//:/__}"
         fi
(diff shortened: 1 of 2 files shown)


rajatjindal:
@eparis sorry for taking so long. I have broken original PR into two different PR:

1. Support bash aliases for the cobra commands (this PR)
2. Support cobra aliases in completions (will open this later as it needs more work for flaghash changes).

Can you please merge this PR if the changes looks ok.

rajatjindal:
Ping @eparis

eparis:
What exactly is this supposed to be giving me that I didn't have before?

rajatjindal:
if you create a command alias without this fix and try to use completion, it does not work.

alias aliasname=origname
complete -o default -F __start_origname aliasname

aliasname <tab><tab> will result in:

aliasname origname

and then

aliasname origname <tab><tab> will give you completion results.


after the fix:

aliasname <tab><tab> will give you correct completion results

this was reported here too: https://github.com/kubernetes/kubectl/issues/120

eparis:
Perfect, want to get that example somehow in the godoc and I'll merge?

rajatjindal:
sure thing. added the documentation for bash aliases

rajatjindal:
thank you for merging the PR

PR #642, opened 2018-02-28
Return error when usage was output due to non-runnable command

### LINKED ISSUE #582: Invalid command usage result code
I may be missing something obvious, but say you have the command structure:
```myapp foo bar```

If I type:
```myapp foo```

It will return the normal Usage output showing (snipped for brevity): ```Available Commands: bar```

This is fine, but in a shell script, I don't see how to catch the error cleanly because triggering ```Usage``` has a result code of 0, instead of a desired >=1.

```bash
myapp foo
if (( rc=$? )); then
  echo "[FAIL] myapp failed with rc==${rc}"
fi
```
This came up because one of the go apps that I use had ```foo```, but now (from an upgrade) it has been extended to ```foo bar``` and I didn't realize it until my administrative script started to fail unexpectedly.  It should have stopped dead at the ```myapp foo``` point, but an error didn't pop up until much later in the code.

chriswhelix:
This allows programs to detect running a non-runnable command as a failure condition and return a non-zero exit code, if they want to. For most existing programs, it will probably result in that behavior by default (assuming they exit non-zero whenever Execute() returns an error).

Usage output in this scenario is also now sent to stderr instead of stdout, which is better for scripting and seems more consistent with surrounding code.

Fixes #582

### DIFF
--- a/command.go
+++ b/command.go
@@ -23,10 +23,12 @@ import (
 	"os"
 	"path/filepath"
 	"sort"
-	"strings"
 
 	flag "github.com/spf13/pflag"
+	"strings"
 )
+
+var NotRunnable = errors.New("Command not runnable; need subcommand.")
 
 // Command is just that, a command for your application.
 // E.g.  'go run ...' - 'run' is the command. Cobra requires
@@ -714,7 +716,7 @@ func (c *Command) execute(a []string) (err error) {
 	}
 
 	if !c.Runnable() {
-		return errors.New("Subcommand required.")
+		return NotRunnable
 	}
 
 	c.preRun()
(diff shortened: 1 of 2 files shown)


bogem:
@chriswhelix Please resolve merge conflicts

bruceadowns:
I've encountered the same issue, and agree that if an incorrect subcommand is specified, a non-zero exit code should be returned. This PR has sat for a year and now has conflicts, however I am willing to create a new PR to restart the conversation. @bogem, what do you think?

bogem:
@bruceadowns it would be great if you reopen the conversation.

bruceadowns:
> @bruceadowns it would be great if you reopen the conversation.

Ok, I will put a PR together.

bruceadowns:
> @bruceadowns it would be great if you reopen the conversation.

Thanks @bogem! See new PR #922.

bogem:
Close in favour of #922

PR #643, opened 2018-02-28
Prefix bash functions with root command name

johnmccabe:
Prior to this PR the autocomplete bash functions were being prefixed with the root command name, but references to those functions from subcommands were having the subcommands prefixed instead - causing the function lookups to fail and error out.

As spotted by @janetkuo in kubernetes/kubernetes#60517:
```
kubectl create -f [Tab] failed with the following message: kubectl
create -f __create_handle_filename_extension_flag: command not found
```
in this case the function being invoked should be:
```
__kubectl_handle_filename_extension_flag
```
This PR switches to using `cmd.Root().Name()` in order to use the root rather than subcommand.

kubectl_autocomplete

### DIFF
--- a/bash_completions.go
+++ b/bash_completions.go
@@ -317,7 +317,7 @@ func writeFlagHandler(buf *bytes.Buffer, name string, annotations map[string][]s
 
 			var ext string
 			if len(value) > 0 {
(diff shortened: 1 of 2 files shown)


eparis:
Should we add a test to make sure that subdir completions will always call the rootdir handler?

johnmccabe:
Sounds reasonable

johnmccabe:
@eparis added some tests for subcommands, that look ok to you

PR #645, opened 2018-03-01
Update the bash completion example for kubectl

bamarni:
The current example is not working.


eparis:
I'm going to merge this, but I'd love love love if someone were to kill all of this and turn it into real golang compiled examples, without the reference the kuberntes. Just create their own program:
https://blog.golang.org/examples

PR #646, opened 2018-03-02
Zsh Completion

babysnakes:
I improved the existing zsh completion with support for flags and some of the custom completion commands from bash completion. A description of what's supported could be found i the [zsh completion readme][zshreadme]. Thanks to everyone in the #107 thread for their help, especially to @eparis for explaining things to me over and over again ...

One important note is that I only tested it on a small project on my laptop. It needs to be tested on a larger project and more important on other people's shell configurations to make sure it works correctly.

[zshreadme]: https://github.com/babysnakes/cobra/blob/zsh-completion/zsh_completions.md


babysnakes:
I encountered an issue (with a utility that I write) that affect completion. Not sure if/how to solve this issue. I have a `Version` configured on my root command so I'll have `--version` flag. To generate the completion I created a `completion` sub-command. The symptom was that when I ran `<mycmd> completion zsh --file ...` It didn't show completion for `--version`. This is because `InitDefaultVersionFlag` operates on the executed command and not on the root command and since my `completion` command doesn't have `Version` it doesn't generate it.

Should I at least point it out in the zsh completion readme?

eparis:
Does that mean that --help doesn't show up either? Do we want to make those late flag init calls a function, which we explicitly call from the generator command tree walk code?

babysnakes:
You're right about the `--help`, I was destructed by the fact that the `help` subcommand did show up but yeah, `InitDefaultHelpCmd` runs on the root command while `InitDefaultHelpFlag` and `InitDefaultVersionFlag` runs on the actual command you run. If there's a good reason it behaves that way and we don't want to change it then we can extract it to a separate function call.

babysnakes:
@eparis, it seems that the whole `--version` and `--help` thing is my error. I need a little time to do some more checks but I believe it's my fault 😞 . Should fix in a short while.

babysnakes:
There's something really strange here. On my tests (last commit: 70f3e35) it behaves correctly but on my real command that I'm using it doesn't. damn.

babysnakes:
Ok, sadly the problem still exist. The problem was in my test. Added a skipped test to show real issue. Back to square one in terms of dealing with `--version` and `--help`. @eparis, sorry for spamming.

breunigs:
I tried your patch for rclone, and it seems much more complete! I believe single quotes `'` in the descriptions are not escaped, which leads to misleading error messages (`/usr/share/zsh/vendor-completions/_rclone:1987: unmatched "` – the line doesn't have `"` and the *actual* broken single quote was somewhere around line 5000). Anyway, after manually replacing all occurrences of single quotes in rclone's descriptions, the completion works.

It seems that all subcommands (e.g. `rclone copy source:path dest:path [flags]`) are missing the file completion option. So, the `--some-flag` will autocomplete, but it will not suggest local files. For the bash completion this seems to work. I still have to checkout if this is maybe setup incorrectly in the command description on rclone's side.

babysnakes:
Thanks for the input @breunigs, Do you have your changes committed in a branch or should I just try modifying master? I'll try to check it later today.

breunigs:
Oh, rclone vendors cobra and updates its dependencies at the start of a new development cycle, so I can’t really commit that. It‘s enough to update the file (https://github.com/ncw/rclone/blob/master/vendor/github.com/spf13/cobra/zsh_completions.go), then run something like this:

```
go build && sudo rclone genautocomplete zsh
```

babysnakes:
@breunigs, can you try now with updated version? this should solve the quotes issue. Regarding filename completion, I'm not sure how bash-completion is implemented - maybe it allows completing filenames by default - however, the way I implemented it it will offer file completion only if you used one of `MarkPersistentFlagFilename` or `MarkFlagFilename`.

breunigs:
With your new version the completion works fine. I can also confirm that it correctly completes files when I annotate a flag properly.

However, I am still unsure how to turn on file completion for non-flag / positional arguments. From my testing with the bash completion it seems that it just uses local files as a fallback. Ideally I would be able to have something like `MarkFlagCustom`, `MarkFlagFilename` and so on for non-flag arguments, but I guess this is out of scope for this patch.

I am not sure what is better, though: the ZSH completion from this patch just offers flags and will autocoomplete a `-` when I press tab after a space. Bash will complete local files and only offer flags if I have `-<Tab>`.

In any case, looks good to me!

eparis:
In the bash completions if a flag is `MarkFlagRequired` then `[space][tab][tab]` will complete the required flag name(s) (or a subcommand). If no flags are marked required then you are right, you need `-[tab][tab]`.

babysnakes:
@breunigs, hehe, you're just in time 😄, I just need something like that for a project I'm working on and I'm doing some tests on it. It should be pretty easy but let me know what you think about this idea. You will have an option (via function call) to:

* Give a specific position argument (zero indexed)  the indication for file completion (e.g. something like `MarkZshArgumentFilename(index i, globs ...string)`.
* Give a specific positional argument a list of valid words to complete (e.g. `MarkZshArgumentWords(i int, words ...string)`)
* The above will be saved in `Command.Annotations`.
* If an index is left un-annotated (e.g. you marked argument 1 but not zero) it will not offer any completions for first argument but will offer completion for second argument.
* If no marking occurred but you have `Command.ValidArgs` it will used them as words to complete first argument.

This could be implemented pretty fast. Does it make sense?

babysnakes:
@breunigs, regarding the completion style for flags, I think in zsh it's also depends on your completion settings. I don't think I use any but there are multiple global options in zsh for all kind of completion styles.

babysnakes:
@breunigs, @eparis I added argument completion. It has the following characteristics (as specified the `zsh-completions.md` file):

* Argument position for all options below starts at `1`. If argument position
  `0` is requested it will raise an error.
* Use `command.MarkZshCompPositionalArgumentFile` to complete filenames. Glob
  patterns (e.g. `"*.log"`) are optional - if not specified it will offer to
  complete all file types.
* Use `command.MarkZshCompPositionalArgumentWords` to offer specific words for
  completion. At least one word is required.
* It's possible to specify completion for some arguments and leave some
  unspecified (e.g. offer words for second argument but nothing for first
  argument). This will cause no completion for first argument but words
  completion for second argument.
* If no argument completion was specified for 1st argument (but optionally was
  specified for 2nd) and the command has `ValidArgs` it will be used as
  completion options for 1st argument.
* Argument completions only offered for commands with no subcommands.

@eparis, one thing to think about is wether we want different functions to configure various aspects of completion in different shells or find a way that's defined once and used by all shells (like the flag annotations). I know nothing about bash completion so I don't really understand how you implemented bash completion for arguments.

breunigs:
Sorry for taking this long to reply. I tested your new version and all your features work as advertised. It definitely allows me to improve the current completion in rclone. However, to get a fully magic system, I think I will eventually need the ability to pass custom functions. Let me show this with an example:

```
rclone mount remote:path /path/to/mountpoint [flags]
↑ binary, works fine
       ↑ cobra command, works fine
             ↑ "remote" is a list of names that the user can configure. I 
               guess I can get away with reading those from the config at
               completion-generation time
             ↑ path depends on the remote chosen and potentially changes
               all the time. I definitely would need a custom function that
               calls into rclone to obtain a list to get those.
                         ↑ I'd like to only complete directories here
                                             ↑ flags, works fine
```

Regarding the directory completion: Interestingly I can pass `-/` and zsh (with my settings) will only complete directories, but also show `-/` as a potential directory entry. So zsh somehow interprets this both as an argument *and* a globbing pattern…? I am not sure what is happening here. Dropping the `-g` makes everything behave as expected.

Regarding the custom path completion: ideally this would work like completion for rsync does. I am not yet sure if this is even feasible from rclone side. It's an edge case for sure, so I'd be fine with being able to specify my own completion function that just gets called if we're at this position.

I guess if I am able to set arbitrary functions, directory only completion comes basically for free – I'd just have to pass it `_files -/` and no actual implementation, since zsh already has it. So maybe custom function is the decent way out of all these different edge cases I will run into with rclone?

Thanks a lot for putting work into this!

breunigs:
also, is there any chance of putting the shared/global flags in a variable and reuse that? rclone has a _lot_ of global flags, resulting in a ~500kB completion file. If you're curious, I uploaded the current export here: https://gist.github.com/breunigs/0b26415ad974114e4f049036935efe8b

babysnakes:
Thanks for the input @breunigs. Let's see:

* Regarding directories; I tried to pass `"-(/)"` as a pattern and it completed only directories on my settings - I actually tried it directly in a complete script, not via the generation but it should work the same.
* Regarding custom functions It's not complicated to add it, however I'm pretty sure it could be abused easily 🙂 to completely break the completion but you could say that this is the problem of the tool developer.
* Regarding shared flags in a variable, I have to check it. Not sure how it will behave.

I'm currently in a very busy couple of weeks but I'll try to check it. One other consideration though is some sort of API unity with bash completion. Currently there are some global _default_ api calls (e.g. `MarkFlagCustom`) and a others specifically for zsh. Maybe from an API standpoint it would be better to try to make it more unified (It would be tricky to do it without breaking backward computability though). @eparis, any thoughts on that?

eparis:
I was, admittedly, thinking only of myself when I wrote the bash completion generator :-(
I know nothing about how zsh completions work.
I also don't know all the places I forced bash-iness into the std API. I'm very sorry about that.

You gave MarkFlagCustom as an example, so I'll look just at that one. Can you point out the other places we need to consider?

I'm not sure why MarkFlagCustom is a bashism aside from the internal annotation name: `BashCompCustom`. Since the function is going to have to be provided by the opaque `BashCompletionFunction` (ooo, found another bashism in the core API) that doesn't seem to mean it couldn't be equally useful for zsh. In both cases I think we could pretty easily make a new annotation and a new API type which don't include `Bash` in the name. The bash completion code could look for the new generic name and then fall back to looking for the `Bash*` names if it didn't find the new one. But their functionality seems generic to me.

`BashCompletionFunction` is replaced with `CompletionFunctions` which does nothing but write the input to the output file.
`BashCompCustom` is replaced with `CompletionFlagFunctionName` or something like that. Which just is another name for the same thing.

Or are you saying that the idea of a custom function to complete a flag is just not a zsh thing at all?

babysnakes:
Sorry @eparis, this is not what I meant (admittedly I wasn't explaining myself very well 😞), and I certainly didn't mean it as a criticism for your work. What I meant was that currently the way I implemented the completion was to almost always choose a different system (function name and implementation) than yours , mainly because I know nothing about bash completion and I didn't want to spend much time trying to figure it out. In general, I think we can merge most of the other shells completions to one function and find a way to differentiate between the different shells (I hope more will be coming). Here are a few examples:

* `MarkFlagFilename` could easily be - _and it is_  - used for both. The caveat is that there are probably some patterns that are only applicable to certain shell, (e.g. following the example above, does `-(/)` pattern applies also to bash?) so I thought about the option of prefixing patterns with `"bash:*pattern"` and `"zsh:*pattern"` to indicate the specific shell patterns.
* `MarkFlagCustom` (or any other custom for that matter) is also the same, we can optionally prefix the string with a shell name so the differentiate if needed) but we may have to change the signature to make `f` variadic just so one will be able to add completion for several shells)
* One place where currently there is a difference is the arguments. As far as I can tell you only have `Command.ValidArgs` as completion options while I tried to offer the option for every positional argument (while using `ValidArgs` if exists and not overwritten). Maybe we can think of common function that we can both leverage from (at least optionally, we'll indicate in the docs what's supported and what's not).

The flip side of that is that it might be more confusing and maybe cause some zsh completions issues with existing applications (I still need to check what is the existing zsh doing and what can break).

So, what I meant was to ask you if you care to invest time with me to try and make it work (considering that I'm not an experienced programmer and I _will need help_)?

babysnakes:
@eparis I want to start implementing what @breunigs requested (custom completion) and I would appreciate your input on my last comment so I can try implementing this in a way that is more common for the two completion systems.

babysnakes:
Thanks for the detailed feedback 😃. I'll take a day or two to look at it in more depth but here are some thoughts that came to mind:

* Breaking backwards compatibility was never an option. However the issue is how to behave on existing code (especially if creating common API). It might be that some code (e.g. custom completion) that was written with bash in mind would break (or at least be un-usable) in zsh without any way to notify the _customer_ of the library (the tool maintainer) that this will not behave correctly unless he's trying it with zsh.
* I really like the idea of having more specific functions (e.g. `MarkFlagDirname`). Continuing with this paradigm I would create `MarkFlagCustomCommandCall` and `MarkArgumentCustomCommandCall` (or better names) to allow custom calling the command while only having to supply the parameters which will be easier on the tools author because it is likely the custom completion commands for _bash_ and _zsh_ would be different.
* I would like to offer the possibility of having shell specific code even if it will only be used on rare occasions. I can certainly see where custom completion code would have to differ between different shells (going forward I hope other shells will join). In such cases prefixing the string argument with the shell name might be possible solution.

Regarding lack of replies, I fully understand the lack of free time part 😃. However, I wouldn't like to spend a lot of time on a solution that I don't know if the maintainers agree with the direction I want to go. Since I know nothing about bash completion (I wrote one bash completion script more then 10 years ago and I remember it as a very un-pleasent experience) I would most certainly need active help and If no one is able to provide it then there's no point in doing half work on merging the _bash_ and _zsh_ APIs.

I'll try to do some work on one of your suggestions as an example and let's see.

breunigs:
You make good points. 

Regarding custom completion code written for bash, breaking on zsh: Hm, but that would only be an issue if we want to automatically reuse existing bash completions for zsh? Otherwise, it seems to me that deprecating  `BashCompletionFunction string` in favor of something like `ShellCompletionFunctions map[string]string` (or reuse `Annotations`) sidesteps the whole issue. Of course, the bash part would have to check both places and warn/fail if both are specified.

Regarding custom shell code: yes, I'm not advocating of not allowing that. I was arguing against allowing custom "code" for stuff like `MarkFlagFilename` and others. Having something like `MarkFlagCustomCommandCall` that accepts a shellname and some shell code seems quite decent.

hoshsadiq:
Any updates on this? I'm hoping to fix the Kubernetes implementation of zsh completion.

babysnakes:
Sorry, I was drowning in a big project at work which got longer and longer.  On my side I'll free up time next weekend to work on it. However I will need cooperation from the maintainers to really finish it.

eparis:
Oh wow, I thought we had merged all this zsh stuff. Whoops. @babysnakes let me know when you think it is good to go!

Note: the bash custom completion code made a mistake I do not want to repeat. In bash we hard code the name of the custom func as `__custom_func` (or something like that). So if you have loaded 2 difference cobra generated completion in your shell, for different binaries, one of them will overwrite the other. We either need the command name settable or we need it to include the Root().Name(), or something like that...

babysnakes:
Thanks @eparis, This is a good point I didn't think of. I'll make it include the command name.

Tomorrow I'll start implementing a merge of one or two of the completions command per @breunigs suggestions. Once I commit them please go over them, see of you like the implementation so we can continue merging the completion API.

babysnakes:
I only had very limited time to work on it but I wanted to provide at least a starting point for merging bash and zsh completion API so I'm providing this for @eparis/everyone to decide if you like it:

I followed @breunigs idea to simplify things for the user and create directory completion as a separate function than filename completions. I created `MarkFlagDirname` and `MarkPersistentFlagDirname` functions with currently only implementation for zsh to complete only directories. The idea is that we'll add _bash_ implementations in the same function. I also moved a few functions that should not be bash specific (e.g. `MakeFlagFilename`, `MakeFlagCustom`, etc) to a new file which should hold non shell specific functions (just so it will be more intuitive to find them). Last, I modified the function docs of a few functions to be more generic and added shell-completion compatibility matrix on each of them.

A few notes regarding going forward with unifying the API:

In the current implementation there are a few exported bash specific constants. While It would make sense in a shell agnostic implementation (like `MakeFlagFilename`) to not use bash specific annotation (`BashCompFilenameExt`) since it's exported it would probably break something to change it so we'll leave them for now.

Flag completions are more or less similar. While not all `MarkFlag...` functions are available in both shells (e.g. I'm not even sure required flag has meaning in zsh) they are similar in concept. Positional argument are currently completely different. From what I understand bash only offer word completion from single source (`Command.ValidArgs`) in _zsh_ I implemented several options for positional arguments (I'm also using `Command.ValidArgs` for word completion on first argument if not specified otherwise). @eparis do you want me to create stubs for all these functions which you can complete later?

In the meantime I'll try to continue working on some of @breunigs older requests.

Please provide feedback so I'll know how to continue.

babysnakes:
Anyone can direct me to an example of `MarkFlagCustom`? I'm trying to figure out how it works and I can't :(

babysnakes:
Implementing `MarkFlagCustom` proved to be a little tricky. From what I figured out (please correct me if I'm wrong) one adds his functions to `<RootCommand>.BashCompletionFunction` and provides the name of the function to call in `MarkCustomFlag`. This is an elegant solution and I wanted to integrate with it as it is. I've added `Command.ZshCompletionFunction` which will hold the _zsh_ custom completion functions. The idea is that one can provide two functions with the same name, one for _zsh_ and one for _bash_ and call `MarkFlagCustom` only once. The problem is implementing this in a way that:

* Will not break _any_ existing code.
* Will allow to implement only for one shell and not the others (hopefully they will come 🙂  ).

I could think of two ways to implement this:

1. In order to not call custom functions that do not exist for specific shell (wether it's because it's an old code or because it was not implemented for all the supported shells) the shell generator should search for the function name in the matching `...CompletionFunction` fields of the root command. If it doesn't exist it will not add it to the completion script. On the one hand this is by far a more elegant solution, however, in the (unlikely?) event of having custom function names implemented with variable (weird but possible) it will break.
2. Far less elegant option is to specify which shells have the named function implemented. This is done by prefixing the function with comma separated list of supported shells: `[zsh,bash]:__function_name`. If this prefix doesn't exist the old (only bash) system will be used. While being less elegant then the first solution, it should never break existing code.

I've implemented the second option but I'm not pushing it yet because I like the first one much more and I hope the maintainers will allow this small chance of breaking changes (and also I need guidance from @eparis modifying bash completion).

Last note for @eparis / the other maintainers. I spent a lot of time on this. I was happy to do it and I'll happily spend more time. However, I don't want to continue spending time without active input regarding the way to go (and active help when modifying bash completion code). I can certainly relate to not having any spare time, however, I also have very little time and I want to spend it where it will make a change...

Waiting for your input

babysnakes:
Hey @eparis, I'm in a vacation for a couple of weeks to work on my own stuff. This means I have free time to work on it now. It would be great if we can discuss the options I described above.

eparis:
For both bash and zsh, could we, at run time, call `declare -f custom_func && custom_func`? I just realized it is actually an array of functions, so that makes it a harder...

I think that will mean replacing https://github.com/spf13/cobra/blob/master/bash_completions.go#L93 and line 108 with something that does
```
funcs="${flags_completion[${index}]}"
OLDIFS="${IFS}"
IFS=';'
funcsExist=true
for f in ${funcs}; do 
  declare -F "${f}" || funcsExist=false
done
IFS="${OLDIFS}"
if [[ "${funcs_exist}" == true ]];
  $funcs
done
```
I'm sorry I keep dropping the ball on you. I'm happy to merge when you are with any solution! You're work has been amazing!

babysnakes:
Thanks @eparis 

I'm not sure I understand what do you mean when you say that it is an array of functions. In the [implementation](https://github.com/spf13/cobra/blob/a114f312e075f65bf30d6d9a1430113f857e543b/bash_completions.go#L582) it sets single element array which mean I can use `declare -f ...`. Not sure how to set the bash completion though 😞.

I'm happy to continue implementing more options but I will need your help with bash completion.

eparis:
The [bash generation code](https://github.com/spf13/cobra/blob/master/bash_completions.go#L335) does:
```
		case BashCompCustom:
			buf.WriteString(fmt.Sprintf("    flags_with_completion+=(%q)\n", name))
			if len(value) > 0 {
				handlers := strings.Join(value, "; ")
				buf.WriteString(fmt.Sprintf("    flags_completion+=(%q)\n", handlers))
			} else {
				buf.WriteString("    flags_completion+=(:)\n")
			}
```
Which looks to like you could, theoretically, manually set multiple custom functions in the annotations array. (I had no idea that code did that until yesterday). Which is why I'm suggesting we have to split on `;` and check that all of the functions exist. If you just did `declare -F func1;func2` it will test for func1 and then run func2. Which obviously isn't intended...

I think that allowing the bash and zsh functions to have the same name, and check at runtime (with declare) is the best we can do...

babysnakes:
Ahhh, I see, Damn 😞. Ok, I'll do my best trying to figure out what's going on there. I'll spend a few days next week on it and we'll see how far it goes.

jls5177:
@babysnakes,
Maybe the best approach to handle multiple completion functions for a single argument is to just build a custom function that handles calling each function and add it to the top of the completion file.

Here is a snippet for an example program I wrote to test this theory:
```
function __custom_funct {
  declare -f __some_random_func1 > /dev/null && __some_random_func1
  declare -f __some_random_func2 > /dev/null && __some_random_func2
}

function _example_test {
  _arguments \
    '(-n --no-run)'{-n,--no-run}'[Dry run]' \
    '(-v --verbose)'{-v,--verbose}'[Enable verbose output]' \
    '1: :__custom_funct'
}
``` 

I will put up a pull request on your repo later today with my proposed solution as I think this is a great solution to zsh completions and would like to use it in an upcoming project.

babysnakes:
Thanks @jls5177, It seems like your sample is handling zsh completion. In my implementation of zsh (not yet committed because of what I described above) I explicitly use only the first element. I don't see the point in having more then one command (except for backward compatibility with bash). If you need to combine the output of more then one command it's better to just add them to the custom command you're using. I'll also documented it in the function doc.

I will also add a simpler form (like `MakeFlagCustomCommandOutput`) in which you will only have to provide the actual command to run (e.g. call your executable with specific parameters) and I'll wrap it in a custom call.

Can you think of a reason to include multiple functions?

jls5177:
@babysnakes,
The question is really about user experience. Do we want the user to have to learn Zsh/Bash completion syntax or is the better solution to provide the hooks to load (and use) library functions that already do what they want. 

My current usecase is to load in the git zsh completion scripts to leverage the existing functions to parse  branch names (local and remote). For example, One of my arguments takes any known branch name (either local or remote). Now I could write this myself but I find it better to just leverage what is already there. 

Here is the snippet of example code that handles this. This is just sample code but the usecase is real.
```
# Import git shell completions
local -a locations
local e
locations=(
  '/usr/share/zsh/functions/Completion/Unix/_git'
  )
for e in $locations; do
  test -f $e && source $e && break
done

function _review {
...
}

function __custom_funct {
  declare -f __git_branch_names > /dev/null && __git_branch_names
  declare -f __git_remote_branch_names > /dev/null && __git_remote_branch_names
}

function _review_test {
  _arguments \
    '(-n --no-run)'{-n,--no-run}'[Dry run]' \
    '(-v --verbose)'{-v,--verbose}'[Enable verbose output]' \
    '1: :__custom_funct'
}
```

I agree we could go either route but I like the idea of having the tool generate this templating code (as it is always better at it).

jls5177:
The way I see this is the shell abstraction layer should hide as much of the gory details from the user as possible until the baked in completions no longer meet their needs. At this point we should try to minimize the users need to entirely switch to writing custom shell completion code as they would have to write code for each completion engine. Which is why I am in favor of having the tool generate the function that is capable of invoking multiple functions.

Now, this also drives a larger change in your Shell abstraction layer as (in its current form) is not generalized enough. I think we should provide a set of common completions and then each completion engine worries about how to handle them. Right now we have a unique set of bash completions and an entirely different set of zsh completions.

babysnakes:
@jls5177,

I think I see what you mean. However, the current implementation of `MarkFlagCustom` only accepts one function argument and the implementation discards the previous values. The only way to add more then one function is by directly adding annotation to the flag. Do you think we should change the `MarkFlagCustom` to accept multiple functions (something like `MarkFlagCustom(args, f ...string)`?)

babysnakes:
> Now, this also drives a larger change in your Shell abstraction layer as (in its current form) is not generalized enough. I think we should provide a set of common completions and then each completion engine worries about how to handle them. Right now we have a unique set of bash completions and an entirely different set of zsh completions.

This is exactly what I'm trying to achieve, but not in one go. If you followed the thread you could see that I had most of the functionality I needed in the first few commits but it was completely external to the bash completion. The main effort is to merge them to one coherent set of features that will hopefully be implemented in these (and maybe other) shells. The problem is that this will not happen in one go. I will try to make the effort to create a set of general functions which contains all existing functions + a few more. However, not all functions will have affect on all shells. It will take some time to implement everything if it will be even possible.

babysnakes:
Sorry everyone, my mother health deteriorated since the beginning of July and I spent my time nursing her. She passed away last week. I'll finish the "Shivah" (jewish mourning) and I'll work on it.

joshuarubin:
> She passed away last week

ברוך דיין האמת

MichaelMure:
@babysnakes as much as I would like to see your work land in Cobra, please take all the time you need. Don't feel obligated to work on it if you don't feel like it.

WGH-:
Thank you for your work. Even though it's not merged yet, your code is already incredibly useful for generating completion definitions, storing them as an asset, and amending them manually.

I noticed one small issue so far. When usage string contains multiple lines, descriptions looks very broken, even though `--help` itself looks fine and properly indented. I think the solution is to only include the first line of the usage in the description.

rsteube:
Awesome work.
I needed custom completion for positional arguments and persistent flags so i added that in a fork.
If it's any help you can see my changes here: https://github.com/babysnakes/cobra/compare/zsh-completion...rsteube:zsh-completion-custom

hoshsadiq:
Any updates on this?

babysnakes:
Hi @hoshsadiq and everyone else.

Sorry for disappearing but I can't seem to find enough time to get into _bash-completion_ (and I can't merge the completion of zsh and bash without it). I'm pretty fluent with zsh completion, but I tried several times already to understand how bash-completion works and I didn't manage to figure it out in the very limited time I had. The good news is that I found a team member who is ready to help so I hope to sync him on the status and hopefully he'll manage to take it further.

juliangilbey:
There's also a forked version of this at [github.com/rstuebe/cobra](https://github.com/rsteube/cobra) which is now being used by [github.com/zaquestion/lab](https://github.com/zaquestion/lab).  So it would be good to do a fair bit of merging when it's feasible...
Thanks!

rsteube:
Can't we get this merged somehow? I don't really understand whats the hold-up with bash completion since this targets the zsh completion and i don't expect it to solve everything at once.
As a user i'm already very happy with what is generated - be it the bash or zsh completion (apart from the small fixes i did).

I would rather like to see this merged in the current state and fixes be done in succeeding smaller MR's than having it lying around another year.
I can try to help if there is something to do.

rsteube:
Maybe a temporary alternative until the PR is merged: https://github.com/rsteube/cobra-zsh-gen

I extracted the zsh_completions.go changes and added a simple wrapper struct.
This way it can be invoked as an additional dependency without changes to the cobra codebase.

```go
import zsh "github.com/rsteube/cobra-zsh-gen
zsh.Wrap(RootCmd).GenZshCompletion(os.Stdout)
```
example: https://github.com/zaquestion/lab/compare/master...rsteube:use-cobra-zsh-gen

It's a bit crude but seems to work so far.

pior:
Thanks @rsteube

spf13:
@babysnakes @eparis Thanks for this great contribution. Sorry I sat on things for such a long time. I'm trying to be a better steward. Merging 👍

babysnakes:
Hey @spf13, 

I wasn't available for a few days. Thank you for merging it. Sadly, I'm not sure it was fully ready yet. I phased out of it when I realised I couldn't get any support with merging bash completion with zsh completion. I tried to find someone from work to help but it didn't work as well. In any case, I'm still ready to do the work on the `zsh` end if someone is ready to help me on the `bash` completion end. 

Thanks

corneliusweig:
Can you remind me why this affects bash completion? If you are aiming for a unified completion code, that's very much worth a separate PR IMHO. As a first step, we should get zsh completion right.

@babysnakes What in your opinion is missing to make it fully ready?

babysnakes:
@corneliusweig, my original idea was to get something working (that was many months ago). One of the first feedbacks I got (which I completely agree with) is that we better merge the completions so users will not have to work hard on configuring _bash_ and _zsh_  separately. I actually did the first steps (on my laptop, never committed it - check the long thread to see why) but really got stuck at trying to figure out bash completion.

The problem with merging this pull request before merging the completions is that now we have to worry about 3 APIs to be backward compatible (_bash_, _zsh_ and unified).

PR #649, opened 2018-03-19
terminates the flags when -- is found in commandline

dixudx:
Fixes kubernetes/kubernetes#61184

When `--` is found in the command line, the args come after should not be treated as regular flags.

Just like the example shown in kubernetes/kubernetes#61184, the `normalizeNameFunc` should be only applicable for args before `--`.

@eparis @bogem PTAL


bogem:
Thank you for your PR!

PR #657, opened 2018-03-27
Add CancelRun method to Command

kevincwells:
This method will nil out the Run and RunE of a command. This can be called from
PreRun-style functions to prevent the command from running.

Example: A user can check if the environment has been initialized in a PreRun
method and, if not, prevent the command from running.

Note that this does not nil out any PostRun-style commands, as these are
typically used for clean-up and tear-down operations that likely still need to
occur.

Signed-off-by: Kevin C. Wells <kevin.c.wells@intel.com>

### DIFF
--- a/command.go
+++ b/command.go
@@ -666,6 +666,13 @@ func (c *Command) ArgsLenAtDash() int {
 	return c.Flags().ArgsLenAtDash()
 }
 
+// CancelRun will nil out the Run and RunE of a command. This can be called from
+// PreRun-style functions to prevent the command from running.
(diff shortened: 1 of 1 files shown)


kevincwells:
Because the `execute` function currently does not re-check if `c.Run == nil` before executing it, the closest similar behavior without this patch would be:

```go
	...
	PreRun: func(cmd *cobra.Command, args []string) {
		if checkSomeCondition() {
			cmd.RunE = nil
			cmd.Run = func(cmd *cobra.Command, args []string) {
				// No-op
			}
		}
	},
```

PR #659, opened 2018-03-28
Fix viper flag

nseps:
Having a bool flag default to true makes no sense.


eparis:
Why? You can do --viper=false.

nseps:
Ah, thank you! I tried --viper false and failed.

PR #664, opened 2018-04-05
Add support for user specified args validation func (#663)

### LINKED ISSUE #663: feature: ValidateArgsFn
It would be great to have a feature where the user could pass an args validation function in the Command object.

In my scenario the user can pass a date string, such as "2018-04-05". Currently, I'm validating in the PreRun stage, but I believe a validation func is more expressive.

willis7:
Add an attribute (which is a function) to Command which is called before any Run steps. If an error is returned the command execution terminates.

Small change, looking for feedback before updating docs.

Closes #663

### DIFF
--- a/command.go
+++ b/command.go
@@ -84,6 +84,7 @@ type Command struct {
 	Version string
 
 	// The *Run functions are executed in the following order:
+	//   * ValidateArgsFn()
 	//   * PersistentPreRun()
 	//   * PreRun()
 	//   * Run()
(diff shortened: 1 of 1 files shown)


bamarni:
Why not using the `Args` field? It takes a `PositionalArgs` function, you can pass to it your own validation functions.

willis7:
Oh man, I completely missed that section in the docs. My bad!

PR #666, opened 2018-04-06
Changes NoArgs to report invalid argument

neurocline:
Updates NoArgs to also report invalid argument
    
NoArgs currently reports "unknown command" in all cases, which
can be is a little misleading to the user if there are no
subcommands. Change it as follows

- if subcommands and one extra arg, say "unknown command"
- otherwise, say "invalid argument(s)" and show all extra args

This should produce more meaningful error messages to users, and
works for both the root and sub-commands.

### DIFF
--- a/args.go
+++ b/args.go
@@ -2,6 +2,7 @@ package cobra
 
 import (
 	"fmt"
+	"strings"
 )
 
 type PositionalArgs func(cmd *Command, args []string) error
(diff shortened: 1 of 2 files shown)


neurocline:
Of course, I'm now having second thoughts. I think the behavior of NoArgs should be a little more nuanced. If you have no subcommands, it should definitely return "invalid argument". If you have subcommands and you have exactly one extra argument, it should probably say "unknown command", otherwise it would say "invalid arguments".

Also, it should show all the extra arguments, not just the first one.

neurocline:
I updated my pull request with the behavior outlined in the previous message.

pior:
👍

PR #667, opened 2018-04-07
Allow to override build date

bmwiedemann:
via environment variable
in order to make builds easily reproducible.
See https://reproducible-builds.org/ for why this is good
and https://reproducible-builds.org/specs/source-date-epoch/
for the definition of this variable.

Also use UTC to be independent of timezones.

Without this change, docker-17.09.1 package man pages
varied over time.


bmwiedemann:
ping. Is there anything I can do to help get this patch merged?

cyphar:
Looks like I just copied this patch for spf13/cobra#735. I can close that one if you prefer.

eparis:
Merged #735 because it didn't have conflicts

Issue #654, opened 2018-03-25
Short --version flag (-v/--version)


prologic:
I know `--version` was implemented originally in #584 and later fixed in #595 .

Can we have the option to have a short flag in addition to the long flag?

I would like (*for consistency's sake*):

```#!bash
mycli -v
```

and:

```#!bash
mycli --version
```

Thanks!

rajatjindal:
Hi @prologic 

-v is used by multiple popular projects to expose verbosity level of logs (kubectl, kops etc), and therefore I think adding -v for version will break them. 

disclaimer: i am not member of cobra project and this is just my view.

Thanks
Rajat Jindal

eparis:
Lots of project use --verbose/-v.  Such projects must continue to work.  (and @rajatjindal has contributed. which makes him a member of the cobra community if he likes it or not!)

You might get away with things by only adding the long+short version if both the long and short flags are not in use. And only creating the long version if `-v` is in use by another flag....

eparis:
https://github.com/spf13/cobra/pull/671 added `-v` unconditionally, I'm closing it. If someone adds -v anytime it is not used, I'll take that patch,

030:
@eparis Based on your comment this issue should be closed.

hairyhenderson:
@eparis I've attempted to implement what you're looking for in #996 - PTAL

PR #671, opened 2018-04-11
Add note about hiding a command from completion

eddiezane:


eparis:
While it is true that setting hidden will hide it from the bash completion list, it also hides it from generated docs, help text, etc. I'd probably rather we update the godoc text https://github.com/spf13/cobra/blob/master/command.go#L74 to describe what all it does rather than scatter it out in the bash completions doc...

eddiezane:
@eparis is there another method to exclude a command from completion? I fumbled around a bit until I found `Hidden`.

eparis:
I can't think of another way....

eparis:
See https://github.com/spf13/cobra/issues/654#issuecomment-381778043 for a mechanism I would accept,

eparis:
Whoops accidental close!

PR #673, opened 2018-04-12
 [fix bug: can not use cobra in project name endwith `cmd`] add func findCmdSuffix

bonfy:
If cobra init a project named endwith 'cmd', will result the problem, when add command, the `command.go` will create in the wrong place

```cmd
$ cobra init tcmd
$ cd tcmd
$ cobra add hello

```
hello.go create in the path `src/t/cmd/hello.go` not `src/tcmd/cmd/hello.go`

to fix this bug, add os.PathSeparator to the findCmdDir func, thus 

`p.absPath = strings.TrimSuffix(absPath, findCmdDir(absPath))`

can get the correct path

### DIFF
--- a/cobra/cmd/project.go
+++ b/cobra/cmd/project.go
@@ -92,7 +92,8 @@ func NewProjectFromPath(absPath string) *Project {
 	}
 
 	p := new(Project)
-	p.absPath = strings.TrimSuffix(absPath, findCmdDir(absPath))
+
+	p.absPath = findCmdPath(absPath)
 	p.name = filepath.ToSlash(trimSrcPath(p.absPath, p.SrcPath()))
 	return p
 }
@@ -128,6 +129,15 @@ func (p *Project) CmdPath() string {
 		p.cmdPath = filepath.Join(p.absPath, findCmdDir(p.absPath))
 	}
 	return p.cmdPath
+}
+
+// findCmdPath returns the cmd path.
+func findCmdPath(path string) string {
+	cmdDir := findCmdDir(path)
+	if filepathHasPrefix(cmdDir, string(os.PathSeparator)) {
+		return strings.TrimSuffix(path, cmdDir)
+	}
+	return strings.TrimSuffix(path, string(os.PathSeparator)+cmdDir)
 }
 
 // findCmdDir checks if base of absPath is cmd dir and returns it or


cobra/cmd/project.go
```diff
@@ -130,6 +132,14 @@ func (p *Project) CmdPath() string {
 	return p.cmdPath
 }
 
+// findCmdSuffix return the cmd dir start with file path separator
```
bogem: // findCmdSuffix returns the cmd dir start with file path separator.

cobra/cmd/project.go
```diff
@@ -92,7 +92,9 @@ func NewProjectFromPath(absPath string) *Project {
 	}
 
 	p := new(Project)
-	p.absPath = strings.TrimSuffix(absPath, findCmdDir(absPath))
+
+	cmdDir := findCmdDir(absPath)
```
umarcor: Since `cmdDir` is used only once, these two lines can be merged:

`p.absPath = strings.TrimSuffix(absPath, findCmdSuffix(findCmdDir(absPath)))`

cobra/cmd/project.go
```diff
@@ -130,6 +132,14 @@ func (p *Project) CmdPath() string {
 	return p.cmdPath
 }
 
+// findCmdSuffix returns the cmd dir start with file path separator.
+func findCmdSuffix(cmdDir string) string {
```
umarcor: This function seems not to find a suffix, but to just ensure that the string starts with `os.PathSeparator`. I think it should be renamed to reflect the functionality. Alternatively `p.absPath = strings.TrimSuffix(absPath, findCmdSuffix(findCmdDir(absPath)))` (see comment above) can be included in this function, so it actually does find and extract a suffix. If so, the line above would be: `p.absPath = findCmdSuffix(absPath)`.

umarcor:
Thanks @bonfy! Upon a further review, I don't know whether this fix applies anymore. The underlying issue might possibly have been solved since you opened this PR.

Precisely, `cobra init tcmd` is not valid now. You need to give a package name, e.g. `cobra init --pkg-name github.com/user/tcmd`. I tried the following:

``` bash
root@ed2404112dc6:~# mkdir tcmd
root@ed2404112dc6:~# cd tcmd
root@ed2404112dc6:~/tcmd# cobra init --pkg-name github.com/user/tcmd
Your Cobra applicaton is ready at
/root/tcmd
root@ed2404112dc6:~/tcmd# cobra add hello
hello created at /root/tcmdroot@ed2404112dc6:~/tcmd# ls -la 
total 28
drwxr-xr-x 3 root root  4096 Jul 12 02:04 .
drwx------ 1 root root  4096 Jul 12 02:04 ..
-rw-r--r-- 1 root root 11358 Jul 12 02:04 LICENSE
drwxr-x--x 2 root root  4096 Jul 12 02:04 cmd
-rw-r--r-- 1 root root   656 Jul 12 02:04 main.go
root@ed2404112dc6:~/tcmd# ls -la cmd
total 16
drwxr-x--x 2 root root 4096 Jul 12 02:04 .
drwxr-xr-x 3 root root 4096 Jul 12 02:04 ..
-rw-r--r-- 1 root root 1596 Jul 12 02:04 hello.go
-rw-r--r-- 1 root root 2814 Jul 12 02:04 root.go
```

Could check that the issue is still present in the latest `master`? If so, please rebase this PR on top of it.

@jharshman, line `hello created at /root/tcmdroot@ed2404112dc6:~/tcmd# ls -la` seems to be a bug. Can you identify at first sight where should it be fixed?

bonfy:
`cobra init tcmd` is not valid now, and the problem no longer exists.

jharshman:
@umarcor yeah it's a newline missing from https://github.com/spf13/cobra/blob/9a31ddff0ea7ee4045c71a113ff135917752383b/cobra/cmd/add.go#L67

PR #677, opened 2018-04-15
implement short -v version flag (issue #654)

ilmanzo:
added some tests to verify


ilmanzo:
Understandable.. but why not take this comment to the issue, so it can be useful to everyone? Thanks

rajatjindal:
thanks, added the comment there. adding disclaimer here too that I am not a member of Cobra project, and this is just my view, which Cobra maintainer may disagree with.

eparis:
See https://github.com/spf13/cobra/issues/654#issuecomment-381778043 for a mechanism I would accept.

hairyhenderson:
I've re-implemented this in #996 in a way that I think would work.

PR #686, opened 2018-04-24
doc: hide hidden parent flags

### LINKED ISSUE #685: Single hidden flag results in empty "Options inherited from parent commands" section in generated documentation.
The pflag FlagSet that generates usage information skips hidden commands. But cobra's Command.HasFlags returns true for flag sets that contain a single hidden command, resulting in sub-optimal doc generation.

The doc generator should (by default) follow in the steps of the FlagSet usage-generator of pflag and only consider those commands that are not hidden. If there are zero non-hidden commands, then the "Options inherited from parent commands" section should be omitted from the generated documentation.

jdef:
* fixes #685


jdef:
@bogem PTAL

Issue #655, opened 2018-03-26
Mandatory Flags not verified before Prerun.


amanhigh:
Presence of Mandatory Flags is not verified before Prerun unlike Args Check which happens before Prerun.

Hence can't safely use Mandatory Flags in Prerun without additional manual checks.

**Sample Code Demonstrating.**
```
package main

import (
	"fmt"
	"github.com/spf13/cobra"
)

var testFlag = ""

var rootCmd = &cobra.Command{
	Use:   "test",
	Short: "Test Command",
	PreRun: func(cmd *cobra.Command, args []string) {
		fmt.Println("Value for Mandator TestFlag is " + testFlag)
	},
	Run: func(cmd *cobra.Command, args []string) {

	},
}

func init() {
	rootCmd.Flags().StringVarP(&testFlag, "test", "t", "Nothing", "Test Flag")
	rootCmd.MarkFlagRequired("test")
}

func main() {
	rootCmd.Execute()
}
```

**Output**
Value for Mandator TestFlag is Nothing
Error: required flag(s) "test" not set
Usage:
  test [flags]

Flags:
  -h, --help          help for test
  -t, --test string   Test Flag (default "Nothing")

mattthym:
#747 Issues are related

kasvith:
I recently got into same error

relloyd:
I have a similar issue in that the configured default values for flags are not set if they are missing from the CLI.  Parsing fails with errors like the one above, "Error: required flag(s) "test" not set".  I would have expected the default value to be assumed for the flag if it's missing from the CLI.

I have to call `Flags.Set(<name>,<default value>)` even though I created the flag with a default value up front. 

Have I missed something obvious?

AdamSLevy:
This is old but still persists in the latest release. Marking a flag as required is effectively useless if you intend to use the flag value in the PreRun functions.

tasdikrahman-deliveryhero:
Hey folks, was wondering if this is by design (apologies if I have missed another issue in case it discusses the same). I came across this issue recently. 

To circumvent this issue, moved the flow added in `PreRun` to the `Run` flow, for the mandatory flags set to be checked.

tasdikrahman:
Hey folks, was wondering if this is by design (apologies if I have missed another issue in case it discusses the same). I came across this issue recently.

To circumvent this issue, moved the flow added in `PreRun` to the `Run` flow, for the mandatory flags set to be checked.

johnSchnake:
[...]

PR #700, opened 2018-06-04
#655 Validate Required Flags Before Prerun

amanhigh:
Validated Required Flags before Running Preruns. Any error in Prerun results in Validation Not Happening. As a result Required Flags can't be used in Preruns Safely.

Refer: #655

### DIFF
--- a/command.go
+++ b/command.go
@@ -736,6 +736,10 @@ func (c *Command) execute(a []string) (err error) {
 		return err
 	}
 
+	if err := c.validateRequiredFlags(); err != nil {
+		return err
+	}
+
 	for p := c; p != nil; p = p.Parent() {
 		if p.PersistentPreRunE != nil {
 			if err := p.PersistentPreRunE(c, argWoFlags); err != nil {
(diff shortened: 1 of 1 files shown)


MarkusFreitag:
Any estimate when this gets merged?

agrimmer:
I would also be interested if this gets merged?

amanhigh:
Can this be merged we want to use cobra in another project and this issue still exists. It would have been good if issue is fixed in official repo for all.

marckhouzam:
This look pretty straight forward. However I was wondering if checking the presence of required flag was purposely done after PreRun() for some reason?  For example, could `PreRun` be used by some programs to add a required flag in some special circumstances?  If so, this change would break such programs I believe.

Any opinion on the backwards-compatibility implications of this change @eparis?

wwillinghamWork:
Any word on whether this will get merged or not?

marckhouzam:
I agree with @johnSchnake's assessment that the risks outweigh the benefits.
We must remember that Cobra is used by hundreds if not thousands of projects; we therefore have to be extremely careful to avoid breaking anything.

Considering there was no disagreement with @johnSchnake's last comment for over 6 weeks now, I will close this.

PR #711, opened 2018-06-27
zsh-completion with description and flags

BrandonRoehl:
This new zsh-completion will give you

- Command description (if present)
- Flag completion
- Short version of flags (if present)
- Flag description (if present)

### DIFF
--- a/zsh_completions.go
+++ b/zsh_completions.go
@@ -1,10 +1,10 @@
 package main
 
 import (
-	"os"
-	"io"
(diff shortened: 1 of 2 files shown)


eparis:
I know #646 is waiting for my review. Any chance you want to look at what is there?

BrandonRoehl:
Never mind then I didn't see that one so you can ignore this version as that version is more robust

PR #712, opened 2018-06-28
fix test build breakage with go1.11

hyangah:
As discussed in golang/go#26109, vet typecheck is enabled
in go1.11 and the command_test.go can't be compiled any
more with go1.11 due to the unread variables in the
command_test.go.

Instead of removing the unused variables, this CL reads the
variables and compares the values against the current
behavior so when the related issue is fixed, this test can
be updated accordingly.

### DIFF
--- a/command_test.go
+++ b/command_test.go
@@ -1164,12 +1164,13 @@ func TestPersistentHooks(t *testing.T) {
 		t.Errorf("Unexpected error: %v", err)
 	}
 
-	// TODO: This test fails, but should not.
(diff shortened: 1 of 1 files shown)


eparis:
How does that cause a 1.11 breakage?

hyangah:
the variables were not read, which gc considers 'unused' vars.
From go1.11, that is a typecheck error and the test will not build.
More precisely speaking, the title is cobra test breaks when compiled with go1.11.

Issue #340, opened 2016-08-31
On error in RunE, do not display usage


vincentbernat:
Hey!

When an error is triggered by `RunE`, usage is displayed. This is a bit odd at this point, the usage is likely to be correct. I am using `RunE` to be able to get the errors from `Execute()` to display a detailed report. I have to set `SilenceUsage` in each `RunE` command. Is that the expected workflow?

eparis:
sadly this is expected because any change might break people. It probably about time someone with time called this v1 of cobra and we start cleaning up mess like this in a v2....

broady:
Yeah, `SilenceUsage: true` in the root command is likely always correct.

vincentbernat:
Thanks for the tips, both `SilenceUsage: true` and `SilenceError: true` on the root works OK for me. I still get usage with `--help` or when subcommand is missing. I print errors myself. Should I close the issue since there is no obvious next step?

eparis:
I'd be fine with a change that made it respect SilenceError on a missing submodule...

apriendeau:
Okay. I will work on a fix for that tonight. 👍

apriendeau:
@vincentbernat I am unable to recreate the issue. When I use `SilentErrors`, it silences the out put of from the missing subcommand. Do you have any example code where this happens?

vincentbernat:
I was unclear. My initial report was about the fact that you have to use `SilenceErrors`/`SilenceUsage` to not display usage when there is an error in `RunE`. By using those, I have to handle error display myself too. However, usage and help are still working for me, either when asked explicitely or when a subcommand is missing.

So far, there is no bug for me anymore. I would prefer to not have to use `SilenceErrors` but I understand we cannot break other people setup.

jmalloc:
Sorry to hijack, but I believe this is related (enough) :) -- As best I can tell, `SilenceUsage` silences the usage text even when the error is produced by the built-in argument and flag validation, before `RunE` is reached. I had incorrectly assumed usage would shown in that case, but not when a "generic application error" is returned.

Perhaps this could be controlled explicitly on a per-error basis by optionally wrapping errors in an error type that is checked alongside the `SilenceUsage` flag in `Command.Execute()`. I'm imagining something that looks like `return cobra.SilenceUsage(err)` from a `RunE` function.

lehors:
Hi all,
Having faced the same issue I'd like to share that in the end I think the current behavior is fine.

[...]

PR #713, opened 2018-07-04
[issue #340] Don't display usage on a returned error by a RunE function

pierre-emmanuelJ:
Signed-off-by: Pierre-Emmanuel Jacquier <pierre-emmanuel.jacquier@epitech.eu>

I want to solve this issue:
https://github.com/spf13/cobra/issues/340

This is a good point ! I have also add a flag `ShowRunErrorUsage` to do like before  !


tmc:
This should make it in -- this is unexpected behavior as-is.

Issue #563, opened 2017-11-02
How can I pass context.Context between pre/post-run hooks?


binary132:
In my pre-run hook, I'd like to set a value (someplace; maybe in the Command?) and have access to it in my `Run`.  This seems like a good use-case for `context.Context`.  Is this supported somehow?  I was thinking of setting a Flag value, but that seems awfully hackish, and I really don't want to use a global variable.

dnephin:
The way I've seen this done is to always wrap the `cobra.Command{}` in a function, which accepts the context object. I haven't seen `context.Context` used before, usually it's a more specific type defined by the application. The function wrapping the creation of the `cobra.Command` accepts an instance of the type, so it exists in the scope for both `Run` and `PreRun`.

```go

func newXCommand(context CustomContext) *cobra.Command {
    return &cobra.Command{
        ...
    }
}
```

In `main()` you would create a new context and pass it to every function that creates a new command.

binary132:
What do you think about offering a Context version of the lifecycle functions (Pre / Post) with a `Command.Context() context.Context` method to retrieve the current Context (or nil)?  The API for that could look a bit like this:

```go
const (
    Response = iota
    Note
)

var preCmd = &cobra.Command {
    // ...
    PersistentPreRunCtx: func(cmd *cobra.Command) context.Context {
        rsp, _ := http.Get("https://google.com")
        return context.WithValue(context.WithTimeout(
            // Response Body must be read within 5 seconds
            cmd.Context(), 5*time.Second,
        ), Response, rsp.Body)
    },
}

// preCmd is the parent of storeCmd
var storeCmd = &cobra.Command {
    // ...
    PreRun: func(cmd *cobra.Command) context.Context {
        return context.WithValue(cmd.Context(), Note, "hi there")
    },
    RunCtx: func(cmd *cobra.Command) context.Context {
        rc := cmd.Context().Value(Response).(io.ReadCloser)
        // db.WriteAll(uuid.NewV4(), rc)
    },
}
```

Each lifecycle hook, if it returns a Context, could be threaded with something like:

```go
func ThreadCtx(cmd *Command, fs ...func(cmd *Command) context.Context) {
    c := context.Background()
    for _, f := range fs {
        c = f(cmd)
        select {
        case <-c.Done():
            // fail gracefully
        case err := <-c.Err():
            // fail less gracefully
        default:
            // keep looping
    }
}
```

flimzy:
[...]

PR #727, opened 2018-08-12
Add minimal Context support to cobra

flimzy:
RFC.  This PR is meant to demonstrate a possible, minimal solution to #563.

I welcome feedback, suggestions, and alternatives.

### DIFF
--- a/command.go
+++ b/command.go
@@ -17,12 +17,14 @@ package cobra
 
 import (
 	"bytes"
+	"context"
 	"fmt"
 	"io"
 	"os"
 	"path/filepath"
 	"sort"
 	"strings"
+	"sync"
 
 	flag "github.com/spf13/pflag"
 )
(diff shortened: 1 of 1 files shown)


aidansteele:
I think this would make a great addition to cobra. When I was playing with the functionality introduced in your PR, something caught me: I had assumed -- perhaps naïvely -- that a parent command's context would "trickle down" into subcommands. I tried accessing a value stored in the root command's context by accessing `child.Context()` and it wasn't there.

What semantics do you think make the most sense? I'm happy to hear that my initial assumptions were silly, but I thought it'd be worth a discussion.

flimzy:
@aidansteele How are you trying to set context in the root command, then access in the child command? Can you show some sample code?  I'm not quite visualizing what you're doing, so it's hard to comment.

marckhouzam:
PR #893 which has been merged has brought context support to Cobra.

This PR can therefore be closed /cc @jpmcb

johnSchnake:
Per comment; closing.

Issue #694, opened 2018-05-21
Multiple commands' custom completion functions collide


Shamus03:
## The issue

Custom completions use a function named `__custom_func`, which does not change between commands.  Each cobra utility defines a `__custom_func` for its own use, but these multiple commands all overwrite the same function.

## Expected behavior

Multiple cobra utilities' `__custom_func` definitions should not collide/overwrite each other.

## Steps to reproduce

Create two cobra utilities with separate `__custom_func` definitions as described [here](https://github.com/spf13/cobra/blob/master/bash_completions.md).  Generate their bash completion scripts and place them in the proper location.  Reload your shell and type `type __custom_func` to get the definition, and notice that one of the commands' completions will not work.  Removing one of the bash completion scripts will ensure the other one works, and vice-versa.

## Suggested solution

In `__handle_reply`, before `__custom_func` is evaluated, the script should search for a function named `__custom_func_COMMAND` where COMMAND is the name of the root command (making it somewhat unique) and evaluate that if it exists.  To ensure backwards compatibility, `__custom_func` should be run if no matching function is defined.

pgwhalen:
@eparis This should probably be closed after my fix in #730

Shamus03:
Thanks!

PR #730, opened 2018-08-20
Qualify custom bash func name

pgwhalen:
- fixes issue where multiple cobra apps using custom bash completion
 would have their __custom_func collide
 - support fallback to plain __custom_func to maintain compatibility

#694 

I tested this with a cobra app I'm developing locally and had success.  I'm not totally sure how write automated tests for it other than the changes I made.

### DIFF
--- a/bash_completions.go
+++ b/bash_completions.go
@@ -129,7 +129,13 @@ __%[1]s_handle_reply()
     fi
 
     if [[ ${#COMPREPLY[@]} -eq 0 ]]; then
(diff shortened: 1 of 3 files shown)


eparis:
I think the best way to test would be to define a custom func and then just do a string search to see if it shows up twice in the output...  (Once for the actual func def, and once for the call to it?)

Also need the CLA.

pgwhalen:
Signed CLA and improved the tests.

PR #734, opened 2018-08-22
opinionated documentation about versioning

ahammond:
It took me a while to figure out how to support both a version cmd and a --version flag as well as the go build flags necessary to get them to work. Maybe this will help the next guy?

### DIFF
--- a/README.md
+++ b/README.md
@@ -291,6 +291,13 @@ import (
   "github.com/spf13/cobra"
 )
 
+var (
+	Version    string
+	BuildStamp string
+	SHA        string
+)
+
+
 func init() {
   rootCmd.AddCommand(versionCmd)
 }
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -300,11 +307,43 @@ var versionCmd = &cobra.Command{
...
+}
+```
+
+And build it passing `-ldflags -X full/path/to/cmd.Version=v0.0.0`.
+For example:
+
+```bash
+PROJECT_LC=$(echo "${CIRCLE_PROJECT_USERNAME}" | tr '[:upper:]' '[:lower:]')
+BASE="github.com/${PROJECT_LC}/${CIRCLE_PROJECT_REPONAME}/cmd"
+LDFLAGS="-X $BASE.Version=$(git describe --tags) -X $BASE.BuildStamp=$(date -u '+%Y-%m-%dT%H:%M:%SZ') -X $BASE.SHA=$(git rev-parse HEAD)"
```
the-gigi: Is that relying on every PR being tagged with the version?

ahammond: If you follow the example code, you'll get a Version that looks like `v2.3.0-2-ga4bf952` which translates to ${last tag}-${number of commits since the last tag}-g${sha or last commit}, unless you are actually on a tag... in which case you get that tag. I think it's pretty commonly accepted best practice to git-tag before building binaries for general consumption, but not doing that doesn't break this. It's ugly, but then so is releasing untagged code.

heindl: I like where this is going, but maybe go:generate with a post commit would be a better option? Then the executable would not rely on remote environment configuration, and you still could use `go get github.com/user/repo/cmd`.

In fact, this could ship by default with the Cobra Compiler, so that a version_static.go file could be generated after each commit.

somewhere.go:
// go:generate cobra --version --root-command-variable=rootCmd

Then:
go generate (includes the same tag generation logic you wrote)
git add /cmd/version_static.go
git commit --amend --no-edit

What do you think? Also, why did you include the number of commits since the last tag. Is that canonical somewhere?

ahammond: @heindl I'm pretty new to golang, so I don't understand your comment about `go:generate`. It would be really nice to have this be environmentally independent.

I know `go get` is convenient, but it's also kinda evil in that it encourages un-pinned dependencies. `curl -fsSL -o /usr/local/bin/dep https://github.com/golang/dep/releases/download/${DEP_VERSION}/dep-linux-amd64` is a lot less likely to yield surprises. My hope would be that people publish their binaries as github release artifacts (or wherever they choose to publish their artifacts) on tag inside their CI pipeline.

The number of commits since the last tag is just part of the default output of `git describe` https://git-scm.com/docs/git-describe So, pretty standard, but maybe not quite canon.

heindl: Thanks for your comment!

Go generate is a tool for generating new code, partially meant to overcome the lack of generics.

I agree that cross-compiled binaries are the best option. Check out https://goreleaser.com/. But since the advent of modules in the last Go release, I think go get is also a safe option.

ahammond: `goreleaser` looks pretty good! I've been using https://github.com/tcnksm/ghr for uploading binaries, but I might switch. Example: https://github.com/ahammond/ipcheck/blob/master/.circleci/config.yml#L53 I have reservations about mixing build and upload though. I think that people might reasonably want to run integration tests after building their binaries.

One thing I do like is that goreleaser is exporting 3 variables by default into `main`. I couldn't figure out how to put these variables in main for `ipcheck` so they're under `cmd`, which is why the huge `BASE` variable in the code above. I'd like to refactor my cobra usage so that I can -L these variables into main instead, but like I said, I'm still pretty new to golang. Do you know how I can do that?

heindl: Hey, sorry for the late reply! Not being notified about comments.

Ja I agree about the benefit of running integration tests on the binaries. I assumed though that Travis or CircleCi would have room for this in the build process. Honestly I only discovered Go releaser last week and haven't used it yet, but looking forward to with new projects. 

Partially that's also why I'm interested in extending Cobra - in order to have a more efficient open source build process. Tools for go static code analysis make a basic code review process almost automatic. And cross-platform compilation may enable a universal package manager.

You're asking how are they updating those three variables? I took a quick scan and not exactly sure, but again, they may be rewriting them directly in the code before rebuilding. It looks like they run go generate in the make process.

ahammond:
@heindl I'm not sure where to go with this or if you're still interested. I'm not writing golang stuff these days, so... I'll leave it to you.

PR #757, opened 2018-10-04
Add flag that gives the ability to disable bash default completion

geoand:
`-o default` supplied to `complete` forces bash to provide the default suggestions when the custom bash completion scripts do not provide any values (as is described [here](https://www.gnu.org/software/bash/manual/html_node/Programmable-Completion-Builtins.html)).
This is a sensible default, but in a of lot does not make sense and can potentially confuse uses of applications using this feature.

To remedy the situation a flag is provided that can turn of the bash default generation when enabled


geoand:
I'm not really sure what to make of the test failures, I can't really tell if they are related or not...

geoand:
@eparis Could you please have a look?

Thank you

marckhouzam:
IIUC this PR gives the option to turn off default completions suggestions all the time.  Default completions are the list of file in the current directory.

I think this ability would be nice but I think it should be on a case by case basis.  Showing local files is wrong  for some commands, but for others it is correct and important.

I've posted a PR that allows to write custom completions in Go (https://github.com/spf13/cobra/pull/1035) and it provides the ability to turn off default completions on a per-command basis: https://github.com/spf13/cobra/pull/1035/files#diff-c07279f68bfb9c503f680286cc72dd88R155

marckhouzam:
With Cobra 1.0, it is now possible to disable file completion on a per command basis.
Please see `ShellCompDirectiveNoFileComp` in this section of the documentation:
https://github.com/spf13/cobra/blob/master/bash_completions.md#1-custom-completions-of-nouns-written-in-go

I believe this PR can therefore be closed.

marckhouzam:
> With Cobra 1.0, it is now possible to disable file completion on a per command basis. Please see `ShellCompDirectiveNoFileComp` in this section of the documentation: https://github.com/spf13/cobra/blob/master/bash_completions.md#1-custom-completions-of-nouns-written-in-go
> 
> I believe this PR can therefore be closed.

Issue #756, opened 2018-09-29
Cobra and VGO?


andrewmeissner:
I wonder if I'm maybe missing something...

When I run `cobra help init`, this is what I see:

```
PS C:\Users\Andrew> cobra help init
Using config file: C:\Users\Andrew\.cobra.yaml
Initialize (cobra init) will create a new application, with a license
and the appropriate structure for a Cobra-based CLI application.

  * If a name is provided, it will be created in the current directory;
  * If no name is provided, the current directory will be assumed;
  * If a relative path is provided, it will be created inside $GOPATH
    (e.g. github.com/spf13/hugo);
  * If an absolute path is provided, it will be created;
  * If the directory already exists but is empty, it will be used.

Init will not use an existing directory with contents.
```
If I'm reading that correctly, I should be able to run `cobra init myApp` and have cobra create the structure in the current directory.  This isn't the case.

```
PS C:\Users\Andrew\Desktop\definitelyNotGOPATH> pwd

Path
----
C:\Users\Andrew\Desktop\definitelyNotGOPATH


PS C:\Users\Andrew\Desktop\definitelyNotGOPATH> cobra init myApp
Using config file: C:\Users\Andrew\.cobra.yaml
Your Cobra application is ready at
C:\Users\Andrew\go\src\myApp.

Give it a try by going there and running `go run main.go`.
Add commands to it by running `cobra add [cmdname]`.

PS C:\Users\Andrew\Desktop\definitelyNotGOPATH> echo $env:GOPATH
C:\Users\Andrew\go

PS C:\Users\Andrew\Desktop\definitelyNotGOPATH> cobra init .\rightHerePlease
Using config file: C:\Users\Andrew\.cobra.yaml
Error: Cobra supports project only within $GOPATH: Rel: can't make C:\Users\Andrew\Desktop\definitelyNotGOPATH\rightHerePlease relative to
```

It seems to always create it in the `GOPATH`.  I'm hoping to be able to use go 1.11's dependency manager (vgo) with cobra, vgo doesn't work in the `GOPATH` and it looks like cobra only works _in_ the `GOPATH`.  Am I missing something blazingly apparent?  If not, are there plans to make cobra compatible with go 1.11's built in dependency manager?

Thanks!

jharshman:
Cobra generator only works in your `GOPATH`.  Since I was moving from dep to modules (vgo), I just generated what I needed on my project in the `GOPATH` and copied it over.  Hacky but it works.

[...]

PR #762, opened 2018-10-11
756-vgo-support

jharshman:
# Moves to resolve Issue #756 
This is a small change set which allows the cobra generator to work outside of the GOPATH by setting the flag `--vgo=<package>`.   If run without the `--vgo` flag, the generator flow remains the same.

## Changes
* Add a `--vgo` root persistent flag taking a string representing the package name.
e.g: `--vgo=github.com/spf14/hugo`
* If `--vgo` flag is set, allow init of project outside of `GOPATH`
* For adding commands, if you are in the project directory when using `--vgo`, there is no need to specify package name.  If you are outside the project directory, you must specify `--package` equal to the relative directory of your current working directory.

## Future improvements
* Cobra relies on the `GOPATH` to be able to derive the location of packages. I would like to see the weight on this dependency lighten as it is phased out of use in the language.
* I think it's not needed to be able to add commands to projects from anywhere on the system. It's not a feature I really use and am curious if removing it would cause kickback from the community.


andrewmeissner:
Hi @jharshman, it looks like circleci is reporting a failing check.  the go tests pass on my local box with your changes.  I'm not sure what else I might be missing, but from my end it should be passing.  :/

jharshman:
@andrewmeissner I'm not sure why there is a failing check there.  Looks like it has something to do with bash completions. Will have to investigate this further.

andrewmeissner:
@jharshman just wanted to check on this.  Did anything show up in the bash completions investigation?

Thanks again!

jharshman:
@andrewmeissner Hadn't had time to do take an in-depth look due to the holidays.  I'll give it a go in the next couple days and see what turns up.

jharshman:
@andrewmeissner ah I see the fix - it's a formatting error.  I'll fix it real quick.

andrewmeissner:
Hi again, @jharshman , just wanted to follow up on this.

tsal:
This could be a valid workaround for #795 ?

andrewmeissner:
@tsal seems like it.  As long as cobra can work with modules - I'm a happy camper.  :)

PR #772, opened 2018-10-19
Bump Travis versions and let patch versions float freely

HaraldNordgren:


HaraldNordgren:
It appears that by fixing the formatting for Go 1.11 I broke it for older versions. Not quite sure how to fix that.

PR #775, opened 2018-10-21
Quote bash variables in bash_completion.go, fixes #766

### LINKED ISSUE #766: bash_completion code should quote variables
The bash code template in https://github.com/spf13/cobra/blob/master/bash_completions.go contains a lot of unquoted variables which will cause many errors for people with spaces in their env variables.  A particular case I ran into was a person with a space character in their $PATH which caused a `/bin/bash: line 0: test: -ge: unary operator expected` error.  

Quoting variables is also recommended in bash style guide like [Google's](https://google.github.io/styleguide/shell.xml?showone=Variable_expansion#Variable_expansion).

Brcrwilliams:
Quote bash variables to prevent errors with variable expansion. Fixes issue #766.

### DIFF
--- a/bash_completions.go
+++ b/bash_completions.go
@@ -24,7 +24,7 @@ func writePreamble(buf *bytes.Buffer, name string) {
 	buf.WriteString(fmt.Sprintf(`
 __%[1]s_debug()
 {
-    if [[ -n ${BASH_COMP_DEBUG_FILE} ]]; then
+    if [[ -n "${BASH_COMP_DEBUG_FILE}" ]]; then
         echo "$*" >> "${BASH_COMP_DEBUG_FILE}"
     fi
 }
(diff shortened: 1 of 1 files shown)


bash_completions.go
```diff
@@ -104,31 +104,31 @@ __%[1]s_handle_reply()
...
     # we are parsing a flag and don't have a special handler, no completion
-    if [[ ${cur} != "${words[cword]}" ]]; then
+    if [[ "${cur}" != "${words[cword]}" ]]; then
         return
     fi
 
     local completions
     completions=("${commands[@]}")
-    if [[ ${#must_have_one_noun[@]} -ne 0 ]]; then
+    if [[ "${#must_have_one_noun[@]}" -ne 0 ]]; then
```
eparis: For my own knowledge, do `${#` really need quoting?

Brcrwilliams: I don't think so, but it doesn't hurt to have them.

PR #782, opened 2018-11-05
Have CalledAs function return the name the cmd was called as; weather or not it was run.

gdey:
Currently `CalledAs` returns the name the command was called as, only
if it was executed; the parent command doesn't get run. This
change allows the child to know how the parent was called as well.

Added new function Called() will let's the user know if the cmd was
called or not.

Updated tests accordingly.


gdey:
For one of our commands, we want to have the following call structure:
```bash
tegola cache seed tile-name z/x/y
```
or

```bash
tegola cache purge tile-name z/x/y
```

Requiring that the `tile-name` command be able to understand if the `seed` command was called as `seed` or `purge.` Unfortunately, the `CalledAs()` function checks to see if the command was run before returning the actual name the command was called with. This pull request changes it so that the function always returns the name the command was called with. If you want to find out if the command was run, the new function `Called()` will return this value.

PR #784, opened 2018-11-06
Add setters for default help strings

ameyer-pivotal:


ameyer-pivotal:
Test failure seems unrelated to this PR. Can someone double check?

brianstrauch:
Any update on this?

brianstrauch:
@spf13

abitrolly:
This will crash for default help, because default help is defined only when command is executed, and will be `nil` before that.

https://github.com/spf13/cobra/blob/e2570c90ff5e22963ec876dd65bf03cdadb5839d/command.go#L957

PR #794, opened 2018-11-27
Fix too many underscore for __custom_func

dtaniwaki:
I found evaluating the function existing for `__custom_func` has too many leading `_`.


dtaniwaki:
I also fixed the typo of too many leading underscores in the docs.

PR #801, opened 2018-12-27
zsh: sort the commands to keep a determinist output

MichaelMure:
It's a minor thing but it has become an annoyance as I have the bash and zsh completion generation hooked in a `go:generate` directive.

Currently the zsh completion doesn't sort the commands. As go doesn't guarantee a deterministic iteration over a map (and actually scramble it on purpose), this means that the zsh completion file can be different with each build.

This PR fix that.

### DIFF
--- a/zsh_completions.go
+++ b/zsh_completions.go
@@ -5,6 +5,7 @@ import (
 	"fmt"
 	"io"
 	"os"
+	"sort"
 	"strings"
 )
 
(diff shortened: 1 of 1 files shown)


MichaelMure:
Test failure are unrelated to this PR.

zsh_completions.go
```diff
@@ -81,16 +82,24 @@ func writeLevel(w io.Writer, root *Command, i int) {
 	commands := filterByLevel(root, i)
 	byParent := groupByParent(commands)
 
-	for p, c := range byParent {
-		names := names(c)
-		fmt.Fprintf(w, "      %s)\n", p)
+	// sort the parents to keep a determinist order
+	parents := make([]string, len(byParent))
```
umarcor: Is it possible to use the following syntax instead?

``` go
	parents := make([]string, 0)
	for _, p := range byParent {
		parents = append(parents, p)
	}
	sort.StringSlice(parents).Sort()
```

spf13:
Can you update this PR to merge cleanly with the improved zsh support?

MichaelMure:
The ZSH completion has been fully rewritten so I'm not sure this apply anymore. It's not obvious to me if the code is deterministic or not now. I'll have to test in my own project.

MichaelMure:
It's no longer needed since the rewrite.

PR #807, opened 2019-01-12
Fix two word flags

dtaniwaki:
Flags like `--namespace` can also be a two-word flag, but the current completion considers only shorthand flags as two-word flags.

I also added test cases for two-word flags.


dtaniwaki:
@spf13 Could you review this PR?

PR #809, opened 2019-01-14
added variable to allow configuration of mousetrap message duration

thewilli:
New variable `MousetrapDisplayDuration` allows to modify the default display duration of 5s, or to completely disable the timeout and wait for the user to press the return key.

This allows for greater flexibility, instead of having to define a custom `preExecHook` for even such a small change. The requirement arose as in a UAT few users weren't able to read the message within the small default time span of only 5s. IMHO, if you wanted to keep the Mousetrap feature included, it should be configurable at least at a bare minimum.

This PR does not change the default behaviour.

### DIFF
--- a/cobra.go
+++ b/cobra.go
@@ -23,6 +23,7 @@ import (
 	"strconv"
 	"strings"
 	"text/template"
+	"time"
 	"unicode"
 )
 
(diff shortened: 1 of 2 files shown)


eparis:
@jharshman reviewed. Thank you!

PR #811, opened 2019-01-18
fixed typo in cobra readme

artlovan:
Fixed typo showing cobra commands. Cobra "add" command is written as "app".


umarcor:
IMHO the README is currently ok and this PR is invalid. However, because `app` and `add` are quite similar, it can be misleading indeed. I suggest renaming `app` to some other word such as `APP` or `myApp`.

jharshman:
The intent was to correct a spelling mistake but I think the author misunderstood the ReadMe.
It intends to convey adding a command to an app called "app".  Not the actual command `cobra add subcommand`.

PR #814, opened 2019-01-24
Custom Help verb

syntaqx:
Allows for the `help` verb used on the default command and flags to be changed.


syntaqx:
I'm a bit confused at what test is failing here, if someone has some more knowledge about what this is that would be helpful, but I'll keep looking.

syntaqx:
Looks like it was an issue with `fmt` on tip vs older versions. Seems that it's fixed in `master`, so it's been pulled in. Do you want me to rebase/squash the commits, or are we good to go?

jharshman:
@syntaqx Is there a custom help verb that you prefer or one that you need to fit a use case?

syntaqx:
I'm building a tool that's basically a joke tool, so this lets me change it to `halp`. It's a weird usecase, but non-invasive to the project so I figured it wouldn't hurt to send a PR for the changeset

jharshman:
While it is a very small change set, I don't think it's the best fit for upstream.
@eparis your opinion?

spf13:
I like that this is being used for a joke, but I can't think of why anyone else would need this. I'd rather keep things simpler. The only think I can think of is part of localization.

Issue #770, opened 2018-10-18
Document a best practice for testing a cobra app


sethcenterbar:
Hey there, I'm quite new to go and cobra, but I'm leveraging the framework to build some tools for work and loving it. 

I've been trying to set up testing for my project, but I'm wondering how other people using cobra are implementing testing for sub-commands. The typical go example led me to believe that in the cmds folder, i would make a command x.go and then create a test x_test.go. Because of the way that all of the commands are passed to the root command, I don't believe this is possible. 

To other users, how would you suggest doing testing on your sub-commands? If this isn't just a noob misunderstanding, maybe it would be good to add a best practice for testing into the main readme?

shilangyu:
Could anyone share how they write tests for sub-commands? I am also stuck i find it really hard to write any kind of test for a more complex sub-command. Thanks

jsenon:
I've found an example here https://chromium.googlesource.com/external/github.com/spf13/cobra/+/refs/heads/master/command_test.go

lpabon:
If you are coming here from Google, in our project we wrapped the var and init sections of the commands in functions so that they can be called by unit tests and re-initialize all variables.

* [Example command](https://github.com/portworx/px/blob/cb2e24b09cd1dbeadd3cf5f736882af56fae7110/cmd/version.go#L30-L55)
* [Registration of the command](https://github.com/portworx/px/blob/cb2e24b09cd1dbeadd3cf5f736882af56fae7110/cmd/main.go#L116-L147)
* [Changed the Execute function to call Main](https://github.com/portworx/px/blob/cb2e24b09cd1dbeadd3cf5f736882af56fae7110/cmd/main.go#L106-L111)
* [Now unit tests can call executeCli](https://github.com/portworx/px/blob/cb2e24b09cd1dbeadd3cf5f736882af56fae7110/cmd/common_test.go#L99-L103) to test their calls

tiagotele:
Hi! I think you can get inspiration from who are using Cobra like [helm]https://github.com/helm/helm/tree/master/cmd/helm) or [kubectl](https://github.com/kubernetes/kubernetes/blob/master/pkg/kubectl/cmd/cmd_test.go)

igomez10:
Hi! @lpabon your links are broken. Could your reupload? :pray:

marcellodesales:
@lpabon Maybe the repo `px` is private?

red8888:
[...]

Issue #763, opened 2018-10-14
Command.SetErrOutput(io.Writer) for more finegrained testing


edio:
> Initially I commented on #607 , but I then found out, that I can't reopen it, hence I decided to create a dedicated issue. I apologize if that's an unwanted noise.

Right now `Command` offers `SetOutput` method to set alternative `io.Writer` implementation for testing stdout/stderr interaction.

I think it would be great to be able to distinct standard output and error output here.

It is not uncommong for commands to report warnings to stderr. For example, take GNU's `ls`

```
$ ls -l /does/not/exist /tmp
ls: cannot access '/does/not/exist': No such file or directory
/tmp:
total 27776
...
```

If I were to reimplement this in go, I'd want to be sure, that `ls: cannot access ...` won't get to stdout, as it may be piped further. So I'd like to have test for that. With current _cobra_ code it is not possible to test this behavior.

It looks like a very straightforward and simple change and we can even make it backward compatible with current API.

If there's interest and/or maintainers are open for this change, I'd be willing to work on it.

jleni:
This is definitely necessary. 
I would even go as far as adding a third method to redirec `sdtin` too. 
Sometimes, CLIs need some input and we definitely need all these three to properly test things.
btw, this other open issue proposes the same. https://github.com/spf13/cobra/issues/658

Issue #658, opened 2018-03-28
Consider backing `OutOrStdout` / `OutOrStderr` with 2 different writers


bamarni:
After stumbling upon `OutOrStdout` and `OutOrStderr`, I thought they'd be great for testing purposes. From my command I would get stdout / stderr writers by calling them and in tests I could just inject some `bytes.Buffer` to assert on outputs.

However there is a single `SetOutput` in the Command, which has the following comment : `SetOutput sets the destination for usage and error messages.` So I assume this refers to stderr.

Would you consider allowing two writers there?

Some others also seem to be bit by this, in kubectl for example they have to pass the 2 writers explicitly (to properly test command outputs I assume) : https://github.com/kubernetes/kubernetes/blob/e00a60283f87db1bd8db0b80bec622353cf88f96/pkg/kubectl/cmd/auth/cani.go#L84

johnSchnake:
Seems fixed by #822 which added a SetErr and an ErrWriter. Closing.

PR #822, opened 2019-02-11
Extending redirection to stdout, stderr, stdin

jleni:
A couple of issues in the past have already requested the ability of setting stdio (in, out, err): https://github.com/spf13/cobra/issues/763 https://github.com/spf13/cobra/issues/658

In addition, there are some discussions about testing cobra based applications https://github.com/spf13/cobra/issues/770

This PR tries to address this by allowing redirection of I/O at the command level allowing for better unit testing.

```
	mockOut := bytes.NewBufferString("")
	mockErr := bytes.NewBufferString("")
	myCmd.SetIn(mockIn)
	myCmd.SetOut(mockOut)
        ....
        err = runMyCmd(myCmd, []string{})
        .... 
 	assert.Equal(t, expectedOut, myCmd.String())
	assert.Equal(t, expectedErr, myCmd.String())
```

`stdin` is also considered as some CLI may request data (such as passwords, confirmations) that may also need to be redirected when testing.

### DIFF
--- a/command.go
+++ b/command.go
@@ -384,13 +384,22 @@ func (c *Command) Help() error {
 	return nil
 }
 
-// UsageString return usage string.
+// UsageString returns usage string.
 func (c *Command) UsageString() string {
+	// Storing normal writers
 	tmpOutput := c.outWriter
+	tmpErr := c.errWriter
+
 	bb := new(bytes.Buffer)
 	c.outWriter = bb
+	c.errWriter = bb
+
 	c.Usage()
+
+	// Setting things back to normal
 	c.outWriter = tmpOutput
+	c.errWriter = tmpErr
+
 	return bb.String()
 }
 
--- a/command.go
+++ b/command.go
@@ -177,8 +177,6 @@ type Command struct {
 	// that we can use on every pflag set and children commands
 	globNormFunc func(f *flag.FlagSet, name string) flag.NormalizedName
 
-	// output is an output writer defined by user.
-	output io.Writer
 	// usageFunc is usage func defined by user.
 	usageFunc func(*Command) error
 	// usageTemplate is usage template defined by user.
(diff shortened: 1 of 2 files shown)


command.go
```diff
@@ -206,7 +211,25 @@ func (c *Command) SetArgs(a []string) {
 // SetOutput sets the destination for usage and error messages.
```
eparis: Is this comment still true?

jleni:
I kept the old behavior (backwards compatibility) and updated the comment to indicate the newer alternative methods.

syndbg:
It would be very nice if this gets merged :)

Testability of stdin is currently a huge issue. And as we know so far, stderr separation is also nice to have.

jackzampolin:
@eparis We would love to get this merged as well so that we can migrate some of our tests to this format.

jackzampolin:
@eparis Looks like this PR has some pretty broad support and a defined usecase! When do you anticipate this will be included in a release? 🙏

jharshman:
This looks good to me but would be nice to have some tests.

eparis:
I guess I said I'd merge this a long time ago. Anyone willing to commit to writing a unit test like @jharshman mentioned?

alessio:
@jharshman @eparis test cases have been added

jharshman:
Thanks for adding some test cases. This looks good to me.

alessio:
Bump (gently)

command.go
```diff
@@ -331,11 +386,11 @@ func (c *Command) Help() error {
 
 // UsageString return usage string.
 func (c *Command) UsageString() string {
-	tmpOutput := c.output
+	tmpOutput := c.outWriter
 	bb := new(bytes.Buffer)
-	c.SetOutput(bb)
+	c.outWriter = bb
```
jamesdphillips: It appears as if Usage could write to Stderr. Should we be setting `errWriter` as well?

jleni: yes, to keep things backwards compatible this is necessary. I have improved this and added a new test

alessio:
@jharshman @eparis @jamesdphillips All comments have been addressed

alessio:
(gently) Bump

We'd appreciate if could review this again now that all comments have been addressed.

We really need this feature to makes our life easier and avoid dirty hacks while writing test cases for our sub-commands. As a last resort, we could fork cobra though we're committed to first try everything to avoid such duplication of code and effort.

Thanks for considering.

@jamesdphillips @eparis @jharshman

fedekunze:
is there anything blocking this PR from being reviewed again ?

alessio:
@eparis bump (gently) :)

PR #828, opened 2019-02-23
Derive zsh completion from bash completion

corneliusweig:
This is a temporary fix to get some zsh completion. The idea is taken from kubectl where the bash completion script is wrapped with some zsh adapter code. The zsh magic is taken from kubectl.

There is an old PR #646 where a native zsh completion should be implemented. Sad thing is, there has been no progress for more than half. Since there should be a working out-of-the box solution, we should use this as an temporary drop-in until there is a better native zsh implementation.

Also see discussion in #107


spf13:
@corneliusweig I merged #646 today. Is this still relevant at all?

corneliusweig:
@spf13 No that completely supersedes my port. Thanks for the reminder!

PR #839, opened 2019-03-18
[Fix] [Travis] Execute 'go vet', use stages, use language conf list, add go 1.12.x, update shellcheck

umarcor:
This PR includes several minor enhancements to `.travis.yml`:

- Execute 'go vet' instead of 'go tool vet', to fix the error with go `>1.11` (i.e. `tip` and/or `1.12`).
- Move the job that gets `diff` of `gofmt` to a separate stage in Travis, because it is not a test.
- Use language configuration list instead of explicit entries in `matrix.include`. This is a more natural way to handle the config in Travis.
- Run tests against go `1.12.x`.
- Update `shellcheck-docker` to `v0.6.0`.

This PR conflicts with #832. I can either rebase on top of @peter-edge's branch, or let him rebase on top of this. @eparis, wdyt?

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,11 +1,16 @@
 language: go
 
+stages:
+  - diff
+  - test
+
 matrix:
   include:
     - go: 1.10.x
     - go: 1.11.x
     - go: tip
-    - go: 1.11.x
+    - stage: diff
+      go: 1.11.x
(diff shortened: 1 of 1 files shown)


umarcor:
Closed in favour of #840.

PR #774, opened 2018-10-20
Print errors to stderr in `cobra init` boilerplate

Brcrwilliams:
This has been a minor annoyance for me with the Cobra codegen tool. Error messages and debugging information ("Using config file:") should be printed to stderr and not stdout.

PR #840, opened 2019-03-18
[TrivialPatches] Typos in README.md, fix and update CI, update projects list...

umarcor:
This PR collects several other PRs which are considered to be trivial, as an effort to help the maintainers in getting some work out of their tasklist.

Close #769 (@ariefrahmansyah)
Close #777 (@jayunit100)
Close #779 (@indradhanush)
Close #806 (@saromanov)
Close #826 (@linzhaoming)
Close #831 and Close #832 and Close #833 (@peter-edge)
Close #839
Close #786 (@praveensastry)
Close #824 (@joshuabezaleel)

Furthermore, the projects that use Cobra are shown separated by commas, instead of a long vertical list.


umarcor:
The latest commit adds colored tests to Travis:

- Before: https://travis-ci.com/umarcor/cobra/jobs/185790913
- After: https://travis-ci.com/umarcor/cobra/jobs/186111030

cobra/cmd/init.go
```diff
@@ -210,7 +210,7 @@ func initConfig() {
 
 	// If a config file is found, read it in.
 	if err := viper.ReadInConfig(); err == nil {
-		fmt.Println("Using config file:", viper.ConfigFileUsed())
+		fmt.Fprintln(os.Stderr, "Using config file:", viper.ConfigFileUsed())
```
eparis: Why is this stderr?

umarcor: That's because of #774:

> This has been a minor annoyance for me with the Cobra codegen tool. Error messages and debugging information ("Using config file:") should be printed to stderr and not stdout.

eparis: that's informational, not an error though...

eparis: I think that one thing is what's preventing me from merging. I don't like those 2 lines.

umarcor: I think that it belong to stderr too, because that info does not belong to the logic of the app/command. Instead, it belong to the *middleware/library* that is handling configs/flags/options.

Nevertheless, since the main point of this PR is to merge trivial non-opinionated patches, I will just remove it from the set. Then, #774 can be handled independently.

umarcor: @eparis, it is removed now.

PR #841, opened 2019-03-18
Generalize ValidArgs: use it implicitly with any validator

### LINKED ISSUE #838: Feature request: support MatchValid(PositionalArgs) OR MinimumValidNArgs, MaximumValidNArgs, RangeValidArgs
For consistency, I think it would be useful to have `MinimumValidNArgs(int)`, `MaximumValidNArgs(int)` and `RangeValidArgs(min, max)` added to the list of supported `Args` formats. The complete list would then be:

- NoArgs
- ArbitraryArgs
- OnlyValidArgs
- MinimumNArgs(int)
- +**MinimumValidNArgs(int)**
- MaximumNArgs(int)
- +**MaximumValidNArgs(int)**
- ExactArgs(int)
- ExactValidArgs(int)
- RangeArgs(min, max)
- +**RangeValidArgs(min, max)**

This is because AFAIK it is not possible ATM to combine e.g. `MinimumNArgs(1)` with `OnlyValidArgs`.

---

An alternative solution is to provide a built-in function such as `matchAll` by @Antolius in [#745](https://github.com/spf13/cobra/issues/745#issuecomment-441326468). For example:

``` go
func MatchValid(p cobra.PositionalArgs) cobra.PositionalArgs {
	return func(cmd *cobra.Command, args []string) error {
		if err := cobra.OnlyValidArgs(cmd, args); err != nil {
			return err
		}
		if err := p(cmd, args); err != nil {
			return err
		}
		return nil
	}
}
```

which would allow the following expressions:

- `Args: cobra.MatchValid(cobra.MinimumNArgs(1)), ValidArgs: []string{...}`
- `Args: cobra.MatchValid(cobra.MaximumNArgs(5)), ValidArgs: []string{...}`
- `Args: cobra.MatchValid(cobra.RangeArgs(1, 5)), ValidArgs: []string{...}`
- `Args: cobra.MatchValid(cobra.ArbitraryArgs), ValidArgs: []string{...}`
- `Args: cobra.MatchValid(cobra.ExactArgs(1)), ValidArgs: []string{...}`

Certainly, `MatchValid` can be used implicitly when `ValidArgs` is defined, so the user is not required to type it. Therefore, `OnlyValidArgs` and `ExactValidArgs` can be deprecated from the public API. The list of formats would be:

- NoArgs
- ArbitraryArgs
- MinimumNArgs(int)
- MaximumNArgs(int)
- ExactArgs(int)
- RangeArgs(min, max)

umarcor:
Fix #838

This PR is based on #840 and 5 commits are added on top of it. 4 of them are about refactoring `args_test.go`. The fifth one (`feat: generalize ValidArgs; use it implicitly with any validator`) moves the validation of `ValidArgs` from `args.go` to `command.go`. As a result:

- Any validator can be used along with `ValidArgs`. `ValidArgs` is checked first, and then the defined validator.
- `OnlyValidArgs` and `ExactValidArgs` are deprecated.

`args_test.go` is updated accordingly:

``` bash
=== RUN   TestNoArgs
--- PASS: TestNoArgs (0.00s)
=== RUN   TestNoArgsWithArgs
--- PASS: TestNoArgsWithArgs (0.00s)
=== RUN   TestNoArgsWithArgsWithValid
--- PASS: TestNoArgsWithArgsWithValid (0.00s)
=== RUN   TestArbitraryArgs
--- PASS: TestArbitraryArgs (0.00s)
=== RUN   TestArbitraryArgsWithValid
--- PASS: TestArbitraryArgsWithValid (0.00s)
=== RUN   TestArbitraryArgsWithValidWithInvalidArgs
--- PASS: TestArbitraryArgsWithValidWithInvalidArgs (0.00s)
=== RUN   TestMinimumNArgs
--- PASS: TestMinimumNArgs (0.00s)
=== RUN   TestMinimumNArgsWithValid
--- PASS: TestMinimumNArgsWithValid (0.00s)
=== RUN   TestMinimumNArgsWithValidWithInvalidArgs
--- PASS: TestMinimumNArgsWithValidWithInvalidArgs (0.00s)
=== RUN   TestMinimumNArgsWithLessArgs
--- PASS: TestMinimumNArgsWithLessArgs (0.00s)
=== RUN   TestMinimumNArgsWithLessArgsWithValid
--- PASS: TestMinimumNArgsWithLessArgsWithValid (0.00s)
=== RUN   TestMinimumNArgsWithLessArgsWithValidWithInvalidArgs
--- PASS: TestMinimumNArgsWithLessArgsWithValidWithInvalidArgs (0.00s)
=== RUN   TestMaximumNArgs
--- PASS: TestMaximumNArgs (0.00s)
=== RUN   TestMaximumNArgsWithValid
--- PASS: TestMaximumNArgsWithValid (0.00s)
=== RUN   TestMaximumNArgsWithValidWithInvalidArgs
--- PASS: TestMaximumNArgsWithValidWithInvalidArgs (0.00s)
=== RUN   TestMaximumNArgsWithMoreArgs
--- PASS: TestMaximumNArgsWithMoreArgs (0.00s)
=== RUN   TestMaximumNArgsWithMoreArgsWithValid
--- PASS: TestMaximumNArgsWithMoreArgsWithValid (0.00s)
=== RUN   TestMaximumNArgsWithMoreArgsWithValidWithInvalidArgs
--- PASS: TestMaximumNArgsWithMoreArgsWithValidWithInvalidArgs (0.00s)
=== RUN   TestExactArgs
--- PASS: TestExactArgs (0.00s)
=== RUN   TestExactArgsWithValid
--- PASS: TestExactArgsWithValid (0.00s)
=== RUN   TestExactArgsWithValidWithInvalidArgs
--- PASS: TestExactArgsWithValidWithInvalidArgs (0.00s)
=== RUN   TestExactArgsWithInvalidCount
--- PASS: TestExactArgsWithInvalidCount (0.00s)
=== RUN   TestExactArgsWithInvalidCountWithValid
--- PASS: TestExactArgsWithInvalidCountWithValid (0.00s)
=== RUN   TestExactArgsWithInvalidCountWithValidWithInvalidArgs
--- PASS: TestExactArgsWithInvalidCountWithValidWithInvalidArgs (0.00s)
=== RUN   TestRangeArgs
--- PASS: TestRangeArgs (0.00s)
=== RUN   TestRangeArgsWithValid
--- PASS: TestRangeArgsWithValid (0.00s)
=== RUN   TestRangeArgsWithValidWithInvalidArgs
--- PASS: TestRangeArgsWithValidWithInvalidArgs (0.00s)
=== RUN   TestRangeArgsWithInvalidCount
--- PASS: TestRangeArgsWithInvalidCount (0.00s)
=== RUN   TestRangeArgsWithInvalidCountWithValid
--- PASS: TestRangeArgsWithInvalidCountWithValid (0.00s)
=== RUN   TestRangeArgsWithInvalidCountWithValidWithInvalidArgs
--- PASS: TestRangeArgsWithInvalidCountWithValidWithInvalidArgs (0.00s)
=== RUN   TestRootTakesNoArgs
--- PASS: TestRootTakesNoArgs (0.00s)
=== RUN   TestRootTakesArgs
--- PASS: TestRootTakesArgs (0.00s)
=== RUN   TestChildTakesNoArgs
--- PASS: TestChildTakesNoArgs (0.00s)
=== RUN   TestChildTakesArgs
--- PASS: TestChildTakesArgs (0.00s)
```


jharshman:
IMO this PR overreaches a little over what was originally intended.  Changing to Go modules is great but not required for this feature.

I think there is a case for improvement here but perhaps it can be approached differently.  Issue https://github.com/spf13/cobra/issues/838 mentions that there is not currently a method of combining these validation functions.  A fix for this could to introduce a method of chaining them.

umarcor:
> IMO this PR overreaches a little over what was originally intended. Changing to Go modules is great but not required for this feature.

This PR was based on #840 because I expected that branch to be merged before this, so it could be fast-forwarded. #840 provides some interesting features, such as fixing CI, which would help evaluate this PR.

Anyway, I have now made this PR independent from #840.
 
> I think there is a case for improvement here but perhaps it can be approached differently. Issue #838 mentions that there is not currently a method of combining these validation functions. A fix for this could to introduce a method of chaining them.

I think that it does not make sense to combine any of NoArgs, ArbitraryArgs, MinimumNArgs, MaximumNArgs, ExactArgs, RangeArgs, because all of them are mutually exclusive. Instead, my proposal is to apply ValidArgs (when defined) to the selected validator. That's why OnlyValidArgs and ExactValidArgs are deprecated.

jharshman:
Thanks for making this independent of #840 .

Maybe some clarification on the goal here was needed because it seems to have strayed away from the original issue you referenced.  That issue proposed the addition of validation functions which were combinations of what is currently available. Hence why I suggested adding a method to allow the chaining of validation functions.

However, I think I see where you're coming from.  You want to remove `OnlyValidArgs` and `ExactValidArgs` and instead have a check within the validation function to decide whether or not the argument slice matched some provided valid arguments.

However, this check inside the validation function contains the same logic as `OnlyValidArgs`.  You could perhaps keep `OnlyValidArgs` and therefore only have to depreciate one function instead of two.

umarcor:
@jharshman, I had the same point of view as you when I started to implement this. As I went forward and I saw the codebase, I concluded that the proposed solution is the best. Some hints to better understand the changes:

- This 7 lines are moved from https://github.com/spf13/cobra/blob/master/args.go#L36-L42 to https://github.com/spf13/cobra/pull/841/files#diff-9c42b524cb14ca001c61267826cbefb1.
- As a result, `OnlyValidArgs` and `ExactValidArgs` are not required anymore (https://github.com/spf13/cobra/pull/841/files#diff-07ee57a49d1c0b692bd79676cc31dd6f) because `ArbitraryArgs` and `ExactArgs` do now check `ValidArgs`.
- Moreover, `ValidArgs` is now checked with no regard to the chosen validator.
- This is explained in the README: https://github.com/spf13/cobra/pull/841/files#diff-04c6e90faac2675aa89e2176d2eec7d8
- All the changes in `arg_test.go` are to validate the points above: https://github.com/spf13/cobra/pull/841/files#diff-748c41438b4135ae7035c2af1228e349 (see also https://travis-ci.org/spf13/cobra/jobs/508608833#L543-L602)

> Maybe some clarification on the goal here was needed because it seems to have strayed away from the original issue you referenced. That issue proposed the addition of validation functions which were combinations of what is currently available. Hence why I suggested adding a method to allow the chaining of validation functions.

The issue requested some method to *combine e.g. MinimumNArgs(1) with OnlyValidArgs*. That's possible with the current proposal:

``` go
Args: cobra.MinimumNArgs(1),
```

compared to the *prototype* suggested in #838:

``` go
Args: cobra.MatchValid(cobra.MinimumNArgs(1)),
```

Moreover, in #838 it was suggested that it would be better not to require the user to type `MatchValid`:

> Certainly, MatchValid can be used implicitly when ValidArgs is defined, so the user is not required to type it. Therefore, OnlyValidArgs and ExactValidArgs can be deprecated from the public API.

---

> However, I think I see where you're coming from. You want to remove `OnlyValidArgs` and `ExactValidArgs` and instead have a check within the validation function to decide whether or not the argument slice matched some provided valid arguments.

All of the *original* validators care about the number of arguments that a command accepts:    NoArgs, ArbitraryArgs, MinimumNArgs(int), MaximumNArgs(int), ExactArgs(int), RangeArgs(min, max). We can consider it an enumeration of 6 items.

When `ValidArgs` was introduced, up to 6 additional validators might have been added, in order to provide functions covering the full combination matrix. Certainly, `ValidArgs` is a boolean condition which is orthogonal to `Args`, because it is focused on the content of the args, not the number of them. Therefore, the full matrix covers `6 x 2 = 12` combinations. However, `NoArgs + ValidArgs == NoArgs + !ValidArgs`, so 11.

Nonetheless, the implementation was partial and a single one was introduced. Later, a second one was added. Still, there are 3 combinations missing, which is what was pointed out in #838 at first.

Hence, the design decission is to provide one explicit validator for each combination of the multiple criterias to filter the arguments (original approach), or to enforce the orthogonality in the code (proposed approach).

> However, this check inside the validation function contains the same logic as `OnlyValidArgs`. You could perhaps keep `OnlyValidArgs` and therefore only have to depreciate one function instead of two.

It is possible to make `OnlyValidArgs` an *alias* of `ArbitraryArgs` and `ExactArgs` an *alias* of `ExactValidArgs`, so that bacward compatibility is retained. I don't think it makes sense to deprecate only one of them (see comment above).

command.go
```diff
@@ -877,6 +877,13 @@ func (c *Command) ValidateArgs(args []string) error {
 	if c.Args == nil {
 		return nil
 	}
+	if len(c.ValidArgs) > 0 {
```
jharshman: What do you think about putting this logic into an un-exported function?  You could then invoke it like:
```
if err := validArgs(c, args); err != nil {
  return err
}
return c.Args(c, args)
```

jharshman:
@umarcor thanks for taking the time to detail your approach and solution.  Looks like one of the tests doesn't like the `go tool vet` command.
```
+vet: invoking "go tool vet" directly is unsupported; use "go vet"
```

umarcor:
> @umarcor thanks for taking the time to detail your approach and solution. Looks like one of the tests doesn't like the `go tool vet` command.
> 
> ```
> +vet: invoking "go tool vet" directly is unsupported; use "go vet"
> ```

This is why I originally based this PR on top of #840. Those CI issues are fixed there, as commented [above](https://github.com/spf13/cobra/pull/841#issuecomment-474590215). The commit that fixes it is precisely: https://github.com/spf13/cobra/pull/840/commits/bbec970304026225a1f5fab1583b2c09d2bd9f0f. Nonetheless, I think that it would be better to use `golangci-lint`. That's what I'm doing in my fork (https://github.com/umarcor/cobra/commit/3b6028420526f3f1efad98a589e59ca3a00b943d), while I wait for maintainers to respond.

command.go
umarcor: I think it is not required for now. Because this block is used in a single place in the code base, adding a branch through a call to a function adds unneccesary complexity to the code. It is likely to be optimized by the compiler, so it should have little to no impact on performance. But I would anyway wait until it is really required. Nonetheless, a comment can be added just before line 880, to let readers know that the block can be considered an atomic/independent operation.

umarcor: @jharshman, after your comment, I think that it can be improved. ATM, two for loops are used. One in `command.go` and another one in `cobra.go` (`func stringInSlice`). `stringInSlice` is only used twice: in `command.go` and `command_test.go`. So, we can merge `stringInSlice` in `command_test.go`, and replace the block you commented about with the following:

``` golang
package main

import (
	"fmt"
	"strings"
)

func main() {
	fmt.Println("Hello, playground")
	
	ValidArgs := []string{"one", "two", "three"}
	Args := []string{"three", "one", "a"}
	
    e := sliceContains(ValidArgs, Args)
    if e != nil {
	  fmt.Println(*e)
	}
}

func sliceContains(s, x []string) *string {
  const p := "$"
  for _, a := range x {
    if !strings.Contains(p+strings.Join(s, p)+p, p+a+p) {
      return &a
    }
  }
  return nil
}
```

jharshman: Looks good but what's the purpose of ` (p+strings.Join(s, p)+p, p+a+p)`? 
Why not `!strings.Contains(strings.Join(s,""), a)`?

umarcor: We need to ensure that the argument matches one of the strings in `ValidArgs` exactly:

``` go
package main

import (
	"fmt"
	"strings"
)

func main() {
	s := []string{"one", "two", "three"}
	a := "re"
	p := "$"	
	fmt.Println(strings.Contains(p+strings.Join(s, p)+p, p+a+p))
	fmt.Println(strings.Contains(strings.Join(s,""), a))
}
```

``` bash
false
true
```

jharshman: I see, you are delimiting with a dollar sign.

umarcor: That's it. Please, let me know if you find any other issue with this PR. Otherwise, I'd be really glad if you could approve it.

umarcor:
#840 is merged now, so I rebased this on top of master. Since all the tests are passing, I believe it is ready for review, @eparis.

command.go
jharshman: 👍

eparis: I have to admit, I don't particularly like this way to search for valid args. First, we are still going to be doing a second loop, only now it's hidden inside Contains() https://golang.org/src/strings/strings.go?#L947 and second the code is less obvious. I STRONGLY prefer obvious to smart. Someone else is going to have to maintain this and simple is always going to be better. Third, I don't see any testing related to the usage of the `$` to prove it is correct. And the last thing that bothers me is that I think it's wrong. What if my validArgs = []string{"aa$bb"}. Wouldn't this code allow both "aa" and "bb" to pass, even though they are not in the ValidArgs?

I think the helper function was correct and clear and would prefer to continue something more obvious.

umarcor: It's ok. The change was in a separate commit, so it is dropped now. 

> Third, I don't see any testing related to the usage of the `$` to prove it is correct. 

There is quite a lot of testing, since may of the tests in `args_test.go`. Nonetheless, you are correct, corner cases such as a ValidArg containing `$` were not properly handled.

Please, let me know if you find any other issue with this PR.

args.go
```diff
@@ -99,3 +75,13 @@ func RangeArgs(min int, max int) PositionalArgs {
 		return nil
 	}
 }
+
+// ExactValidArgs is DEPRECATED. Use ExactArgs instead.
```
eparis: I'm sorry, I should have looked more closely the first time. So these aren't just deprecated, you literally broke them? While my code would still compile it wouldn't work correctly anymore. Let's say my code uses the arg to look up a value in a slice. After this patch my code would start to panic() on invalid input instead of failing nicely.

That's not a deprecation, that's API breakage and should only be done until cobra v2. Deprecation means you keep it working, but you tell people you are going to break it eventually.

OnlyValidArgs and ExactValidArgs are exported functions which are part of our API. They must remain working and must remain tested.

Please add a comment line that starts with `// Deprecated: ` similarly to what the core library uses to indicate deprecation. While there aren't many tools that look for it at least a couple do and it makes things a little more consistent for users.

umarcor: > I'm sorry, I should have looked more closely the first time. So these aren't just deprecated, you literally broke them? While my code would still compile it wouldn't work correctly anymore.

That was definitely not expected. I will have a look.

At first I wanted to remove them, not just deprecate. So, yes, this PR was meant to be released in v2. However, since @jharshman suggested it, I then tried to keep then functional.

> Let's say my code uses the arg to look up a value in a slice. After this patch my code would start to panic() on invalid input instead of failing nicely.

I assume that this use case is not covered by any of the existing tests. Precisely, I now added `TestOnlyValid*` and `TestExactValid*` tests again, and all of them are successful. Is it possible to have a MWE to work on?
 
> That's not a deprecation, that's API breakage and should only be done until cobra v2. Deprecation means you keep it working, but you tell people you are going to break it eventually.

As said, I'm ok with any of them. Nonetheless, in this case I think that deprecation without breakage is possible.

> OnlyValidArgs and ExactValidArgs are exported functions which are part of our API. They must remain working and must remain tested.

As commented above, I am keeeping the tests now, and all of them are successful.

> Please add a comment line that starts with `// Deprecated: ` similarly to what the core library uses to indicate deprecation. While there aren't many tools that look for it at least a couple do and it makes things a little more consistent for users.

I didn't find any example in the repo (`grep -r ./ -e '// Deprecated: '`). Do you mean [golang/go/wiki/Deprecated](https://github.com/golang/go/wiki/Deprecated)?

command.go
jharshman: I'm all for a helper function here.

args.go
jharshman: > At first I wanted to remove them, not just deprecate. So, yes, this PR was meant to be released in v2. However, since @jharshman suggested it, I then tried to keep then functional.

Yes is it important to keep them functional.

command.go
```diff
@@ -875,6 +875,13 @@ func (c *Command) ValidateArgs(args []string) error {
...
```
jharshman: Just to sum up, @umarcor , please replace this logic with an un-exported validation function in args.go.  If anything it keeps the method cleaner.  The logic of this function should be that of the `OnlyValidArgs` function you deprecated.

README.md
```diff
@@ -382,28 +382,33 @@ rootCmd.MarkFlagRequired("region")
...
   Short: "hello",
   Args: func(cmd *cobra.Command, args []string) error {
-    if len(args) < 1 {
-      return errors.New("requires a color argument")
+/*
+    // Optionally run one of the validators provided by cobra
+    if err := cobra.MinimumNArgs(1)(cmd args); err != nil {
+        return err
     }
+*/
```
jharshman: Please fix the indentation on the outside comment block or remove them all together.

args.go
```diff
@@ -99,3 +75,18 @@ func RangeArgs(min int, max int) PositionalArgs {
 		return nil
 	}
 }
+
+// ExactValidArgs returns an error if there are not exactly N positional args OR
+// there are any positional args that are not in the `ValidArgs` field of `Command`
+//
+// Deprecated: now `ExactArgs` honours `ValidArgs`, when defined and not empty
```
jharshman: Typo: "honours" assume should be honors.

args.go
```diff
@@ -99,3 +75,18 @@ func RangeArgs(min int, max int) PositionalArgs {
...
+func ExactValidArgs(n int) PositionalArgs {
+	return ExactArgs(n)
+}
+
+// OnlyValidArgs returns an error if any args are not in the list of `ValidArgs`.
+//
+// Deprecated: now `ArbitraryArgs` honours `ValidArgs`, when defined and not empty
```
jharshman: Same here. Typo "honours" should be honors.

args_test.go
```diff
@@ -5,231 +5,208 @@ import (
...
 	if err != nil {
 		t.Fatalf("Unexpected error: %v", err)
 	}
 }
 
-func TestNoArgsWithArgs(t *testing.T) {
-	c := &Command{Use: "c", Args: NoArgs, Run: emptyRun}
-
-	_, err := executeCommand(c, "illegal")
+func expectError(err error, t *testing.T, ex string) {
```
jharshman: Not a fan of this helper, the keys in the map are basically magic values outside this function.

args_test.go
```diff
@@ -5,231 +5,208 @@ import (
...
-	}
+func TestMaximumNArgsWithMoreArgsWithValid(t *testing.T) {
+	_, e := executeCommand(newCommand(MaximumNArgs(2), true), "one", "three", "two")
+	expectError(e, t, "max")
 }
 
-func TestExactArgsWithInvalidCount(t *testing.T) {
-	c := &Command{Use: "c", Args: ExactArgs(2), Run: emptyRun}
-	_, err := executeCommand(c, "a", "b", "c")
+func TestMaximumNArgsWithMoreArgsWithValidWithInvalidArgs(t *testing.T) {
```
jharshman: I understand you're trying to be very descriptive with these names but IMHO it's very difficult to read.

command.go
umarcor: How should it be named? `validateValidArgs`?

args.go
umarcor: > Yes is it important to keep them functional.

Could you reproduce the `panic` that @eparis commented?

args.go
umarcor: Isn't it a difference between british (honour) and american (honor)?

args_test.go
umarcor: This is fixed in #842, where underscores are introduced. E.g. `TestMinimumNArgsWithValidWithInvalidArgs` is renamed to `TestArgs_MinimumN_wValid_InvalidArgs`, which is explained as:

- `TestArgs`: we are testing the args
- `MinimumN`: this is the validator we are testing
- `wValid`: the command does contain a `ValidArgs` field. This can be `wArgs`, `wValid` or `wArgswValid`.
- `InvalidArgs`: execution arguments.

See: https://github.com/umarcor/cobra/compare/feat-matchvalid...umarcor:feat-validinhelp#diff-748c41438b4135ae7035c2af1228e349

I think that the format with underscores is both easy to read and useful. See output: https://travis-ci.com/umarcor/cobra/jobs/186549357

Note that using underscores in test function names does not produce any linting issue.

If you want, I can add the underscores in this PR too.

command.go
jharshman: Meh that sounds a bit repetitive. `validateArgs` is fine with me.

args.go
jharshman: Haha, well learn something new everyday.  Indeed it is. 
Honor is the preferred spelling in American English.  Are there other places in the ReadMe that use British variation then i don't mind. Otherwise please update it since these comments become part of the GoDoc and should have a certain continuity to them.

args_test.go
umarcor: The keys in the map correspond to the validator that we expect to get an error from:

- `validateValidArgs`: `valid`
- `NoArgs`: `no`
- `ArbitraryArgs`: does not produce errors
- `MinimumNArgs`: `min`
- `MaximumNArgs`: `max`
- `ExactArgs`: `exact`
- `RangeArgs`: `range`

I can make it more explicit by just stripping `Args`, if you want:

- `validateValidArgs`: `validateValid`
- `NoArgs`: `No`
- `ArbitraryArgs`: does not produce errors
- `MinimumNArgs`: `MinimumN`
- `MaximumNArgs`: `MaximumN`
- `ExactArgs`: `Exact`
- `RangeArgs`: `Range`

args_test.go
jharshman: The underscores do help but still long and hard to read. I think the tests could be greatly improved by implementing them as sub-tests and grouping them per validation function.  The nitty gritty descriptive parts then go in the subtests.

Perhaps an improvement for another PR.  For now, imma call the name fine since after looking back at the tests_args.go file it seems a lot of these names are too long including ones you haven't changes/added.

umarcor: > I think the tests could be greatly improved by implementing them as sub-tests and grouping them per validation function. The nitty gritty descriptive parts then go in the subtests.
> Perhaps an improvement for another PR. 

That sounds good. This is my first experience with golang tests, tho. So, I'd like to do those changes in a different PR (probably #842).

> For now, if you could add the underscores that would be great. After looking back at the args_test.go file it seems a lot of these names are too long including ones you haven't changes/added.

Done.

args_test.go
jharshman: Well first off, comparing error text is always a bad idea even in tests. It's just brittle.
But that is pretty much what the code was already doing anyways.  I get that you wanted to reduce duplication, but in this case I think some error constants would be more clear and you can nix, the helper function.

umarcor: The main objective was not to remove code duplication, but to make two characteristics explicit and easy to spot:

- Whether the test is expected to success or fail.
- Whether multiple test cases are expected to return exactly the same result.

For example:

``` go
func TestMinimumNArgs_WithValid_WithInvalidArgs(t *testing.T) {
...
	expectError(e, t, "valid")
...
func TestMinimumNArgs_WithLessArgs(t *testing.T) {
...
	expectError(e, t, "min")
...
func TestMinimumNArgs_WithValid_WithLessArgs(t *testing.T) {
...
	expectError(e, t, "min")
...
func TestMinimumNArgs_WithValid_WithLessArgsWithInvalidArgs(t *testing.T) {
...
	expectError(e, t, "valid")
```


``` go
func TestMinimumNArgs_WithValid_WithInvalidArgs(t *testing.T) {
...
	expectError(e, t, `invalid argument "a" for "c"`)
...
func TestMinimumNArgs_WithLessArgs(t *testing.T) {
...
	expectError(e, t, "requires at least 2 arg(s), only received 1")
...
func TestMinimumNArgs_WithValid_WithLessArgs(t *testing.T) {
...
	expectError(e, t, "requires at least 2 arg(s), only received 1")
...
func TestMinimumNArgs_WithValid_WithLessArgsWithInvalidArgs(t *testing.T) {
...
	expectError(e, t, `invalid argument "a" for "c"`)
```

jharshman: > * Whether the test is expected to success or fail.
> * Whether multiple test cases are expected to return exactly the same result.

The first can be accomplished just by checking for the presence of an error value then.
The second is trickier as currently the only way of telling different validation failures apart is by the error text.  This wasn't an issue until now because the ValidateArgs function only ran 1 validator.

args_test.go
umarcor: In the end, I used sub-tests. I think that the implementation can be improved, because the items in the map are executed in random order and it would be desirable to group them according to the first 'field'. But I hope you find it better than the previous approach.

args_test.go
umarcor: > You could approach the checks a few ways,
>     1. use constants for the error text

I would really like to do this. Not just because of this PR, but specially because of #842. The idea there is to slighly rework the behaviour when execution errors are produced either because the command is not runnable, because it does not accept args, because args are invalid or because the arg is neither a valid arg nor a valid subcommand.

- `invalid argument "a" for "c"`
- `command "c" is not runnable`
- `command "c" is not runnable; please provide a subcmd`
- `unknown command "one" for "c"`
- `"one" rejected; "c" does not accept args`
- `requires at least 2 arg(s), only received 1`
- `accepts at most 2 arg(s), received 3`
- `accepts 2 arg(s), received 3`
- `accepts between 2 and 4 arg(s), received 1`

Note that the management of the first five error strings above is directly related to deciding when `[args]` should be shown in the usage/help.

However, the error strings above are not constant. `"a"`, `"c"`, `"one"`, `2`, `1`, `3`, `4`... all of those values are generated when the error is produced. Should I define 'error constants' as `invalid argument %s for %s` and:

- Use the 'constant string template' to generate the error.
- When I need to identify it, execute the template with the current `cmd` and compare it with the err.

How can I ensure that the `cmd` when the error was generated is the same as the `cmd` in the function where it is being evaluated?

>     2. use sentinel errors.

I don't know what are 'sentinel errors'. Isn't it the same as the procedure described above?

https://dave.cheney.net/2016/04/07/constant-errors

>     3. Rely on error behavior (see github.com/pkg/errors)

This seems to be exactly what we need.

> Latter option is best but entails a larger change-set, constants or sentinel errors probably the easiest.

I'll try to add it to #842. But I think it should not be a requirement for this PR.

> Whatever solution, we shouldn't be comparing error text in the tests.

Agree. But we are already doing it everywhere:

- https://github.com/spf13/cobra/blob/master/args_test.go
- https://github.com/spf13/cobra/blob/master/bash_completions_test.go
- https://github.com/spf13/cobra/blob/master/cobra_test.go
- https://github.com/spf13/cobra/blob/master/command_test.go
- https://github.com/spf13/cobra/blob/master/zsh_completions_test.go

I think it is out of the scope of this PR to fix it. Moreover, I think that 'errStrings' is a step forward towards implementing 3, since ot allows the list of expected errors to be listed somewhere.

args_test.go
```diff
@@ -5,230 +5,104 @@ import (
...
+		if o != "" {
+			t.Errorf("Unexpected output: %v", o)
+		}
+		if e != nil {
+			t.Fatalf("Unexpected error: %v", e)
+		}
+	}
+}
+
+func TestArgs(t *testing.T) {
```
jharshman: The addition of the subtests are great, much cleaner and provides an easy to read matrix of cases. If you could organize them based the args validator that would be even better.

i.e 
```
func TestNoArgs(t *testing.T) {}
func TestArbitrary(t *testing.T) {}
func TestMinimum(t *testing.T) {}
... etc.. 
```

args_test.go
jharshman: Agreed that it overreaches the scope. It should probably be a standalone PR.
Going to go through some of this again this evening and then ill give it my 👍 . 

Check my comment below about organization of the test structure.

args_test.go
umarcor: Done.

umarcor:
*ping*

umarcor:
*ping*

umarcor:
@jharshman, is it possible to have this merged or will we wait until some other maintainer reviews it?

umarcor:
*ping*

umarcor:
@hilary, I saw that you merged this PR on your fork a couple of months ago. Is it working ok for you? Did you need to apply any additional fix?

umarcor:
Request to be included in [1.0.0](https://github.com/spf13/cobra/milestone/2). Ref #959.

jharshman:
API changes are not on the roadmap for the 1.0.0.  Bugs, cleanup, and documentation tasks are being targeted for 1.0.0.

umarcor:
> API changes are not on the roadmap for the 1.0.0. Bugs, cleanup, and documentation tasks are being targeted for 1.0.0.

Agree. Precisely, I tried to be careful not to bump PRs with breaking changes.

This PR is the completion of a partial implementation and it is explicitly backwards compatible. I believe that 1.0.0 should include the full implementation, or remove the incomplete parts.

umarcor:
This PR is now rebased on top of #1010.

umarcor:
This PR is now rebased, to fix conflicts after #1048 was merged.

README.md
```diff
@@ -391,28 +391,32 @@ rootCmd.MarkFlagRequired("region")
 ## Positional and Custom Arguments
 
 Validation of positional arguments can be specified using the `Args` field
-of `Command`.
+of `Command`. The following validators are built in:
 
-The following validators are built in:
+- `NoArgs` - report an error if there are any positional args.
+- `ArbitraryArgs` - accept any args.
+- `MinimumNArgs(int)` - report an error if leass N positional args are provided.
```
marckhouzam, line 398: "leass" -> "less than"

umarcor: Thanks! Now fixed.

danmx:
@umarcor it seems you can merge your PR now

umarcor:
> @umarcor it seems you can merge your PR now

Yes. This PR is ready to be merged since March 2019. I've been keeping it updated/rebased for 17 months. Unfortunately, I am not a maintainer, so I cannot decide when to merge it.

marckhouzam (commented):
@umarcor I like the idea behind this change.  As a user, the field `ValidArgs` does make me think that my positional args must be part of that list.  There are a couple of things I'm wondering about.

- `ValidArgs` will only be checked if `Args` is defined.  This probably should be mentioned in the documentation when it mentions that `ValidArgs` is used to check positional args.

- Say I want to have the behaviour of the deprecated `OnlyValidArgs`, how would I achieve this now?  IIUC I would have to set `Args: ArbitraryArgs`?  I find this a little confusing as my intent is to only allow `ValidArgs`.

umarcor:
> * `ValidArgs` will only be checked if `Args` is defined.  This probably should be mentioned in the documentation when it mentions that `ValidArgs` is used to check positional args.

I need to check it explicitly, but I wrote the following in the README:

> If `Args` is undefined or `nil`, it defaults to `ArbitraryArgs`.

Therfore, my understanding is that `ValidArgs` will always be checked, if defined, regardless of `Args`.

> * Say I want to have the behaviour of the deprecated `OnlyValidArgs`, how would I achieve this now?  IIUC I would have to set `Args: ArbitraryArgs`?  I find this a little confusing as my intent is to only allow `ValidArgs`.

You would not need to set it, because it would default to that. If you wanted to define `Args` explicitly, you'd need to use `ArbitraryArgs`. I agree that it might be confusing to have `Args: ArbitraryArgs` and `ValidArgs` at the same time. However, "Arbitrary" in that context means "any number of arguments". The naming is unfortunate, but it is preserved so that this PR is backwards compatible. If maintainers wanted to introduce a breaking change, I'd be ok with renaming it to `AnyNumberOfArgs` or any other more descriptive/specific name.

command.go
```diff
@@ -979,10 +979,15 @@ func (c *Command) ExecuteC() (cmd *Command, err error) {
 	return cmd, err
 }
 
+// ValidateArgs returns an error if any any positional args that are not in
```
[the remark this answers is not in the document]
umarcor: Thanks!

umarcor:
@marckhouzam I fixed the order of the validations, so that `ValidArgs` is checked regardless of `Args`. I also added tests for it. Thanks for catching that!

command.go
```diff
@@ -979,7 +979,12 @@ func (c *Command) ExecuteC() (cmd *Command, err error) {
 	return cmd, err
 }
 
+// ValidateArgs returns an error if any positional args are not in
+// the `ValidArgs` field of `Command`
```
marckhouzam, line 983: The comment should probably also include that the `Args` validator is also executed, of specified.

umarcor: Done.

mikeschinkel:
Checking on the status of this, please?  

I am looking for suggestions on subcommands and it appears this and #842 are needed to provide them, if I read #981 correctly?

umarcor:
@mikeschinkel the status is the same as explained in https://github.com/spf13/cobra/issues/981#issuecomment-647784636.

umarcor:
A PR was created with the first three commits from this one: #1426. Then, this PR was cleaned up. Hope that helps move forward...

tmc:
Could/should this change take into account ValidArgsFunction?

marckhouzam:
> Could/should this change take into account ValidArgsFunction?

Please see #1298

umarcor:
@tmc, since ValidArgs and ValidArgsFunction are mutually exclusive, I don't see how to change this to take the latter into account. If you use ValidArgsFunction, it should be handled by the completion logic, as far as I understand. Maybe you can provide a reproducer of what you are trying to achieve?

umarcor:
Please, remove the stale label from this PR.

Refs:

- https://github.com/spf13/cobra/issues/1496#issuecomment-989131001
- https://github.com/spf13/cobra/issues/1598#issuecomment-1038566503

johnSchnake:
This has an approval and green tests; is it waiting for anything else? I know this PR  is connected to a lot of issues and if everyones happy I'm just curious what the hold is right now? [FWIW I haven't reviewed this, just cleaning up the backlog and kept finding links to this PR]

PR #842, opened 2019-03-20
Add '[args]' and 'Valid Args:' to --help

### LINKED ISSUE #571: Getting --help to print descriptions for arguments
Just wanted to check if there is a way to get Cobra to print help text for ARGS.

Say I have a CLI whose usage is:
`do [SOMETHING]`

I was wondering if it was possible to get `do --help` to print:
`
Usage:
    do [SOMETHING]
    [SOMETHING]: Allowed values are laundry, dishes or dinner

Flags:
   ....
`
The readme doesn't mention anything like this, and there are no examples on SetHelpTemplate, so I'm unsure if that is what needs to be used here. Would someone be able clarify if this is possible? And if not, are there any other options?

Thanks!

### LINKED ISSUE #576: Missing ValidArgs on help
If I describe ValidArgs in a command when I execute -h flag in some command cobra don't show ValidArgs in help print

umarcor:
Close #571 and Close #576.

This PR is based on #841, and it must be merged after that.

The added feature is to modify the output of `Usage()` to print `[args]` if the command accepts args:

``` bash
Usage:
  app [args]              # instead of 'app'
  app [flags] [args]      # instead of 'app [flags]'
  app [command]

Valid Args:
  one two three
```

- `app [args]` and `app [flags] [args]` are not shown at the same time.
- `[args]` is only added if the command is runnable, and `Args != NoArgs`, or `Args == nil && (!c.HasAvailableSubCommands() || c.HasParent())`, etc.
- `Valid Args` is only shown if `ValidArgs` is not empty and `[args]` is shown.

---

As suggested in [#571](https://github.com/spf13/cobra/issues/571#issuecomment-473910246), when `MinimumNArgs`, `ExactArgs` or `RangeArgs` are used, `args` should be shown instead of `[args]`. This has not been implemented yet, because I don't know any easy mechanism to detect the type of `c.Args`.


umarcor:
Ref #666 and Ref #635.

README.md
```diff
@@ -407,28 +407,32 @@ rootCmd.MarkFlagRequired("region")
...
+Moreover, it is possible to set any custom validator that satisfies `func(cmd *cobra.Command, args []string) error`. For example:
 
 ```go
 var cmd = &cobra.Command{
   Short: "hello",
-    if len(args) < 1 {
-      return errors.New("requires a color argument")
+    // Optionally run one of the validators provided by cobra
+    if err := cobra.MinimumNArgs(1)(cmd args); err != nil {
```
zwpaper: should be: (cmd`,` args)?

umarcor: You are correct. Good catch! The bug was introduced in #841, which this PR is based on. I fixed and rebased both PRs now.

umarcor:
This is still WIP, but it depends on #841 being merged.

umarcor:
This PR is not stale, but blocked by #841. Please, remove the `kind/stale` label.
See https://github.com/spf13/cobra/issues/1496#issuecomment-989131001 and https://github.com/spf13/cobra/issues/1496#issuecomment-989287359.

PR #846, opened 2019-03-28
Return nil error in validateRequiredFlags when help command

marco2704:
Hi guys!

When setting required flags to the root commands:
```
rootCmd.MarkPersistentFlagRequired("user")
```
seems that the help command throws an error before showing the help when those flags are not set:
```
$ ./cli help
Error: required flag(s) "user" not set
...
```
It does not make sense to me that required flags have to be validated when calling the `help` command. I do not know whether this would have more impact than skipping the validation.

In any case, thanks for your time!

### DIFF
--- a/command.go
+++ b/command.go
@@ -879,6 +879,10 @@ func (c *Command) ValidateArgs(args []string) error {
 }
 
 func (c *Command) validateRequiredFlags() error {
+	if c.Name() == "help" {
+		return nil
+	}
+
 	flags := c.Flags()
 	missingFlagNames := []string{}
 	flags.VisitAll(func(pflag *flag.Flag) {


umarcor:
This should not be required, because 'help' is checked before calling validateRequiredFlags (https://github.com/spf13/cobra/blob/master/command.go#L704) so validateRequiredFlags should never be called. I.e., it should not be required to patch validateRequiredFlags, because it should not apply in this context. Maybe the error is produced in ParseFlags? https://github.com/spf13/cobra/blob/master/command.go#L689

marco2704:
@umarcor As you can see, that only checks whether the help **flag**, not the help command. I'm not sure whether this should be checked anywhere else before calling `validateRequiredFlags`, but I'm pretty sure that `validateRequiredFlags` is currently being called when using the help command :grimacing:

umarcor:
> @umarcor As you can see, that only checks whether the help **flag**, not the help command.

I was precisely suggesting that the command should be checked too.

> I'm not sure whether this should be checked anywhere else before calling `validateRequiredFlags`, but I'm pretty sure that `validateRequiredFlags` is currently being called when using the help command 😬

I think it should be checked somewhere between https://github.com/spf13/cobra/blob/master/command.go#L689-L703, even though it is not done now, as you say. `validateRequiredFlags` should not be executed at all, if no flag is to be considered because it is a 'help' command.

dzciemix:
Hi!

I think I still expirance this bug.

```
package main

import (
	"log"

	"github.com/spf13/cobra"
)

var RootCmd = &cobra.Command{
	Use: "program",
	SilenceUsage:  true,
	SilenceErrors: true,
}

var createCmd = &cobra.Command{
	Use:  "create",
	Args: cobra.ExactArgs(1),
		return nil
	},
}

func main() {
	RootCmd.PersistentFlags().StringP("url", "u", "", "Server address")
	RootCmd.MarkPersistentFlagRequired("url")
	RootCmd.ValidateRequiredFlags()

	RootCmd.AddCommand(createCmd)

	err := RootCmd.Execute()
	if err != nil {
		log.Fatalf("Error!\n%v", err)
	}
}
```

```
$ go run . help 
2024/06/25 12:12:34 Error!
required flag(s) "url" not set
exit status 1
```

PR #853, opened 2019-04-18
Use os.UserHomeDir() for Go version 1.12+

mgrachev:
Hi there!

Use `os.UserHomeDir()` instead of `go-homedir` for Go version 1.12+.

### DIFF
--- a/README.md
+++ b/README.md
@@ -212,7 +212,6 @@ import (
   "fmt"
   "os"
 
-  homedir "github.com/mitchellh/go-homedir"
   "github.com/spf13/cobra"
   "github.com/spf13/viper"
 )
@@ -238,7 +237,7 @@ func initConfig() {
     viper.SetConfigFile(cfgFile)
   } else {
     // Find home directory.
-    home, err := homedir.Dir()
+    home, err := os.UserHomeDir()
     if err != nil {
       fmt.Println(err)
       os.Exit(1)
--- a/cobra/cmd/init_go1.10_test.go
+++ b/cobra/cmd/init_go1.10_test.go
@@ -1,3 +1,5 @@
+// +build !go1.12
+
 package cmd
 
 import (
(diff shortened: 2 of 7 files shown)


pior:
👍

jharshman:
@mgrachev is there an issue with using the https://github.com/mitchellh/go-homedir package?

mgrachev:
@jharshman No, there isn't. This is a dependency and I prefer to minimize dependencies in my projects.

jharshman:
@mgrachev Trimming back dependencies is always good, however, the `go-homedir` package itself is very small and not dependent on Go version.  I could see this comment giving the impression to a user that if s/he is using Go 1.12+ that s/he would need to edit the code to use `os.UserHomeDir()`.

spf13:
@mgrachev I'm confused about this. How does adding this comment to the source help? Are we expecting users to change their dependency source if they are using Go 1.12+?

mgrachev:
@spf13 Yes, you're right. I've added this comment because I didn't want to drop support for previous versions of Go.

spf13:
I'm not sure what value there is in adding a comment. I do think there is some value in removing dependencies. One way to do this without breaking older users of Go is to incorporate build tags. 
That is, create two alternative files, with a single function in each, one using the dependency and one calling os.UserHomeDir and use build tags to manage which one is included.

mgrachev:
@spf13 Done.

jharshman:
I'm going to close this because it introduces complication without enough benefit.

jpeach:
Commenting on a closed PR is pretty bad form, but `os.UserHomeDir` depends on the "$HOME" environment variable being set, whereas `go-homedir` uses a number of different strategies and basically always works whether "$HOME" is set or not.

marckhouzam:
Looks like this has been done by #1337  and released in Cobra 1.2.

PR #855, opened 2019-05-02
Add tendermint and cosmos-sdk to the list of projects using cobra

alessio:
Both tendermint and cosmos's binaries provide command line interfaces designed and implemented with cobra.

Thanks for considering.


alessio:
CC'ing @jpmcb @eparis @jharshman @marckhouzam

jpmcb (approved):

PR #860, opened 2019-05-14
Changed default usage command to just "help"

albertvaka:
Cobra's `InitDefaultHelpCmd()` sets `help` and not `--help` as the command to print the usage. It makes more sense for the default `usageTemplate` to point to the default help command.


albertvaka:
Woha, it broke over nine thousand tests. Let me know if you think the change makes sense and I will fix the tests :)

albertvaka:
Closing since there seems to be no interest on this.

PR #864, opened 2019-05-17
Add DisableHelpSubCommand option

### LINKED ISSUE #587: Add ability to remove "help" subcommand
Currently, if an application has any subcommands, a `help` subcommand is automatically added. There may be applications that don't want the `help` subcommand, but as it currently stands there is no easy way to disable it.

* You can't simply remove the command, because `InitDefaultHelpCmd` will re-add it
* Using `SetHelpCommand` to set the command to `nil` doesn't work, because `InitDefaultHelpCmd` will instantiate the default help command when the command is `nil`

The best I have come up with is something like:

```go
rootCmd.SetHelpCommand(&cobra.Command{
		Use: "help",
		Hidden: true,
			return fmt.Errorf(`unknown command "help" for "%s"`, cmd.Root().Name())
		},
	})
```

but even this isn't perfect because it goes down a different error path since the subcommand is actually still executed.

I think it would be nice to have some mechanism to disable the `help` command -- even if it's not an explicit flag or setting, having some way to remove this subcommand would be nice.

josephlr:
This option stops the creation of a "help" subcommand for when a CLI
tool only wants to have "-h" and "--help" to get examples/usage.

This can (in certain situations) help reduce abiguity in documentation.
Instead of having two ways to get help about a subcommand ("app help foo"
and "app foo --help") there can be only one.

As this option defaults to false, the default behavior remains the same.

Closes #587

### DIFF
--- a/command.go
+++ b/command.go
@@ -130,6 +130,10 @@ type Command struct {
 	// line of a command when printing help or generating docs
 	DisableFlagsInUseLine bool
 
+	// DisableHelpSubCommand will disable the addition of the "help" subcommand.
+	// The "-h" and "--help" flags will still be available.
+	DisableHelpSubCommand bool
+
 	// DisableSuggestions disables the suggestions based on Levenshtein distance
 	// that go along with 'unknown command' messages.
 	DisableSuggestions bool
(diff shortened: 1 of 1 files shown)


aristosathens:
Apologies if this isn't the proper place, but will this PR get merged? It would be super useful to use.

josephlr:
Closing is this PR is stale, and it seems like the workarounds in the attached issue are the canonical way to solve this problem.

Issue #604, opened 2017-12-27
Subcommand conflict


jiasir:
For example, I want to achieve the following scenario:
```
$ ./cobra-starter config delete
$ ./cobra-starter serve delete
```
Then the file name conflict:
```
macbook:cobra-starter macbook$ cobra add delete -p 'configCmd'
delete created at /Users/macbook/Documents/code/src/localhost/jiasir/cobra-starter/cmd/delete.go
macbook:cobra-starter macbook$ cobra add delete -p 'serveCmd'
Error: /Users/macbook/Documents/code/src/localhost/jiasir/cobra-starter/cmd/delete.go already exists

```

PR #871, opened 2019-06-01
Parent Handling improved

### LINKED ISSUE #266: Ability to add nested commands through `cobra add`

Wulfheart:
Fixes #266, #273 (should be sufficient), #604.

In short it checks if there is another parentCmd thant `rootCmd` and constructs from this information a unique cmd Variable. For example if there is a command like `<appName> compile clean` the var is compileCleanCmd.

It is needed to edit the tests.

### DIFF
--- a/cobra/cmd/add.go
+++ b/cobra/cmd/add.go
@@ -17,6 +17,7 @@ import (
 	"fmt"
 	"os"
 	"path/filepath"
+	"strings"
 	"unicode"
 
 	"github.com/spf13/cobra"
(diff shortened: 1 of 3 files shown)


bogem:
I'm sorry, I can't review it, because I'm out of context.

bogem:
@jharshman can you please take a look at it?

PR #817, opened 2019-01-31
Vgo

jharshman:
# Moves to resolve Issue #756 
This was a re-work of the cobra code generator to allow it to work outside of the GOPATH and hence be compatible with Go modules (vgo) introduced in Go 1.11.  This change introduces a new flag to the `init` command and deprecates the `--package` flag from the `add` command.  With all these changes the flows remains relatively the same.  Any exceptions are outlined below.

## Changes
* Add a `--pkg-name` flag to the init command which takes a string representing the package name.
e.g: `--pkg-name github.com/spf14/hugo`
* For adding commands, you should be inside your project root.
* Init will no longer fail on non-empty directories.

PR #878, opened 2019-06-10
Update README.md

NickBusey:
Current command example fails with error `Error: required flag(s) "pkg-name" not set`

This fixes that, as well as fixing the package name to be consistent with the name lower down in the instructions.

### DIFF
--- a/cobra/README.md
+++ b/cobra/README.md
@@ -20,7 +20,7 @@ Cobra init is pretty smart. You can provide it a full path, or simply a path
 similar to what is expected in the import.
 
 ```
-cobra init github.com/spf13/newApp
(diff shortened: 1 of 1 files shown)


Ghostwritten:
I thought it was my fault

Ghostwritten:
> Current command example fails with error `Error: required flag(s) "pkg-name" not set`
> 
> This fixes that, as well as fixing the package name to be consistent with the name lower down in the instructions.

how long?

jharshman:
Bit of context: https://github.com/spf13/cobra/pull/817

jharshman:
covered by #904

PR #883, opened 2019-06-12
Add framework for dynamic tab completions

PapaCharlie:
By setting the COBRA_FLAG_COMPLETION environment variable, the normal execution path of the command
is short circuited, and instead the function registered by `MarkCustomFlagCompletion` is executed.

All flags other than the one being completed get parsed according to whatever type they are defined
as, but the flag being completed is parsed as a raw string and passed into the custom compeltion.

### DIFF
--- a/bash_completions.go
+++ b/bash_completions.go
@@ -15,6 +15,7 @@ import (
 const (
 	BashCompFilenameExt     = "cobra_annotation_bash_completion_filename_extensions"
 	BashCompCustom          = "cobra_annotation_bash_completion_custom"
+	BashCompDynamic         = "cobra_annotation_bash_completion_dynamic"
 	BashCompOneRequiredFlag = "cobra_annotation_bash_completion_one_required_flag"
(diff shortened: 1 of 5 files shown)


PapaCharlie:
The failing tests are being fixed in #889. After it is checked in, this commit will be all green

marckhouzam:
This is pretty cool!
But Cobra 1.0 added support for Go completions not only for flags but also for arguments, which I believe supersedes this PR.
See 
https://github.com/spf13/cobra/blob/master/bash_completions.md#1-custom-completions-of-flags-written-in-go 
and 
https://github.com/spf13/cobra/blob/master/bash_completions.md#1-custom-completions-of-nouns-written-in-go

I believe this can be closed.

PapaCharlie:
Hey Marc! Let me take a look at the new features, I'll close this for now and open a new branch on top of 1.0 in case I find anything missing, but it looks like I won't have to

PR #888, opened 2019-06-17
`cobra init` no longer works as documented

DazWilkin:
`cobra init github.com/spf13/newApp` results in `Error: required flag(s) "pkg-name" not set`

This alternative works but it's not intuitive to me as a noob user that it's correct.

In order to test what it does, I tried `cobra init a --pkg-name=b` but then `main.go` attempts to (incorrectly) `import "b/cmd"`

I think (!?) I'd expect what you original had: `cobra init github.com/spf13/newApp` creating a directory under `{GOPATH}/src/github.com/spf13/newApp` with `github.com/spf13/newApp` as the package name.

### DIFF
--- a/cobra/README.md
+++ b/cobra/README.md
@@ -20,7 +20,7 @@ Cobra init is pretty smart. You can provide it a full path, or simply a path
 similar to what is expected in the import.
 
 ```
-cobra init github.com/spf13/newApp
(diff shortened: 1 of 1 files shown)


jharshman:
@DazWilkin https://github.com/spf13/cobra/pull/817

umarcor:
Ref #878.

jharshman:
covered by #904

PR #889, opened 2019-06-19
bash: fix shellcheck errors

rsteube:
Fixed shellcheck errors according to documentation:

https://github.com/koalaman/shellcheck/wiki/SC2207
https://github.com/koalaman/shellcheck/wiki/SC2164

**untested** - tests are ok but someone should verify this is correct

### DIFF
--- a/bash_completions.go
+++ b/bash_completions.go
@@ -72,7 +72,9 @@ __%[1]s_handle_reply()
             else
                 allflags=("${flags[*]} ${two_word_flags[*]}")
             fi
-            COMPREPLY=( $(compgen -W "${allflags[*]}" -- "$cur") )
+            while IFS='' read -r c; do
+                COMPREPLY+=("$c")
+            done < <(compgen -W "${allflags[*]}" -- "$cur")
             if [[ $(type -t compopt) = "builtin" ]]; then
                 [[ "${COMPREPLY[0]}" == *= ]] || compopt +o nospace
             fi
@@ -122,10 +124,14 @@ __%[1]s_handle_reply()
     if [[ ${#must_have_one_flag[@]} -ne 0 ]]; then
         completions+=("${must_have_one_flag[@]}")
     fi
-    COMPREPLY=( $(compgen -W "${completions[*]}" -- "$cur") )
+    while IFS='' read -r c; do
+        COMPREPLY+=("$c")
+    done < <(compgen -W "${completions[*]}" -- "$cur")
 
     if [[ ${#COMPREPLY[@]} -eq 0 && ${#noun_aliases[@]} -gt 0 && ${#must_have_one_noun[@]} -ne 0 ]]; then
-        COMPREPLY=( $(compgen -W "${noun_aliases[*]}" -- "$cur") )
+        while IFS='' read -r c; do
+            COMPREPLY+=("$c")
+        done < <(compgen -W "${noun_aliases[*]}" -- "$cur")
     fi
 
     if [[ ${#COMPREPLY[@]} -eq 0 ]]; then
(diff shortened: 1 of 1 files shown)


bash_completions.go
```diff
@@ -72,7 +72,7 @@ __%[1]s_handle_reply()
             else
                 allflags=("${flags[*]} ${two_word_flags[*]}")
             fi
-            COMPREPLY=( $(compgen -W "${allflags[*]}" -- "$cur") )
+            mapfile -t COMPREPLY < <(compgen -W "${allflags[*]}" -- "$cur")
```
PapaCharlie: `mapfile` is not present in all bash versions. Namely, the out-of-the-box bash for MacOS doesn't have it (it's at a 3.0 version, it looks like `mapfile` was introduced in 4.0).

`read -a COMPREPLY` looks like it'll do the trick

bash_completions.go
```diff
@@ -160,7 +160,7 @@ __%[1]s_handle_filename_extension_flag()
 __%[1]s_handle_subdirs_in_dir_flag()
 {
     local dir="$1"
-    pushd "${dir}" >/dev/null 2>&1 && _filedir -d && popd >/dev/null 2>&1
+    pushd "${dir}" >/dev/null 2>&1 && _filedir -d && popd >/dev/null 2>&1 || exit
```
PapaCharlie: Completions are run as functions in the current shell. If `popd` or `pushd` fail, `|| exit` will exit the entire shell, possibly closing your terminal. You need to use `|| return` instead

rsteube: changed to `return`

bash_completions.go
rsteube: I think i need a project where i can test this. Disabling escape characters using `-r` is required as well (shellcheck warning) and the IFS seems to be required as well for it to work. So it's probably one of:
```
# For bash 3.x+
array=()
while IFS='' read -r line; do array+=("$line"); done < <(mycommand)
```
or
```
# For bash
IFS=" " read -r -a array <<< "$(mycommand)"
```
Some tests `read -a` in a bash shell didn't give me the right result (or i simply used it wrong).

PapaCharlie: Yeah it seems `read -a` really doesn't work as expected. I've tested this on multiple bash versions, and it seems to be the most consistent:
```
while read -s -d " " c ; do
  COMPREPLY+=("$c")
done < <(compgen -W "${allflags[*]}" -- "$cur")
```

rsteube: Does it need to be cleared first?: `COMPREPLY=()`

PapaCharlie: Bash will automatically create the array variable if you start appending to it:
```
unset a
a+=("foo")
echo "${a[@]}"
foo
```
But it wouldn't hurt to make sure it's empty anyway

rsteube: Was just worried that it is defined before already, but should be fine then.
Just compiled https://github.com/gohugoio/hugo with it but seems there is still sth. wrong with the completion (https://gist.github.com/rsteube/7647a9a3456cdb102f0b88ba004d74f8).
Currently looking into it.

rsteube: `while IFS='' read -r c; do` works in my shell (5.0.7).
https://gist.github.com/rsteube/6bb3336adf902193997bba048094b870

rsteube: 3.2.57 looks ok as well: https://hub.docker.com/r/albers/bash-completion-mac

umarcor:
@sp13, @eparis, can we please have this merged so that Travis CI jobs are fixed for every other PR based on master?

jharshman:
@rsteube, while I review this, can you squash your commits?

rsteube:
@jharshman done

jharshman:
@rsteube thanks, I'm just going to do a bit of testing of the bash completions before I merge this in. I would encourage you to do the same if you haven't yet.  I'm not _seeing_ any breaking changes here but who knows.

rsteube:
Recompiled hugo with the squashed version and no changes to https://gist.github.com/rsteube/6bb3336adf902193997bba048094b870 which seemed to be fine in bash `5.0.7` and `3.2.57`.

jharshman:
Validated by recompiling kubectl.

PR #893, opened 2019-06-21
Add support for context.Context

burdiyan:
This PR adds support for `context.Context`. It could help people struggling with passing context to the Run functions, that previously had to wrap `cobra.Command` initialization inside another function that would accept context.

So now instead of doing:

```go
func createFooCommand(ctx context.Context) *cobra.Command {
    return &cobra.Command{
        Use: "foo",
            // Use context here
            // ...
        }
    }
}
```

you would do:

```go
cmd := &cobra.Command{
    Use: "foo",
        ctx := cmd.Context()
        // Use context here
        // ...
    }
}


// Get ctx from somewhere else.
cmd.ExecuteContext(ctx)
```

This PR is fully backward-compatible, and doesn't require users to use context at all.

It renders this PR https://github.com/spf13/cobra/pull/727 (which seems to have conficlits, and also goes against the idiomatic way of using context) irrelevant.

### DIFF
--- a/command.go
+++ b/command.go
@@ -209,7 +209,7 @@ type Command struct {
 }
 
 // Context returns underlying command context. If command wasn't
-// executed with ExecuteContext the returned context will be nil.
+// executed with ExecuteContext Context returns Background context.
 func (c *Command) Context() context.Context {
 	return c.ctx
 }
@@ -886,6 +886,10 @@ func (c *Command) Execute() error {
 
 // ExecuteC executes the command.
 func (c *Command) ExecuteC() (cmd *Command, err error) {
+	if c.ctx == nil {
+		c.ctx = context.Background()
+	}
+
 	// Regardless of what command execute is called on, run on Root only
 	if c.HasParent() {
 		return c.Root().ExecuteC()
--- a/command.go
+++ b/command.go
@@ -17,6 +17,7 @@ package cobra
 
 import (
 	"bytes"
+	"context"
 	"errors"
 	"fmt"
 	"io"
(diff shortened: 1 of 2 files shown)


cpuguy83:
This does seem like a decent solution for shoe-horning context support into cobra.Command.

command.go
```diff
@@ -202,6 +205,12 @@ type Command struct {
 	errWriter io.Writer
 }
 
+// Context returns underlying command context. If command wasn't
+// executed with ExecuteContext the returned context will be nil.
+func (c *Command) Context() context.Context {
+	return c.ctx
```
cpuguy83: Generally you do not want to have a nil context.
Instead I'd recommend we make sure that context is at least `context.Background()`.

burdiyan: That's fair, I was thinking about that but didn't have strong opinions on that. But I changed it to use context.Background instead of nil. Thank you!

burdiyan:
I rebased this PR against origin, so that build pass (seems like there were some issues upstream previously). So hope this can be merged soon.

burdiyan:
Fixed merge conflicts again.

burdiyan:
Can this be merged?

korya:
The lack of support of `context.Context` is a huge headache for us. What blocks this PR from getting merged?

PS @burdiyan thanks a lot for the work!

snovichkov:
@cpuguy83 could you approve and merge this is PR?

cpuguy83:
I am not a maintainer on this project.

snovichkov:
@cpuguy83 sorry... @spf13 could you approve and merge this is PR?

gpaul:
Bump?

hairyhenderson:
This'd be pretty useful - I've just found myself abusing some global variables pretty badly because of the lack of `context.Context` support 😉

/cc @bep or @eparis maybe?

umarcor:
@hairyhenderson, see #959.

hypergig:
thank you!

I think something in the root README would be helpful here

aventurella-honey:
Is it possible to update the context in one command Run/PersistentPreRun so that the altered context is what's passed down to the subcommands?

burdiyan:
@aventurella-honey I believe it's not possible. Because context is immutable and you'll have to store altered context somewhere to pass it on to the subcommands, and currently the only way to pass the context is to call ExecuteContext.

PR #894, opened 2019-06-24
Fix stderr printing functions

alessio:
Follow-up of #822

This fixes stderr redirection.

@spf13 @eparis

### DIFF
--- a/command.go
+++ b/command.go
@@ -1139,12 +1139,12 @@ func (c *Command) PrintErr(i ...interface{}) {
 
 // PrintErrln is a convenience method to Println to the defined Err output, fallback to Stderr if not set.
 func (c *Command) PrintErrln(i ...interface{}) {
-	c.Print(fmt.Sprintln(i...))
+	c.PrintErr(fmt.Sprintln(i...))
 }
 
 // PrintErrf is a convenience method to Printf to the defined Err output, fallback to Stderr if not set.
 func (c *Command) PrintErrf(format string, i ...interface{}) {
-	c.Print(fmt.Sprintf(format, i...))
+	c.PrintErr(fmt.Sprintf(format, i...))
 }
 
 // CommandPath returns the full path to this command.


alessio:
CC'ing @jleni

amdprophet:
Should https://github.com/spf13/cobra/blob/master/command.go#L926 use `PrintErr`? We're seeing errors output to stdout after calling `SetOut` and `SetErr`.

alessio:
Yes, I think it should. Will amend my PR ASAP

alessio:
@amdprophet done, please review

jleni:
@alessio for some reason, the changes seem to break `TestBashCompletions`

alessio:
Will try to reproduce and fix shortly

PapaCharlie:
The shellcheck warnings are being addressed in #889

alessio:
Point taken: `mapfile` is available in Bash >= 4. Happy to incorporate #889 commits instead of going for my solution

alessio:
Conflicts have been resolved, please review

portertech:
Please review 🙏

amdprophet:
@spf13 @eparis @jleni any chance of this getting reviewed soon? 🙏

alessio:
Hi!

Any news? @spf13 @eparis @jleni

jleni:
Sorry @alessio.. but there isn't much I can do myself! I don't have any permissions in this repo

alessio:
@jharshman bump :)

portertech:
@jharshman bump :)

alessio:
@jharshman Hi! Any news on this? :)

alessio:
@marckhouzam bump :)

marckhouzam:
> @marckhouzam bump :)

I'm nobody 😉

But if I may suggest something based on my experience. To help get any PR merged (in any project) you should make the review as easy as possible for the reviewer.  So, in the PR description, you really need to explain what problem this fixes. If you can mention why this is valuable (the impact of the bug), it would also help prioritize this PR over the many others.  Finally, mentioning if there are any backwards-compatibility issues or not would probably help build confidence in your change (that's what makes me unsure of this change, considering my limited experience with Cobra).

Just my opinion though.

command.go
```diff
@@ -903,8 +903,8 @@ func (c *Command) ExecuteC() (cmd *Command, err error) {
 			c = cmd
 		}
 		if !c.SilenceErrors {
-			c.Println("Error:", err.Error())
-			c.Printf("Run '%v --help' for usage.\n", c.CommandPath())
+			c.PrintErrln("Error:", err.Error())
+			c.PrintErr(fmt.Sprintf("Run '%v --help' for usage.\n", c.CommandPath()))
```
marckhouzam, line 907: Why not use `PrintErrf()`, which you've fixed further below?

jharshman, line 907: As @marckhouzam has mentioned, seems like `PrintErrf()` should be used here.

alessio: Seems legit. Will amend accordingly.

alessio:
I amended the PR as requested. I've just tweaked the README.me file (error messages should always go to STDERR, even when a program fails).

@jharshman @marckhouzam please review again when you spare some time.

Thanks for considering.

marckhouzam:
This looks correct of course.

But I don't know if this should be considered a breaking change?  Could a script look for one of these errors on stdout and would stop detecting them if they go to stderr?
How has Cobra handled such changes before? And now that 1.0 is released?

jharshman:
> Could a script look for one of these errors on stdout and would stop detecting them if they go to stderr?

Yes this is a concern and undoubtedly will cause issues to arise. Now that 1.0.0 is out, we are working on a release cycle and will have to work out how to incorporate breaking changes into that cycle such that they are as unsurprising as possible.

@alessio thank you for your contribution. Because this will be a breaking change (albeit a small one), incorporation into master is going to be put on hold until myself and the other maintainers have hashed out a release cycle.

alessio:
Hi there! Any news on this?

itaysk:
can we please merge this? it's undoubtedly a bug, and the fix is very simple.

johmanx10:
I'm running into the same problem and I'm not sure the argument made about fixing this is breaking code for end-users is valid.

- If you have no problem with mixed outputs, this DOES NOT affect you
- If you have written a fix around this issue, by using `ErrOrStderr()` or `PrintErr()`, fixing this DOES NOT affect you, because you are no longer using those methods.
- If you are writing new code, this DOES affect you.

I would urge the maintainers to reconsider their argument for breaking backward compatibility. This is clearly a bug that has crept into the major release. No shame there. And yeah, there is potential someone is going to have a bad day if this is fixed. But that amounts to the same bad day all end users have when trying to implement it correctly. IMHO the cost-benefit of breaking this piece of compatibility is way worth it.

Also, consider what is written in the documentation for this version:
> PrintErrf is a convenience method to Printf to the defined Err output, fallback to Stderr if not set.
- https://godoc.org/github.com/spf13/cobra#Command.PrintErrf

This means there is a mismatch between what is documented behavior for this version and what is actual behavior for this version.

ethanfrey:
Thank you, this would really help us with proper output so we can use the cli tools in scripts more easily.

jgimeno:
I have been waiting this solution for a long time, how come is still not merged? I hope this is being merged soon.

bukowa:
````golang

func TestCommand_Print(t *testing.T) {
	errBuff, outBuff := bytes.NewBuffer(nil), bytes.NewBuffer(nil)

	wantErr := []byte("PrintErrPrintErr line\nPrintErr")
	wantOut := []byte("PrintPrint line\nPrint")

	root := &Command{

			cmd.PrintErr("PrintErr")
			cmd.PrintErrln("PrintErr", "line")
			cmd.PrintErrf("PrintEr%s", "r")

			cmd.Print("Print")
			cmd.Println("Print", "line")
			cmd.Printf("Prin%s", "t")
		},
	}

	root.SetErr(errBuff)
	root.SetOut(outBuff)

	if err := root.Execute(); err != nil {
		t.Error(err)
	}

	errBytes, err := ioutil.ReadAll(errBuff)
	if err != nil {
		t.Error(err)
	}

	outBytes, err := ioutil.ReadAll(outBuff)
	if err != nil {
		t.Error(err)
	}

	if !bytes.Equal(errBytes, wantErr) {
		t.Errorf("got: '%s' want: '%s'", errBytes, wantErr)
	}

	if !bytes.Equal(outBytes, wantOut) {
		t.Errorf("got: '%s' want: '%s'", outBytes, wantOut)
	}
}
````

alessio:
@bukowa I've integrated your test case in this PR. Thanks!

jpmcb (approved):
This is great - thanks so much for the patience and time spent on this. And thanks so much for adding a test case!

PR #885, opened 2019-06-12
zsh: extend flag description quotation

rsteube:
Added `[` and `]` to quotation.

PR #754, opened 2018-09-26
add fish completion support

0robustus1:
This adds generation of fish autocompletion as suggested in #350.

It should support the following types of completion:

* subcommands
  * including the deactivation of subcommand completion after a local non-persistent flag
* long and short flags (with boolean flags not taking arguments and all other flags requiring an argument)
* arguments to subcommands (per `ValidArgs` and `ArgAliases`)

@JHK tested the completion by integrating it with [kubectl](https://github.com/kubernetes/kubernetes/compare/master...JHK:fish_completion). We are planning to integrate fish completion with kubectl after it is available in cobra.

PR #899, opened 2019-07-04
Make sure we quote brackets when generating zsh completion

chmouel:
zsh completion would fail when we have brackets in description, for example :

```
% source   <(./tkn completion zsh)
% tkn task list[TAB]
_arguments:comparguments:319: invalid option definition: --template[Template string or path to template file to use when -o=go-template, -o=go-template-file. The template format is golang templates [http://golang.org/pkg/text/template/#pkg-overview].]:filename:_files
```

### DIFF
--- a/zsh_completions.go
+++ b/zsh_completions.go
@@ -332,5 +332,7 @@ func zshCompFlagCouldBeSpecifiedMoreThenOnce(f *pflag.Flag) bool {
 }
 
 func zshCompQuoteFlagDescription(s string) string {
-	return strings.Replace(s, "'", `'\''`, -1)
+	return strings.NewReplacer("'", `'\''`,
+		"[", `\[`,
+		"]", `\]`).Replace(s)
 }
--- a/zsh_completions.go
+++ b/zsh_completions.go
@@ -25,6 +25,7 @@ var (
 		"extractFlags":                zshCompExtractFlag,
 		"genFlagEntryForZshArguments": zshCompGenFlagEntryForArguments,
 		"extractArgsCompletions":      zshCompExtractArgumentCompletionHintsForRendering,
+		"escapeText":                  zshCompQuoteFlagDescription,
 	}
 	zshCompletionText = `
 {{/* should accept Command (that contains subcommands) as parameter */}}
@@ -41,7 +42,7 @@ function {{$cmdPath}} {
   case $state in
   cmnds)
     commands=({{range .Commands}}{{if not .Hidden}}
-      "{{.Name}}:{{.Short}}"{{end}}{{end}}
+      "{{.Name}}:{{.Short | escapeText}}"{{end}}{{end}}
     )
     _describe "command" commands
     ;;
@@ -334,5 +335,7 @@ func zshCompFlagCouldBeSpecifiedMoreThenOnce(f *pflag.Flag) bool {
 func zshCompQuoteFlagDescription(s string) string {
 	return strings.NewReplacer("'", `'\''`,
 		"[", `\[`,
-		"]", `\]`).Replace(s)
+		"]", `\]`,
+		"$(", `\$(`,
+	).Replace(s)
 }
--- a/zsh_completions_test.go
+++ b/zsh_completions_test.go
@@ -158,6 +158,17 @@ func TestGenZshCompletion(t *testing.T) {
 			},
 		},
 		{
+			name: "flag description with brackets ([]) shouldn't break the completion file",
+			root: func() *Command {
+				r := genTestCommand("root", true)
+				r.Flags().Bool("level", false, "[ALERT]")
+				return r
+			}(),
+			expectedExpressions: []string{
+				`--level[\[ALERT\]]`,
+			},
+		},
+		{
 			name: "argument completion for file with and without patterns",
 			root: func() *Command {
 				r := genTestCommand("root", true)
--- a/zsh_completions_test.go
+++ b/zsh_completions_test.go
@@ -240,6 +240,25 @@ func TestGenZshCompletion(t *testing.T) {
 				`--ptest\[ptest]:filename:_files -g "-\(/\)"`,
 			},
 		},
+		{
+			name: "escape text in subcommand description",
+			root: func() *Command {
+				r := &Command{
+					Use:  "rootcmd",
+					Long: "Long rootcmd description",
+				}
+				d := &Command{
+					Use:   "subcmd1",
+					Short: "$(echo foo)",
+					Run:   emptyRun,
+				}
+				r.AddCommand(d)
+				return r
+			}(),
+			expectedExpressions: []string{
+				`\$\(echo foo\)`,
+			},
+		},
 	}
 
 	for _, tc := range tcs {


chmouel:
CLA link doesnt seem to work 😢

chmouel:
I am not really sure why unittest fails, the spellchecking is only done on bash_completion

rsteube:
The shellcheck simply changed: #889  - problably an update or similar.
Duplicate of #885, this one's got a test though which is good.

JohnnyVim:
#889 has been merged into master. A rebase will fix the tests.

chmouel:
thanks for the heads up i rebased!

umarcor:
@chmouel, would you mind merging #885 into this PR?

chmouel:
@umarcor sure! Done.

chmouel:
I guess that's ready for merge now, isntit ?

rhuss:
Can we get this merged, please ? The changes look reasonable simple and valid (and it also bites us for our ZSH completion business).

Thanks !

corneliusweig:
@chmouel Can you please include the following patch in your PR to fix #989 :

```patch
diff --git zsh_completions.go zsh_completions.go
index b58a5d6..e2476b6 100644
--- zsh_completions.go
+++ zsh_completions.go
@@ -25,6 +25,7 @@ var (
 		"extractFlags":                zshCompExtractFlag,
 		"genFlagEntryForZshArguments": zshCompGenFlagEntryForArguments,
 		"extractArgsCompletions":      zshCompExtractArgumentCompletionHintsForRendering,
+		"escapeText":                  zshCompQuoteFlagDescription,
 	}
 	zshCompletionText = `
 {{/* should accept Command (that contains subcommands) as parameter */}}
@@ -41,7 +42,7 @@ function {{$cmdPath}} {
   case $state in
   cmnds)
     commands=({{range .Commands}}{{if not .Hidden}}
-      "{{.Name}}:{{.Short}}"{{end}}{{end}}
+      "{{.Name}}:{{.Short | escapeText}}"{{end}}{{end}}
     )
     _describe "command" commands
     ;;
@@ -334,5 +335,7 @@ func zshCompFlagCouldBeSpecifiedMoreThenOnce(f *pflag.Flag) bool {
 func zshCompQuoteFlagDescription(s string) string {
 	return strings.NewReplacer("'", `'\''`,
 		"[", `\[`,
-		"]", `\]`).Replace(s)
+		"]", `\]`,
+		"$(", `\$(`,
+	).Replace(s)
 }
diff --git zsh_completions_test.go zsh_completions_test.go
index 8079f6c..a9405fe 100644
--- zsh_completions_test.go
+++ zsh_completions_test.go
@@ -240,6 +240,25 @@ func TestGenZshCompletion(t *testing.T) {
 				`--ptest\[ptest]:filename:_files -g "-\(/\)"`,
 			},
 		},
+		{
+			name: "escape text in subcommand description",
+			root: func() *Command {
+				r := &Command{
+					Use:  "rootcmd",
+					Long: "Long rootcmd description",
+				}
+				d := &Command{
+					Use:   "subcmd1",
+					Short: "$(echo foo)",
+					Run:   emptyRun,
+				}
+				r.AddCommand(d)
+				return r
+			}(),
+			expectedExpressions: []string{
+				`\$\(echo foo\)`,
+			},
+		},
 	}
 
 	for _, tc := range tcs {
```

chmouel:
Hello @corneliusweig, I could surely do but what is the point if authors does not seem to want to merge this ?

corneliusweig:
@chmouel I think a new release is due in December and PRs are usually merged shortly before that. But clearly, the maintainers have other responsibilities so lets bundle these improvements in a single PR.

chmouel:
@corneliusweig sure, let's hope this gets merged, I have updated the PR, thanks

rhuss:
> @chmouel I think a new release is due in December and PRs are usually merged shortly before that. But clearly, the maintainers have other responsibilities so lets bundle these improvements in a single PR.

At least some feedback would be nice about the timing and/or roadmap (this PR is now open since ~ 7 months without any feedback).

corneliusweig:
@eparis Maybe you know about a release schedule for cobra?

For context: this PR is a bugfix for zsh command completion and has been waiting for ~7 months.

corneliusweig:
@chmouel Looks like there was a hickup in the CI. Can you trigger the build once more?

umarcor:
@corneliusweig, see #959, #1028, https://github.com/spf13/cobra/pull/841#issuecomment-568546227 . Milestone [1.0.0](https://github.com/spf13/cobra/milestone/2) has been postponed several times during these last months, and it is currently past due by 5 days. Hence, unfortunately, I wouldn't expect this PR to be merged soon. E.g., #754, which is similar, has been waiting for almost 16 months now...

corneliusweig:
I see, thanks for the context @umarcor!

rhuss:
@umarcor thanks a lot for taking care and the context around the maintenance issue. Considering that the last commit on master is from January 6th (an update of license headers), I can't see any progress towards a 1.0 release "early this year".

Maybe if stability is such a huge concern, just lets tag the current master as 1.0 and then start merging the pending PRs afterwards, so that a user has the choice, whether to keep the current state (which is the status quo since some years already, so those folks don't requite any bug fixes anyway I guess) by depending on 1.0 or want to try out the fixes, being aware that this also might involve change.

umarcor:
@rhuss, I agree. IMHO branch 2.0 should have been created long ago, and the vast majority of open PRs should have been merged. That'd allow a large amount of users to slowly move forward, test for bugs, and prepare it for projects that seek stability to have a reference to look at. Actually, there are ~15 PRs that could be merged without risking stability, and another ~15, at least, that could be merged quickly.

Nevertheless, https://github.com/spf13/cobra/issues/959#issuecomment-531612127 was never responded and the issue was blocked. I opened #1028 to hopefully bring the discussion, but no one replied or reacted (except you). Hence, I believe we can assume this project to be archived de facto.

rhuss:
I don't think this PR should be considered stale.

umarcor:
@rhuss >70% of the open issues are already labeled as 'kind/stale' and it seems not possible to tell the bot otherwise. I believe it is a misconfiguration.

corneliusweig:
@umarcor If the bot doesn't obey, just do a `git commit --amend` and force-push the result. Then the PR won't be stale anymore :)

umarcor:
@corneliusweig, are you sure about that? That's what I would expect from any regular bot that handles stale issues. Some do also detect edition of comments, so it is not even required to force-push (any participant can help). However, the bot that is being used in this repo seems not to unmark anything: https://github.com/actions/stale. Unfortunately, most of the actions in https://github.com/actions feel like placeholders for marketing purposes. Some months will be required until quality catches up. See https://github.com/spf13/cobra/issues/952#issuecomment-613017612.

corneliusweig:
No, I'm not sure. I was merely guessing. But have you tried? I'm curious what will happen 😄

umarcor:
> No, I'm not sure. I was merely guessing. But have you tried? I'm curious what will happen 😄

I *preventively* ensured that all the PRs I proposed are updated and have a recent comment: https://github.com/spf13/cobra/pulls?q=is%3Apr+is%3Aopen+sort%3Aupdated-desc+author%3Aumarcor 😄

chmouel:
I gave up updating this PR, afer eight months being hopeful the maintainers just don't seem to care about those fixes

marckhouzam:
The zsh completion logic has been revamped by #1070 and this problem has been addressed.

@jpmcb I believe we can close this one now that #1070 has been merged.

navidshaikh:
@marckhouzam : Thanks for the fix!

Noticed a difference in the way auto-completion of flags starting with `-` or `--` are handled.

With the patch in this PR #899 , it also allows to auto-complete the flags, for eg:
```
➜  client git:(master) kn service create [tab]
```
would add `-` after create and another [tab] after `-` lists all the flags
```
kn service create -[tab]
zsh: do you wish to see all 129 possibilities (43 lines)? [tab]
--annotation               -a  -- Service annotation to set. name=value; you may provide this flag any number of times to set multiple annotations. To unset, specify the annotation name fo
--arg                          -- Add argument to the container command. Example: --arg myArg1 --arg --myArg2 --arg myArg3=3. You can use this flag multiple times.                         
[..]
```

While if I try using https://github.com/spf13/cobra/pull/1070, and [tab] after `kn service create` it appends `:0` to the command
```
kn service create [tab]
```
gives

```
kn service create :0
```
while if I *type* `-` after `kn service create`, it would list the available flags.

So the difference is, this patch also takes care of auto-completing the flags, while #1070 would add `:0`.

Reproduce:
1. With the patch in this PR
```
git clone https://github.com/knative/client && cd client
./hack/build.sh 
source <(./kn completion zsh)
compdef _kn kn
```

2. With #1070 
```
git clone -b pr/cobra-upgrade https://github.com/navidshaikh/client && cd client
./hack/build.sh 
source <(./kn completion zsh)
compdef _kn kn
```

marckhouzam:
@navidshaikh Thank you for trying the new zsh completion early, it will help confirm there is no issues.

> Noticed a difference in the way auto-completion of flags starting with `-` or `--` are handled.
> With the patch in this PR #899 , it also allows to auto-complete the flags, for eg:
> ```
> ➜  client git:(master) kn service create [tab]
> ```
> would add `-` after create and another [tab] after `-` lists all the flags

This change is on purpose to align it with bash-completion behaviour.  It is important that shell-completion behaves the same across all shells, so we had to make this change since bash completion only completes flags if the user types the first `-`.  I also believe this is the better behaviour.

I realize now that I didn't document it in: https://github.com/spf13/cobra/blob/master/zsh_completions.md
Do you want to post a PR to add this information?

> While if I try using #1070, and [tab] after `kn service create` it appends `:0` to the command
> ```
> kn service create [tab]
> ```
> gives
> ```
> kn service create :0
> ```

This is not right.
But thanks to the instructions you gave on how to reproduce I was able to debug this and it looks like it is a particularity of `kn`.
First, to help see the problem you can do:
```
$ ./kn __complete s<ENTER>
:0
service Manage Knative services
source  Manage event sources
:4
```
You'll notice that **two** `ShellCompDirective` are printed.  This should not happen.  This is where the `:0` comes from, as the zsh completion script does not expect to have two directives.  In fact if you complete
```
$ ./kn <tab>
broker      -- Manage message broker
completion  -- Output shell completion code
help        -- Help about any command
options     -- Print the list of flags inherited by all commands
plugin      -- Manage kn plugins
revision    -- Manage service revisions
route       -- List and describe service routes
service     -- Manage Knative services
source      -- Manage event sources
trigger     -- Manage event triggers
version     -- Show the version of this client
:0
```
you will notice the extra `:0` at the end which should not be there.

From what I can see, `kn` executes an extra command at the start:
https://github.com/knative/client/blob/217a8133446acfadd19600a768b2be52fb7afd19/cmd/kn/main.go#L119
https://github.com/knative/client/blob/217a8133446acfadd19600a768b2be52fb7afd19/cmd/kn/main.go#L144

I believe this is what is causing the extra `ShellCompDirective` to be printed when the `__complete` command is called.
I didn't spend more time to try to understand why `kn` runs such a command, so I'm not sure if it is something to fix in `kn` or if we should put something in Cobra to protect against such a case.

Could you give more information about that?

marckhouzam:
> This change is on purpose to align it with bash-completion behaviour. It is important that shell-completion behaves the same across all shells, so we had to make this change since bash completion only completes flags if the user types the first `-`. I also believe this is the better behaviour.
> 
> I realize now that I didn't document it in: https://github.com/spf13/cobra/blob/master/zsh_completions.md
> Do you want to post a PR to add this information?

I had to make a documentation PR for something else, so I did this also.  See #1169.

navidshaikh:
> I believe this is what is causing the extra ShellCompDirective to be printed when the __complete command is called.
I didn't spend more time to try to understand why kn runs such a command, so I'm not sure if it is something to fix in kn or if we should put something in Cobra to protect against such a case.

> Could you give more information about that?

@marckhouzam : thank you for debugging this!
kn offers plugins support. We need to find the exact set of args (valid/invalid) without flags (and their values).
The computed set of args is then used to check if there is matching kn-plugin available which implements the given command. 

For example:
Given a plugin `kn-source-kafka` exists (in the path kn looks for) and provided command is
`kn --namespace foo source kafka create -h` ; 
it would compute the args as `[kn source kafka create]` and then combine the args to find the matching plugins (longest path first). If found the plugin among combination of these paths, kn will execute the plugin.
This happens before we execute the actual kn root command, and in order to extract the args without the flags (and their values), we inject an `extractCommand` as you [observed](https://github.com/knative/client/blob/217a8133446acfadd19600a768b2be52fb7afd19/cmd/kn/main.go#L119) and execute it. 
This way the `extractCommand` leverages cobra parsing logic for flags, etc and returns the list of args we are looking for.

rhuss:
Of course, there might be a better way to let cobra parse the arguments without temporarily creating a command. If so, please let us know.

marckhouzam:
> Of course, there might be a better way to let cobra parse the arguments without temporarily creating a command. If so, please let us know.

I'm not sure if it's the best way, but this seems to strip all the flags and their values and leaves you with args only:
```
	rootCmd.FParseErrWhitelist = cobra.FParseErrWhitelist{UnknownFlags: true}
	if err := rootCmd.ParseFlags(os.Args[1:]); err != nil {
		fmt.Printf("Error while parsing flags from args %v: %s\n", os.Args[1:], err.Error())
	}
	finalArgs := rootCmd.Flags().Args()
	fmt.Printf("args %v\n", finalArgs)

```

rsteube:
Yeah, that's probably the right way - i used `Traverse` / `ParseFlags` as well: https://github.com/rsteube/carapace/blob/master/carapace.go#L194

navidshaikh:
Thanks @marckhouzam @rsteube for the help! we could remove the extra command execution and use inbuilt parsing utils, and the extra directive / `:0` issue seems to be resolved.

> This change is on purpose to align it with bash-completion behaviour. It is important that shell-completion behaves the same across all shells, so we had to make this change since bash completion only completes flags if the user types the first -. I also believe this is the better behaviour.

Is it shell-completion limitation to have `-` typed for auto-completing the flags ? 
IMO, different shells bring distinct capabilities which could be leveraged to provide users with more power.

rsteube:
No, as far as i know that depends how you configure it. I think cobra had something like a `required` option for flags, which is meant for those to be shown during completion even without preceeding `-`. Still, that working depends on how the completion is generated. Seems to be the common default to show flags only with preceeding `-` though. Personally i normally prefer it that way, and intentionally generated my completion this way so far (although there's a good reason to do it the other way on some occasions).

marckhouzam:
I believe @rsteube has explained things very well.  Here is the doc for marking a flag "required": https://github.com/spf13/cobra/blob/master/shell_completions.md

marckhouzam:
Whoops, more precise URL for required flags
https://github.com/spf13/cobra/blob/master/shell_completions.md#mark-flags-as-required

PR #884, opened 2019-06-12
zsh: added custom completions

rsteube:
Added custom completions for zsh.

PR #900, opened 2019-07-05
Allow a default function to argument

chmouel:
If user add an annotation with the
`cobra_annotations_zsh_completion_default_function` key and a value to a
function it will be used as a default completion for that command.


chmouel:
closing since I just saw this, https://github.com/spf13/cobra/pull/884

PR #876, opened 2019-06-07
Add golangci-lint, deprecate go 1.10.x and 1.11.x, cleanup/fixes

umarcor:
Close #774
Close #853
Close #933

In this PR, golangci-lint is added to Travis CI. Then, multiple linting issues are fixed: errors are handled, unused variables/functions are removed, unnecessary types are removed, frequently used strings are defined as constants, etc.

As a result of using `io.StringWriter` and `os.UserHomeDir`, merging this PR implies dropping support for golang 1.10.x and 1.11.x.

PR #906, opened 2019-07-14
Use os.UserHomeDir() in document

micnncim:
Related to https://github.com/spf13/cobra/pull/853, replaced `mitchellh/go-homedir` with `os.UserHomeDir()` in README.md.

### DIFF
--- a/README.md
+++ b/README.md
@@ -213,7 +213,6 @@ import (
   "fmt"
   "os"
 
-  homedir "github.com/mitchellh/go-homedir"
   "github.com/spf13/cobra"
   "github.com/spf13/viper"
 )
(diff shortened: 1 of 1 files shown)


umarcor:
Hi @micnncim! I am not sure about this PR making sense on its own.

In #853, this change is applied along with making the codebase compatible for go 1.10, 1.11 and 1.12.

In #876, this change is applied along with deprecating go 1.10 and 1.11, thus making cobra >=1.12 only.

Since this PR replaces `mitchellh/go-homedir` with `os.UserHomeDir` without further modifications, I assume that you agree with #876. If so, I'd suggest to close this issue and review #876 instead.

micnncim:
Thank you @umarcor! I don't notice #876 so will close this issue.

PR #912, opened 2019-07-19
Used goimports to fix import order

muesli:
Keeps the list sorted and prevents future merge conflicts.


umarcor:
This PR is already part of #876 (see [first commit](https://github.com/spf13/cobra/pull/876/commits/e9ea59bd041c0997db53effa0f16bd78bb7d3516)). However, I think it's ok to merge it independently, since #876 deprecates golang 1.10 and 1.11.

PR #913, opened 2019-07-19
Simplified assignment operation

muesli:
`extras +=` just looks neater than `extras = extras +`.

### DIFF
--- a/zsh_completions.go
+++ b/zsh_completions.go
@@ -318,7 +318,7 @@ func zshCompGenFlagEntryExtras(f *pflag.Flag) string {
 		case BashCompFilenameExt:
 			extras = ":filename:_files"
 			for _, pattern := range values {
-				extras = extras + fmt.Sprintf(` -g "%s"`, pattern)
+				extras += ` -g "` + pattern + `"`
 			}
 		}
 	}


zsh_completions.go
```diff
@@ -318,7 +318,7 @@ func zshCompGenFlagEntryExtras(f *pflag.Flag) string {
 		case BashCompFilenameExt:
 			extras = ":filename:_files"
 			for _, pattern := range values {
-				extras = extras + fmt.Sprintf(` -g "%s"`, pattern)
+				extras += fmt.Sprintf(` -g "%s"`, pattern)
```
umarcor: Might this work too?

``` golang
extras += ` -g "` + pattern + `"`
```

See https://gist.github.com/dtjm/c6ebc86abe7515c988ec

muesli: It would, yeah, even though I guess the speed difference is negligible here. We should probably go with whatever we consider more readable.

umarcor: Agree. This is just a matter of style. That's why I approved the PR anyway. I added this comment for the sake of discussion only.

I find avoiding 'sprintf' more readable, because it makes it explicit that no data transformation/representation is implied; this is just a concatenation.

muesli: I tend to agree, yeah. I'll update the PR.

jharshman:
I'm reluctant to approve this.  There's no change in functionality and everyone has their own preferences and styles.

muesli:
Conveying the same information in shorter, clearer form seems always beneficial to me (when it comes to programming anyway), but I fully understand others may have different preferences.

jharshman:
Striving for clarity is great, but I don't think there's a huge case for the change here.
Take for example, your first commit, I thought that style was perfectly fine and easy to read. @umarcor suggested a different style, one which I personally find less clear.  All three styles, get the same job done.

Because there's no change in functionality and the change comes down to preference, I'm going to close this.

PR #904, opened 2019-07-10
update generator docs

jharshman:
Update Generator Docs to reflect changes brought in #817 

Covers PRs:
#878
#888

PR #916, opened 2019-07-22
Fix cobra generator readme file

xsb:
When following the current instructions I found the following error:

```
Error: required flag(s) "pkg-name" not set
```

This change updates the _cobra init_ step to clarify that a flag is required to specify the package name.


xsb:
Closing myself as this is a duplicated from https://github.com/spf13/cobra/pull/904

PR #924, opened 2019-08-01
fix undefined cfgFile in documentation

gregoiremahe:
Someone forgot to initialize cfgFile variable on documentation example which leads to error when doing go get


jharshman:
It should be noted that copying/pasting from the README into your source code isn't advised as the examples are just code snippets. You should use the cobra generator instead.

PR #936, opened 2019-08-23
Update .gitignore to track cobra subdirectory

rhcarvalho:
The 'cobra' entry on .gitignore was meant to ignore a potential 'cobra'
executable generated by 'go build' during development. However, 'cobra'
is a directory containing the command line tool, and such a generic
entry in .gitignore ignores too much.

Instead of the generic 'cobra' entry, ignore only a specific
'/cobra/cobra' path.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -32,7 +32,7 @@ Session.vim
 tags
 
 *.exe
-cobra
+/cobra/cobra
 cobra.test
 
 .idea/


umarcor:
Hi @rhcarvalho! I just sumitted a similar PR (#946). The difference is that you ignore an explicit location `cobra/cobra`, while I suggest to ignore any `cobra` which is not `cobra/`. wdyt?

rhcarvalho:
Hi @umarcor. I had considered initially a similar approach as what you wrote in your PR. However, on second thought and seeing how that line was introduced, I realized that ignoring `/cobra/cobra` is simpler and clearer, does what it needs to do and nothing more.

It is not possible to have a "cobra" executable in the top-level (because of the directory with the same name), and there is no good reason to generically ignore other files or directories named "cobra" in other paths.

umarcor:
Thanks for clarifying! I'm closing #946 in favour of this.

jharshman:
closed by https://github.com/spf13/cobra/pull/976

umarcor:
@jharshman, #976 seems to work around this issue, but not fix it. Using the makefiles will not produce `cobra/cobra`. However, users that build cobra explicitly will still find it, isn't it?

PR #946, opened 2019-09-05
fix: gitignore should not ignore subdir 'cobra'

umarcor:


umarcor:
Closed in favour of #936.

PR #948, opened 2019-09-05
fix: ensure that testproject is removed even after a failure

umarcor:
In https://github.com/spf13/cobra/commit/984374f5b622efc94388e2872915c2d97404adba a `defer func()` was added after a possible `t.Fatal(err)`. This PR moves it before, to ensure that testproject files are removed after a failure.

### DIFF
--- a/cobra/cmd/init_test.go
+++ b/cobra/cmd/init_test.go
@@ -7,24 +7,23 @@ import (
 	"testing"
 )
 
-func TestGoldenInitCmd(t *testing.T) {
-
+func getProject() *Project {
 	wd, _ := os.Getwd()
-	project := &Project{
+	return &Project{
 		AbsolutePath: fmt.Sprintf("%s/testproject", wd),
 		Legal:        getLicense(),
 		Copyright:    copyrightLine(),
-
-		// required to init
-		AppName: "testproject",
-		PkgName: "github.com/spf13/testproject",
-		Viper:   true,
+		AppName:      "testproject",
+		PkgName:      "github.com/spf13/testproject",
+		Viper:        true,
 	}
+}
+
+func TestGoldenInitCmd(t *testing.T) {
+	project := getProject()
 	defer os.RemoveAll(project.AbsolutePath)
 
-	// init project first
-	err := project.Create()
-	if err != nil {
+	if err := project.Create(); err != nil {
 		t.Fatal(err)
 	}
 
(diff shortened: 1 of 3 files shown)


cobra/cmd/project.go
```diff
@@ -75,6 +75,7 @@ func (p *Project) createLicenseFile() error {
 	if err != nil {
 		return err
 	}
+	defer licenseFile.Close()
```
rhcarvalho: Good catch!

umarcor: Thanks! It was actually hard to spot because all the CI tests pass, but this was making them fail in a `golang` docker container on a Windows 10 host. I'm yet not sure about this 'fix' being required for all versions of golang; or whether it is specific to windows hosts. Nevertheless, I believe it is better to have it.

jharshman:
Looks like a good bug fix @umarcor.  Got my 👍but will give others a chance to review the change as well.  I know it's not part of your change-set but If you don't mind, I think that `defer` statement can be simplified as well.

umarcor:
Thanks @jharshman!

>  I know it's not part of your change-set but If you don't mind, I think that `defer` statement can be simplified.

I wouldn't mind to do so, but I'm sleepy now and I don't understand what you mean. How do you suggest to simplify it? Note that I would apply the simplification to `add_test.go` too (which is the other file that was modified in https://github.com/spf13/cobra/commit/984374f5b622efc94388e2872915c2d97404adba).

jharshman:
>How do you suggest to simplify it?

According to the package docs, `os.RemoveAll(path string)` doesn't fail when the path doesn't exist.  Therefore you should be able to simplify to `defer os.RemoveAll(project.AbsolutePath)`.

ref: https://golang.org/pkg/os/#RemoveAll

umarcor:
Thanks for the hint @jharshman! I updated the PR accordingly.

cobra/cmd/init_test.go
```diff
@@ -12,24 +12,22 @@ func TestGoldenInitCmd(t *testing.T) {
 	wd, _ := os.Getwd()
 	project := &Project{
 		AbsolutePath: fmt.Sprintf("%s/testproject", wd),
-		PkgName:      "github.com/spf13/testproject",
 		Legal:        getLicense(),
 		Copyright:    copyrightLine(),
-		Viper:        true,
-		AppName:      "testproject",
+
+		// required to init
```
jharshman: I think `AbsolutePath` is also required.  I'm sure others could be as well.  Pls remove comment and group separation.  Unless, that is, there is a really good reason that I'm not seeing.

umarcor: I just copied it from `add_test.go` so that the similarity between sources was easier to spot. Nonetheless, I added a helper function now (`getProject`) that is used by both `TestGoldenInitCmd` and `TestGoldenAddCmd`.

PR #953, opened 2019-09-08
update viper to v1.4.0

umarcor:
Fix #938
Fix #978

PR #954, opened 2019-09-08
update go-yaml to v3

umarcor:


jharshman:
@umarcor is there a reason for the update?
PR descriptions should never be blank.

umarcor:
@jharshman, there is no special reason, but the preference to slowly keep up to date, instead of having large conflicts afterwards. That's why there is no description. Unlike #953, this is just a *chore*.

BTW, spf13/pflag v1.0.5 was released two weeks ago. I didn't open a PR yet, because of the *maintenance status*.

jharshman:
> That's why there is no description. Unlike #953, this is just a chore.

PR descriptions should never be blank, even for a "chore".  This causes maintainers to guess _why_ this PR has been opened.  i.e: bug fixes? new feature?

jharshman:
Looks like they have not tagged a v3.0.0 release yet.
Gonna hold off on upgrading for the time being.
https://github.com/go-yaml/yaml/issues/487

umarcor:
Sure. There is also v2.2.3. I could update the PR accordingly.

Nevertheless, I think that #953 is prioritary. By the time we have other prioritary PRs merged, there might be a tagged v3.0.0 for go-yaml.

PR #964, opened 2019-09-19
Added ability to do custom stuff when command is not found

mjovanovic0:
Provide a way to do custom code when command is not found.


umarcor:
Ref #841 and #842.

Issue #967, opened 2019-10-01
Undefined variables in README.md example


srybacki:
The variables `projectBase` and `userLicense` are used  in the [README.md example](https://github.com/spf13/cobra#example), but the variables are not declared in the example code.  Should the variables be declared as package-level variables like `cfgFile`?

umarcor:
Actually, `cfgFile` was added just a couple of months ago. As @jharshman [commented](https://github.com/spf13/cobra/pull/924#issuecomment-518292844):

> It should be noted that copying/pasting from the README into your source code isn't advised as the examples are just code snippets. You should use the cobra generator instead.

Issue #910, opened 2019-07-17
How can I include Cobra in my actual project structure ?


fallais:
I have this structure (not exhaustive) :

```
models/
services/
cmd/
  server/
    routes/
    Dockerfile
    main.go
  collector/
    Dockerfile
    main.go
```

**server** and **collector** are my two micro-services. They both have a **Dockefile**. That means I need to create two images at each build. Boring. That is why I want to migrate to Cobra.

But it seems that my vision of `cmd` interferes with the Cobra's one.

How can I do please ? I think about putting a `main.go` at the root of `cmd/`. But how can I link the two `commands` to my two existing `main.go` ?

Then, the Dockerfile, I assume that it will also be placed at the root of `cmd/` ?

In fact, problem is that my `cmd` does not contain commands, but the whole micro-services.

jharshman:
@fallais I think I need a bit more clarity on your question.
What exactly do you want to utilize Cobra for?

umarcor:
@fallais, thanks for the clarification. Yes, cobra is suitable for your use case. However, let's ignore the Dockerfiles for now; it is not relevant for the golang codebase, and it adds unnecessary complexity to the question. I would rephrase the question as:

*I have two independent golang apps which are built to separate binaries. Now, I want to integrate/merge both of them in a single binary. How can I use cobra to achieve it?*

I suggest to start by creating a completely new app structure, using cobra's generator:

``` bash
go get -u github.com/spf13/cobra/cobra
mkdir prjdir
cd prjdir
cobra init --pkg-name github.com/user/prjdir
cobra add server
cobra add collector
go mod init github.com/user/prjdir
```

You will get the following structure:

```
prjdir/
LICENSE
main.go
cmd/
  root.go
  collector.go
  server.go
```

Now, check the codebase, and play with it:

```
root@e05d1fadee50:/go/prjdir# go run main.go
A longer description that spans multiple lines and likely contains
examples and usage of using your application. For example:

Cobra is a CLI library for Go that empowers applications.
This application is a tool to generate the needed files
to quickly create a Cobra application.

Usage:
  prjdir [command]

Available Commands:
  collector   A brief description of your command
  help        Help about any command
  server      A brief description of your command

Flags:
      --config string   config file (default is $HOME/.prjdir.yaml)
  -h, --help            help for prjdir
  -t, --toggle          Help message for toggle

Use "prjdir [command] --help" for more information about a command.

root@e05d1fadee50:/go/prjdir# go run main.go server
[...]

PR #973, opened 2019-10-04
Update cmd/root.go example in README.md

jharshman:
The cmd/root.go example in the readme is out of date.  This PR updates the example with what is currently in cmd/root.go.

Resolves https://github.com/spf13/cobra/issues/967

### DIFF
--- a/README.md
+++ b/README.md
@@ -210,53 +210,72 @@ You will additionally define flags and handle configuration in your init() funct
 For example cmd/root.go:
 
 ```go
-import (
-  "fmt"
-  "os"
+package cmd
 
-  homedir "github.com/mitchellh/go-homedir"
-  "github.com/spf13/cobra"
-  "github.com/spf13/viper"
+import (
+	"fmt"
+
+	homedir "github.com/mitchellh/go-homedir"
+	"github.com/spf13/cobra"
+	"github.com/spf13/viper"
 )
 
-var cfgFile string
+var (
+	// Used for flags.
+	cfgFile     string
+	userLicense string
+
+	rootCmd = &cobra.Command{
+		Use:   "cobra",
+		Short: "A generator for Cobra based Applications",
+		Long: `Cobra is a CLI library for Go that empowers applications.
+This application is a tool to generate the needed files
+to quickly create a Cobra application.`,
+	}
+)
+
+// Execute executes the root command.
+func Execute() error {
+	return rootCmd.Execute()
+}
 
 func init() {
-  cobra.OnInitialize(initConfig)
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -210,53 +210,72 @@ You will additionally define flags and handle configuration in your init() funct
...
-  viper.BindPFlag("author", rootCmd.PersistentFlags().Lookup("author"))
-  viper.BindPFlag("projectbase", rootCmd.PersistentFlags().Lookup("projectbase"))
-  viper.BindPFlag("useViper", rootCmd.PersistentFlags().Lookup("viper"))
-  viper.SetDefault("author", "NAME HERE <EMAIL ADDRESS>")
-  viper.SetDefault("license", "apache")
+	cobra.OnInitialize(initConfig)
+
+	rootCmd.PersistentFlags().StringVar(&cfgFile, "config", "", "config file (default is $HOME/.cobra.yaml)")
+	rootCmd.PersistentFlags().StringP("author", "a", "YOUR NAME", "author name for copyright attribution")
+	rootCmd.PersistentFlags().StringVarP(&userLicense, "license", "l", "", "name of license for the project")
```
umarcor: I not convinced about providing such example flags. This is extracted from a very specific tool/app, which is the cobra generator/tool. The fact that both the library and the cmd tool have the same name and are located in the same repo might already be confusing for new users. Providing an example that is based on this so special *inception* might lead users to think that flags `author` or `userLicense` have any effect at all on the license of the app they are building, which is not the case.

If we want to copy some "real world" example, I think we'd better pick any other (i.e. hugo). Furthermore, I think we should have a proper "examples" or "guide" subdir with MWEs. A cobra project can be described in a single markdown file. Ref #910.

Of course, this is not something you introduced. Hence, I completely agree if you consider it to be out of scope of this PR.

jharshman: This PR is just to update the documentation that already exists.  It could be useful to have some sort of more in-depth documentation, however, a list of projects that Cobra is used in is already in the readme so examples such as Hugo are only a few clicks away.

umarcor: So, don't you think that all the flags and viper binds in this `init` function can be misleading rather than helpful? I mean to just remove lines 246-252 from the current proposal.

jharshman: I honestly don't believe this example to be all that confusing.  To be clear, this does not add _more_ flags to the init function example, but rather just updates it with a more current reflection of Cobra's own `cmd/root.go'.

I'm not saying there aren't changes to be made to improve documentation.  I'm saying that because Cobra is widely used, documentation changes need to be well thought through.

PR #975, opened 2019-10-05
Check if root cmd is runnable and run if it is

### LINKED ISSUE #974: RunE on rootCmd is never being called.
I am trying to parse Persistent Flags that I pass to my rootCmd. I have next flags:
```
rootCmd.PersistentFlags().IntVar(&logLevel, "log", 2, "Set log level 0-NONE, 1-INFO, 2-DEBUG")
rootCmd.PersistentFlags().StringVar(&confPath, "config", "config/default.conf", "Pass path to UPM config file")                   
``` 
And I have rootCmd.RunE which handles this flags. Everything would be perfect but RunE is never being called. If I call it explicitly, flags are being parsed successfully.
Why is that?

I will attach a folder with CLI handling after this message.

h0tw4t3r:
Should resolve #974 
"Should" because I am not sure if it resolves it, I tried cloning repo and copying it to my project folder as module - didn't work out.
Though I didn't change much, it should now save rootCmd to cmd and Run it to be able to parse flags as in #974.

### DIFF
--- a/command.go
+++ b/command.go
@@ -892,7 +892,10 @@ func (c *Command) ExecuteC() (cmd *Command, err error) {
 	}
 
 	var flags []string
-	if c.TraverseChildren {
+	if c.Runnable() {
+		cmd = c
+		flags = args
+	} else if c.TraverseChildren {
 		cmd, flags, err = c.Traverse(args)
 	} else {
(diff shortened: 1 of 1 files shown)


h0tw4t3r:
OK, I understood how to test it, I will update branch asap.

PR #968, opened 2019-10-02
Add golangci-lint and GitHub Actions workflow

umarcor:
Close #952

In this PR<strike>, Travis CI and CircleCI are deprecated in favour of</strike> a GitHub Actions workflow is added:

- As proposed in #952, GNU/Linux (ubuntu-latest), Windows and macOs platforms are tested. 
- Regarding golang versions, the proposed matrix includes `1.10.x`, `1.11.x`, `1.12.x`, `1.13.x` and `1.14.x`. 
- <strike>Compared to the current Travis matrix, `1.10.x`, `1.11.x` and `tip` are removed and `1.13.x` and `1.14.x` are added. Note that `tip` is not supported in GitHub Actions by default yet (actions/setup-go#21).</strike>

<strike>Execution example: https://github.com/dbhi/cobra/commit/9bdf771ff8e6f415a56446759031a9864e4d6c50/checks?check_suite_id=248869625</strike>

PR #976, opened 2019-10-07
Add Makefile to project

jharshman:
Add Makefile(s) to project.  Tweak TravisCI to use the makefile targets.  Remove Circle CI as it currently only duplicates work.

* Developers test and build using the same commands the CI system will execute.
* Leverage github.com/mitchellh/gox to cross compile for different architectures.
* Generator builds drop into ./bin directory.

Resolves:
https://github.com/spf13/cobra/pull/936
https://github.com/spf13/cobra/pull/876


.travis.yml
```diff
@@ -10,19 +10,20 @@ go:
...
+    - stage: test
+      script: make test
 
 script:
-  - richgo test -v ./...
-  - go build
+  - make
   - if [ -z $NOVET ]; then
       diff -u <(echo -n) <(go vet . 2>&1 | grep -vE 'ExampleCommand|bash_completions.*Fprint');
     fi
```
umarcor: Shouldn't this be added as another make target?

cobra/Makefile
```diff
@@ -0,0 +1,23 @@
+XC_OS="linux darwin"
```
umarcor: Why linux and darwin only? Is there any issue with windows?

jharshman: The intention is to provide defaults.  These variables can be overridden when calling `make`.  We do not currently build for Windows, but this is laying the groundwork to build and publish binaries for various operating systems and architectures.

.travis.yml
jharshman: yes, take note of "WIP" in the title of the PR.

cobra/Makefile
umarcor: Do you mean to provide an example about how users could publish their tools? Or to actually publish cobra cli tool periodically? In the second case, and just because of curiosity, may I know why not https://github.com/goreleaser/goreleaser, which is quite popular?

jharshman: I mean to publish the CLI tool periodically. Honestly was not aware of the tool you have mentioned.  I'll look into it but for now gox works fine as all I want to do is build a few binaries in parallel and stick them in the bin directory.

umarcor:
@jharshman, how about adding `1.13.x`?

jharshman:
@umarcor Well right now it's testing tip which I assume is the latest and greatest version of Go.
Currently I' considering the following - 
* Removing Circle CI as it seems repetitive.
* To what degree do we care about running tests against older versions of Go? i.e: 1.13.x is out, do we care about testing 1.10.x?

umarcor:
> @umarcor Well right now it's testing tip which I assume is the latest and greatest version of Go.

`1.13.x` was released last month: https://blog.golang.org/go1.13. `tip` is currently `1.14.x`.

> * Removing Circle CI as it seems repetitive.

Agree. Precisely, I would remove Travis too, and go with GitHub Actions. I can merge this with #968.

> * To what degree do we care about running tests against older versions of Go? i.e: 1.13.x is out, do we care about testing 1.10.x?

Officially, the two latest versions are supported: https://golang.org/doc/devel/release.html#policy. Hence, neither `1.10.x` nor `1.11.x` are supported. Precisely, that's the motivation to deprecate both of them in #876. Should you decide to remove them, please consider adding #876 to [1.0.0](https://github.com/spf13/cobra/milestone/2).

Issue #933, opened 2019-08-22
Generated files are not `goimports`-ed


andig:
See https://travis-ci.org/andig/ingress/builds/575241832#L232 for an example. It seems as if cobra templates prefer spaces for indent while goimport wants tabs.

Would it make sense to enhance the cobra utility command to goimport or gofmt the generated assets?

umarcor:
I think this could be solved by fixing the template (https://github.com/spf13/cobra/blob/master/cobra/tpl/main.go), instead of requiring some specific lint/format tool. It seems that `AddCommandTemplate` uses tabs only, but spaces are used in `MainTemplate` and `RootTemplate`. The specific error you referenced is because of line 13.

andig:
> BTW, this is fixed in #876.

I see. #876 seems to contains too many changes at once to actually make it?

umarcor:
> I see. #876 seems to contains too many changes at once to actually make it?

Sure. I just added it as a reference. Go ahead, open #935 and ping @jharshman. This minor change is likely to be merged easily on it's own.

umarcor:
Should this be closed now that #980 is merged?

PR #980, opened 2019-10-19
Apply formatting to templates

jharshman:
### Apply Go formatting to templates
Currently, after templates are generated, `gofmt` has to be run against the files in order to properly format them.

Goal here is to have the output of the templates not need additional formatting to fall in line with Go standard format.


jharshman:
Failing on the comparison with the "golden" files.  Fixing...

umarcor:
Close #933
Ref #876

PR #1001, opened 2019-12-12
Fix regression when calling *_custom_func

marckhouzam:
PR #889 introduced a small regression where the global variable `$c` is no longer set when `*custom_func` is called.  This is because `$c` is re-used by mistake in the new `read` loop.

This PR simply changes the name of the variable used in the loop.

I've proposed a PR to the `helm/helm` projet which uses the `$c` variable in `__helm_custom_func` and when testing with the `master` branch of Cobra, things started failing.  With this PR applied, the `helm/helm` PR works again.

See https://github.com/helm/helm/pull/7078/files#diff-84d051076d41e254e011a57e40b82a34R279 for the line using `$c` in the PR.

/cc @rsteube @jharshman

### DIFF
--- a/bash_completions.go
+++ b/bash_completions.go
@@ -61,6 +61,7 @@ __%[1]s_contains_word()
 __%[1]s_handle_reply()
 {
     __%[1]s_debug "${FUNCNAME[0]}"
+    local comp
     case $cur in
         -*)
             if [[ $(type -t compopt) = "builtin" ]]; then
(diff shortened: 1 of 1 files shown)


marckhouzam:
BTW, the helm project has a set of automated regression tests for shell auto-completion, which can help test such changes.
https://github.com/helm/acceptance-testing/blob/master/scripts/completion-tests/completionTests.sh

rsteube:
LGTM

PR #1003, opened 2019-12-14
Add groups for commands in help

### LINKED ISSUE #836: Add ability to group commands in help message
Wondering if:

1. Is this possible via [custom help commands](https://github.com/spf13/cobra/tree/v0.0.3#defining-your-own-help)? i.e. `cmd.SetHelpCommand(cmd *Command)`
2. Is this something you (the author) would consider? Or you (the reader) who might have come across this by searching through the issues?

We don't have to group all commands. Maybe just the root command. A good example

```console
$ kubectl help
kubectl controls the Kubernetes cluster manager.

Find more information at: https://kubernetes.io/docs/reference/kubectl/overview/

Basic Commands (Beginner):
  create         Create a resource from a file or from stdin.
  expose         Take a replication controller, service, deployment or pod and expose it as a new Kubernetes Service
  run            Run a particular image on the cluster
  set            Set specific features on objects

Basic Commands (Intermediate):
  explain        Documentation of resources
  get            Display one or many resources
  edit           Edit a resource on the server
  delete         Delete resources by filenames, stdin, resources and names, or by resources and label selector

Deploy Commands:
  rollout        Manage the rollout of a resource
  scale          Set a new size for a Deployment, ReplicaSet, Replication Controller, or Job
  autoscale      Auto-scale a Deployment, ReplicaSet, or ReplicationController

Cluster Management Commands:
  certificate    Modify certificate resources.
  cluster-info   Display cluster info
  top            Display Resource (CPU/Memory/Storage) usage.
  cordon         Mark node as unschedulable
  uncordon       Mark node as schedulable
  drain          Drain node in preparation for maintenance
  taint          Update the taints on one or more nodes

Troubleshooting and Debugging Commands:
  describe       Show details of a specific resource or group of resources
  logs           Print the logs for a container in a pod
  attach         Attach to a running container
  exec           Execute a command in a container
  port-forward   Forward one or more local ports to a pod
  proxy          Run a proxy to the Kubernetes API server
  cp             Copy files and directories to and from containers.
  auth           Inspect authorization

Advanced Commands:
  diff           Diff live version against would-be applied version
  apply          Apply a configuration to a resource by filename or stdin
  patch          Update field(s) of a resource using strategic merge patch
  replace        Replace a resource by filename or stdin
  wait           Experimental: Wait for a specific condition on one or many resources.
  convert        Convert config files between different API versions

Settings Commands:
  label          Update the labels on a resource
  annotate       Update the annotations on a resource
  completion     Output shell completion code for the specified shell (bash or zsh)

Other Commands:
  api-resources  Print the supported API resources on the server
  api-versions   Print the supported API versions on the server, in the form of "group/version"
  config         Modify kubeconfig files
  plugin         Provides utilities for interacting with plugins.
  version        Print the client and server version information

Usage:
  kubectl [flags] [options]

Use "kubectl <command> --help" for more information about a given command.
Use "kubectl options" for a list of global command-line options (applies to all commands).
```

### LINKED ISSUE #1271: subcommand categories ?
As a tool grow in complexity and number of sub-commands, the help text can become confusing. I find that grouping sub-commands in categories helps a lot because:
- it groups related commands together
- it add additional easily understood semantic

As an example, here is the go-ipfs help text reproduced in full:
```
USAGE
  ipfs  - Global p2p merkle-dag filesystem.

SYNOPSIS
  ipfs [--config=<config> | -c] [--debug | -D] [--help] [-h] [--api=<api>] [--offline] [--cid-base=<base>] [--upgrade-cidv0-in-output] [--encoding=<encoding> | --enc] [--timeout=<timeout>] <command> ...

OPTIONS

  -c, --config               string - Path to the configuration file to use.
  -D, --debug                bool   - Operate in debug mode.
  --help                     bool   - Show the full command help text.
  -h                         bool   - Show a short version of the command help
                                      text.
  -L, --local                bool   - Run the command locally, instead of using
                                      the daemon. DEPRECATED: use --offline.
  --offline                  bool   - Run the command offline.
  --api                      string - Use a specific API instance (defaults to
                                      /ip4/127.0.0.1/tcp/5001).
  --cid-base                 string - Multibase encoding used for version 1
                                      CIDs in output.
  --upgrade-cidv0-in-output  bool   - Upgrade version 0 to version 1 CIDs in
                                      output.
  --enc, --encoding          string - The encoding type the output should be
                                      encoded with (json, xml, or text).
                                      Default: text.
  --stream-channels          bool   - Stream channel output.
  --timeout                  string - Set a global timeout on the command.

SUBCOMMANDS
  BASIC COMMANDS
    init          Initialize ipfs local configuration
    add <path>    Add a file to IPFS
    cat <ref>     Show IPFS object data
    get <ref>     Download IPFS objects
    ls <ref>      List links from an object
    refs <ref>    List hashes of links from an object
  
  DATA STRUCTURE COMMANDS
    block         Interact with raw blocks in the datastore
    object        Interact with raw dag nodes
    files         Interact with objects as if they were a unix filesystem
    dag           Interact with IPLD documents (experimental)
  
  ADVANCED COMMANDS
    daemon        Start a long-running daemon process
    mount         Mount an IPFS read-only mount point
    resolve       Resolve any type of name
    name          Publish and resolve IPNS names
    key           Create and list IPNS name keypairs
    dns           Resolve DNS links
    pin           Pin objects to local storage
    repo          Manipulate the IPFS repository
    stats         Various operational stats
    p2p           Libp2p stream mounting
    filestore     Manage the filestore (experimental)
  
  NETWORK COMMANDS
    id            Show info about IPFS peers
    bootstrap     Add or remove bootstrap peers
    swarm         Manage connections to the p2p network
    dht           Query the DHT for values or peers
    ping          Measure the latency of a connection
    diag          Print diagnostics
  
  TOOL COMMANDS
    config        Manage configuration
    version       Show ipfs version information
    update        Download and apply go-ipfs updates
    commands      List all available commands
    cid           Convert and discover properties of CIDs
    log           Manage and show logs of running daemon
  
  Use 'ipfs <command> --help' to learn more about each command.
  
  ipfs uses a repository in the local file system. By default, the repo is
  located at ~/.ipfs. To change the repo location, set the $IPFS_PATH
  environment variable:
  
    export IPFS_PATH=/path/to/ipfsrepo
  
  EXIT STATUS
  
  The CLI will exit with one of the following values:
  
  0     Successful execution.
  1     Failed executions.

  For more information about each command, use:
  'ipfs <subcmd> --help'
```

Is that something you would consider?

aawsome:
Adds the possibility to group the commands in the help message.

closes #836 
closes #1271

### DIFF
--- a/command.go
+++ b/command.go
@@ -332,7 +332,7 @@ func (c *Command) SetHelpCommandGroupID(groupID string) {
 
 // SetCompletionCommandGroup sets the group id of the completion command.
 func (c *Command) SetCompletionCommandGroupID(groupID string) {
-	// completionCommandGroupID is used if no completion commmand is defined by the user
+	// completionCommandGroupID is used if no completion command is defined by the user
 	c.Root().completionCommandGroupID = groupID
 }
 
@@ -544,13 +544,16 @@ Aliases:
   {{.NameAndAliases}}{{end}}{{if .HasExample}}
 
 Examples:
-{{.Example}}{{end}}{{if .HasAvailableSubCommands}}{{$cmds := .Commands}}
+{{.Example}}{{end}}{{if .HasAvailableSubCommands}}{{$cmds := .Commands}}{{if eq (len .Groups) 0}}
 
-Available Commands:{{range $cmds}}{{if (and (eq .GroupID "") (or .IsAvailableCommand (eq .Name "help")))}}
-  {{rpad .Name .NamePadding }} {{.Short}}{{end}}{{end}}{{range $group := .Groups}}
+Available Commands:{{range $cmds}}{{if (or .IsAvailableCommand (eq .Name "help"))}}
+  {{rpad .Name .NamePadding }} {{.Short}}{{end}}{{end}}{{else}}{{range $group := .Groups}}
 
 {{.Title}}{{range $cmds}}{{if (and (eq .GroupID $group.ID) (or .IsAvailableCommand (eq .Name "help")))}}
-  {{rpad .Name .NamePadding }} {{.Short}}{{end}}{{end}}{{end}}{{end}}{{if .HasAvailableLocalFlags}}
+  {{rpad .Name .NamePadding }} {{.Short}}{{end}}{{end}}{{end}}{{if not .AllChildCommandsHaveGroup}}
+
+Additional Commands:{{range $cmds}}{{if (and (eq .GroupID "") (or .IsAvailableCommand (eq .Name "help")))}}
+  {{rpad .Name .NamePadding }} {{.Short}}{{end}}{{end}}{{end}}{{end}}{{end}}{{if .HasAvailableLocalFlags}}
 
 Flags:
 {{.LocalFlags.FlagUsages | trimTrailingWhitespaces}}{{end}}{{if .HasAvailableInheritedFlags}}
@@ -1217,7 +1220,7 @@ func (c *Command) AddCommand(cmds ...*Command) {
 		cmds[i].parent = c
 		// if Group is not defined let the developer know right away
 		if x.GroupID != "" && !c.ContainsGroup(x.GroupID) {
-			panic(fmt.Sprintf("Group id %s is not defined", x.GroupID))
+			panic(fmt.Sprintf("Group id '%s' is not defined for subcommand '%s'", x.GroupID, cmds[i].CommandPath()))
 		}
 		// update max lengths
 		usageLen := len(x.Use)
@@ -1244,6 +1247,16 @@ func (c *Command) AddCommand(cmds ...*Command) {
 // Groups returns a slice of child command groups.
 func (c *Command) Groups() []*Group {
 	return c.commandgroups
+}
+
+// AllChildCommandsHaveGroup returns if all subcommands are assigned to a group
+func (c *Command) AllChildCommandsHaveGroup() bool {
+	for _, sub := range c.commands {
+		if (sub.IsAvailableCommand() || sub == c.helpCommand) && sub.GroupID == "" {
+			return false
+		}
+	}
+	return true
 }
 
 // ContainGroups return if groupID exists in the list of command groups.
--- a/command.go
+++ b/command.go
@@ -35,6 +35,12 @@ const FlagSetByCobraAnnotation = "cobra_annotation_flag_set_by_cobra"
 // FParseErrWhitelist configures Flag parse errors to be ignored
 type FParseErrWhitelist flag.ParseErrorsWhitelist
 
+// Structure to manage groups for commands
+type Group struct {
+	ID    string
+	Title string
+}
+
 // Command is just that, a command for your application.
 // E.g.  'go run ...' - 'run' is the command. Cobra requires
 // you to define the usage and description as part of your command
@@ -60,6 +66,9 @@ type Command struct {
 
 	// Short is the short description shown in the 'help' output.
 	Short string
+
+	// The group id under which this subcommand is grouped in the 'help' output of its parent.
+	GroupID string
 
 	// Long is the long message shown in the 'help <this-command>' output.
 	Long string
@@ -128,6 +137,9 @@ type Command struct {
 	// PersistentPostRunE: PersistentPostRun but returns an error.
 	PersistentPostRunE func(cmd *Command, args []string) error
 
+	// groups for subcommands
+	commandgroups []*Group
+
 	// args is actual args parsed from flags.
 	args []string
 	// flagErrorBuf contains all error messages from pflag.
@@ -160,6 +172,12 @@ type Command struct {
 	// helpCommand is command with usage 'help'. If it's not defined by user,
 	// cobra uses default help command.
 	helpCommand *Command
+	// helpCommandGroupID is the group id for the helpCommand
+	helpCommandGroupID string
+
+	// completionCommandGroupID is the group id for the completion command
+	completionCommandGroupID string
+
 	// versionTemplate is the version template defined by user.
 	versionTemplate string
 
@@ -301,6 +319,21 @@ func (c *Command) SetHelpFunc(f func(*Command, []string)) {
 // SetHelpCommand sets help command.
 func (c *Command) SetHelpCommand(cmd *Command) {
 	c.helpCommand = cmd
+}
+
+// SetHelpCommandGroup sets the group id of the help command.
+func (c *Command) SetHelpCommandGroupID(groupID string) {
+	if c.helpCommand != nil {
+		c.helpCommand.GroupID = groupID
+	}
+	// helpCommandGroupID is used if no helpCommand is defined by the user
+	c.helpCommandGroupID = groupID
+}
+
+// SetCompletionCommandGroup sets the group id of the completion command.
+func (c *Command) SetCompletionCommandGroupID(groupID string) {
+	// completionCommandGroupID is used if no completion commmand is defined by the user
+	c.Root().completionCommandGroupID = groupID
 }
 
 // SetHelpTemplate sets help template to be used. Application can use it to set custom template.
@@ -511,10 +544,13 @@ Aliases:
   {{.NameAndAliases}}{{end}}{{if .HasExample}}
 
 Examples:
-{{.Example}}{{end}}{{if .HasAvailableSubCommands}}
+{{.Example}}{{end}}{{if .HasAvailableSubCommands}}{{$cmds := .Commands}}
 
-Available Commands:{{range .Commands}}{{if (or .IsAvailableCommand (eq .Name "help"))}}
-  {{rpad .Name .NamePadding }} {{.Short}}{{end}}{{end}}{{end}}{{if .HasAvailableLocalFlags}}
+Available Commands:{{range $cmds}}{{if (and (eq .GroupID "") (or .IsAvailableCommand (eq .Name "help")))}}
+  {{rpad .Name .NamePadding }} {{.Short}}{{end}}{{end}}{{range $group := .Groups}}
+
+{{.Title}}{{range $cmds}}{{if (and (eq .GroupID $group.ID) (or .IsAvailableCommand (eq .Name "help")))}}
+  {{rpad .Name .NamePadding }} {{.Short}}{{end}}{{end}}{{end}}{{end}}{{if .HasAvailableLocalFlags}}
 
 Flags:
 {{.LocalFlags.FlagUsages | trimTrailingWhitespaces}}{{end}}{{if .HasAvailableInheritedFlags}}
@@ -1140,6 +1176,7 @@ Simply type ` + c.Name() + ` help [path to command] for full details.`,
 					CheckErr(cmd.Help())
 				}
 			},
+			GroupID: c.helpCommandGroupID,
 		}
 	}
 	c.RemoveCommand(c.helpCommand)
@@ -1178,6 +1215,10 @@ func (c *Command) AddCommand(cmds ...*Command) {
 			panic("Command can't be a child of itself")
 		}
 		cmds[i].parent = c
+		// if Group is not defined let the developer know right away
+		if x.GroupID != "" && !c.ContainsGroup(x.GroupID) {
+			panic(fmt.Sprintf("Group id %s is not defined", x.GroupID))
+		}
 		// update max lengths
 		usageLen := len(x.Use)
 		if usageLen > c.commandsMaxUseLen {
@@ -1198,6 +1239,26 @@ func (c *Command) AddCommand(cmds ...*Command) {
 		c.commands = append(c.commands, x)
 		c.commandsAreSorted = false
 	}
+}
+
+// Groups returns a slice of child command groups.
+func (c *Command) Groups() []*Group {
+	return c.commandgroups
+}
+
+// ContainGroups return if groupID exists in the list of command groups.
+func (c *Command) ContainsGroup(groupID string) bool {
+	for _, x := range c.commandgroups {
+		if x.ID == groupID {
+			return true
+		}
+	}
+	return false
+}
+
+// AddGroup adds one or more command groups to this parent command.
+func (c *Command) AddGroup(groups ...*Group) {
+	c.commandgroups = append(c.commandgroups, groups...)
 }
 
 // RemoveCommand removes one or more commands from a parent command.
--- a/user_guide.md
+++ b/user_guide.md
@@ -490,6 +490,13 @@ command and flag definitions are needed.
 Help is just a command like any other. There is no special logic or behavior
 around it. In fact, you can provide your own if you want.
 
+### Grouping commands in help
+
+Cobra supports grouping of available commands. Groups must be explicitly defined by `AddGroup` and set by
+the `GroupId` element of a subcommand. The groups will appear in the same order as they are defined.
+If you use the generated `help` or `completion` commands, you can set the group ids by `SetHelpCommandGroupId`
+and `SetCompletionCommandGroupId`, respectively.
+
 ### Defining your own help
 
 You can provide your own Help command or your own template for the default command to use
--- a/README.md
+++ b/README.md
@@ -23,6 +23,7 @@ Cobra provides:
 * Global, local and cascading flags
 * Intelligent suggestions (`app srver`... did you mean `app server`?)
 * Automatic help generation for commands and flags
+* Grouping help for subcommands
 * Automatic help flag recognition of `-h`, `--help`, etc.
 * Automatically generated shell autocomplete for your application (bash, zsh, fish, powershell)
 * Automatically generated man pages for your application
--- a/command_test.go
+++ b/command_test.go
@@ -1767,6 +1767,79 @@ func TestEnableCommandSortingIsDisabled(t *testing.T) {
 	EnableCommandSorting = defaultCommandSorting
 }
 
+func TestUsageWithGroup(t *testing.T) {
+	var rootCmd = &Command{Use: "root", Short: "test", Run: emptyRun}
+	rootCmd.CompletionOptions.DisableDefaultCmd = true
+
+	rootCmd.AddGroup(&Group{ID: "group1", Title: "group1"})
+	rootCmd.AddGroup(&Group{ID: "group2", Title: "group2"})
+
+	rootCmd.AddCommand(&Command{Use: "cmd1", GroupID: "group1", Run: emptyRun})
+	rootCmd.AddCommand(&Command{Use: "cmd2", GroupID: "group2", Run: emptyRun})
+
+	output, err := executeCommand(rootCmd, "--help")
+	if err != nil {
+		t.Errorf("Unexpected error: %v", err)
+	}
+
+	// help should be ungrouped here
+	checkStringContains(t, output, "\nAvailable Commands:\n  help")
+	checkStringContains(t, output, "\ngroup1\n  cmd1")
+	checkStringContains(t, output, "\ngroup2\n  cmd2")
+}
+
+func TestUsageHelpGroup(t *testing.T) {
+	var rootCmd = &Command{Use: "root", Short: "test", Run: emptyRun}
+	rootCmd.CompletionOptions.DisableDefaultCmd = true
+
+	rootCmd.AddGroup(&Group{ID: "group", Title: "group"})
+	rootCmd.AddCommand(&Command{Use: "xxx", GroupID: "group", Run: emptyRun})
+	rootCmd.SetHelpCommandGroupID("group")
+
+	output, err := executeCommand(rootCmd, "--help")
+	if err != nil {
+		t.Errorf("Unexpected error: %v", err)
+	}
+
+	// now help should be grouped under "group"
+	checkStringOmits(t, output, "\nAvailable Commands:\n  help")
+	checkStringContains(t, output, "\nAvailable Commands:\n\ngroup\n  help")
+}
+
+func TestUsageCompletionpGroup(t *testing.T) {
+	var rootCmd = &Command{Use: "root", Short: "test", Run: emptyRun}
+
+	rootCmd.AddGroup(&Group{ID: "group", Title: "group"})
+	rootCmd.AddGroup(&Group{ID: "help", Title: "help"})
+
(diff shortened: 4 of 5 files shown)


jharshman:
@aawsome I think this is a good addition.  It's something I can put on the roadmap after 1.0.0 is released.

mhewedy:
Any updates on this? or any workarounds exist?

aawsome:
I rebased this PR.
Now that 1.0.0 is out, is there anything I can do to get this PR merged?

itskingori:
Maybe we can nudge @jharshman for some visibility. I think should be able to provide some direction.

umarcor:
@aawsome would you mind adding some tests for this feature?

aawsome:
Added some tests. Almost all new code is covered by these tests. 

The exception is `AddCommandGroup` when 'helpCommand' is manually set.
(In this case most people would anyway directly specify the `Group` within the `Command` definition)

aawsome:
Is there anything I can do to help getting this PR merged?

nacx:
This has been open for a while and would be pretty nice to have this merged. Any chance anyone could have a look at it?

itskingori:
@aawsome Branch has conflicts. Could you rebase? So that if it's ready by the time @jharshman has a another look?

aawsome:
> @aawsome Branch has conflicts. Could you rebase? So that if it's ready by the time @jharshman has a another look?

is rebased.

umarcor:
@yuvalsosman, see #1496.

aawsome:
From my side, I'm fine with getting this merged. Actually realized that this PR is now waiting now for more than two years - that's quite some time...

Anyhow, if there is anything I can do, please tell me!

itskingori:
@jpmcb Any chance you could look at this?

RainbowMango:
@aawsome Seems we are ready to go, but you might have to solve the conflicts first.

aawsome:
I rebased and fixed the test that was failing bacause of the automatically-added `completion` command.

This PR should be now ready for merging.

ssoroka:
@marckhouzam @jpmcb Sorry to ping, but can we get this merged now?

umarcor:
@ssoroka see #1626.

aawsome:
Sorry if this sounds rude.

This PR is now open for 2 1/2 years. I always almost immediately answered to any question or request. Which was in last history mostly rebasing because of code changes in master. This PR was on the roadmap for 1.0.0, 1.1.0, 1.2.0, 1.3.0, 1.4.0, 1.5.0 (or maybe not for all, but I'm not connected to release planing). Sorry, but this is really annoying!

I'm not working seriously with go for about one year now. This PR is still mergeable. @maintainers: Please decide if you like to merge this PR or close it if not. When the next request for rebasing comes I might simply close it instead of investing (most likely wasted) work into this PR.

fruitbutt:
@marckhouzam @jharshman adding to the next milestone has occurred numerous times over the 2.5 years this PR has been open. Many of us would like to see this merged so we can stop maintaining forks, and or ugly templating hacks.

jpmcb:
Hi all - I'm sorry (again) this PR has gotten stale and I understand your frustration; I really do. I've been trying now for years to get the different places I've worked at to support me and my work for this extremely important open source project.

But, long story short, it all comes down to time; I just have no time to support this project on my own and the companies I've worked for are not interested in allocating any of my time to cobra. But it often feels like if I don't, no one else will, and [I deeply fear what could happen if this project falls into disarray](https://www.youtube.com/watch?v=YBsDnXXW_d8). If you didn't know already, cobra is a core dependency in Kubernetes, Helm, Tanzu, Cluster API, Linkerd, Grafana, Istio, etc. etc. and sits at a really important intersection of the Go software supply chain. This is generally why my posture towards new changes is a hesitant one.

So, I give corba what little extra time I have in order to ensure we don't have [another faker.js situation on our hands.](https://news.ycombinator.com/item?id=25032105)

Please don't take your frustration out on the maintainers; we are doing our best with what few resources we have but we understand how frustrating this is for the Go community. If we should be frustrated at anyone, it's the endless number of companies that are more than happy to consume these open source libraries while not giving back.

---

I will make this a priority for the next release which I hope to cut within the next few weeks. Thanks all.

umarcor:
Maybe not introducing any relevant change for years in order to satisfy the needs of all those big companies gave the idea that they don't need to care about this project, because maintainers will take care of gatekeeping potential problems (even in development). I don't mean John's employers only, since this was happening far before he joined the project. Should the project evolve as an open source project which is not funded by any of those companies, maybe they would understand the need to do so.

user_guide.md
```diff
@@ -447,6 +447,11 @@ command and flag definitions are needed.
 Help is just a command like any other. There is no special logic or behavior
 around it. In fact, you can provide your own if you want.
 
+### Grouping commands in help
+
+Cobra supports grouping of available commands. Groups can either be explicitly defined by `AddGroup` and set by
+the `Group` element of a subcommand. If Groups are not explicitly defined they are implicitly defined.
```
jpmcb, line 453: > If Groups are not explicitly defined they are implicitly defined.

I think this is a bit confusing. Instead, it seems that the commands that don't have a group get bundled into the "Available commands" which is the current default behavior. When I originally read this, I wasn't sure what "implicitly defined" meant. Can we simplify this chunk?

jpmcb (approved):
Overall, this looks good. Tested this out with a large cobra based utility, added a few groups and it works well.

Just a small documentation nit and this will be good to merge. And if @aawsome doesn't have the bandwidth to address that, we can just get this in and look at it latter

Might also be good to get @marckhouzam's eyes on this since it's a big community ask!

command.go
```diff
@@ -31,6 +31,12 @@ import (
 // FParseErrWhitelist configures Flag parse errors to be ignored
 type FParseErrWhitelist flag.ParseErrorsWhitelist
 
+// Structure to manage groups for commands
+type Group struct {
+	Group string
```
marckhouzam, line 36: How about calling this field `ID` or something like that?  I find that `Group.Group` is confusing.

command.go
```diff
@@ -31,6 +31,12 @@ import (
...
+	Title string
+}
```
marckhouzam, line 38: I assume we create this struct to allow users to create a group once with a title, and then use the `Group.Group` field as a short id to be used multiple times?  Was that the idea behind this struct?

command.go
```diff
@@ -1147,6 +1175,10 @@ func (c *Command) AddCommand(cmds ...*Command) {
 			panic("Command can't be a child of itself")
 		}
 		cmds[i].parent = c
+		// if Group is not defined generate a new one with same title
+		if x.Group != "" && !c.ContainsGroup(x.Group) {
+			c.AddGroup(&Group{Group: x.Group, Title: x.Group})
+		}
```
marckhouzam, line 1181: I wonder what is better for the user?  
Implicitly creating the group seems nice at first impression.  However, I wonder if it makes typos harder to notice?
Say I want a group of commands called "System:" (notice the colon), and for one of my commands I type `Group: "System"` with a missing colon.  This will implicitly create a new group.  It will be obvious looking at the help, but if this is a couple of levels of subcommands down, it may go unnoticed.

What we could do instead is `panic` if the group was not created in advance.  That way, the user will not miss such subtle errors.

The flip side of course is that if forces users to call `AddGroup` for each group.  Is that too heavy?

A benefit is that it forces users to specify a `Title` for the group, instead of sometimes using the id (`Group.Group`) as the title.

And a nice side-effect is that it would make the order of groups more obvious: groups will be displayed in the order they were created with `AddGroup()`

Any opinions on that?

marckhouzam (commented):
This is really great!  Thanks @aawsome for your efforts and sorry it took so long to get reviewed.

I think it is worth putting just a little more effort to make sure the API we provide is good for users.  Once the API is released, we won't be able to change it anymore.

Here are a couple of thoughts: 
1. We will need a way to set the group for the default `completion` command. ~We could add a field to the existing `CompletionOptions` structure to set the Group~ Let's use the same technique as for the `help` command so that it is obvious for users.  So having a function `SetCompletionCommandGroup`.
1. If all commands are assigned to groups, we still see "Available Commands:" at the top but there are no commands under it.  Do we want that, or should we not show it in this case?  In truth, it is not that bad as it looks like it is a heading for all the groups.  However, I wonder if users will like that? (I've seen users refuse to upgrade to a Cobra version because the new `completion` command had its help start with a lower case instead of an uppercase 😄 )
2. If a command is not part of a group it ends up under "Available Commands:", with the Groups listed after.  I personally find "Available Commands:"  a poor title in this case as it will surely cause confusion compared to the group names.  How about we call the section "Additional commands:" and put it after all the groups? 
3. The ordering of the Groups in the help output is not obvious to a user, especially when using implicit creating of groups.  We would at least need to document it.  But let's see if we keep implicit creation or not (see in-line comments).
4. And a couple of questions inline.

marckhouzam (changes requested):
Marking that I request changes because we need at a minimum to allow to put the `completion` command in a group

command.go
jpmcb: I'm also a bit confused on this. Can't we just match on a string instead having to allocate a new "group" structure in memory?

command.go
jpmcb: > What we could do instead is panic if the group was not created in advance. That way, the user will not miss such subtle errors.

I like the idea of not implicitly creating groups now that you bring it up but is panicing the right move here? Do we have a panic president elsewhere in cobra to surface user error?

> And a nice side-effect is that it would make the order of groups more obvious: groups will be displayed in the order they were created with AddGroup()

Yes, I think that's the obvious benefit. The unknown ordering of groups in the help messaging would be a poor user experience for people creating CLIs. My vote would be to be more explicit.

jpmcb:
> If all commands are assigned to groups, we still see "Available Commands:" at the top but there are no commands under it. Do we want that, or should we not show it in this case?

Good point: we should not show "Available Commands" (or maybe something like Marc suggested "Additional commands") if there are no commands under it (aka, all commands have been added to groups. That would be the least confusing, best user experience.

command.go
marckhouzam: > I like the idea of not implicitly creating groups now that you bring it up but is panicing the right move here? 
> Do we have a panic president elsewhere in cobra to surface user error?

Two lines above 😉 
https://github.com/spf13/cobra/blob/212ea4078323771dc49b6f25a41d84efbaac3a4c/command.go#L1177-L1179
I rarely have original thoughts 🤣 

I have to admit that using `panic()` is something I used to think was exaggerated.   However, I now feel that as long as we know that the panic would be triggered immediately, then we know a developer would immediately realize their error and correct it.

user_guide.md
aawsome: I agree we maybe should remove the ability to implicitly define groups. I'll remove this passage here.

command.go
aawsome: I'll change it to `Id`.

command.go
aawsome: Yes, the idea was to use groups by an ID within the `Command` struct and have the title just set once. This also allows to rather easily change the title on just one place.

If you wish, I can remove the ID, but then we still need to save an array of Strings and call `AddGroup` for all groups to set the order of the groups.

command.go
aawsome: I'll change that such that it panics. I agree that implicitly defining groups was maybe a bad idea.

aawsome:
>     1. We will need a way to set the group for the default `completion` command. ~We could add a field to the existing `CompletionOptions` structure to set the Group~ Let's use the same technique as for the `help` command so that it is obvious for users.  So having a function `SetCompletionCommandGroup`.

Will be added.

>     2. If all commands are assigned to groups, we still see "Available Commands:" at the top but there are no commands under it.  Do we want that, or should we not show it in this case?  In truth, it is not that bad as it looks like it is a heading for all the groups.  However, I wonder if users will like that? (I've seen users refuse to upgrade to a Cobra version because the new `completion` command had its help start with a lower case instead of an uppercase smile )

I thought to leave this as a header. It can be of course removed, but that would require a bit of computation to check if all commands have the Group set. Should I do this?

>     3. If a command is not part of a group it ends up under "Available Commands:", with the Groups listed after.  I personally find "Available Commands:"  a poor title in this case as it will surely cause confusion compared to the group names.  How about we call the section "Additional commands:" and put it after all the groups?

Ok, I can also add this, but I think we should keep the exiting behavior if the user defines no group at all. That is, if no group is defined, the "group title" of a command without a `Group` set is "Available Commands" and if at least one group is defined, it would be "Additional Commands".
Also, if users wants to see "Additional Commands", they can also get this behavior by defining a new group titled "Additional Commands" and put these commands into that group.
So, do we really want this?

>     4. The ordering of the Groups in the help output is not obvious to a user, especially when implicitly creating groups.  We would at least need to document it.  But let's see if we keep implicit creation or not (see in-line comments).

After removing the implicit definition, the order will be the order of the `AddGroup` commands. I'll add this in the manual.

command.go
marckhouzam: I think the separation between ID and Title may be good.  Let's first keep it, reviewe the other changes, and then see how we feel about it.

marckhouzam:
> > ```
> > 2. If all commands are assigned to groups, we still see "Available Commands:" at the top but there are no commands under it.  Do we want that, or should we not show it in this case?  In truth, it is not that bad as it looks like it is a heading for all the groups.  However, I wonder if users will like that? (I've seen users refuse to upgrade to a Cobra version because the new `completion` command had its help start with a lower case instead of an uppercase smile )
> > ```
> 
> 
> I thought to leave this as a header. It can be of course removed, but that would require a bit of computation to check if all commands have the Group set. Should I do this?

Leave it for now and we'll see how we feel on the next review.

> 
> > ```
> > 3. If a command is not part of a group it ends up under "Available Commands:", with the Groups listed after.  I personally find "Available Commands:"  a poor title in this case as it will surely cause confusion compared to the group names.  How about we call the section "Additional commands:" and put it after all the groups?
> > ```
> 
> 
> Ok, I can also add this, but I think we should keep the exiting behavior if the user defines no group at all. That is, if no group is defined, the "group title" of a command without a `Group` set is "Available Commands" and if at least one group is defined, it would be "Additional Commands".

Good point. This is actually important to keep the help the same for existing programs 👍

> Also, if users wants to see "Additional Commands", they can also get this behavior by defining a new group titled "Additional Commands" and put these commands into that group.
> So, do we really want this?

The "Additional Commands" automatic group is just to catch commands that were not put into any group.  It should not actually happen but experience shows we better play it safe.  It may happen fire the `help` and `completion` command where devs may forget to give them a group or not realize they can.

If it's complicated to implement, leave it and we can see on the next review.

> 
> 
> > ```
> > 4. The ordering of the Groups in the help output is not obvious to a user, especially when implicitly creating groups.  We would at least need to document it.  But let's see if we keep implicit creation or not (see in-line comments).
> > ```
> 
> 
> After removing the implicit definition, the order will be the order of the `AddGroup` commands. I'll add this in the manual.

Perfect.

Looking forward to the next iteration! 🎉

command.go
aawsome: is changed.

user_guide.md
aawsome: is changed.

aawsome:
I rebased this PR to the current master and made the mentioned changes.
I now get a linting error `Error: struct of size 720 bytes could be of size 712 bytes (maligned)` which I don't know how to solve. And IMO I will add a test of `SetCompletionCommandGroup`.

When these two points (and squashing the commits) are tackled, I think this PR  will be ready.

command.go
```diff
@@ -1200,6 +1241,26 @@ func (c *Command) AddCommand(cmds ...*Command) {
 	}
 }
 
+// Groups returns a slice of child command groups.
+func (c *Command) Groups() []*Group {
+	return c.commandgroups
+}
+
+// ContainGroups return if group is in command groups.
```
marckhouzam, line 1249:
```suggestion
// ContainGroups return if groupId exists in the list of command groups.
```

command.go
```diff
@@ -1200,6 +1241,26 @@ func (c *Command) AddCommand(cmds ...*Command) {
...
+func (c *Command) ContainsGroup(group string) bool {
```
marckhouzam, line 1250:
```suggestion
func (c *Command) ContainsGroup(groupId string) bool {
```

marckhouzam (commented):
Some small suggestions to cleanup a bit.

I haven't test yet.

marckhouzam:
> I rebased this PR to the current master and made the mentioned changes. 
> I now get a linting error `Error: struct of size 720 bytes could be of size 712 bytes (maligned)` which I don't know how to solve.

If you move the new `completionCommandGroup` declaration slightly higher right under the declaration of `helpCommandGroup`, the linter error goes away.  It has to do with memory alignment, although I haven't looked exactly at what was not well aligned.

> And IMO I will add a test of `SetCompletionCommandGroup`.

Yes please.

marckhouzam (changes requested):
Ok, I like this a lot.  Nice job @aawsome.  I put a bunch of minor suggestions about improving comments and such, but you should just be able to accept them in Github which should be very easy.

I believe the only thing remaining is when a command is not part of a group.  I have created the "Administration Commands:" and "System Commands:" groups but "forgot" to assign the `login` command to any group.
Currently we see this:
```
My CLI

Usage:
  prog [command]

Available Commands:
  login             Login to the platform

System Commands:
  config            Configuration for the CLI
  init              Initialize the CLI
  update            Update the CLI
  version           Version information

Administration Commands:
  completion        Generate the autocompletion script for the specified shell
  help              Help about any command

Flags:
  -h, --help   help for prog
```

I think the section "Available Commands:" looks bad.
However, I would personally be ok with leaving it as is and saying that it is up to developers to properly define their groups and not leave commands without a group.

@jpmcb Thoughts?

@aawsome I'm finished with the review.  Whenever you find time, if you can update, and then we'll be pretty much ready to merge 🎉 .

command.go
```diff
@@ -1178,6 +1215,10 @@ func (c *Command) AddCommand(cmds ...*Command) {
 			panic("Command can't be a child of itself")
 		}
 		cmds[i].parent = c
+		// if Group is not defined generate a new one with same title
+		if x.Group != "" && !c.ContainsGroup(x.Group) {
+			panic("Group is not defined")
```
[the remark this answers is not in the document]
aawsome: is added.

command.go
```diff
@@ -61,6 +67,9 @@ type Command struct {
 	// Short is the short description shown in the 'help' output.
 	Short string
 
+	// The group under which the command is grouped in the 'help' output.
+	Group string
```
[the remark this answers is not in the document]
aawsome: The linter says `GroupID`, so I used that. I also added the `ID` at all places where the id is actually used, e.g. `SetHelpCommandGroupID` etc.

aawsome:
Thanks a lot for your review, @marckhouzam 
I resolved all comments, added a test for the `completion` command and rebased this PR.
I hope now everything is fine :wink:

jpmcb (commented):
> I would personally be ok with leaving it as is and saying that it is up to developers to properly define their groups and not leave commands without a group.

I think this is should be the desired behavior since it essentially makes this feature backwards compatible: if someone doesn't define _any_ groups, I'd expect this the help section to continue to work like it does today.

---

I've pondered the design decisions this PR carries and while I think this implementation is good, I wonder if this would be due for a refactor down the road since people may want a way to do more interesting things with groups beyond matching on strings (i.e., a group struct with a superset of cobra commands). But we'd need to see if people would even want that. 

TLDR: I think this is good for now and we can merge it to get this behavior in

marckhouzam:
I tested the latest version and it looked great.
But I kept being bothered by the "Additional commands:" section, mostly when it was empty.
So I went ahead and pushed a new commit that fixed it.  This is the result.

If no groups are defined at all we remain backwards-compatible and get:
```
$ prog -h
My CLI

Usage:
  prog [command]

Available Commands:
  completion        Generate the autocompletion script for the specified shell
  config            Configuration for the CLI
  help              Help about any command
  version           Version information

Flags:
  -h, --help   help for prog

Use "prog [command] --help" for more information about a command.
```
If all subcommands have a group, we no longer get an empty "Available Commands:" line:
```
$ prog -h
My CLI

Usage:
  prog [command]

System
  config            Configuration for the CLI
  version           Version information

Administration
  completion        Generate the autocompletion script for the specified shell
  help              Help about any command

Flags:
  -h, --help   help for prog

Use "prog [command] --help" for more information about a command.
```
And if some commands have groups but some do not we get a new "Additional Commands:" section at the bottom:
```
$ prog -h
My CLI

Usage:
  prog [command]

System
  config            Configuration for the CLI

Administration
  version           Version information

Additional Commands:
  completion        Generate the autocompletion script for the specified shell
  help              Help about any command

Flags:
  -h, --help   help for prog

Use "prog [command] --help" for more information about a command.
```

@aawsome @jpmcb do you agree this is better?  If not, we can just remove the top commit that I pushed.
If you like it, I'm ready to merge.

> I wonder if this would be due for a refactor down the road since people may want a way to do more
> interesting things with groups beyond matching on strings (i.e., a group struct with a superset of cobra commands).
>  But we'd need to see if people would even want that.

That is interesting and may indeed have potential for improvement.  But I also can't tell at the moment what that would look like and if it will be in demand.  So I agree with @jpmcb that we should go ahead with this version, which we know people want, and see where it takes us.

marckhouzam (approved):

aawsome:
> So I went ahead and pushed a new commit that fixed it.

Just had a look at your commit. This is how I would have added these functionalities. So,your changes look good for me!

jpmcb (approved):
Amazing work @marckhouzam!!

Airblader:
I'd just like to say thank you from a backseat watcher and user of the library to @aawsome for bringing this over the finish line even after such a long time! And of course everyone else who helped along the way.

marckhouzam:
Yes thank you @aawsome !  You have shown great professionalism in what can be considered a frustrating situation 🙂

The community will reap the rewards 🎉

Airblader:
If you don't mind, a follow-up question while adopting this: the groups appear in the order they are defined, but what about the commands in the group? Can we order them? It seems that currently this is alphabetical, which isn't usually very meaningful.

PR #922, opened 2019-07-30
Emit non-zero exit code upon non-runnable subcommand

bruceadowns:
This is a refresh of PR #642 and fixes issue #582.

PR #1004, opened 2019-12-22
Revert change so help is printed on stdout again

### LINKED ISSUE #1002: Possible breaking change: help now goes to STDERR
As part of #922 the help message was recently changed from being printed to STDOUT to now being printed to STDERR. 

This sounds like a breaking change.  Some scripts could be written with the expectation that the help message is on stdout, and changing it would break such scripts.  I know the change broke one of Helm's acceptance testing scripts (https://github.com/helm/acceptance-testing/pull/64).  We've discussed this at the Helm developper call and we felt it was worth bringing up to the Cobra project.

Doing a little digging I found that sending the help message to STDOUT was a conscious decision done in #303, 3 years ago.  Changing back to STDERR however seemed more of a "why not" decision, as can be interpreted from the original code change text from https://github.com/spf13/cobra/pull/642#issue-171823645 from which #922 was based, which states:
> Usage output in this scenario is also now sent to stderr instead of stdout, which is better for scripting and seems more consistent with surrounding code.

(Note that although the text says "Usage output" it is actually the Help output that was changed.)

The specific change in question is not part of a COBRA release yet so could easily be reverted to have the help message be sent to STDOUT again and avoid breaking automation scripts.

I can provide a PR if there is agreement to revert.

Thank you.

/cc @bruceadowns @bogem

marckhouzam:
Fixes #1002
For backwards compatibility reasons, and to follow the need of
https://github.com/kubernetes/kubernetes/pull/26077#issuecomment-230818900
the help message should be printed on stdout.

/cc @bruceadowns @bogem @jharshman

### DIFF
--- a/command.go
+++ b/command.go
@@ -372,7 +372,9 @@ func (c *Command) HelpFunc() func(*Command, []string) {
 	}
 	return func(c *Command, a []string) {
 		c.mergePersistentFlags()
-		err := tmpl(c.OutOrStderr(), c.HelpTemplate(), c)
+		// The help should be sent to stdout
+		// See https://github.com/spf13/cobra/issues/1002
+		err := tmpl(c.OutOrStdout(), c.HelpTemplate(), c)
 		if err != nil {
 			c.Println(err)
 		}


jharshman:
@marckhouzam, since @bogem was the one who merged #922, I am going to give him a chance to weigh in here.

eparis:
I completely agree. Help, especially when requested by the user, belong on stdout.

PR #962, opened 2019-09-17
Bump module import from github.com/spf13/pflag@v1.0.3

cfromknecht:
Update to pflag@v1.0.5, running now to execute tests. Will remove replace statement once https://github.com/spf13/pflag/pull/219 or a similar fix lands and has a real tag

PR #1012, opened 2020-01-04
update viper, pflag and yaml

### LINKED ISSUE #1013: Update viper dependency
Viper 1.3.2 is not goroutine safe and causes race conditions for any code using Set/Get from multiple goroutines.  Packages that use both cobra and viper are forced to use the unsafe version of viper.

umarcor:
Close #1013

This PR updates pflag, viper and yaml to the latest stable releases.

Refs: #962, #954, #1091, #1128


umarcor:
Updated and rebased.

umarcor:
yaml was updated separatedly in #1153. This PR now updates viper and pflag only.

PR #1014, opened 2020-01-07
Add Aliases completions on ZSH

chmouel:
This add ZSH completion on command aliases

### DIFF
--- a/zsh_completions.go
+++ b/zsh_completions.go
@@ -23,6 +23,7 @@ var (
 	zshCompFuncMap = template.FuncMap{
 		"genZshFuncName":              zshCompGenFuncName,
 		"extractFlags":                zshCompExtractFlag,
+		"genAliases":                  zshGenAliases,
 		"genFlagEntryForZshArguments": zshCompGenFlagEntryForArguments,
 		"extractArgsCompletions":      zshCompExtractArgumentCompletionHintsForRendering,
 	}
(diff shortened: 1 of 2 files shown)


chmouel:
It's not stalled fyi, just nobody wants to merge this 😭

marckhouzam:
Bash completion does not complete command aliases.
I'm not sure why this is.  Maybe the idea is that we don't want to guide the users to use aliases, but only to use the main command?

Also, be aware that the current zsh script may get replaced by #1070

marckhouzam:
The zsh completion script has been completely change by #1070 so this PR no longer applies.

@chmouel Note that Cobra's shell completion, including the new zsh logic, does not complete aliases.  This seems to be done by design. If you feel aliases should be completed, I recommended opening new PR and arguing the value for it.

@jpmcb I believe we can close this one now that #1070 has been merged.

jpmcb:
Thanks for letting me know @marckhouzam!! Glad we are getting some things cleaned up

@chmouel thank you so much for the contributions and PRs, apologies for not being able to get them in before the zsh revamp.

chmouel:
@marckhouzam why would you assume this is "by design" ?

marckhouzam:
@chmouel You are right, I don't actually have a confirmation it was by design.

Personally I feel command aliases should be completed somehow, either fully like you suggest or only when there are no other completions, as is done for argument aliases. Not sure which is best.

I encourage you to post a PR (since this one was no longer applicable).

Issue #1019, opened 2020-01-16
`LocalFlags().Visit()` doesn't visit set flags when in child command's `Run()`


bcongdon:
If you call `cmd.LocalFlags().Visit()` in a child command's `Run` function, it always visits an empty list of flags.

Consider the following test:

```go
func TestLocalFlagsInChildRun(t *testing.T) {
	root := &Command{
		Use: "root",
		Run: emptyRun,
	}

	var setFlags []string
	c := &Command{
		Use: "child",
		Run: func(cmd *Command, args []string) {
			cmd.LocalFlags().Visit(func(f *pflag.Flag) {
				setFlags = append(setFlags, f.Name)
			})
		},
	}
	c.Flags().BoolP("boolb", "b", false, "a boolean flag")
	root.AddCommand(c)

	_, err := executeCommand(root, "child", "--boolb")
	if err != nil {
		t.Fatal("unexpected error: ", err.Error())
	}

	if len(setFlags) != 1 || setFlags[0] != "boolb" {
		t.Errorf(`expected setFlags to be ["boolb"], but was: %v`, setFlags)
	}
}
```

Which fails on master with the following error message:

```
--- FAIL: TestLocalFlagsInChildRun (0.00s)
    command_test.go:1801: expected setFlags to be ["boolb"], but was: []
```

---

It looks like this is because when a flag is added to the `c.lflags` set in `LocalFlags()`, `pflag.FlagSet.AddFlag()` does not copy the fact that the flag may have been already parsed.

WuShaoQiang:
Why not use `Flags`?

bcongdon:
If I understand correctly, `Flags` also includes flags from the parent command, which I don't want.

(i.e. if `root` has a `boola` flag, I wouldn't want to see that in the child `LocalFlags()`)

WuShaoQiang:
```
cmd.LocalFlags().VisitAll(func(f *pflag.Flag) {
			if f.Changed {
				result = append(result, f.Name)
			}
		})
```

Try this.

bcongdon:
Thanks! That seems to solve the immediate issue. I'd still argue that this is a bug, though

PR #1020, opened 2020-01-16
Copy parsed flag values when constructing `LocalFlags()`

bcongdon:
This checks in the failing test that #1019 references, and proposes a fix to `LocalFlags()` not copying information about whether a flag has been parsed or not.

If preferable, I can open a PR against spf13/pflag to change `flagSet.AddFlag` to copy the fact that a flag already has a value instead of performing that logic here.

### DIFF
--- a/command.go
+++ b/command.go
@@ -1418,6 +1418,13 @@ func (c *Command) LocalFlags() *flag.FlagSet {
 	addToLocal := func(f *flag.Flag) {
 		if c.lflags.Lookup(f.Name) == nil && c.parentsPflags.Lookup(f.Name) == nil {
 			c.lflags.AddFlag(f)
+
+			// If `f` has already been changed, re-add its value to the flag set,
+			// otherwise `c.lflags` has no record of the flags value.
+			if f.Changed {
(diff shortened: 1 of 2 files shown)


bcongdon:
Pinging this, CLA is signed

command.go
```diff
@@ -1418,6 +1418,13 @@ func (c *Command) LocalFlags() *flag.FlagSet {
 	addToLocal := func(f *flag.Flag) {
 		if c.lflags.Lookup(f.Name) == nil && c.parentsPflags.Lookup(f.Name) == nil {
 			c.lflags.AddFlag(f)
+
+			// If `f` has already been changed, re-add its value to the flag set,
+			// otherwise `c.lflags` has no record of the flags value.
+			if f.Changed {
+				f.Changed = false
```
johnSchnake, line 1425: Why is it that you need to unset the `Changed` field? I thought the point was to just copy over the new value, I didn't expect to see this modified as well.

johnSchnake:
I'm so sorry its taken so long to respond to this PR. We are actively trying to tackle the backlog.

PR #1035, opened 2020-02-16
Custom completions coded in Go (instead of Bash)

marckhouzam:
This PR allows programs using Cobra to code their custom completions in **Go** directly inside the program itself.

It is a port to Cobra of the custom completion solution introduced in Helm 3.1 (https://github.com/helm/helm/pull/7323).

It is fully-backwards compatible.
It also allows to use both the new Go custom completions and the existing Bash custom completions at the same time.

The PR **introduces a new hidden sub-command: `__complete`** that is to be called by the auto-completion script. I chose the name with a double underscore prefix to make sure this command will not collide with any command of a program.

Along the lines of the existing `ValidArgs` field, support is added for the **`ValidArgsFunction`** field.  The program can register a function that will provide custom completion through this newly added field. 

For flags, the new function **`Command.RegisterFlagCompletionFunc()`** is added.  The program can register a custom completion function for a flag using this new function.

The bash completion script, before calling the completion `*_custom_func`, will now verify if `Command.ValidArgsFunction` has been specified.  If so, the completion script will call the `__complete` command passing it all command-line arguments; the `__complete` command will call the function specified by `Command.ValidArgsFunction` to obtain completions from the program itself.

To be as feature-rich as custom completions written in Bash, the PR also allows Go completions to control:
- if a single completion should be followed by a space
- if file completion should be performed when no completions are available
 
Some advantages of using Go instead of Bash for custom completions:
- Cobra users are Go developers
- Access to the full program code base to perform completions instead of being limited to relying on externally-exposed functionality.
- Debugging and development much easier because developers can directly call e.g., `<program> __complete <cmd> ""<ENTER>` to test the corresponding custom completion code
- The Go debugger can even be used to troubleshoot custom completion code
- Go tests can be written for the custom completion code
- Flag parsing done automatically by Cobra (instead of having to be coded in bash in the completion script)
- No need to make sure the custom completion code works for both `bash` and `zsh` (for those programs that share completions between the two)
- No more issues with the use of shell tools (e.g., `echo`, `cat`, `tail`) that work differently on different OSes

To illustrate a working program using this PR, I've ported Helm from its own solution to using this PR in Cobra: https://github.com/VilledeMontreal/helm/tree/feat/compInGoWithCobra
Here is an example of using the new `ValidArgsFunction` field:
https://github.com/VilledeMontreal/helm/blob/2c88891a96eb6cb27d124d5a6d750eee85abcc8c/cmd/helm/status.go#L55
and an example of using `RegisterFlagCompletionFunc()`:
https://github.com/VilledeMontreal/helm/blob/2c88891a96eb6cb27d124d5a6d750eee85abcc8c/cmd/helm/root.go#L86

Go tests have been added in Cobra for this PR and the documentation for bash_completion has been updated.

### DIFF
--- a/bash_completions.go
+++ b/bash_completions.go
@@ -58,6 +58,67 @@ __%[1]s_contains_word()
     return 1
 }
 
+__%[1]s_handle_go_custom_completion()
+{
+    __%[1]s_debug "${FUNCNAME[0]}: cur is ${cur}, words[*] is ${words[*]}, #words[@] is ${#words[@]}"
+
+    local out requestComp lastParam lastChar comp directive args
+
+    # Prepare the command to request completions for the program.
+    # Calling ${words[0]} instead of directly %[1]s allows to handle aliases
+    args=("${words[@]:1}")
+    requestComp="${words[0]} %[2]s ${args[*]}"
+
+    lastParam=${words[$((${#words[@]}-1))]}
+    lastChar=${lastParam:$((${#lastParam}-1)):1}
+    __%[1]s_debug "${FUNCNAME[0]}: lastParam ${lastParam}, lastChar ${lastChar}"
+
+    if [ -z "${cur}" ] && [ "${lastChar}" != "=" ]; then
+        # If the last parameter is complete (there is a space following it)
+        # We add an extra empty parameter so we can indicate this to the go method.
+        __%[1]s_debug "${FUNCNAME[0]}: Adding extra empty parameter"
+        requestComp="${requestComp} \"\""
+    fi
+
+    __%[1]s_debug "${FUNCNAME[0]}: calling ${requestComp}"
+    # Use eval to handle any environment variables and such
+    out=$(eval "${requestComp}" 2>/dev/null)
+
+    # Extract the directive integer at the very end of the output following a colon (:)
+    directive=${out##*:}
+    # Remove the directive
+    out=${out%%:*}
+    if [ "${directive}" = "${out}" ]; then
+        # There is not directive specified
+        directive=0
+    fi
+    __%[1]s_debug "${FUNCNAME[0]}: the completion directive is: ${directive}"
+    __%[1]s_debug "${FUNCNAME[0]}: the completions are: ${out[*]}"
+
+    if [ $((directive & %[3]d)) -ne 0 ]; then
+        # Error code.  No completion.
+        __%[1]s_debug "${FUNCNAME[0]}: received error from custom completion go code"
+        return
+    else
+        if [ $((directive & %[4]d)) -ne 0 ]; then
+            if [[ $(type -t compopt) = "builtin" ]]; then
+                __%[1]s_debug "${FUNCNAME[0]}: activating no space"
+                compopt -o nospace
+            fi
+        fi
+        if [ $((directive & %[5]d)) -ne 0 ]; then
+            if [[ $(type -t compopt) = "builtin" ]]; then
+                __%[1]s_debug "${FUNCNAME[0]}: activating no file completion"
+                compopt +o default
+            fi
+        fi
+
+        while IFS='' read -r comp; do
+            COMPREPLY+=("$comp")
+        done < <(compgen -W "${out[*]}" -- "$cur")
+    fi
+}
+
 __%[1]s_handle_reply()
 {
     __%[1]s_debug "${FUNCNAME[0]}"
@@ -121,6 +182,10 @@ __%[1]s_handle_reply()
     completions=("${commands[@]}")
     if [[ ${#must_have_one_noun[@]} -ne 0 ]]; then
         completions=("${must_have_one_noun[@]}")
+    elif [[ -n "${has_completion_function}" ]]; then
+        # if a go completion function is provided, defer to that function
+        completions=()
+        __%[1]s_handle_go_custom_completion
     fi
     if [[ ${#must_have_one_flag[@]} -ne 0 ]]; then
         completions+=("${must_have_one_flag[@]}")
@@ -279,7 +344,7 @@ __%[1]s_handle_word()
     __%[1]s_handle_word
 }
 
-`, name))
+`, name, CompRequestCmd, BashCompDirectiveError, BashCompDirectiveNoSpace, BashCompDirectiveNoFileComp))
 }
 
 func writePostscript(buf *bytes.Buffer, name string) {
@@ -304,6 +369,7 @@ func writePostscript(buf *bytes.Buffer, name string) {
     local commands=("%[1]s")
     local must_have_one_flag=()
     local must_have_one_noun=()
+    local has_completion_function
     local last_command
     local nouns=()
 
@@ -404,7 +470,22 @@ func writeLocalNonPersistentFlag(buf *bytes.Buffer, flag *pflag.Flag) {
 	buf.WriteString(fmt.Sprintf(format, name))
 }
 
+// Setup annotations for go completions for registered flags
+func prepareCustomAnnotationsForFlags(cmd *Command) {
+	for flag := range flagCompletionFunctions {
+		// Make sure the completion script calls the __*_go_custom_completion function for
+		// every registered flag.  We need to do this here (and not when the flag was registered
+		// for completion) so that we can know the root command name for the prefix
+		// of __<prefix>_go_custom_completion
+		if flag.Annotations == nil {
+			flag.Annotations = map[string][]string{}
+		}
+		flag.Annotations[BashCompCustom] = []string{fmt.Sprintf("__%[1]s_handle_go_custom_completion", cmd.Root().Name())}
+	}
+}
+
 func writeFlags(buf *bytes.Buffer, cmd *Command) {
+	prepareCustomAnnotationsForFlags(cmd)
 	buf.WriteString(`    flags=()
     two_word_flags=()
     local_nonpersistent_flags=()
@@ -468,6 +549,9 @@ func writeRequiredNouns(buf *bytes.Buffer, cmd *Command) {
 	sort.Sort(sort.StringSlice(cmd.ValidArgs))
 	for _, value := range cmd.ValidArgs {
 		buf.WriteString(fmt.Sprintf("    must_have_one_noun+=(%q)\n", value))
+	}
+	if cmd.ValidArgsFunction != nil {
+		buf.WriteString("    has_completion_function=1\n")
 	}
 }
 
--- a/custom_completions.go
+++ b/custom_completions.go
@@ -1,0 +1,299 @@
+package cobra
+
+import (
+	"errors"
+	"fmt"
+	"os"
+	"strings"
+
+	"github.com/spf13/pflag"
+)
+
+// CompRequestCmd is the name of the hidden command that is used to request
+// completion results from the program.  It is used by the shell completion script.
+const CompRequestCmd = "__complete"
+
+// Global map of flag completion functions.
+var flagCompletionFunctions = map[*pflag.Flag]func(cmd *Command, args []string, toComplete string) ([]string, BashCompDirective){}
+
+// BashCompDirective is a bit map representing the different behaviors the shell
+// can be instructed to have once completions have been provided.
+type BashCompDirective int
+
+const (
+	// BashCompDirectiveError indicates an error occurred and completions should be ignored.
+	BashCompDirectiveError BashCompDirective = 1 << iota
+
+	// BashCompDirectiveNoSpace indicates that the shell should not add a space
+	// after the completion even if there is a single completion provided.
+	BashCompDirectiveNoSpace
+
+	// BashCompDirectiveNoFileComp indicates that the shell should not provide
+	// file completion even when no completion is provided.
+	// This currently does not work for zsh or bash < 4
+	BashCompDirectiveNoFileComp
+
+	// BashCompDirectiveDefault indicates to let the shell perform its default
+	// behavior after completions have been provided.
+	BashCompDirectiveDefault BashCompDirective = 0
+)
+
+// RegisterFlagCompletionFunc should be called to register a function to provide completion for a flag.
+func (c *Command) RegisterFlagCompletionFunc(flagName string, f func(cmd *Command, args []string, toComplete string) ([]string, BashCompDirective)) error {
+	flag := c.Flag(flagName)
+	if flag == nil {
+		return fmt.Errorf("RegisterFlagCompletionFunc: flag '%s' does not exist", flagName)
+	}
+	if _, exists := flagCompletionFunctions[flag]; exists {
+		return fmt.Errorf("RegisterFlagCompletionFunc: flag '%s' already registered", flagName)
+	}
+	flagCompletionFunctions[flag] = f
+	return nil
+}
+
+// Returns a string listing the different directive enabled in the specified parameter
+func (d BashCompDirective) string() string {
+	var directives []string
+	if d&BashCompDirectiveError != 0 {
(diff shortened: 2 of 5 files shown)


AdamSLevy:
You may be interested in this golang package as it is related to your intent here: https://github.com/posener/complete

I use it alongside cobra to achieve the same end as this PR.

jharshman:
@marckhouzam this is great work! 

Can this be extended to generate completions for other shells. We currently support ZSH but have had requests for others like FiSH etc.. 

There's a lot to take in on this PR so please keep in mind that the review process for this one might be more lengthy. That being said I have slated this for inclusion in our 1.0.0 release.

marckhouzam:
> @marckhouzam this is great work!

Thanks for having a look! 

> Can this be extended to generate completions for other shells. We currently support ZSH but have had requests for others like FiSH etc..

I didn't want to come on too strong, so I started with posting the functionality of this PR.  However you are absolutely right! This can be extended to provide full completions to other shells. In fact, I'm about to post a PR to the helm project that will support full Fish completion (command names, flag names *and* custom completions). And the fish script is less than 100 lines.

> There's a lot to take in on this PR so please keep in mind that the review process for this one might be more lengthy. 

No problem at all. I'll do my best to be responsive to any comment you may have.

> That being said I have slated this for inclusion in our 1.0.0 release.

🎉

marckhouzam:
Here's the Helm PR for full Fish completion: 
https://github.com/helm/helm/pull/7690

I'm going to work with the other helm maintainers to finalize it and then I can try to port it to Cobra.

bash_completions.go
```diff
@@ -58,6 +58,67 @@ __%[1]s_contains_word()
...
+    __%[1]s_debug "${FUNCNAME[0]}: the completions are: ${out[*]}"
+
+    if [ $((directive & %[3]d)) -ne 0 ]; then
+        # Error code.  No completion.
+        __%[1]s_debug "${FUNCNAME[0]}: received error from custom completion go code"
+        return
+    else
+        if [ $((directive & %[4]d)) -ne 0 ]; then
+            if [[ $(type -t compopt) = "builtin" ]]; then
+                __helm_debug "${FUNCNAME[0]}: activating no space"
```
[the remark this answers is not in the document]
marckhouzam: Good find!  There are two instances where I forgot to change this when I ported from Helm.
I will push an update.

custom_completions.go
```diff
@@ -0,0 +1,299 @@
...
+	// BashCompDirectiveDefault indicates to let the shell perform its default
+	// behavior after completions have been provided.
+	BashCompDirectiveDefault BashCompDirective = 0
+)
+
+// RegisterFlagCompletionFunc should be called to register a function to provide completion for a flag.
+func (c *Command) RegisterFlagCompletionFunc(flagName string, f func(cmd *Command, args []string, toComplete string) ([]string, BashCompDirective)) {
+	flag := c.Flag(flagName)
+	if flag == nil {
+		log.Fatal(fmt.Sprintf("RegisterFlagCompletionFunc: flag '%s' does not exist", flagName))
```
marckhouzam: Is it ok to abort the program when this API is not used correctly?  Or should this method return an `error` to give more flexibility to the program?

bash_completions.go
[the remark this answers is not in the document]
marckhouzam: Fixed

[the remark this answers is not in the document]
marckhouzam: I don't understand why the tests are failing now...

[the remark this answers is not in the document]
jharshman: Rebase against master and the Circle CI tests should go away.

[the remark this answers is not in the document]
jharshman: Ah - actually I need to talk to @spf13 about removing the Circle CI integration.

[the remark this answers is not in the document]
jharshman: @marckhouzam - ok ignore the circleCI check. The integration has been disabled for future PRs.
It's just failing because there is no CircleCI configuration for this project anymore.

custom_completions.go
jharshman: I think we should return an error from these methods instead of aborting the program.

bash_completions.go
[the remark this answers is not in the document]
marckhouzam: OK thanks.

custom_completions.go
marckhouzam: Done

jharshman:
@marckhouzam I'm trying to validate this on my end and can't seem to get the statically defined autocompletion functioning.

i.e:
```
var rootCmd = &cobra.Command{
	ValidArgs: []string{"pod", "rs", "ds"},
	Use:       "testprog1",
		fmt.Println("rootCmd called")
	},
}

	Use: "pod",
		fmt.Println("child called")
	},
}
```

However, when i compile it and run `./testprog1 [tab][tab]` I don't get any completion suggestions where I assume I should get "pod", "rs", or "ds".
Am I mistaken or missing something key here?

marckhouzam:
> However, when i compile it and run `./testprog1 [tab][tab]` I don't get any completion suggestions where I assume I should get "pod", "rs", or "ds".
> Am I mistaken or missing something key here?

Thanks @jharshman for trying things out!

There are a couple things going on here:
1. You will always need to source a completion script first, before completions can work
1. This PR only introduces Go Custom Completions (the ones code by the program using cobra).  It does not modify the way Cobra provides command, flag, or ValidArgs completions. So to give this code a spin, you need try the new `ValidArgsFunction` field.

Here is a test program and how to test with it.  That's how I ran it anyway.

main.go:
```
package main

import (
	"fmt"
	"os"

	"github.com/spf13/cobra"
)

var rootCmd = &cobra.Command{
	ValidArgs: []string{"pod", "rs", "ds"},
	Use:       "testprog1",
		fmt.Println("rootCmd called")
	},
}

	Use: "pod",
	ValidArgsFunction: func(cmd *cobra.Command, args []string, toComplete string) ([]string, cobra.BashCompDirective) {
		return []string{"these", "are", "custom", "completions"}, cobra.BashCompDirectiveDefault
	},
		fmt.Println("child called")
	},
}

	Use: "completion",
	},
}

func main() {
	rootCmd.AddCommand(childCmd)
	rootCmd.Execute()
}
```
go.mod (you'll need to modify the path to your cobra git repo)
```
go 1.13
replace github.com/spf13/cobra => /Users/marckhouzam/git/cobra
```

To build and test:
```
bash

GO111MODULE=on go build -o testprog1 main.go
source <(./testprog1 completion)

./testprog1 <TAB><TAB>
# Should show: ds   pod  rs
# The completions above are done by cobra's existing script

./testprog1 pod <TAB><TAB>
# Should show: are          completions  custom       these
# The completions above are done by this PR through the ValidArgsFunction field called by the cobra completion script
```

Let me know if I can help in any other way, and thanks again.

jharshman:
Ah - thank you for clarifying that. Dumb mistake on my part.  I'll continue my review of this change.

marckhouzam:
BTW, allowing the `__complete` command to also complete flags, commands, and ValidArgs is added as part of #1048; specifically in commit https://github.com/spf13/cobra/commit/828c60bddd1109bf4a2ad5c578769881c6c00c61
However, the bash completion script of Cobra won't make use of that (not needed), but the Fish one will.

But you could test it manually (using #1048) like this:
```
./testprog1 __complete ""
# will show sub commands and validArgs
```

jharshman:
@marckhouzam so for the `ValidArgsFunction`, it looks like static `ValidArgs` takes precedence over `ValidArgsFunction` if both are defined on a command. Is this right?

custom_completions_test.go
```diff
@@ -0,0 +1,349 @@
...
+
+	// Test completing an empty string
+	output, err := executeCommand(rootCmd, CompRequestCmd, "")
+	if err != nil {
+		t.Errorf("Unexpected error: %v", err)
+	}
+
+	expected := `one
+two
+:0
```
jharshman: Would there be a way to format these expected outputs a bit better?

marckhouzam:
> @marckhouzam so for the `ValidArgsFunction`, it looks like static `ValidArgs` takes precedence over `ValidArgsFunction` if both are defined on a command. Is this right?

@jharshman Good observation.  With this PR, we are only focusing on Custom Completions so I chose to follow the way the existing Bash completion script triggers custom completions.  That meant that if there are `ValidArgs` then `ValidArgsFunction` does not get called.  To be honest, this was the simplest choice, which avoided messing too much with the Bash script code.  But from a user's perspective, this may not be the right choice.

In fact, in #1048 for Fish completion, where `ValidArgs` are handled in Go through the `__completion` command, I actually chose to allow both `ValidArgs` and `ValidArgsFunction` to be used together.  You made me to realize now this would cause incompatibilities between bash completions and fish completions.

So, what do we want to offer the user?  For consistency, I think we have two options:
1. allow to use both `ValidArgs` and `ValidArgsFunction` together
1. disallow using both together and cause some configuration error if they are used together.

Allowing both gives more flexibility to the user, but could cause confusion if they don't realize they have used both... What do you think we should support?

jharshman:
@marckhouzam my opinion on this is that a command can either use one or the other but not both. While being able to set both on a command might add more flexibility, I think it would be at the expense of simplicity.

So let's make these fields mutually exclusive and going with your second option. Also please add some documentation for the function and in the readme.

marckhouzam:
> @marckhouzam my opinion on this is that a command can either use one or the other but not both. 

@jharshman Agreed. I will post an update and make matching changes to #1048.

This brings up a similar situation.  Cobra allows a command to have sub-commands ***and*** arguments at the same time (your test program even does it, although it would need the `Args:` field on `rootCmd` to actually accept the specified `ValidArgs`).  So we can have a situation where `ValidArgs` is specified as well as sub-commands. When doing completion, it seems that currently `ValidArgs` has precedence over sub-commands.  To explain:

In `testprog1` the `rootCmd` specifies `ValidArgs` of `pod, rs, ds`, but also a sub-command `pod`.  In testing I saw that the completion script only proposes the values of `ValidArgs`.  If `ValidArgs` did not include `pod`, the completions would not propose the sub-command `pod` at all.   Cobra should probably do better to help the user by automatically including non-hidden sub-commands in the completions.  But that is something else (does this deserve an issue?). For this PR:

What shall we do if `ValidArgsFunction` is specified at the same time as sub-commands?  I assume we should propose both the sub-commands and the results of calling `ValidArgsFunction`?  What do you say?

custom_completions_test.go
marckhouzam: When improving the format, I hesitated between:
`expected := fmt.Sprintf("one\ntwo\n:0\n")`
and 
`expected := strings.Join([]string{"one", "two", ":0", ""}, "\n")`
I chose the latter because I didn't like the repeating `\n`

bash_completions.go
```diff
@@ -469,6 +551,9 @@ func writeRequiredNouns(buf *bytes.Buffer, cmd *Command) {
 	for _, value := range cmd.ValidArgs {
 		buf.WriteString(fmt.Sprintf("    must_have_one_noun+=(%q)\n", value))
 	}
+	if cmd.ValidArgsFunction != nil {
```
marckhouzam: > disallow using both together and cause some configuration error if they are used together.

I couldn't find a good way to send an error when both `ValidArgs` and `ValidArgsFunction` are present.  The closest I could find is here, in `writeRequiredNouns()` , which runs when generating the completion script.  Although there is no error being returned, so all we could do is `log.Fatal()`. 

Any suggestions @jharshman ?

marckhouzam:
> So let's make these fields mutually exclusive and going with your second option please also add some documentation for the function and in the readme.

I pushed a commit with a comment added next to the `ValidArgsFunction` field, and mentioned it in the completion readme.

The code behaves as we want, only using `ValidArgs` if both are present, but I couldn't find a good way to warn the user at runtime.

bash_completions.go
jharshman: @marckhouzam I'm kinda leaning towards just documenting the behavior rather than bailing out. `ValidArgs` already takes precedence so if it's sufficiently documented then I think it's good enough.

jharshman:
> What shall we do if ValidArgsFunction is specified at the same time as sub-commands? I assume we should propose both the sub-commands and the results of calling ValidArgsFunction? What do you say?

Hm... this is a good point.  The intention of `ValidArgsFunction` is to allow dynamic population of completions for arguments. I would say here we want to match the behavior of `ValidArgs`.

bash_completions.go
marckhouzam: ok with me.

marckhouzam:
> The intention of `ValidArgsFunction` is to allow dynamic population of completions for arguments. I would say here we want to match the behavior of `ValidArgs`.

I have pushed a commit that makes `ValidArgsFunction` be treated like `ValidArgs` when there are also sub-commands, which is to show completions matching `ValidArgsFunction` but not sub-commands.  I have however also opened #1076 to improve this behaviour.

@jharshman I believe I have addressed all comments for the moment.

marckhouzam:
The force-push was rebase to master, as there were conflicts.

marckhouzam:
The tests fail because of something to do with stderr output.  Maybe because of the rebase to master.  I will investigate.

jharshman:
> The tests fail because of something to do with stderr output

@marckhouzam yes I merged in a small change to start using StdErr and StdOut functions in the test code. 
Ref: #1053

marckhouzam:
> @marckhouzam yes I merged in a small change to start using StdErr and StdOut functions in the test code.

@jharshman Thanks for the pointer.  It is now fixed.  I made a mistake rebasing.

jharshman:
@marckhouzam awesome - gonna give this one more pass and then I think we can call it good and start looking at #1048

jharshman:
Okay one observation that I have made is that the `ValidArgs` behavior actually _does_ return both the contents of `ValidArgs` and the subcommand. So If it makes sense, I think the `ValidArgFunction` should match this behavior.
What do you think @marckhouzam.  I think we were mistaken in our interpretation of the behavior of `ValidArgs`.  Can you also validate the behavior I'm seeing?

marckhouzam:
> Okay one observation that I have made is that the `ValidArgs` behavior actually _does_ return both the contents of `ValidArgs` and the subcommand. 

When I tested it, `ValidArgs` prevented subcommands from being completed.  In the code it looks like that too. If you look at https://github.com/spf13/cobra/blob/138b98f39dedc35ae01b827fb9bceacaf30b0880/bash_completions.go#L121-L124

Notice that if there are `ValidArgs` (`must_have_one_noun`), then the `completions` array is overwritten completely, which removes the subcommands.  It is just a question of changing the `=` to `+=` to fix that (which makes me wonder if this was an oversight from the very start), but I think it should be done in a separate PR, for traceability, which is why I opened #1076. 

How did you test on your side?  Is it with the `testprog1` we mentionned above?  If so, could it be that you see the subcommand name (`pod`) being completed because it is also listed in the list of `ValidArgs`?  

> @marckhouzam awesome - gonna give this one more pass and then I think we can call it good and start looking at #1048

Awesome!  I'll have to rebase #1048 on the new master once this PR is merged as #1048 is not currently based on the very latest version of this PR.

jharshman:
> could it be that you see the subcommand name (pod) being completed because it is also listed in the list of ValidArgs?

Yep, that was the reason, thanks for double checking!

Alright - I think this PR is ready to rock then!

jharshman:
@marckhouzam actually can you squash your commits first. This has quite a large commit message.

marckhouzam:
> @marckhouzam actually can you squash your commits first. This has quite a large commit message.

@jharshman All cleaned up.  Thanks again!

PR #1040, opened 2020-02-20
chore(ci): add release note generation

fallion:
Allows for better changelogs for apps such as dependabot and renovate.

Please compare the release notes sections here: https://github.com/outillage/release-notary/pull/180 and here: https://github.com/outillage/release-notary/pull/120

This PR will change to look like the latter. Non-conventional commit messages will go into `other` section meaning this does not break any flow. The advantage however is that people using Renovate will see the changes directly in the `Release Notes` section and don't have to manually check.

References #259

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -1,0 +1,18 @@
+name: Release
+on:
+  push:
+    tags:
+      - "v*"
+
+jobs:
+  release-notary:
+    name: Release
+    runs-on: ubuntu-latest
+    steps:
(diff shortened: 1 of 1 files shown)


umarcor:
Ref #968

jpmcb:
For now, we will continue with doing release notes by hand. We've removed the changelog #1634 since we hardly used it.

Thank you so much for the contribution!

fallion:
Makes sense. Thanks for the update! :)

PR #1044, opened 2020-02-26
Use golangci-lint

### LINKED ISSUE #952: GitHub Actions
It seems that GitHub Actions (beta) is enabled for @spf13's repos, so it works here: https://github.com/spf13/cobra/actions

I think we should migrate from Travis CI (and maybe circleci) to GitHub Actions, and add macOS and Windows jobs too.

Ref #907

umarcor:
Close #952
Close #968

golangci-lint is added to Travis CI. Then, multiple linting issues are fixed: errors are handled, unused variables/functions are removed, unnecessary types are removed, frequently used strings are defined as constants, etc.

Ref #876


.gitignore
```diff
@@ -37,3 +37,6 @@ bin
 
 .idea/
 *.iml
+
+# test-generated files
+cobra/cmd/testproject
```
jharshman: Unnecessary - the `testproject` dir should delete itself.

.travis.yml
```diff
@@ -18,12 +20,13 @@ matrix:
   allow_failures:
     - go: tip
   include:
-    - stage: diff
+    - stage: lint
       go: 1.13.x
-      script: make fmt
+      before_install: curl -sfL https://install.goreleaser.com/github.com/golangci/golangci-lint.sh | sh -s -- -b $(go env GOPATH)/bin latest
+      script: golangci-lint run -v
```
jharshman: Should be triggered via makefile target.

README.md
```diff
@@ -267,7 +266,7 @@ func initConfig() {
 		viper.SetConfigFile(cfgFile)
 	} else {
 		// Find home directory.
-		home, err := homedir.Dir()
+		home, err := os.UserHomeDir()
```
jharshman: This would force users to upgrade Go in order to use this package making this a potentially breaking change.

umarcor: This is supported in `>=1.12`. Versions `<=1.11` are deprecated. I think it is good to enforce users not to use non-maintained software.

jharshman: Forcing a breaking change is not something we are going to include in a 1.0.0 release.

umarcor: It's ok. At this point, I'm just waiting for any v1.0.0 to be released (not matter what is included), so that we can move get past the blocking factor that is keeping tens of PRs on hold for months/years.

umarcor:
This PR is now rebased, and the issues introduced in the latest contributions to master are fixed. The failure of the labeler seems to be external.

umarcor:
I rebased and fixed the issues introduced in the latest contributions to master. Since the failing labeler was removed recently, all tests are green now.

umarcor:
This PR was rebased again and some issues introduced in `*completions*` where fixed. 

/cc @marckhouzam [`39d3f42` (#1044)](https://github.com/spf13/cobra/pull/1044/commits/39d3f4259f1d3f6ea44dd7be5c19538394e2711f)

bash_completions_test.go
```diff
@@ -74,26 +75,26 @@ func TestBashCompletions(t *testing.T) {
 		Run:                    emptyRun,
 	}
 	rootCmd.Flags().IntP("introot", "i", -1, "help message for flag introot")
-	rootCmd.MarkFlagRequired("introot")
+	er(rootCmd.MarkFlagRequired("introot"))
```
wfernandes, line 78: For all instances of `er` could we use the exported function `Er` instead? 
Although it is part of the same package, I think we should strive to use our exported methods. What do you think?

cobra.go
```diff
@@ -205,3 +206,18 @@ func stringInSlice(a string, list []string) bool {
 	}
 	return false
 }
+
+func er(msg interface{}) {
+	if msg != nil {
+		fmt.Println("Error:", msg)
```
wfernandes, line 212: If this is intended to print out errors, I wonder if we should do something like 
```suggestion
		fmt.Fprintln(os.Stderr, "Error:", msg)
```

command.go
```diff
@@ -410,7 +412,7 @@ func (c *Command) UsageString() string {
 	c.outWriter = bb
 	c.errWriter = bb
 
-	c.Usage()
+	er(c.Usage())
```
wfernandes, line 415: Since it's in cobra pkg, could we use `Er`?

custom_completions_test.go
```diff
@@ -720,17 +720,17 @@ func TestRequiredFlagNameCompletionInGo(t *testing.T) {
 	rootCmd.AddCommand(childCmd)
 
 	rootCmd.Flags().IntP("requiredFlag", "r", -1, "required flag")
-	rootCmd.MarkFlagRequired("requiredFlag")
+	er(rootCmd.MarkFlagRequired("requiredFlag"))
```
wfernandes, line 723: Do we need these `er` wrappers here? Since it's a test, if we are not expecting an error we should specifically assert on them.

bash_completions_test.go
```diff
@@ -74,26 +76,26 @@ func TestBashCompletions(t *testing.T) {
...
 
 	// Filename.
 	rootCmd.Flags().String("filename", "", "Enter a filename")
-	rootCmd.MarkFlagFilename("filename", "json", "yaml", "yml")
+	er(rootCmd.MarkFlagFilename("filename", "json", "yaml", "yml"))
```
wfernandes, line 82: So I know that these show up as part of `errcheck` linter, but maybe the correct thing to do here is actually assert that there is no error since these are all in tests. 
Currently we aren't using any assertion libraries, maybe we write a helper like this for now. WDYT?
```suggestion
	assertNoErr(t, rootCmd.MarkFlagFilename("filename", "json", "yaml", "yml"))
```

```
func assertNoErr(t *testing.T, e error) {
  if e != nil {
    t.Error(e)
  }
}
```

powershell_completions_test.go
```diff
@@ -109,7 +109,7 @@ func TestPowerShellCompletion(t *testing.T) {
 	for _, tc := range tcs {
 		t.Run(tc.name, func(t *testing.T) {
 			buf := new(bytes.Buffer)
-			tc.root.GenPowerShellCompletion(buf)
+			er(tc.root.GenPowerShellCompletion(buf))
```
wfernandes, line 112: Similar, since this is a test, we should probably do something like `assertNoErr(t, ...)`

.golangci.yml
```diff
@@ -0,0 +1,25 @@
...
+    - maligned
+    - megacheck
+    - structcheck
+    - unconvert
+    - varcheck
+  enable-all: false
+  disable-all: true
+#  presets:
+#    - bugs
+#    - unused
```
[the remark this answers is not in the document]
umarcor: TBH, I just picked the default/minimal configuration file. My main purpose was/is to have golangci-lint added to the repo, so specific tweaks and additions can be done progessively. I now remove `enable-all` and `presets` and added the list of linters used in golangci-lint: https://golangci-lint.run/usage/configuration/

However, I commented all those that are not tested in this PR yet. Please, let me know if you want them removed.

cobra/cmd/helpers.go
```diff
@@ -60,109 +57,8 @@ func init() {
...
-		} else {
-			if line == "" {
-				newlines = append(newlines, "//")
-			} else {
-				newlines = append(newlines, "// "+line)
-			}
-		}
+	if msg != nil {
+		fmt.Println("Error:", msg)
+		os.Exit(1)
```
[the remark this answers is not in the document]
umarcor: Do you mean to use `cobra.Er` explicitly or to write the following?

```go
func er(msg interface{}) {
	cobra.Er(msg)
}
```

bash_completions_test.go
umarcor: At first, I didn't want to export it, because it feels like an internal utility. However, then I had to export it (I don't remember why). Once it is exported, I guess it makes sense to remove the unexported version. I did that now.

umarcor:
@wfernandes thanks a lot for taking care of this PR! ❤️ 

I addressed your concerns/comments regarding `er` and `assertNoErr`. I'm unsure about a couple of details, tho. Please, see the comments above.

cobra.go
```diff
@@ -205,3 +206,17 @@ func stringInSlice(a string, list []string) bool {
 	}
 	return false
 }
+
+// Er prints the msg with the prefix 'Error:' and exits with error code 1. If the msg is nil, it does nothing.
+func Er(msg interface{}) {
```
jharshman, line 211: Why are we exporting this? Looks like we are only using it internally.
Also `Er` doesn't _mean_ anything. Can we rename it to something a bit more telling?

umarcor: We are exporting it because it is used in different internal packages. One is cobra itself, the other ones are `cmd` and `doc`. These do not inherit functions from cobra. Hence, we need to export it if we want to reuse it without rewriting the same function three times. See review comment by @wfernandes above.

Regarding the name `er`, it's been in the codebase since @bogem added it three years ago: https://github.com/spf13/cobra/blame/master/cobra/cmd/helpers.go#L62. I just reused it first and then uppercased it. Last, after @wfernandes suggestion, I replaced it with `assertNoErr` in `*_test.go` sources. See also https://github.com/spf13/cobra/commit/21cab29ef902d36da074e8c940437c9f536b2283. Nevertheless, I'm ok with renaming it if you tell me what name to use instead.

jharshman: So IMO, if we are to export this method, it needs a better name. I wouldn't be a stickler if it was internal to the code base. But exporting it means it will be public facing in the docs so we should attempt to name it a bit better.
@wfernandes opinion?

cobra/cmd/helpers.go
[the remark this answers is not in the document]
wfernandes: Yeah...I was thinking of using `cobra.Er` directly instead of adding another level of indirection.

cobra.go
wfernandes: Thinking about it, I agree with @jharshman. Maybe we can call it `PrintErr`. WDYT?

umarcor: I renamed `Er` to `CheckErr`, as @marckhouzam wrote in https://github.com/spf13/cobra/blame/master/shell_completions.md#L90. Although I think it might have been a typo when copying it from somewhere else, the name feels good.

I also removed the indirections for both `Er` and `WrStringAndCheck`, as @wfernandes commented above.

marckhouzam: > I renamed `Er` to `CheckErr`, as @marckhouzam wrote in https://github.com/spf13/cobra/blame/master/shell_completions.md#L90

That was a copy paste from the old doc 😁.  I believe`util.CheckErr()` is code from `kubectl`. It's a good name in my opinion also.

cobra.go
```diff
@@ -205,3 +206,17 @@ func stringInSlice(a string, list []string) bool {
...
+// CheckErr prints the msg with the prefix 'Error:' and exits with error code 1. If the msg is nil, it does nothing.
+func CheckErr(msg interface{}) {
+	if msg != nil {
+		fmt.Fprintln(os.Stderr, "Error:", msg)
+		os.Exit(1)
+	}
+}
+
+// WrStringAndCheck writes a string into a buffer, and checks if the error is not nil.
+func WrStringAndCheck(b io.StringWriter, s string) {
```
jharshman, line 219: Let's rename this to `WriteStringAndCheck`. Be descriptive w/o being overly wordy for exported functions and methods.

jharshman:
One more minor comment and then I think we can call this one good.

umarcor:
@jharshman done. Just remember that this is a breaking change, since it requires go >1.11 (https://github.com/spf13/cobra/pull/1044#discussion_r384624132).

jharshman:
>  Just remember that this is a breaking change, since it requires go >1.11 (#1044 (comment)).
@umarcor ah yes let's not do that.

umarcor:
@jharshman is there any plan about when to deprecate older versions of go in cobra? v1.11 was EOL in september 2019, and v1.12 in february 2020.

jharshman:
@umarcor no current plan to do that.

doc/man_docs_test.go
```diff
@@ -133,7 +133,7 @@ func TestGenManSeeAlso(t *testing.T) {
 func TestManPrintFlagsHidesShortDeperecated(t *testing.T) {
 	c := &cobra.Command{}
 	c.Flags().StringP("foo", "f", "default", "Foo flag")
-	c.Flags().MarkShorthandDeprecated("foo", "don't use it no more")
+	cobra.CheckErr(c.Flags().MarkShorthandDeprecated("foo", "don't use it no more"))
```
wfernandes, line 136: nit:
Should this be an `assertNoErr`?

README.md
```diff
@@ -193,7 +193,6 @@ import (
 	"fmt"
 	"os"
 
-	homedir "github.com/mitchellh/go-homedir"
```
jharshman, line 196: @umarcor removing this will force older versions of Go to stop working. Let's not do that in this PR.

umarcor: @jharshman, I thought about that. If that was the only breaking change, I would agree. However, there is another breaking change that requires >1.11. IIRC `io.StringWriter` is not supported in deprecated versions of the language.

jharshman: I would say that my comment also applies to that change as well.
I think there is potential for a large amount of fallout from depreciating Go versions. @wfernandes opinion?

umarcor: @jharshman, note that I used `io.StringWriter` because golangci-lint would fail otherwise. In order to split it, the rule would need to be skipped per line, or per file, or the linter would need to be disabled.

I think it is better to keep this PR on hold until cobra is ready to deprecate older versions of go. Or to start a `v2` branch.

wfernandes: Hmm...this is definitely a hard one. 
Ideally, I'd be all for `cobra` to declare compatible versions of go. Maybe we can revise our support for unsupported golang versions in future major releases.
However, for now it wouldn't make sense for us to break users who may be using older versions of golang and have already upgraded to v1.x.x.

I also like the use of the interface rather than passing the `*bytes.Buffer` so maybe in the interim we can replicate the interface to be something like this.
```golang
// Using a duplicate of the io.StringWriter interface so that we can support golang version <1.11
type StringWriter interface{
  WriteString(s string) (n int, err error)
}
```
The above interface type can be located at the `cobra` package level. 

Then in the future, if we ever do decide to stop supporting older versions of golang, we can simply switch out the references from `cobra.StringWriter` to `io.StringWriter`. Or if there are public methods that somehow depend on the `cobra.StringWriter` we can also do something like 
```golang
type StringWriter interface{
  io.StringWriter
}
```
This is just a suggestion. WDYT?

wfernandes: Regarding the homedir stuff, maybe we just revert to the original version since it feels like it would be more work trying to work around it. 🤷

umarcor: All of the projects in https://github.com/spf13/cobra/blob/master/projects_using_cobra.md (except [delve](https://github.com/go-delve/delve)) are using >=1.12, most of them use 1.13 or 1.14 and several do require >=1.12. Honestly, I don't think that many actively maintained projects would be affected by this. Is it possible to get the +20k items in https://pkg.go.dev/github.com/spf13/cobra?tab=importedby through some API?

@wfernandes, while I think that such a solution would work now, I don't think it is acceptable in the mid-long term. I believe that having all sorts of intermediate functions to handle different versions of the language, or conditionally compiling files are acceptable solutions if required to support non EOL versions. Otherwise, it adds a severe maintainability burden and it makes it difficult to cleanup/refactor in the future. If cobra is going to support versions of golang longer than Google supports the language, I'd like to have a limit defined. For example, Debian Buster ships [1.11](https://packages.debian.org/buster/golang), so I think it would make sense to support >=1.11 but not <=1.10. By defining a limit, we can add deprecation warnings/messages for users to be aware.

umarcor: > Regarding the homedir stuff, maybe we just revert to the original version since it feels like it would be more work trying to work around it. 🤷

It is 300 loc (including sources and tests) and it has not been updated in 18 months. We might well copy it into cobra's codebase and remove the dependency.

jharshman: > or the linter would need to be disabled.

This seems like the answer here. Which linter in particular is flagging this?

wfernandes: @umarcor Apologies. Didn't see your reply before I had posted mine. 
@jharshman I think that would be the `interfacer`. https://github.com/mvdan/interfacer

jharshman: I have no problem keeping the dependency a dependency.
@umarcor I think the best path forward here is to disable the linter and revert the breaking changes.

When we do eventually drop support for older Go versions, we will have to give it a bit more thought and ensure that our user impact is minimal and the change doesn't come as a surprise.

jpmcb: While dropping support for older versions of Go may not be the ideal path here, I don't believe that it should be the responsibility of cobra to be the versioning "bottle neck". Any Go "1" project should in theory be compatible with all future versions of Go 1. Per the Go compatibility guarantee:
> Go programs that work today should continue to work even as future "point" releases of Go 1 arise (Go 1.1, Go 1.2, etc.). 

[Go 1 and the Future of Go Programs](https://golang.org/doc/go1compat)

Although this guarantee really only includes the standard Go API packages, cobra users who are still on go 1.11 should _in theory_ have an easy upgrade path.

In the case that cobra authors are not able to upgrade their go version, cobra 1.0.0 is feature complete and currently backwards compatible with earlier versions of go.

So my question is; why not include these? We can upgrade the version of go that cobra uses (getting all the security and performance fixes) and if users need a version of cobra that works with older versions of go, they can use the v1.0.0 tag. I do not think it should be our mandate to keep cobra (and all it's future features) compatible with all user scenarios and implementations. Rather, we should offer an upgrade path, noting the changes we make to the cobra API that are no longer able to be used with older versions of Go. Further, I would question why projects are being maintained with a version of Go that is technically EOL. There are plenty of security and performance fixes that justify cobra and it's users keeping up with the stable Go release channel.

I also recognize all the challenges this would bring. Having multiple "branches" of supported versions and compatibility creates release challenges all their own. We would likely also need to mandate a "semver" contract that would indicate how long versions are "supported".

jharshman: You have good points @jpmcb 

> ... if users need a version of cobra that works with older versions of go, they can use the v1.0.0 tag

The issue I have with doing this is that it seems like we are considering deprecating Go versions _just_ to satisfy this linter. We could say "If you want backward compat with older Go versions use v1.0.0. But then that cuts out those users of benefiting from some of the features that have been added since.

> Further, I would question why projects are being maintained with a version of Go that is technically EOL

You'd be surprised. :D 

IMO if we want to deprecate older Go versions that sounds good to me. I just want to make sure we prep and do it properly as to not take users by surprise. :)

umarcor: > cobra 1.0.0 is feature complete and currently backwards compatible with earlier versions of go

In fact, I thought that releasing v1.0.0 was a priority in order to then allow merging small but justified breaking changes. I've been rebasing this PR and #841 (although not strictly breaking) for +1y. I don't mind keeping doing it until the decision to accept those changes is taken.

umarcor: > IMO if we want to deprecate older Go versions that sounds good to me. I just want to make sure we prep and do it properly as to not take users by surprise. :)

Shall we open an issue to discuss it? Or will you maintainers handle it?

wfernandes:
@umarcor  After a quick chat with @jharshman and @jpmcb we decided to make plans to decide on which versions of Golang to support moving forward. This **may** result in a new version (probably an alpha) being cut to start working towards a more updated cobra moving forward.

In regards to this PR, we can either 
1. hold off on merging it until we cut a release that will support newer Golang versions or
1. modify this PR such that no breaking changes as suggested by @jharshman above are included and have that be merged in. A separate PR can be made with the breaking changes such as `homedir` stuff etc... which can be targeted for the future version.

Again, thanks for the contribution and we'll try and keep y'all posted on future changes. Thanks.

umarcor:
This PR is now based on #968. Here, changes that require golang `>=1.12` are included.

jharshman:
@umarcor , can you submit a trimmed down version of this PR that *only* adds support for golangci-lint and the changes to the source code to make that pass the linters?
I am going to move forward with deprecating old versions of Go and would love to incorporate some of these changes.

umarcor:
@jharshman, this is ready. Shall I move the last two commits to new PRs?

jharshman:
@umarcor yes that would be great.

umarcor:
Done.

jharshman:
@umarcor can you please squash your commits?

umarcor:
@jharshman I think that preserving the commits is meaningful in this case, because each of them fixes a different linter warning/error. Anyway, GitHub's web UI should allow you to resolve this PR by Merge, Squash or Rebase, as you wish.

Issue #1057, opened 2020-03-11
Feature request to support multi-word usage for root command


brianpursley:
Currently, cobra assumes that the command to display for usage in help is the first word.  However, sometimes the root command can actually be multiple words.  This is the case for kubectl plugins for example.

There is a [related issue opened in the krew repository](https://github.com/kubernetes-sigs/krew/issues/521) and it was proposed that the preferred fix may actually be to see if cobra can be changed to show the entire value of `Use` for the root command and not just the first word.

---

Here is an example illustrating the problem:

Consider this simple cobra application, **main.go**.  Imagine that this code implements a kubectl plugin called myplugin:
```
package main

import (
	"github.com/spf13/cobra"
)

func main() {
	var rootCmd = &cobra.Command{
		Use: "kubectl myplugin",
	}

	rootCmd.AddCommand(&cobra.Command{
		Use: "dosomething",
		Run: func(cmd *cobra.Command, args []string) {
		},
	})

	_ = rootCmd.Execute()
}
```

**PROBLEM 1:** When you run it and the default help is displayed, it says the usage is just `kubectl`.  But the code above says it should be `kubectl myplugin`.
```
go run main.go
```

```
Usage:
  kubectl [command]

Available Commands:
  dosomething 
  help        Help about any command

Flags:
  -h, --help   help for kubectl

Use "kubectl [command] --help" for more information about a command.
```

**PROBLEM 2:** When you run help for a specific subcommand, it similarly only shows the first word of the command:
```
go run main.go dosomething --help
```

```
Usage:
  kubectl dosomething [flags]

Flags:
  -h, --help   help for dosomething
```

---

I'm looking for any thoughts and ideas on this problem.  

I was thinking about trying to make a PR that would support something like this, but I don't want to do anything that would break existing apps. 

One idea would be to allow "escaping" spaces in the Use property.

I'm looking specifically at this function in command.go:
```
func (c *Command) Name() string {
	name := c.Use
	i := strings.Index(name, " ")
	if i >= 0 {
		name = name[:i]
	}
	return name
}
```

Instead of looking for the first space, can we change this so that I can specify something like `kubectl\ myplugin` and it will not consider `\ ` to be a space in this case because it is escaped, but it would still render as a space (without the `\`) in help output.

brianpursley:
I think we found a workaround using a custom usage template.

I’ll close the issue and reopen later if needed.

PR #1058, opened 2020-03-11
Allow Use property to have an escaped space

brianpursley:
This PR is a possible change to implement https://github.com/spf13/cobra/issues/1057

It allows you to specify a command name that has a space by passing in an escaped space.  This is needed for some commands that might have a space in them, as is the case for kubectl plugins for example.

My attempt was to preserve the existing functionality, so no existing behaviors are changed, and just add the optional ability to use an escaped space in the `Use` property.

See the linked issue above for more details and a code example of the problem this is attempting to solve.

With this PR, you would be able to do something like this
```
package main

import (
	"spf13/cobra"
)

func main() {
	rootCmd := &cobra.Command{
		Use: "kubectl\\ myplugin",  // <------ Notice the escaped space
	}

	rootCmd.AddCommand(&cobra.Command{
		Use: "dosomething",
		},
	})

	_ = rootCmd.Execute()
}
```

and the output will be:
```
go run main.go --help
```
```
Usage:
  kubectl myplugin [command]

Available Commands:
  dosomething 

Flags:
  -h, --help   help for kubectl myplugin

Use "kubectl myplugin [command] --help" for more information about a command.
```

I'm open to feedback, or if this is not in line with cobra's philosophy, let me know, but I wanted to make an attempt at proposing a solution for the issue I opened.  

Thanks!

### DIFF
--- a/command.go
+++ b/command.go
@@ -1203,9 +1203,9 @@ func (c *Command) CommandPath() string {
 func (c *Command) UseLine() string {
 	var useline string
 	if c.HasParent() {
-		useline = c.parent.CommandPath() + " " + c.Use
+		useline = c.parent.CommandPath() + " " + replaceEscapedSpaces(c.Use)
 	} else {
-		useline = c.Use
+		useline = replaceEscapedSpaces(c.Use)
 	}
 	if c.DisableFlagsInUseLine {
 		return useline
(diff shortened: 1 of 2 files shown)


brianpursley:
I think we found a workaround using custom usage templates.  I’m going to close this PR for.

PR #1062, opened 2020-03-18
Fix #1056

### LINKED ISSUE #1056: Additional help topic commands now return "subcommand is required" error
#922 changed the behavior of all non-runnable commands, including additional help topic commands.

They now return the new `ErrSubCommandRequired` error, which by default results in `subcommand is required` being displayed and exiting with status 1. Both are incorrect for additional help topic commands, whose entire purpose is to display a help message.

Can we restore the previous behavior for additional help topic commands? Naively, something like this:

```patch
	if !c.Runnable() {
-		return ErrSubCommandRequired
+		if c.IsAdditionalHelpTopicCommand() {
+			return flag.ErrHelp
+		} else {
+			return ErrSubCommandRequired
+		}
	}
```

cc @bruceadowns

ob-stripe:
Restore previous behavior for additional help topic commands.

Fixes #1056.

cc @bruceadowns

### DIFF
--- a/command.go
+++ b/command.go
@@ -801,7 +801,11 @@ func (c *Command) execute(a []string) (err error) {
 	}
 
 	if !c.Runnable() {
-		return ErrSubCommandRequired
+		if c.IsAdditionalHelpTopicCommand() {
+			return flag.ErrHelp
+		} else {
+			return ErrSubCommandRequired
+		}
 	}
 
 	c.preRun()
--- a/command_test.go
+++ b/command_test.go
@@ -897,10 +897,35 @@ func TestFlagsInUsage(t *testing.T) {
 	checkStringContains(t, output, "[flags]")
 }
 
-func TestHelpExecutedOnNonRunnableChild(t *testing.T) {
+func TestHelpExecutedOnAdditionalHelpTopicCommand(t *testing.T) {
 	rootCmd := &Command{Use: "root", Run: emptyRun}
 	childCmd := &Command{Use: "child", Long: "Long description"}
+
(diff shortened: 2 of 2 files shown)


jharshman:
@ob-stripe - thanks for finding this bug.  Quick question, what do you think of just reverting to the behavior that we had before #922 ?

ob-stripe:
Good question. It's debatable whether not providing a subcommand to a command that expects one should return an error or not.

I'm one of the maintainers of the [Stripe CLI](https://github.com/stripe/stripe-cli) which makes heavy use of nested subcommands. For example, running this:
```
$ stripe charges
Usage:
  stripe charges <operation> [parameters...]

Available Operations:
  capture
  create
  list
  retrieve
  update

...
```

For our case at least, I think simply displaying the help message and exiting with status code 0 is preferable. The `charges` command did its job: display the help message listing available subcommands.

If that's an option, I would prefer reverting to the previous behavior entirely. Maybe we can have the best of both worlds by making the new behavior available on demand?

jharshman:
I'm reviewing the issues that brought the change in #922 about.  From what I can tell, it was a debate about whether or not Cobra should emit an exit code for a command that doesn't exist.

My preference here is to lean on `RunE`. To handle the return of an error code.  From my experience, this handles the case nicely.

Since this change really just landed in 0.0.6, I'm definitely okay reverting it.

ob-stripe:
Awesome. Feel free to close this PR then.

jharshman:
closing in favor of #1068

PR #1048, opened 2020-02-29
Fish completion (including support for Go custom completion)

marckhouzam:
This is an implementation of Fish completion based on the Custom Go Completions of PR https://github.com/spf13/cobra/pull/1035

The PR takes a different approach to completion.  Instead of coding the entire completion in shell script, it delegates to the Go program the responsibility of determining the completion choices.  This makes it simpler to implement completion support, as Cobra provides all the necessary info to easily determine completions.

It automatically supports custom completions once they have been migrated to the Custom Go Completion method of PR #1035.

And it supports automatically adding the "usage" string for command and flag completions to describe these completions.  This is something bash completion doesn't have.

I haven't yet tried to support the other annotations `BashCompFilenameExt`, `BashCompOneRequiredFlag` or `BashCompSubdirsInDir`.

To easily try this out I have adapted `kubectl` in my repo to use this PR. Bear in mind that I haven't migrated the custom completions for `kubectl` (e.g., `kubectl get <TAB>`) to use the new Custom Go Completions of #1035, so custom completions will not yet work with this PR.  The `kubectl` that works with basic completions is here: https://github.com/VilledeMontreal/kubernetes/tree/feat/fishCompletion

My experience with Fish is limited, so it would be good for others to try this out and provide feedback.

PR #1065, opened 2020-03-23
Don't exclude 'help' from bash completions

### LINKED ISSUE #1000: Bash completions shouldn't exclude `help` command
There appears to be an explicit check in the bash completion code that excludes the help command: https://github.com/spf13/cobra/blob/b04b5bfc50cbb6c036d2115ed55ea1bccdaf82a9/bash_completions.go#L326

This means that the resulting completion script doesn't show `help` as a valid cobra command, which seems unnecessary? This is also not present in the zsh completions generator.

voithos:
Fixes #1000.

### DIFF
--- a/bash_completions.go
+++ b/bash_completions.go
@@ -324,7 +324,7 @@ fi
 func writeCommands(buf *bytes.Buffer, cmd *Command) {
 	buf.WriteString("    commands=()\n")
 	for _, c := range cmd.Commands() {
-		if !c.IsAvailableCommand() || c == cmd.helpCommand {
+		if !c.IsAvailableCommand() && c != cmd.helpCommand {
 			continue
 		}
 		buf.WriteString(fmt.Sprintf("    commands+=(%q)\n", c.Name()))
@@ -496,7 +496,7 @@ func writeArgAliases(buf *bytes.Buffer, cmd *Command) {
 
 func gen(buf *bytes.Buffer, cmd *Command) {
 	for _, c := range cmd.Commands() {
-		if !c.IsAvailableCommand() || c == cmd.helpCommand {
+		if !c.IsAvailableCommand() && c != cmd.helpCommand {
 			continue
 		}
 		gen(buf, c)


marckhouzam:
Thanks @voithos, this is nicer than before.

I've tested this change with Helm and I confirm that the `help` command is now proposed as a completion.  I've updated #1048 to match this better behaviour (https://github.com/spf13/cobra/pull/1048/commits/da428b554c6ad6fe523a7c9bc5869dec2875c79a).

However, this introduces an unexpected change for completion.  If I do `helm help [tab][tab]`, I now get file completion instead of a list of all sub-commands.  Before, I would get the list of sub-commands.  I believe the previous behaviour was a fluke because the completion script ignored the unknown `help` argument (or something along those lines), but now it knows about `help` but `help` does not actually have all the real sub-commands as its own sub-commands.

In my mind the previous behaviour provided better completion for the user.  I'm not sure we can do something reasonable to fix this right now; however, it may be possible, once #1035 is merged, to add a `ValidArgsFunction` field to the `help` command which would provide sub-commands as completions.

Of course, I don't have the final say in this, so let's see what @jharshman thinks about it.

P.S. I believe the circleCI test failures can be ignored because they are due to a missing config file in Cobra and not related to this PR.

marckhouzam:
BTW, the current Cobra zsh completion has the same limitation of not showing sub-commands when doing `<program> help [tab][tab]`

jharshman:
> I believe the circleCI test failures can be ignored because they are due to a missing config file in Cobra and not related to this PR.

Yeah these go away on next commit

@marckhouzam thanks for testing this one out. I think once we have #1035 in we can re-evaluate.

voithos:
Thanks for testing this! I'll hold off, then, until #1035 resolves.

marckhouzam:
@voithos Now that #1035 has been merged, do you want to try this PR again?  This idea is to add the `ValidArgsFunction` field to the `help` command to provide each sub-command name as a completion.

marckhouzam:
Fixing this properly will also allow the new Fish completion to complete sub-commands after `help`.  Currently `<program> help [tab]` does not provide command names for Fish shell.

voithos:
Sorry for the slow response, but I likely won't get much time to work on this for a while, so if anyone else wants to tackle it, feel free!

marckhouzam:
> Sorry for the slow response, but I likely won't get much time to work on this for a while, so if anyone else wants to tackle it, feel free!

I've posted #1136 which supersedes this PR and adds the completion of the `help` command.
@voithos I've kept your original commit and add my own on top.

voithos:
@marckhouzam Excellent, thank you!

marckhouzam:
#1136 was merged which supercedes this one.

@voithos do you mind closing this one?

voithos:
You got it, thanks for doing all that!

PR #887, opened 2019-06-12
zsh: added compdef call to support direct sourcing

rsteube:
as suggested by @corneliusweig

fixes #881

Issue #886, opened 2019-06-12
Zsh completion never suggests files but only options


corneliusweig:
An overhauled zsh completion was merged with #646. One drawback of that script is that it only ever suggests options. Say for example, I have a command `cmd` with option `--foo`. When I type
```zsh
cmd <Tab><Tab>
```
I would expect to see a suggestion list of all files in the working dir. However, it actually adds a `-` and shows the `--foo` suggestion.

It would be much better to trigger option completion only when typing
```zsh
cmd -<Tab><Tab>  # <-- note the dash
```
and otherwise show the standard list of files in the cwd.

/cc @rsteube @babysnakes

rsteube:
Since you added a specific completion for `cmd` you have overriden the default `_files` function.
Now unless you tell it that your command accept files it won't complete it for you.

`command.MarkZshCompPositionalArgumentFile(1)` might be the solution (complete files for first argument).

There's also the possibilty filtering by patterns:
https://github.com/spf13/cobra/blob/master/zsh_completions.go#L125

corneliusweig:
Thanks @rsteube, I was not aware that this had to be enabled manually. :+1:

blacktop:
@rsteube can you show a small example on how to add them?  I am trying and it doesn't seem to work for sub commands?

rsteube:
There is a test using it: https://github.com/spf13/cobra/blob/e2c45ac9eb0cdb823b89c1e868d698dc74822797/zsh_completions_test.go#L164

You probably know, but if it's a sub command it needs to be called on the sub command and not on the root command. Can't think of any other reason right now why it's not working.

blacktop:
I added it here:

https://github.com/blacktop/ipsw/blob/master/cmd/ipsw/cmd/device.go#L49

It doesn't seem to have any impact on what is output by running:

```bash
ipsw completion zsh
```

I would have expected to see "DeviceTree*" show up somewhere or `_file`

😕

rsteube:
I think it's because of the sub sub command "list": https://github.com/blacktop/ipsw/blob/9b307d745c4a7b27f8567d281c0fa7ce0169c600/cmd/ipsw/cmd/device_list.go#L15

That should set the list commond as positional argument 1 and thus override your file argument (both is probably not possible). When that is commented the Devicetree argument is rendered in the output.

blacktop:
Ah thank you! I should simplify my sub sub subs into a more typical cli.  I also confirmed if I remove the sub subs it does in fact work 👍

PR #1070, opened 2020-03-30
Zsh completion V2

marckhouzam:
This is a new implementation of Zsh completion but based on the Custom Go Completions of PRs #1035 and #1048.

The current Zsh completion has a major limitation in the fact that it does not support custom completions.  Furthermore, if it ever did, it would require the program using Cobra to implement a second, zsh-flavoured version of its bash completion code.

This v2 version allows to automatically re-use custom completions once they have been migrated to the Custom Go Completion solution of PR #1035.

*Advantages:*

* support custom completions (in Go only) (as mentioned above)
* proper support for  using `=` for flags
* fixed size completion script of around 100 lines, so very fast on shell startup
* allows disabling of descriptions (to feel like bash)
* when there are no other completions, provides file completion automatically. This can be turned off on a per command basis


mike-stewart:
Is this going to be merged? I'm very interested in being able to apply the changes from PR 1035 for zsh completion.

mike-stewart:
@marckhouzam I tried your last commit rebased onto master, with one modification to allow it to build (53b3354af71147874264fbd163c0fdc56f7c09a7). The custom completions work great, but all other completions are broken - not even completing the commands themselves. Is that expected or a bug?

Let me know if I can contribute in any way - I'm new to Go but could take a stab at it if you can point me in the right direction.

marckhouzam:
Thanks @mike-stewart for trying it out.  
This PR is dependant on #1048. That's why completions don't work for you.

I'm about to post an updated version that will work.

marckhouzam:
Ok, so this new version should work.
Here is a small test program I use:
```
package main

import (
	"fmt"
	"os"

	"github.com/spf13/cobra"
)

var rootCmd = &cobra.Command{
	ValidArgs: []string{"replica\tanother", "rs\tdescription for rs", "ds"},
	Args:      cobra.ExactArgs(1),
	Use:       "testprog1",
		fmt.Println("rootCmd called")
	},
}

	Use:   "pod",
	Short: "Start the pod",
	ValidArgsFunction: func(cmd *cobra.Command, args []string, toComplete string) ([]string, cobra.BashCompDirective) {
		return []string{"these", "are\tverb", "custom\tadjective", "completions\tnoun, plural"}, cobra.BashCompDirectiveDefault
	},
		fmt.Println("child called")
	},
}

var childCmd2 = &cobra.Command{
	Use:   "svc",
	Short: "Launch the service",
	ValidArgsFunction: func(cmd *cobra.Command, args []string, toComplete string) ([]string, cobra.BashCompDirective) {
		return []string{"second", "level"}, cobra.BashCompDirectiveDefault
	},
		fmt.Println("child called")
	},
}

	Use:                   "completion [bash|zsh1|zsh2|fish] [--no-descriptions]",
	Short:                 "Generate completion script",
	DisableFlagsInUseLine: true,
	ValidArgs:             []string{"bash", "zsh1\tOriginal zsh completion", "zsh2\tV2 zsh completion", "fish"},
		switch args[0] {
		case "bash":
			break
		case "zsh1":
			break
		case "zsh2":
			cmd.Root().GenZshCompletionV2(os.Stdout, !completionNoDesc)
			break
		case "fish":
			break
		}
	},
}

var (
	completionNoDesc bool
	testFlag         string
)

func main() {
	completionCmd.PersistentFlags().BoolVar(
		&completionNoDesc,
		"no-descriptions", false,
	completionCmd.PersistentFlags().StringVar(
		&testFlag,
		"test-flag", "",
		"just to test")
	rootCmd.AddCommand(childCmd)
	childCmd.AddCommand(childCmd2)
	rootCmd.Execute()
}
```

mike-stewart:
@marckhouzam This works great on my (fairly simple) app w/ custom completions.

> when there are no other completions, provides file completion automatically. This can be turned off on a per command basis

What is the correct way to turn off this file completion?

marckhouzam:
I'll direct you to the documentation so you can let us know if it's understandable or not.

Look for the description for `BashCompDirective` in https://github.com/spf13/cobra/blob/master/bash_completions.md#1-custom-completions-of-nouns-written-in-go

mike-stewart:
@marckhouzam Yes, that works. I had implemented that already for some commands, but thought your comment suggested a way to disable file completion for commands where custom completion is *not* being used.

I worked around that by adding this to commands that I don't want any completion on:
```
ValidArgsFunction: func(cmd *cobra.Command, args []string, toComplete string) ([]string, cobra.BashCompDirective) {
	return nil, cobra.BashCompDirectiveNoFileComp
},
```

Thanks again for getting this branch working.

marckhouzam:
> I worked around that by adding this to commands that I don't want any completion on:
> 
> ```
> ValidArgsFunction: func(cmd *cobra.Command, args []string, toComplete string) ([]string, cobra.BashCompDirective) {
> 	return nil, cobra.BashCompDirectiveNoFileComp
> },
> ```

That's exactly how you should do it.  Should we add an example in the docs for this case, which I expect will be common or will users figure it out easily enough? As the first user, what do you think?

mike-stewart:
@marckhouzam The workaround was obvious upon reading the code and documentation, but it right not have been immediately apparent to a casual reader unless they specifically set out to disable file completion. Adding it to the documentation would make it clearer that it's a "feature" that is available to users.

marckhouzam:
I will rebase this later today to adapt to the `BashCompDirective` to `ShellCompDirective` change of #1082 .

marckhouzam:
>  Adding it to the documentation would make it clearer that it's a "feature" that is available to users.

@mike-stewart would you like to make a small PR to add this explanation to the documentation?

marckhouzam:
I rebased this PR after the `BashCompDirective` to `ShellCompDirective` change.

itskingori:
> **This PR is not totally finished, but before spending more time on it, I wanted to know if there is some interest for such a change?**

@marckhouzam There's definitely interest in this change! 😅 I've been following your other PRs and was glad when they got merged and [v1.0.0 released](https://github.com/spf13/cobra/releases/tag/v1.0.0). The only thing missing is this PR.

For context I've used custom Go completions to implement completion of flag values using `RegisterFlagCompletionFunc` but if I understand correctly, that would work for Bash and Fish only at the moment.

marckhouzam:
I will soon get back to this PR. But first I want migrate the Helm project to using Cobra 1.0.0 🎉

marckhouzam:
I looked a bit into the existing zsh completion support to see if we really need a V2 version or if we could completely fulfill the current zsh completion API with this new solution.  IIUC the current zsh solution has some APIs of its own that are not relevant to bash completion.  This seems confirmed by this comment from the original author of the current zsh completion solution:
https://github.com/spf13/cobra/pull/646#issuecomment-500731127

As that comment mentions, there really should not be different APIs for different shells but we should strive for a single API working for all shells.

I think we can get very close to that goal with the current PR.  But that confirms we do need to introduce a V2 version for zsh completion as this PR should not support the zsh-only API of the other zsh solution.

Now that I'm more comfortable about having to introduce a V2, I'll be able to get a polished version ready soon.

marckhouzam:
I've rebased on master and updated the documentation. To my knowledge, all types of completions are supported except the ones enabled by `BashCompFilenameExt` (filtering by file extension), `BashCompOneRequiredFlag` (required flags) and `BashCompSubdirsInDir` (filtering by directory).

I'm not removing the WIP label because I'm still thinking about how to avoid having a V2 version.

The only two APIs that are zsh-specific are `MarkZshCompPositionalArgumentWords`
and `MarkZshCompPositionalArgumentFile` .  I believe the first one corresponds to `ValidArgs` so would be easy to handle.  IIUC, the second one does not have a corresponding one in Bash-completion, but maybe we should add one to standardize, and then be backwards-compatible for zsh.

I'll keep thinking about it.

zsh_completions_v2.go
```diff
@@ -0,0 +1,145 @@
...
+        __%[1]s_debug "Activating nospace."
+        # We can use compadd here as there is no description when
+        # there is only one completion.
+        compadd -S '' "${lastComp}"
+    else
+        _describe "completions" completions
+    fi
+}
+
+compdef _%[1]s %[1]s
```
marckhouzam, line 124: This line allows to source the zsh completion script as is currently done for `helm` and `kubectl`.  I like the idea of providing that option.

However, I need to point out that @jharshman disagreed with this in https://github.com/spf13/cobra/pull/887#issuecomment-536621700, so he may prefer to have this removed, once we get to review time.

jharshman: @marckhouzam yeah I still maintain my previous opinion about this.

marckhouzam: @jharshman is right.  I found this article to be a good explanation of why sourcing a completion script should not be recommended: https://medium.com/@jzelinskie/please-dont-ship-binaries-with-shell-completion-as-commands-a8b1bcb8a0d0

For projects that need to keep backwards-compatibility, they can insert that one extra line themselves.  I will need to do that for Helm.

I will remove this line in my next update.

marckhouzam:
I've continued investigating the existing zsh-completion support.
As explained by the original author (https://github.com/spf13/cobra/pull/646#issuecomment-500821302), the current zsh support was designed independently of bash completion and is not well aligned to it.

There are multiple differences but one major difference is that the current zsh completion does not do file completion by default while bash does (see #886).  Say I have a program `testprog` that has no sub-commands and no `ValidArgs` then `./testprog [tab][tab]` will list all files for bash, but will not for zsh.  

To get file completion (first the first argument only) for zsh, the program developer would need to add the line
```
command.MarkZshCompPositionalArgumentFile(1)
```
On the other hand, to turn off file completion for bash, the developer would need to use the new `ShellCompDirectiveNoFileComp` with `ValidArgsFunction`.

This means that a developer that wants to support both bash and zsh must write code to handle zsh completion (current version) and different code for the bash completion.

I think we should strive to align completion behavior across all shells or else it becomes a nightmare for the program developers.

This leads me to backwards-compatibility.  I strongly feel that the new zsh completion of this PR should provide file completion by default like bash, as justified above.  This would technically be a breaking-change compared to the current zsh completion script.

So I think we are at the point of making a decision: do we do a V2 version for zsh like this PR is doing right now, or do we modify the current behavior of zsh completion to align with bash?

 @jharshman I'm looking for guidance from the maintainers, as this PR is close to finished and mostly need to know which direction to take with respect to V2 or not.  Thanks in advance.

jharshman:
@marckhouzam Having all the shell completion work the same is definitely appealing.
@wfernandes @jpmcb what do you guys think?

jpmcb:
> I think we should strive to align completion behavior across all shells or else it becomes a nightmare for the program developers.

I agree. Otherwise, I think we end up with "drift" between the different shells which creates pain for the cobra application developers. They end up having to maintain any number of different completions that all have different APIs for similar behavior. 

I'm in favor of introducing breaking changes for zsh completions that strives for closer parity among the various shell completions. And maybe we begin to deprecate the old version or announce that this is the new way to go about completions. But I'm not sure how this project in the past has approached deprecating features so I'll leave that up to @jharshman

jharshman:
@marckhouzam alright let's align the behavior even it it introduces some breaking changes.
We will handle the announcement and release.

marckhouzam:
Alright, I like that too.  Thanks.
What I'll do is document whatever "breaking" change this PR introduces.

marckhouzam:
I'm now working on supporting Required flags (the `BashCompOneRequiredFlag` annotation) in Go completions for this PR.  Whenever time allows, I need a review on the small fix of #1095 to get that working.  Thanks in advance.

marckhouzam:
Still a WIP but I've made good progress.  It is not fully tested yet.

This latest version **replaces** the current Zsh completion with a Zsh completion solution based on Go completions (without introducing a V2 API). 

Also, Go completions now support the bash completion annotations:
* BashCompFilenameExt (including Command.MarkFlagFilename() family)
    * this if for file filtering
    * supported by zsh as before
* BashCompSubdirsInDir
    * this is for filtering directories
    * newly supported by zsh
 * BashCompOneRequiredFlag (including Command.MarkFlagRequired() family)
    * this is for required flags
    * newly supported by zsh and fish

Still pending:
* update the documentation
* deal with the zsh-specific APIs in an sufficiently safe manner

marckhouzam:
This is finally ready for review.

The changes to zsh completion behaviour were not as dramatic as I expected and mostly surround file completion being done by default now.

The API was kept backwards-compatible, with 2 functions being deprecated.

I've also standardized the documentation (in the last commit) to talk about "shell completion" instead of "bash completion".

Below is the description of the `zsh` changes, as taken from the proposed documentation:
https://github.com/VilledeMontreal/cobra/blob/feat/zshCompletionV2/zsh_completions.md#zsh-completions-standardization

### Deprecation summary

See further below for more details on these deprecations.

* `cmd.MarkZshCompPositionalArgumentFile(pos, []string{})` is no longer needed.  It is therefore **deprecated** and silently ignored.
* `cmd.MarkZshCompPositionalArgumentFile(pos, glob[])` is **deprecated** and silently ignored.
  * Instead use `ValidArgsFunction` with `ShellCompDirectiveFilterFileExt`.
* `cmd.MarkZshCompPositionalArgumentWords()` is **deprecated** and silently ignored.
  * Instead use `ValidArgsFunction`.

### Behavioral changes

**Noun completion**
|Old behavior|New behavior|
|---|---|
|No file completion by default (opposite of bash)|File completion by default; use `ValidArgsFunction` with `ShellCompDirectiveNoFileComp` to turn off file completion on a per-argument basis|
`cmd.MarkZshCompPositionalArgumentFile(pos, []string{})` used to turn on file completion on a per-argument position basis|File completion for all arguments by default; `cmd.MarkZshCompPositionalArgumentFile()` is **deprecated** and silently ignored|
|`cmd.MarkZshCompPositionalArgumentFile(pos, glob[])` used to turn on file completion **with glob filtering** on a per-argument position basis (zsh-specific)|`cmd.MarkZshCompPositionalArgumentFile()` is **deprecated** and silently ignored; use `ValidArgsFunction` with `ShellCompDirectiveFilterFileExt` for file **extension** filtering (not full glob filtering)|
|`cmd.MarkZshCompPositionalArgumentWords(pos, words[])` used to provide completion choices on a per-argument position basis (zsh-specific)|`cmd.MarkZshCompPositionalArgumentWords()` is **deprecated** and silently ignored; use `ValidArgsFunction` to achieve the same behavior|

**Flag-value completion**

|Old behavior|New behavior|
|---|---|
|No file completion by default (opposite of bash)|File completion by default; use `RegisterFlagCompletionFunc()` with `ShellCompDirectiveNoFileComp` to turn off file completion|
|`cmd.MarkFlagFilename(flag, []string{})` and similar used to turn on file completion|File completion by default; `cmd.MarkFlagFilename(flag, []string{})` no longer needed in this context and silently ignored|
|`cmd.MarkFlagFilename(flag, glob[])`  used to turn on file completion **with glob filtering** (syntax of `[]string{"*.yaml", "*.yml"}` incompatible with bash)|Will continue to work, however, support for bash syntax is added and should be used instead so as to work for all shells (`[]string{"yaml", "yml"}`)|
|`cmd.MarkFlagDirname(flag)` only completes directories (zsh-specific)|Has been added for all shells|
|Completion of a flag name does not repeat, unless flag is of type `*Array` or `*Slice` (not supported by bash)|Retained for `zsh` and added to `fish`|
|Completion of a flag name does not provide the `=` form (unlike bash)|Retained for `zsh` and added to `fish`|

**Improvements**

* Custom completion support (`ValidArgsFunction` and `RegisterFlagCompletionFunc()`)
* File completion by default if no other completions found
* Handling of required flags
* File extension filtering no longer mutually exclusive with bash usage
* Completion of directory names *within* another directory
* Support for `=` form of flags

bash_completions.md
```diff
@@ -1,206 +1,14 @@
...
-backend        frontend       database 
-```
-
-Note that without declaring `rc` as an alias, the completion algorithm would show the list of nouns
-in this example again instead of the replication controllers.
-
-### Dynamic completion of nouns
-
-In some cases it is not possible to provide a list of possible completions in advance.  Instead, the list of completions must be determined at execution-time.  Cobra provides two ways of defining such dynamic completion of nouns. Note that both these methods can be used along-side each other as long as they are not both used for the same command.
+## Bash legacy dynamic completions
```
marckhouzam, line 5: I've referred to the original bash custom completions (the one that inserts bash code provided by the program into the script) as "legacy".  I don't want to offend anyone though, so it can be reformulated.

custom_completions.go
```diff
@@ -147,79 +171,138 @@ func (c *Command) getCompletions(args []string) (*Command, []string, ShellCompDi
...
+		// If we have not found any required flags, only then can we show regular flags
+		if len(completions) == 0 {
+			// We cannot use finalCmd.Flags() because we may not have called ParsedFlags() for commands
+			// that have set DisableFlagParsing; it is ParseFlags() that merges the inherited and
+			// non-inherited flags.
+
+			doCompleteFlags := func(flag *pflag.Flag) {
+				if !flag.Changed ||
+					strings.Contains(flag.Value.Type(), "Slice") ||
+					strings.Contains(flag.Value.Type(), "Array") {
```
marckhouzam, line 234: This prevents from suggesting a flag name that is already on the command-line (unless it is an Array or Slice). This is how zsh completion was doing things but not done for `bash`.  However, it seems to be a good idea, as mentioned in https://github.com/spf13/cobra/issues/107#issuecomment-369291283

custom_completions.go
```diff
@@ -147,79 +171,138 @@ func (c *Command) getCompletions(args []string) (*Command, []string, ShellCompDi
...
 
+	var completions []string
+	directive := ShellCompDirectiveDefault
 	if flag == nil {
 		// Complete subcommand names
-		for _, subCmd := range finalCmd.Commands() {
-			if subCmd.IsAvailableCommand() && strings.HasPrefix(subCmd.Name(), toComplete) {
-				completions = append(completions, fmt.Sprintf("%s\t%s", subCmd.Name(), subCmd.Short))
+		if len(finalArgs) == 0 {
+			// We only complete sub-commands if there are no arguments
```
marckhouzam, line 269: This was a bug compared to bash.  We should not suggest a command if there is a noun already present.

custom_completions.go
```diff
@@ -147,79 +171,138 @@ func (c *Command) getCompletions(args []string) (*Command, []string, ShellCompDi
...
+		// Always complete ValidArgs, even if we are completing a subcommand name.
+		// This is for commands that have both subcommands and ValidArgs.
 		if len(finalCmd.ValidArgs) > 0 {
-			// Always complete ValidArgs, even if we are completing a subcommand name.
-			// This is for commands that have both subcommands and ValidArgs.
-			for _, validArg := range finalCmd.ValidArgs {
-				if strings.HasPrefix(validArg, toComplete) {
-					completions = append(completions, validArg)
+			if len(finalArgs) == 0 {
+				// ValidArgs is only for the first argument
```
marckhouzam, line 287: This was a bug compared to bash. We should only suggest a ValidArg for the first argument, not keep suggesting it.

custom_completions.go
```diff
@@ -249,11 +332,18 @@ func getFlagNameCompletions(flag *pflag.Flag, toComplete string) []string {
...
-			flagName += "="
-			completions = append(completions, fmt.Sprintf("%s\t%s", flagName, flag.Usage))
-		}
+		// Why suggest both long forms: --flag and --flag= ?
+		// This forces the user to *always* have to type either an = or a space after the flag name.
+		// Let's be nice and avoid making users have to do that.
+		// Since boolean flags and shortname flags don't show the = form, let's go that route and never show it.
+		// The = form will still work, we just won't suggest it.
+		// This also makes the list of suggested flags shorter as we avoid all the = forms.
+		//
```
marckhouzam, line 341: This deviates from bash completion.
I find it more user-friendly and don't understand why it was not done like that from the start.

marckhouzam:
Here is a bit of a test program that I've been modifying back and forth during development.
```
package main

import (
	"fmt"
	"os"
	"strings"

	"github.com/spf13/cobra"
)

var completionNoDesc bool

	Use:   "completion [bash|zsh|fish|powershell]",
	Long: `To load completions:

Bash:

$ source <(yourprogram completion bash)

Linux:
  $ yourprogram completion bash > /etc/bash_completion.d/yourprogram
MacOS:
  $ yourprogram completion bash > /usr/local/etc/bash_completion.d/yourprogram

Zsh:

$ source <(yourprogram completion zsh)

$ yourprogram completion zsh > "${fpath[1]}/_yourprogram"

Fish:

$ yourprogram completion fish | source

$ yourprogram completion fish > ~/.config/fish/completions/yourprogram.fish
`,
	DisableFlagsInUseLine: true,
	ValidArgs:             []string{"bash", "zsh", "fish", "powershell"},
		switch args[0] {
		case "bash":
		case "zsh":
			if !completionNoDesc {
			} else {
			}
		case "fish":
		case "powershell":
		}
	},
}

var rootCmd = &cobra.Command{
	Use: "testprog",
		fmt.Println("rootCmd called")
	},
}

	Use:       "vargs",
	Short:     "Cmd with vargs and subcmd",
	Args:      cobra.MinimumNArgs(1),
	},
}

	Use:   "childVargs",
	Short: "Child command",
	},
}

	Use:   "filterFile",
	Short: "filterFile [*.log,*.txt]",
		if len(args) == 0 {
		}
	},
	},
}

	Use:   "nofile",
	Short: "nofile",
	},
		fmt.Println("nofile called")
	},
}

	Use:   "fullFile",
	Short: "fullFile [filename]",
	},
}

	Use:   "dirOnly",
	Short: "dirOnly [dirname]",
		if len(args) == 0 {
			return nil, cobra.ShellCompDirectiveFilterDirs
		}
	},
		fmt.Println("dirOnly called")
	},
}

	Use:   "subdir",
	Short: "subdir [sub-dirname]",
		if len(args) == 0 {
		}
	},
		fmt.Println("subdir called")
	},
}

	Use:   "compAndFile",
	Short: "compAndFile [sun|moon|<filename>]",
		var completions []string
		// If we don't filter on "toComplete", zsh and fish will not do file completion
		// even if the prefix typed by the user does not match the returned completions
		// e.g., ./testprog compAndFile f<TAB>
		// will only get file completion for bash, not zsh and fish, if we blindly return
		// the non-matching completions ("sun", "moon")
		for _, comp := range []string{"sun", "moon"} {
			}
		}
	},
	},
}

func setFlags() {
	rootCmd.Flags().String("theme", "", "theme to use (located in /themes/THEMENAME/)")
	rootCmd.Flags().SetAnnotation("theme", cobra.BashCompSubdirsInDir, []string{"themes"})

	rootCmd.Flags().String("theme2", "", "theme to use (located in /themes/THEMENAME/)")
	rootCmd.RegisterFlagCompletionFunc("theme2", func(cmd *cobra.Command, args []string, toComplete string) ([]string, cobra.ShellCompDirective) {
		return []string{"themes"}, cobra.ShellCompDirectiveFilterDirs
	})

	rootCmd.Flags().String("customComp", "", "test custom comp for flags")
	rootCmd.RegisterFlagCompletionFunc("customComp", func(cmd *cobra.Command, args []string, toComplete string) ([]string, cobra.ShellCompDirective) {
		return []string{"firstComp\tthe first value", "secondComp\tthe second value", "forthComp\tthe forth value"}, cobra.ShellCompDirectiveNoFileComp
	})

	rootCmd.Flags().StringP("file", "f", "", "list files")

	rootCmd.Flags().BoolP("bool", "b", false, "bool flag")
	rootCmd.Flags().BoolSliceP("bslice", "B", nil, "bool slice flag")

	// rootCmd.Flags().String("required", "", "required flag")
	// rootCmd.MarkFlagRequired("required")
}

func main() {
		&completionNoDesc,
		"no-descriptions", false,

	rootCmd.AddCommand(filterFileCmd)
	rootCmd.AddCommand(noFileCmd)
	rootCmd.AddCommand(fullFileCmd)
	rootCmd.AddCommand(dirOnlyCmd)
	rootCmd.AddCommand(subdirCmd)
	rootCmd.AddCommand(compAndFileCmd)

	rootCmd.AddCommand(vargsCmd)

	setFlags()
	rootCmd.Execute()

	// childCmd1.AddCommand(childCmd2)

	// childCmd1.Flags().String(
	// 	"test-file2", "",
	// 	"just to test2")
	// childCmd1.RegisterFlagCompletionFunc("test-file2", func(cmd *cobra.Command, args []string, toComplete string) ([]string, cobra.ShellCompDirective) {
	// 	return []string{"test1", "test2"}, cobra.ShellCompDirectiveDefault
	// })
	// childCmd1.Flags().String(
	// 	"test-file", "",
	// 	"just to test")
	// childCmd1.RegisterFlagCompletionFunc("test-file", func(cmd *cobra.Command, args []string, toComplete string) ([]string, cobra.ShellCompDirective) {
	// 	return []string{"test1\tthis is test1", "test2\tthis is test2"}, cobra.ShellCompDirectiveDefault
	// })

	// rootCmd.Flags().String("filter", "", "only show log and txt files")
	// rootCmd.MarkFlagFilename("filter", "log", "txt")

	// rootCmd.Flags().StringArray("array", []string{"allo", "toi"}, "string array flag")

	// childCmd1.PersistentFlags().IntP("requiredPersistent", "p", -1, "required persistent int flag")
	// childCmd1.MarkPersistentFlagRequired("requiredPersistent")
	// childCmd2.Flags().Bool("required", false, "required bool flag")
	// childCmd2.MarkFlagRequired("required")
	// childCmd2.Flags().String("notRequired", "", "not required string flag")
}
```

itskingori:
@marckhouzam A couple of questions, based on this work:

1. Say I have a program `exampletool` that has a command `examplecommand` **with sub-commands** i.e. `command1` & `command2`. If you happen to define `ValidArgs` set to `command1` & `command2` (this isn't entirely necessary and probably isn't a sensible thing to do), you get a duplicate row at the bottom. Is that expected or should it de-duplicate? And should we have the mixed vertical and horizontal listing?
   ```console
   $ exampletool examplecommand
   command1      -- Example description something1
   command2      -- Example description something2
   command1  command2
   ```
2. Say I have a program `exampletool` that has a command `examplecommand` **with no sub-commands** but it accepts args i.e. `arg1` & `arg2`. If you happen to define `ValidArgs` set to `arg1` & `arg2`, you get the single row at the bottom listed horizontally. Is that expected behaviour or should all Zsh auto-completion output be listed vertically as above?
   ```console
   $ exampletool examplecommand
   arg1  arg2
   ```

marckhouzam:
Thanks @itskingori for trying it out.

About #1, this is zsh's behavior when there are descriptions (vertical) and when there are none (horizontal). If you were to add descriptions to your ValidArgs, they would show vertically. E.g., `ValidArgs: []string{"command1\tdescription one", "command2\tanother description"}`.

It may seem surprising that `command1` and `command2` are duplicated, but since they have different descriptions, zsh shows both.
To confirm, my guess is that if you turn off descriptions by using `GenZshCompletionNoDesc()` there would be no duplicates anymore.

#2 is caused by the lack of descriptions in your ValidArgs.

itskingori:
@marckhouzam ...

> Thanks @itskingori for trying it out.
> 
> About point 1, this is zsh's behavior when there are descriptions (vertical) and when there are none (horizontal). ...

Ah, noted. New to Zsh.

> ...If you were to add descriptions to your ValidArgs, they would show vertically. E.g., `ValidArgs: []string{"command1\tdescription one", "command2\tanother description"}`.

This is good to know. I didn't know how to add descriptions to `ValidArgs`.

> It may seem surprising that `command1` and `command2` are duplicated, but since they have different descriptions, zsh shows both.
> To confirm, my guess is that if you turn off descriptions by using `GenZshCompletionNoDesc()` there would be no duplicates anymore.
> 
> For point 2, similarly, it is caused by the lack of descriptions in your ValidArgs.

Understood. It's all good now ... gave this a spin late last night and it's working well. 👌

marckhouzam:
I found a problem with the current solution: it does not support completion from the middle of the command-line.  For example:
```
./testprog [tab][tab] zsh
```
will not show any completions, while the old zsh completion would show all possible sub-commands of `testprog`.

I'm hoping it is easy to fix.  Something along the lines of not passing the entire command-line to the `__complete` command, but only the part before the cursor.  I will work on it.

marckhouzam:
> ```
> ./testprog [tab][tab] zsh
> ```

I've fixed this.
It was also not working in fish so I fixed it too.

I believe everything is ready.

marckhouzam:
I just forced pushed a tiny change because while working on something else I noticed two small bugs for bash Go completion for the new file filtering and sub-directory filtering.

This PR is again ready for review @jharshman, @jpmcb.
The changes may seem pretty big, but almost 2/3 of the changes are tests and documentation.
I'm available to clarify any question you may have.

itskingori:
@jharshman @marckhouzam Just wanted to mention that I've been using this branch for a while and it's been working quite well. And that was before the fix for the tiny issue mentioned in https://github.com/spf13/cobra/pull/1070#issuecomment-631848238.

marckhouzam:
I did another forced push as I realized I had forgot to include the functionality of #288 in Go completions.

This particular change checks for the presence of a local, non-persistent flag on the command-line, and when present, no longer completes sub-command names, since a sub-command doesn't inherit such a flag and would fail.

I then went through the entire `bash-completions.go` file to make sure everything it supported was also supported in Go completions.  I found one more thing missing, which is the proper handling of `ArgAliases`.  I'll be fixing this soon.  After that this PR will be done done 😉 .

marckhouzam:
Last forced-push added support for `ArgAliases` as done for bash completion.  This means that if there are no other valid completions, `ArgAliases` are completed.

@jharshman Sorry for the repeated fixed. This is done done.

marckhouzam:
Murphy's law.  Just after I wrote I was done done, I noticed that a completion without an available description would be shown with a trailing `--` but no description.  I fixed it so that the `--` would not be shown.

That's it. I'm putting down my laptop.

custom_completions.go
edsantiago: Yes oh yes. Showing two duplicate forms of the same flag adds visual clutter and makes it much harder to find what one needs. Thank you for this!

Luap99: I think this is the most user-friendly way, but I think it would be best to leave this as an option to the _Gen[Bash/Zsh/Fish]Completion_ function. With such an option each project could decide what it should be like for them.

stipx: I don't really get the point of having an option to bring duplication (or the form with =) back. Completion should ease the use of the command. I would definitely hide the fact that there is the form with the equals character.

marckhouzam: > I think this is the most user-friendly way, but I think it would be best to leave this as an option to the Gen[Bash/Zsh/Fish]Completion function. With such an option each project could decide what it should be like for them.

But if we add an option, what would it be?  Should the project be able to enable both forms? Or maybe they just want the equal form instead of the non-equal form? So we'd need to support a three choice option.  I get the feeling this may be complicating things too much.

I suggest we start by removing the equal form and see if the community requests anything more.

edsantiago: I haven't thought this out completely, but what if you use the `=` form only for flags that accept/require an option?
```
$ mycommand --<tab>
--binaryflag  --config-file=  --force --name= ...
```
Sometimes a program grows organically, and option names are not as intuitive as if they had all been designed on Day One. This could add immediate visual feedback to help a user figure out, say, "is it `--tmpdir` that takes a path option, or `--use-tmp`"? (contrived example)

Luap99: > Or maybe they just want the equal form instead of the non-equal form? So we'd need to support a three choice option. I get the feeling this may be complicating things too much.

Yes that would result in three choice option.
> I suggest we start by removing the equal form and see if the community requests anything more.

I'm fine with this. I was just pointing out that others might want it differently and having control  is not a bad idea.

marckhouzam: > what if you use the = form only for flags that accept/require an option?

I like that! The user would then immediately know when a flag value is required and when it is not.
My only hesitation is that I'm under the impression most people don't use the `=` form when adding a flag value; they may not like that we "force" them to use it, which comes back to @Luap99's point.

Let's wait until @jharshman weighs in before making any change (also, I don't want to touch the code anymore to facilitate a review).

marckhouzam:
I was thinking about how to make this PR easier to review. It is pretty massive as it stands and I admit I wouldn't like to be on the receiving end.

Would splitting it into smaller increments, each in its own PR be something that would be desirable?

We could gradually build up the go completion logic with targeted PRs that are easier to test and review. They'd have to be tested with Fish completion though. Then at the very end, we would replace the current zsh logic with the new one once everything is in.
Would that make things easier @jharshman, @jpmcb?

jharshman:
> Would splitting it into smaller increments, each in its own PR be something that would be desirable?

@marckhouzam Thank you for the consideration. This is a large but valuable PR but we will work through it. No need to split it up into smaller ones.

austintraver:
This is a feature I passionately desire to have. @marckhouzam @jharshman is there any work I that still needs to be done before this P.R. is merged into master?

jharshman:
@marckhouzam could you rebase against master to resolve the conflicts?

marckhouzam:
> @marckhouzam could you rebase against master to resolve the conflicts?

@jharshman Woops.  The conflict was caused by me with the completion of the help command PR (#1136).
I forgot I would need to rebase after it.

Done.

jpmcb:
I will take a look at this sometime this week - so sorry about falling off the face of the earth 🌎 

Review incoming soon!

custom_completions.go
jharshman: Personally I don't expect to see `=` signs in completions. It's an implicit understanding that you can use `=` signs in assignment but it is not required.

jharshman:
Code looks pretty good to me. Haven't tested it out yet but once I do and things look good this should be ready to merge.

itskingori:
> Code looks pretty good to me. Haven't tested it out yet but once I do and things look good this should be ready to merge.

Not sure if this counts but we resorted to using this branch because we really wanted these changes and we haven't had any issues whatsoever.

jharshman:
@marckhouzam can you squash your commits?

marckhouzam:
> can you squash your commits?

@jharshman I can if you prefer, but I actually worked at keeping each commit self contained and logical, with tests passing, with the goal of providing more information about these changes for future lookup.

But I leave it up to you. In light of this explanation, just let me know if you still prefer to have just one commit and I'll squash.

jharshman:
@marckhouzam Keeping it trimmed down to one commit makes it easier to roll back changes if needed later on.

marckhouzam:
@jharshman The squash is done.

While you guys are knee-deep into completion code, there is a small PR for Fish completion, #1122, that fixes a bug that also needs approval 😄 .  But I will surely need to rebase it if this PR goes in first, but I would do that ASAP.

Thanks for the continued reviews!

marckhouzam:
Thank you @jpmcb and @jharshman for your help on getting this done!

I will rebase #1122 now.

jakrosh:
@marckhouzam I am having trouble getting custom dynamic flag completions for zsh working for my project. The custom flag completions work in bash, but when I hit <tab><tab> after typing in the flag, nothing happens. I went through the code in custom_completions.go, bash_completions.go, and zsh_completions.go, and noticed that the var flagCompletionFunctions is used in bash_completions.go, but not zsh_completions.go. Is that why I can't get custom flag completions for zsh working, or is something else the problem?

marckhouzam:
@jakrosh could you give an example to reproduce the problem?

Also, how did you setup the zsh completion script in your environment?

marckhouzam:
@jakrosh actually, the first step to debug your custom completions is to call the `__complete` command manually.  So let's say the following doesn't work:
```
program --flag <TAB>
```
Then you should run
```
program __complete --flag "" <ENTER>
```
to see if you get the proper completions.

jakrosh:
running 
`program __complete --flag "" <ENTER>` 
gives my custom completions. however, running
`program __complete --flag jen<ENTER>`
does not give only the flag that starts with "jen" as it is described in the docs.
Screen Shot 2020-07-15 at 11 53 12 AM


I set up the zsh completion script with this line:  
`dctl completion zsh > "$ZSH/completions/_dctl"`

marckhouzam:
> `program __complete --flag jen<ENTER>`
> does not give only the flag that starts with "jen"

This is because the function you registered with `RegisterFlagCompletionFunc()` does not filter based on the `toComplete` filter.  But that should not be a problem as zsh filters things itself (at least for me).
If you want to filter yourself, you just need to add something like `if strings.HasPrefix(completion, toComplete) {` in your function in `RegisterFlagCompletionFunc()`

What exactly is the problem you are seeing with zsh compared to bash? 

> I set up the zsh completion script with this line:
> `dctl completion zsh > "$ZSH/completions/_dctl"`

That should be ok.

jakrosh:
Sorry for being unclear, but my problem is when I type 
`dctl -a <TAB><TAB>`
in bash, the custom completion list shows up:
Screen Shot 2020-07-15 at 1 06 43 PM  

but in zsh nothing happens:
Screen Shot 2020-07-15 at 1 16 25 PM

but as you saw in my previous comment, when I use `__complete` the custom completion list does show up in zsh

marckhouzam:
It may be your zsh that is not configured to list all the possible completions where there are more than one.
If you try `ls <TAB>` do you see all the choices of files listed for you?

jakrosh:
yes, I can see the files listed: 
Screen Shot 2020-07-15 at 1 43 43 PM

marckhouzam:
Then you can try to see what the zsh completion script generated by Cobra does. For example:
```
export BASH_COMP_DEBUG_FILE /tmp/zshdebug
dctl -a <TAB>
cat /tmp/zshdebug
```
You'll see a bunch of traces following what the script does.

austintraver:
Thank you for all that hard work @marckhouzam 

I'm really excited for this feature, @jharshman how far out are we from the next release?

brianpursley:
+1 for @austintraver's question, when will this be available in a release?

marckhouzam:
1.1.0 has just been released!
Thanks @jharshman.

For people that will want to start using zsh completion from Cobra be aware that  custom completions will only work for zsh if they are implemented in Go (using `ValidArgsFunction` and `RegisterFlagCompletionFunc()`).  They will not work for the ones using bash scripts injected through `BashCompletionFunction`.

PR #1074, opened 2020-04-01
Add Labeler Actions

xchapter7x:
adds labeling actions
when certain parts of the code are modified by a 
PR then the corresponding label is added

### DIFF
--- a/.github/labeler.yml
+++ b/.github/labeler.yml
@@ -5,4 +5,8 @@
 "area/lib": ./*.go
 
 # changes to the zsh completion
-"area/*sh completion": ./zsh*
+"area/*sh completion":
+  - ./zsh_*
+  - ./shell_*
+  - ./powershell_*
+  - ./bash_*
--- a/.github/labeler.yml
+++ b/.github/labeler.yml
@@ -1,0 +1,8 @@
+# changes to documentation
+"area/documentation": doc/**/*
+
+# changes to the core lib package
+"area/lib": ./*.go
+
+# changes to the zsh completion
+"area/*sh completion": ./zsh*
--- a/.github/workflows/label.yml
+++ b/.github/workflows/label.yml
@@ -1,0 +1,19 @@
+# This workflow will triage pull requests and apply a label based on the
+# paths that are modified in the pull request.
+#
+# To use this workflow, you will need to set up a .github/labeler.yml
+# file with configuration.  For more information, see:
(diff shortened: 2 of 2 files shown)


xchapter7x:
@jharshman

.github/labeler.yml
```diff
@@ -0,0 +1,8 @@
+# changes to documentation
+"area/documentation": doc/**/*
+
+# changes to the core lib package
+"area/lib": ./*.go
+
+# changes to the zsh completion
+"area/*sh completion": ./zsh*
```
jharshman, line 8: Can we match on the bash completion as well?

.github/workflows/label.yml
```diff
@@ -0,0 +1,19 @@
...
+
+jobs:
+  label:
+
+    runs-on: ubuntu-latest
+
+    steps:
+    - uses: actions/labeler@v2
+      with:
+        repo-token: "${{ secrets.GITHUB_TOKEN }}"
```
jharshman, line 19: Is this secret something I need to configure?

.github/labeler.yml
xchapter7x: 👍 
just added the matchers for the other autocompletion shells

.github/workflows/label.yml
xchapter7x: I don't believe you need to do anything no.

jharshman: okay perfect

.github/labeler.yml
```diff
@@ -0,0 +1,12 @@
+# changes to documentation
+"area/documentation": doc/**/*
+
+# changes to the core lib package
+"area/lib": ./*.go
+
```
jpmcb, line 6: What about one for the Cobra Generator itself?

```
# changes to the cobra generator
"area/cobra-generator": cobra/**/*
```

xchapter7x: I didn't see that label in the current set.

Maybe we have a enhancement PR to add more labels and matchers?
@jpmcb

.github/workflows/label.yml
umarcor: FWIW, from https://github.com/spf13/cobra/issues/952#issuecomment-532200393:

> Better integration of permissions. Since Actions is built in GitHub, **it is possible to use secret tokens without explicitly managing/setting them**.

PR #1075, opened 2020-04-01
add support for autolabel stale PR

xchapter7x:
automatically labels stale PRs

### DIFF
--- a/.github/workflows/stale.yml
+++ b/.github/workflows/stale.yml
@@ -1,0 +1,19 @@
+name: Mark stale issues and pull requests
+
+on:
+  schedule:
+  - cron: "0 0 * * *"
+
+jobs:
+  stale:
+
+    runs-on: ubuntu-latest
+
+    steps:
+    - uses: actions/stale@v1
+      with:
+        repo-token: ${{ secrets.GITHUB_TOKEN }}
+        stale-issue-message: 'Stale issue message'
+        stale-pr-message: 'Stale pull request message'
+        stale-issue-label: 'no-issue-activity'
+        stale-pr-label: 'no-pr-activity'
--- a/.github/workflows/stale.yml
+++ b/.github/workflows/stale.yml
@@ -15,5 +15,5 @@ jobs:
         repo-token: ${{ secrets.GITHUB_TOKEN }}
         stale-issue-message: 'Stale issue message'
         stale-pr-message: 'Stale pull request message'
-        stale-issue-label: 'no-issue-activity'
-        stale-pr-label: 'no-pr-activity'
+        stale-issue-label: 'kind/stale'
+        stale-pr-label: 'kind/stale'
--- a/.github/workflows/stale.yml
+++ b/.github/workflows/stale.yml
@@ -17,3 +17,5 @@ jobs:
         stale-pr-message: 'This PR is being marked as stale due to a long period of inactivity'
         stale-issue-label: 'kind/stale'
         stale-pr-label: 'kind/stale'
+        exempt-issue-label: 'kind/stale'
+        exempt-pr-label: 'kind/stale'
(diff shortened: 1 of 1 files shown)


xchapter7x:
@jharshman

.github/workflows/stale.yml
```diff
@@ -0,0 +1,19 @@
...
+
+    runs-on: ubuntu-latest
+
+    steps:
+    - uses: actions/stale@v1
+      with:
+        repo-token: ${{ secrets.GITHUB_TOKEN }}
+        stale-issue-message: 'Stale issue message'
+        stale-pr-message: 'Stale pull request message'
+        stale-issue-label: 'no-issue-activity'
```
jharshman, line 18: We have a label for stale `kind/stale` can that be used here?

Can we change the messages to relay the reason that it is being marked stale?

jharshman:
Will this just apply the stale label or also close the issue?

.github/workflows/stale.yml
xchapter7x: i've updated the label `kind/stale` to apply to both stale issues and PRs. 

I can add messaging around what it means to be stale ie. "This issue/PR is being marked as stale due to a long period of inactivity"
thoughts @jharshman ?

xchapter7x:
> Will this just apply the stale label or also close the issue?

@jharshman 

For now this is just adding labels. We can, I believe, modify that later to auto-close things.

.github/workflows/stale.yml
jharshman: @xchapter7x Yes I think that would be helpful.

jharshman:
> For now this is just adding labels. We can, I believe, modify that later to auto-close things.

Okay great - step in the right direction.  This will give us an ability to prioritize combing through some of the stale issues and either create action items or close them out.

wfernandes:
I had a question about the default behavior and found my answer so I thought I'd share it here:

This github action runs midnight every day to mark issues as stale. By default if the issue/PR has no activity for 60 days it will mark it as stale and if it remains stale for another week it then closes the issue/PR automatically. More info: https://github.com/actions/stale

xchapter7x:
> I had a question about the default behavior and found my answer so I thought I'd share it here:
> 
> This github action runs midnight every day to mark issues as stale. By default if the issue/PR has no activity for 60 days it will mark it as stale and if it remains stale for another week it then closes the issue/PR automatically. More info: https://github.com/actions/stale

@wfernandes thanks for detailing the behavior. I was mistaken when I said it wouldnt close the issues. Great catch. 

I guess the question for @jharshman, is the behavior described by @wfernandes desired?

If not we can modify the config such that issues/prs do not get auto-closed.

.github/workflows/stale.yml
xchapter7x: message is not updated to be more relevant.

jharshman:
@xchapter7x 
> If not we can modify the config such that issues/prs do not get auto-closed.

Yes let's do that so they don't auto-close. There are some issues that I know are stale but are still relevant.

xchapter7x:
> @xchapter7x
> 
> > If not we can modify the config such that issues/prs do not get auto-closed.
> 
> Yes let's do that so they don't auto-close. There are some issues that I know are stale but are still relevant.

@jharshman 
there are 2 general ways. We can add an exempt label, which would allow us to filter the stale items we do not want auto-closed.
or
we can just make nothing get auto-closed via a really long config ie `days-before-close: 999999`

thoughts on which workflow you'd prefer?

xchapter7x:
@jharshman 
where i landed for the implementation is i created an exempt label, which matches the stale label. 
This way, once the issues/prs are labelled stale by the action, they will from then on be ignored by the action and will not be auto-closed. 

moving forward we can simply change or remove the exempt label to support our desired workflow in auto-closing or not.

jharshman:
@xchapter7x sounds like a plan. Lmk if I need to create a new label.

xchapter7x:
> @xchapter7x sounds like a plan. Lmk if I need to create a new label.
@jharshman 
the only label this action uses is `kind/stale`. as long as that exists we should be fine.

Issue #1071, opened 2020-04-01
Replace deprecated SetOutput func with SetOut and SetErr everywhere


xchapter7x:
Follow on for this PR: https://github.com/spf13/cobra/pull/1053

We should remove all usage of deprecated `SetOutput` and replace with `SetOut` & `SetErr`

```
command.go|221 col 4| // SetOutput sets the destination for usage and error messages.
command.go|224 col 19| func (c *Command) SetOutput(output io.Writer) {
command.go|1403 col 11| c.flags.SetOutput(c.flagErrorBuf)
command.go|1431 col 12| c.lflags.SetOutput(c.flagErrorBuf)
command.go|1457 col 12| c.iflags.SetOutput(c.flagErrorBuf)
command.go|1485 col 12| c.pflags.SetOutput(c.flagErrorBuf)
command.go|1495 col 10| c.flags.SetOutput(c.flagErrorBuf)
command.go|1497 col 11| c.pflags.SetOutput(c.flagErrorBuf)
command.go|1614 col 19| c.parentsPflags.SetOutput(c.flagErrorBuf)

doc/md_docs.go|31 col 8| flags.SetOutput(buf)
doc/md_docs.go|39 col 14| parentFlags.SetOutput(buf)

doc/rest_docs.go|31 col 8| flags.SetOutput(buf)
doc/rest_docs.go|40 col 14| parentFlags.SetOutput(buf)
```

PR #1078, opened 2020-04-02
Remove/replace SetOutput on Command - deprecated

jpmcb:
Addresses https://github.com/spf13/cobra/issues/1071

This removes the `SetOutput` method on `*Command` as it is deprecated and replaces those methods with `SetOut` and `SetErr`. 

I didn't touch any of the `SetOutput` methods on `*FlagSet` as that appears to be apart of the `github.com/spf13/pflag` library. Do we think it's valuable to change that library as well?

If we don't want to entirely remove the function just yet, let me know!

### DIFF
--- a/command.go
+++ b/command.go
@@ -218,13 +218,6 @@ func (c *Command) SetArgs(a []string) {
 	c.args = a
 }
 
-// SetOutput sets the destination for usage and error messages.
-// If output is nil, os.Stderr is used.
-// Deprecated: Use SetOut and/or SetErr instead
-func (c *Command) SetOutput(output io.Writer) {
-	c.outWriter = output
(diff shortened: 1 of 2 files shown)


jharshman:
> If we don't want to entirely remove the function just yet, let me know!
@jpmcb yeah let's not remove it entirely. This is an exported method and we have no way of knowing the extent to which it is used in other projects.

command.go
```diff
@@ -218,14 +218,6 @@ func (c *Command) SetArgs(a []string) {
 	c.args = a
 }
 
-// SetOutput sets the destination for usage and error messages.
```
jharshman, line 221: Let's hold off on removing this.

PR #1082, opened 2020-04-06
API (urgent): Rename BashCompDirectives to ShellCompDirectives

marckhouzam:
Rename `BashCompDirectives` to `ShellCompDirectives`

When answering questions in #1070 about Zsh completion, I realized that the naming for the Completion Directives could be improved.  Since those completion directives will be used for different shells, and that these names will be consumed by users, it is more appropriate not to include `bash` in the name.

If this is worth doing, it should be done right away before the old names get locked as a released API.  Assuming it is still ok to make this change.

@jharshman what do you think?

### DIFF
--- a/bash_completions.go
+++ b/bash_completions.go
@@ -344,7 +344,7 @@ __%[1]s_handle_word()
     __%[1]s_handle_word
 }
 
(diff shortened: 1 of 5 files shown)


marckhouzam:
Thank. you @jharshman !

PR #1089, opened 2020-04-12
env variable available for git outh key

Salamandastron1:
Hey!

Auto deploy configuration for Travis complete 🖖 

There will be a little bit of setup necessary in to use a github oauth token. I'll link the docs below. The setup should all be here though. It will only deploy automagically on tags, which seem to be deliberately used in the repo so it should be muy bien.

https://docs.travis-ci.com/user/deployment/releases/#authenticating-with-an-oauth-token

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -25,5 +25,15 @@ matrix:
       go: 1.13.x
       script: make cobra_generator
 
-script: 
- - make test
+script:
+  - make test
+
+deploy:
+  provider: releases
+  api_key:
+    secure: $YOUR_API_KEY_ENCRYPTED
+  file_glob: true
+  file: ./bin/*
+  skip_cleanup: true
+  on:
+    tags: true


umarcor:
I suggest using GitHub Actions instead, since the default GITHUB_TOKEN does already allow to upload artifacts and/or creating releases.

Ref #952 #968

Salamandastron1:
Travis suggests making a machine user for this to work, and generating api key for via the information in the docs I linked. All for either implementation of Travis or Github actions but I was asked to keep this in scope of the current CI/CD method.

umarcor:
> Travis suggests making a machine user for this to work, and generating api key for via the information in the docs I linked. All for either implementation of Travis or Github actions but I was asked to keep this in scope of the current CI/CD method.

Yes, Travis requires explicitly generating an API key/token, and because of limited configurability of scopes, it is suggested to create a machine user. AFAIK, doing so requires to create an e-mail account for the machine user. Conversely, the GITHUB_TOKEN available in GHA by default does already have a limited scope, so creation of a machine user or a custom token is not required. Anyway, if you are constrained to use Travis, there is little you can do...

PR #1094, opened 2020-04-20
Metal Stack CLI is using cobra as well

majst01:
We appreciate your library and want thank by pointing to our usage.


jharshman:
Not sure why the CI build failed... probably a fluke; re-running now.

jharshman:
yep, looks like it passed now.

PR #1095, opened 2020-04-20
Ignore required flags when DisableFlagParsing is true

### LINKED ISSUE #1093: Go completion broken when MarkPersistentRequiredFlag() called on root command
While working on supporting required flags for Go completions (Fish and Zsh v2) I hit the following bug.

If a persistent flag on the root command is marked as required, e.g., 
```
rootCmd.PersistentFlags().String("flag", "", "required persistent")
rootCmd.MarkPersistentFlagRequired("flag")
```
it breaks the `__complete` command.

This is because the `__complete` command, being a child of the root command, inherits the required persistent flag, and since that flag is not specified when calling `__complete`, the `__complete` command fails to execute.
```
$ ./testprog __complete ""
Error: required flag(s) "flag" not set
```
I think it is and unlikely case though; if the root command would have a required persistent flag, it would imply that that flag would need to be specified in every single case of calling the program; I have trouble imagining a use case for that.

Still, it should probably be fixed.
I don't have a solution right now, but I will be investigating if defining a `PreRun` on `__complete`  can allow us to turn off required flags for `__complete`.

In the unlikely case of this bug, the impacts are:
- bash: Go custom completions (`ValidArgsFunction` and `RegisterFlagCompletionFunc()`) would not work (but everything else would)
- fish: all completions would not work

marckhouzam:
Fixes #1093

When a command requests to DisableFlagParsing, it should not fail due to a missing required flag. 
In fact, such a check would always fail since flags weren't parsed!

To reproduce the problem, one can use this program:
```
package main

import (
	"fmt"

	"github.com/spf13/cobra"
)

var rootCmd = &cobra.Command{
	Use: "testprog",
		fmt.Println("rootCmd called")
	},
}
	Use:                "child",
	DisableFlagParsing: true,
		fmt.Println("childCmd2 called")
	},
}

func main() {
	rootCmd.PersistentFlags().Bool("flag", false, "required persistent")
	// Setting a persistent flag as a required flag
	// breaks sub-commands that set DisableFlagParsing==true
	rootCmd.MarkPersistentFlagRequired("flag")

	rootCmd.AddCommand(childCmd)
	rootCmd.Execute()
}
```
and then test as follows:
```
$ ./testprog child --flag
Error: required flag(s) "flag" not set
```
As shown, even when the flag is specified, the command fails because flag parsing is disabled.

I've added a test to make sure this does not break again.

### DIFF
--- a/command.go
+++ b/command.go
@@ -979,6 +979,10 @@ func (c *Command) ValidateArgs(args []string) error {
 }
 
 func (c *Command) validateRequiredFlags() error {
+	if c.DisableFlagParsing {
+		return nil
+	}
+
 	flags := c.Flags()
 	missingFlagNames := []string{}
 	flags.VisitAll(func(pflag *flag.Flag) {
--- a/command_test.go
+++ b/command_test.go
@@ -785,6 +785,37 @@ func TestPersistentRequiredFlags(t *testing.T) {
 	}
 }
 
+func TestPersistentRequiredFlagsWithDisableFlagParsing(t *testing.T) {
+	// Make sure a required persistent flag does not break
+	// commands that disable flag parsing
+
+	parent := &Command{Use: "parent", Run: emptyRun}
+	parent.PersistentFlags().Bool("foo", false, "")
+	flag := parent.PersistentFlags().Lookup("foo")
+	parent.MarkPersistentFlagRequired("foo")
+
+	child := &Command{Use: "child", Run: emptyRun}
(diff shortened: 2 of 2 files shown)


marckhouzam:
This is a small one that is required to move forward with #1070 (zsh completion).
If someone can have a look, it's 3 lines + tests.
/cc @jharshman @jpmcb

Issue #1092, opened 2020-04-15
[Bug][Contributor workflow] Github actions labeler doesn't work for forks


jpmcb:
I've noticed that the labeler Github action isn't working. [Here's an example](https://github.com/spf13/cobra/pull/1044/checks?check_run_id=583192945)

It looks like the `actions/labeler` Github action won't work with PRs opened from forks (for security reasons): https://github.com/actions/labeler/issues/12

And looks like this was recently clarified: https://github.com/actions/labeler/commit/2b75d22987baa8b9fb023106fe3a4d90648b7104

A few workarounds:
- https://github.com/kaxil/boring-cyborg in use by [apache airflow](https://github.com/apache/airflow/blob/master/.github/boring-cyborg.yml)
- https://github.com/marketplace/actions/periodic-labeler in use by [cloud alchemy](https://github.com/cloudalchemy/ansible-prometheus/blob/master/.github/workflows/labeler.yml)

@xchapter7x I know you looked at the labeler previously in https://github.com/spf13/cobra/pull/1074 - Any thoughts here?

xchapter7x:
seems like the `periodic-labeler` would wrap what we have now, and make it work as intended.

the `boring-cyborg` might be better long term, as it seems to have more logic and features. 

id maybe implement `periodic-labeler` as a bugfix, and then explore future enhancements to the feature set with things like `boring-cyborg` or anything else that might be available in the marketplace.

xchapter7x:
This PR should address this bug: https://github.com/spf13/cobra/pull/1097

PR #1097, opened 2020-04-21
replace labeler action with periodic-labeler

xchapter7x:
this replaces the default labeler as it
does not label PRs from forks properly.

with periodic-labeler it should apply labels on
a cron to any PR and resolve the below bug:

https://github.com/spf13/cobra/issues/1092


xchapter7x:
@jharshman @jpmcb 

Thoughts on this PR as a fix for https://github.com/spf13/cobra/issues/1092 ?

jharshman:
Not sure what I think about this. It's a Github Action created by a 3rd party. Does this give anyone else pause?

xchapter7x:
fair point, @jharshman .

It is open source, so we can see exactly what it's doing here: https://github.com/paulfantom/periodic-labeler/blob/master/main.go
Doesn't seem to be leaking anything, just calling the Github API.

Also the container image used is `distroless` maintained by Google: https://github.com/GoogleContainerTools/distroless

We are pinning to a specific version of this Action, so it changing out from under us and compromising secrets somehow seems unlikely. 

I guess we could fork it and maintain our own version of the Action, but that might be more burden than the risk is worth. 

Your call :)

jharshman:
> We are pinning to a specific version of this Action, so it changing out from under us and compromising secrets somehow seems unlikely. 

It's easy for a pinned version to get updated at some point in the future w/o undergoing another code review.

Can you add a comment in the action right above the image? It should indicate that if the version is to be updated, a code review of the 3rd party software should be performed to ensure that nothing malicious is going to be run.

xchapter7x:
> > We are pinning to a specific version of this Action, so it changing out from under us and compromising secrets somehow seems unlikely.
> 
> It's easy for a pinned version to get updated at some point in the future w/o undergoing another code review.
> 
> Can you add a comment in the action right above the image? It should indicate that if the version is to be updated, a code review of the 3rd party software should be performed to ensure that nothing malicious is going to be run.

latest commit adds the requested comment. @jharshman 

Let me know if there are any other concerns we should address :)

jharshman:
There's still a flaw in this that makes me want to fork it.
There is nothing stopping a person from overwriting a tag. Let me know if I'm being too paranoid about this.
@xchapter7x @jpmcb

xchapter7x:
> There's still a flaw in this that makes me want to fork it.
> There is nothing stopping a person from overwriting a tag. Let me know if I'm being too paranoid about this.
> @xchapter7x @jpmcb

Agreed, the author of this action could retag and swap out the pinned version from under us.

The same, however, can be said for every go module or open source import. 

Granted in go there is a `go.sum` to check if a modification happened somewhere in the chain, which guards against tag swapping. Which, as you point out, we're lacking here. 

There is another aspect to this, in that its on the github marketplace. We can report wrong doing. We can see the code. We can subscribe and be notified when the code changes in the repo we're referencing. We can also see who the maintainers are, and in this case they seem to be associated with reputable companies and contributors to reputable OSS projects. 

I see the risk as being rather low. 

If tag swapping is a really high concern, we could consider pinning to a specific commit ref. That should work the same as pinning to a tag does, but i don't believe one can as easily swap that out from under us. 

i'm guided by what you're comfortable with @jharshman :)

jharshman:
@xchapter7x alrighty I think I'm good with this.
:+1:

Issue #741, opened 2018-09-06
Timestamp still shows in man documentation


speedingdaemon:
I thought that this issue was resolved with https://github.com/spf13/cobra/issues/142
But even after specifying in my root cmd:
        DisableAutoGenTag: true,

I see this text in my man documentation:
*Auto generated by spf13/cobra on 6-Sep-2018*

markdown and ReST documentation don't have this text. 
Am I doing something wrong?

Here is my code:
var rootCmd = &cobra.Command{
        Use:               "hello",
        Short:             "test",
        Long:              "my test",
        DisableAutoGenTag: true,
}

func genManDocs() {
        header := &doc.GenManHeader{
                Title:   "MINE",
                Section: "3",
        }
        err := doc.GenManTree(rootCmd, header, "./")

        if err != nil {
                fmt.Println(err)
        }
}

Skarlso:
Yep, second that. It's still in there.

Simran-B:
I want to add that it's also showing in the Markdown documentation, so likely a general issue of `DisableAutoGenTag` not working properly (not passed on correctly?)

kamalgrover:
Guys, Is there a workarround for the same ?

Skarlso:
Not that I know off, sorry. :/

kamalgrover:
I used the "source" member to overcome this.

kamalgrover:
Here is the code snippet.
`func main() {

	header := &doc.GenManHeader{
		Title:   "shorthand",
		Section: "1",
		Manual:  "My Product Manual",
		Source: "Copyright (C) 2020 All Rights Reserved.",

	}
	//Disables the autogenerated comment at the end of the man page.
	cmd.RootCmd.DisableAutoGenTag = true
	err := doc.GenManTree(cmd.RootCmd, header, "doc/")
	if err != nil {
		log.Fatal(err)
	}
}`

jpmcb:
@jharshman I did some investigation on this and seems that `cmd.DisableAutoGenTag` will only skip placing the auto gen tag in the body while the header "source" will get auto populated with the auto gen tag if nothing is specified.

I was able to work around this by specifying something other than an empty string:
```
header := &doc.GenManHeader{
	Title: "someTitle",
	Section: "1",
	Source: " ",
}
```

[...]

PR #1104, opened 2020-04-25
Man pages won't have auto gen tag when option is disabled

jpmcb:
This addresses https://github.com/spf13/cobra/issues/741

When `cmd.DisableAutoGenTag = true`, the man pages will no longer contain `Auto generated by spf13/cobra`

Let me know if this needs anything else!


jpmcb:
I've added documentation for this in a new file `docs/README.md` which will also be viewable in Github when a user views the `docs/` folder in their browser. 

I've also added a changelog in a new file `CHANGELOG.md` which includes the v1.0.0 changes and captures this PRs new behavior in a `Pending` section. This should enable the maintainers to easily track changes, make releases, and request that contributors document their changes. 

If the `CHANGELOG.md` file is out of scope for this PR, I can remove it, let me know!

Issue #1107, opened 2020-04-27
cobra init cmd documentation seems to be incorrrect


wfernandes:
As I was playing around with `cobra init` I noticed the help documentation seems to be at odds with the [cobra/README](https://github.com/spf13/cobra/blob/11ba63fc3bc37818c471fc7912783e6909da4645/cobra/README.md#cobra-init) documentation.

The `cobra init --help` prints out the following:
```
  * If a name is provided, it will be created in the current directory;
  * If no name is provided, the current directory will be assumed;
  * If a relative path is provided, it will be created inside $GOPATH
    (e.g. github.com/spf13/hugo);
  * If an absolute path is provided, it will be created;
  * If the directory already exists but is empty, it will be used.

Init will not use an existing directory with contents.
```

However, as I was using the command and reading through the documentation, the following statements seem to be incorrect. 
```
  * If a relative path is provided, it will be created inside $GOPATH
    (e.g. github.com/spf13/hugo);
  * If an absolute path is provided, it will be created;

Init will not use an existing directory with contents.
```

Upon inspection of the code, we seem to be prepending the current working directory to the argument passed in. 
https://github.com/spf13/cobra/blob/11ba63fc3bc37818c471fc7912783e6909da4645/cobra/cmd/init.go#L46-L58

So I guess the question here is: **what is the expected behavior? Is to initialize a project in the current working directory? Or should we be supporting relative paths in GOPATH and absolute paths?**

jharshman:
I think the text in the readme does a good job depicting it's behavior:
> ... You can either run it in your current application directory or you can specify a relative path to an existing project. If the directory does not exist, it will be created for you.

However, as you have brought up, these comments should be fixed. Namely the comment that `init` will not use an existing directory with contents. As it is, init will overwrite conflicting contents in a directory.
>   * If a relative path is provided, it will be created inside $GOPATH
    (e.g. github.com/spf13/hugo);
>  * If an absolute path is provided, it will be created;
>
> Init will not use an existing directory with contents.

PR #1108, opened 2020-04-27
Fix cobra init cmd help documentation

wfernandes:
Created this PR to address current behavior. That is all projects will be initialized within the current directory. I also added tests for other cases where we do not expect projects to be created
for absolute paths and relative paths in GOPATH.

Fixes https://github.com/spf13/cobra/issues/1107

### DIFF
--- a/cobra/cmd/init.go
+++ b/cobra/cmd/init.go
@@ -32,42 +32,17 @@ var (
 		Long: `Initialize (cobra init) will create a new application, with a license
 and the appropriate structure for a Cobra-based CLI application.
 
-  * If a name is provided, it will be created in the current directory;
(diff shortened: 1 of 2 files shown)


cobra/cmd/init.go
```diff
@@ -76,3 +51,31 @@ func init() {
 	initCmd.Flags().StringVar(&pkgName, "pkg-name", "", "fully qualified pkg name")
 	initCmd.MarkFlagRequired("pkg-name")
 }
+
+func initializeProject(args []string) (string, error) {
+	wd, err := os.Getwd()
+	if err != nil {
+		er(err)
```
jharshman, line 58: @wfernandes perhaps this should return the error.

PR #1110, opened 2020-05-07
Fix PrintErrln and PrintErrf

ian-howell:
This makes PrintErrln and PrintErrf print their output to the command's
Err output as documented.


ian-howell:
Looks like this is already handled in #894

Issue #1109, opened 2020-04-27
Command context is not properly passed on to sub-commands


lewisparma:
When using 'ExecuteContext' on the parent command, the context is not properly passed on to sub commands. It's passed on only the first time, then on the following invocations the first passed value is used. The culprit seems to be this section from commands.go:

`
	// We have to pass global context to children command
	// if context is present on the parent command.
	if cmd.ctx == nil {
		cmd.ctx = c.ctx
	}
`

the parent context should always be passed on to the child command, not just when the child context is null.

utku-caglayan:
Here is a workaround for anyone struggling with this:
```
type CobraMutableCtx struct {
	Internal context.Context
}
func (c *CobraMutableCtx) Deadline() (deadline time.Time, ok bool) {
	return c.Internal.Deadline()
}
func (c *CobraMutableCtx) Done() <-chan struct{} {
	return c.Internal.Done()
}
func (c *CobraMutableCtx) Err() error {
	return c.Internal.Err()
}
func (c *CobraMutableCtx) Value(key interface{}) interface{} {
	return c.Internal.Value(key)
}
...
var cobraCtx CobraMutableCtx
cobraCtx.Internal = context.TODO() // your context
cobraCmd.ExecuteContext(&cobraCtx) // this will be passed to subcommands
cobraCtx.Internal, cmdCancel = context.WithCancel(ctx) // your context
cmdCancel()
cobraCmd.ExecuteContext(&cobraCtx) // this time subcommands will use cancelled context
```

johnSchnake:
Is there any counterargument for the context not being passed to the child when non-nil?

Unfortunately it seems like it would be a breaking change for anyone who already worked around it or began to depend on this code flow.

Out of curiosity, whats the situation with the multiple invocations? Is it tests that are causing the problem? I wonder if that was part of the reason why the context wasn't to be overwritten: so that you could specify it manually to avoid the inheritance.

marckhouzam:
Very similar to #1469

PR #1118, opened 2020-05-19
Implement SetContext() for passing values between hooks

edufschmidt:
- Implement SetContext() function for passing context between hooks, as proposed by @flimzy  on issue #563. A test routine was too implemented, and validates that by using the newly introduced function together with persistent hooks, it is possible to pass values between hooks, and between parent and child commands.

- Remove condition for passing down context to child command, as pointed out in issue #1109.

### DIFF
--- a/command.go
+++ b/command.go
@@ -949,9 +949,7 @@ func (c *Command) ExecuteC() (cmd *Command, err error) {
 
 	// We have to pass global context to children command
 	// if context is present on the parent command.
-	if cmd.ctx == nil {
-		cmd.ctx = c.ctx
-	}
+	cmd.ctx = c.ctx
 
 	err = cmd.execute(flags)
 	if err != nil {
--- a/command.go
+++ b/command.go
@@ -216,6 +216,12 @@ func (c *Command) Context() context.Context {
 	return c.ctx
 }
 
+// SetContext replaces the underlying command context so that parent
+// commands can pass down values to their subcommands.
+func (c *Command) SetContext(ctx context.Context) {
+	c.ctx = ctx
+}
+
 // SetArgs sets arguments for the command. It is set to os.Args[1:] by default, if desired, can be overridden
 // particularly useful when testing.
 func (c *Command) SetArgs(a []string) {
--- a/command_test.go
+++ b/command_test.go
@@ -148,6 +148,38 @@ func TestSubcommandExecuteC(t *testing.T) {
 	}
 }
 
+func TestSetContext(t *testing.T) {
+	ctx := context.TODO()
+
+	aKey := "akey"
+	aVal := "aval"
+
+	parentRun := func(cmd *Command, args []string) {
+		ctx := context.WithValue(cmd.Context(), aKey, aVal)
+		cmd.SetContext(ctx)
+	}
+
+	childRun := func(cmd *Command, args []string) {
+		if val := cmd.Context().Value(aKey); val != aVal {
+			t.Errorf(`Context attribute not found in child command. Expected: "%+v". Have: "%+v"`, aVal, val)
+		}
+	}
+
+	rootCmd := &Command{Use: "root", Run: parentRun, PersistentPreRun: parentRun}
+	childCmd := &Command{Use: "child", Run: childRun}
+
+	rootCmd.AddCommand(childCmd)
+
+	if _, err := executeCommandWithContext(ctx, rootCmd, ""); err != nil {
+		t.Errorf("Root command must not fail: %+v", err)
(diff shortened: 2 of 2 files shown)


Divya063:
When this PR is going to be merged?

edufschmidt:
Judging by the number of open PRs, and the time it's been taking for the maintainers to merge them, I'm guessing this might take a while.

aaronc:
Would be great to see this get merged!

jackzampolin:
Would really love to see this merged cc @eparis !

ori2694:
I will also be glad to see this merged :)

edufschmidt:
lol

AWoelfel:
I just stumbeled on this and would love to see this merged!

command_test.go
```diff
@@ -148,6 +148,38 @@ func TestSubcommandExecuteC(t *testing.T) {
 	}
 }
 
+func TestSetContext(t *testing.T) {
```
jpmcb, line 151: Thanks for adding a test case for this!! 👍

command.go
```diff
@@ -943,9 +949,7 @@ func (c *Command) ExecuteC() (cmd *Command, err error) {
 
 	// We have to pass global context to children command
 	// if context is present on the parent command.
-	if cmd.ctx == nil {
-		cmd.ctx = c.ctx
-	}
+	cmd.ctx = c.ctx
```
jpmcb, line 952: Why was the `if` removed here?

I can imagine, even with the addition of the new `SetContext` API, there still being the need for this if catch. The way I read this code now is that _on every execution,_ a passed in command will _always_ have it's context set.

This could even be a breaking change if during command execution, the users rely on the `cmd.ctx` not trickling down to subsequent child commands. This change makes the assumption that _all_ sub-command executions should have the same context as the parent.

command.go
```diff
@@ -216,6 +216,12 @@ func (c *Command) Context() context.Context {
 	return c.ctx
 }
 
+// SetContext replaces the underlying command context so that parent
+// commands can pass down values to their subcommands.
+func (c *Command) SetContext(ctx context.Context) {
```
jpmcb, line 221: How is this any different from `ExecuteContext`? With using `ExecuteContext`, a user could pass in the necessary parent context, and then it is executed on the child sub-command.

edufschmidt: If I'm not mistaken (it's been a while since this PR was submitted), by using `ExecuteContext` you're not able to set the context in the scope of a command so that nested commands have access to the new, modified context. In other words, the context is only passed to the root command, and then it is simply forwarded without modification to any subcommands. If you want parent commands to be able to act as middleware to their children, you need to have a structure that can be modified within the root context, which might not be desirable. That's where `SetContext` comes in handy. It allows to modify the context at will.

command_test.go
edufschmidt: :+1:

command.go
edufschmidt: Please see  #1109 for the motivation behind this change.

command.go
stapelberg: @jpmcb Any update? :) Thanks!

dfreilich:
Any update on this PR? I was recently looking into this problem, and this would help my use case as well (ensuring that contexts are passed appropriately to deeply nested subcommands)

congon4tor:
I believe this would  also be useful for my use case. I want to initialise a grpc client that is common to all sub commands but some values to initialise it come from flags in the root command. The way I'm trying to achieve this is by having a `PersistentPreRun` on the root which initialises the grpc client and stores it in the context (with `SetContext()`). Then access the client from the sub commands with `cmd.Context()`.

davidmdm:
I too would really like this to get merged soon. it would allow for parent pre-runs to set values that child commands would need in a way that would really make my api less verbose.

alexykot:
Hmm, seems like there were multiple attempts to fix this, but all PRs are either closed or still dangling. E.g.  #1517, #1551. 

I didn't look into details of competing solutions to have a preference here, but would be great to merge either of them to have this feature available.

stapelberg:
Thanks very much! I’m still interested in having this functionality, so this is great news :)

marckhouzam:
The ability to set a context has just been merged thanks to #1551.  That PR, although opened more recently than this one, was limited to adding the `SetContext()` function, so we preferred that one.

If you are still interested in addressing #1109, which I don't fully grasp all the possible impacts with respect to backwards-compatibility, please open a new PR just for that issue and we will try to evaluate those impacts.

Thanks for you patience.

PR #1123, opened 2020-05-25
Include EnablePersistentHookOverride option to control Persistent*Run override behavior

bartdeboer:
Currently Cobra will only run the last Persistent*Run functions that are defined within the command chain. This introduces an EnablePersistentHookOverride option which allows to run all the functions throughout the command chain.

Resolves: #252, #219
Related PR's: #714, #220


bartdeboer:
PR #1142 is perhaps a better approach

PR #1135, opened 2020-06-12
Add example for making persistent flags required

ademuri:
I just started using Cobra, and was confused as to why MarkFlagRequired didn't seem to be working. It was because I was using a persistent flag, and called MarkFlagRequired instead of MarkPersistentFlagRequired. I think this change makes this more clear.


alessio:
Good addition to the doc :+1:

alessio:
CC'ing @jharshman @eparis @jpmcb

jpmcb (approved):
Thanks for contributing!

PR #1136, opened 2020-06-14
Add completion for help command

### LINKED ISSUE #1000: Bash completions shouldn't exclude `help` command

marckhouzam:
Supersedes #1065.
Fixes #1000 

This PR does two things:
 
1) adds completion *of* the `help` command so that
```
<program> [tab][tab]
```
will include `help` in the completion choices.  This is done by the first commit provided by @voithos taken from #1065.

2) adds completion *for* the `help` command using `ValidArgsFunction` so that  
```
<program> help [tab][tab]
```
will show all sub-commands (including `help` since `<program> help help` is a valid command), and
```
<program> help <subCmd1> [tab][tab]
```
will show all sub-commands of `<subCmd1>`.  And so forth down the command tree.

### DIFF
--- a/bash_completions.go
+++ b/bash_completions.go
@@ -389,7 +389,7 @@ fi
 func writeCommands(buf *bytes.Buffer, cmd *Command) {
 	buf.WriteString("    commands=()\n")
 	for _, c := range cmd.Commands() {
-		if !c.IsAvailableCommand() || c == cmd.helpCommand {
+		if !c.IsAvailableCommand() && c != cmd.helpCommand {
 			continue
 		}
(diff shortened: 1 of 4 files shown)


marckhouzam:
Thanks @jharshman!

PR #1142, opened 2020-06-22
Extend Persistent*Run behavior to allow multiple hooks throughout the execution chain

bartdeboer:
Supersedes: PR #1123 
Resolves: #252, #219
Related PR's: #714, #220

* Introduces On*Run() methods to register hooks
* Introduces OnPersistent*Run() methods to register hooks that are always executed (childs not overriding parents).
* Allows to register multiple *Run hooks
* Keeps current Persistent*Run behavior unaffected (childs still override parents when set as a property)
* Introduces EnablePersistentRunOverride option to control if child Persistent*Run hooks should override their parent's hooks

### DIFF
--- a/cobra.go
+++ b/cobra.go
@@ -39,6 +39,9 @@ var templateFuncs = template.FuncMap{
 
 var initializers []func()
 
+// EnablePersistentRunOverride ensures Persistent*Run* functions in childs override their parents
+var EnablePersistentRunOverride = true
+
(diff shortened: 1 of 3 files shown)


bartdeboer:
Alternatively I've created a package [Cobrahooks](https://github.com/bartdeboer/cobrahooks) that attempts to introduce the functionality from this PR as close as possible.

PR #1144, opened 2020-06-25
Add customizable suggestions template

jbrunton:
I was looking into how to style suggestions and that didn't seem possible (unless I missed something?) so this PR adds a simple way to replace the suggestions template, similar to SetUsageTemplate() and related methods.

I still need to update the docs, but I wanted to get feedback on the feature and implementation first.

### DIFF
--- a/command.go
+++ b/command.go
@@ -546,8 +546,7 @@ func (c *Command) SuggestionsTemplate() string {
 	return `
 
 Did you mean this?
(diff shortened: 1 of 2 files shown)


jbrunton:
I realized it may make more sense to implement a SetUnknownCommandErrorFunc() method a bit like the SetFlagErrorFunc() mechanism. This would allow greater customization, given the SuggestionsFor() is already part of the command API.

PR #1145, opened 2020-06-25
Add UnknownCommandErrorFunc API to override default unknown command behavior

jbrunton:
I wanted to customize the appearance of the unknown command output, but that didn't seem possible (unless I missed something?) so this PR provides a way to override it.


umarcor:
Ref #842.

PR #1157, opened 2020-07-11
fix(args): Correct `legacyArgs` logical errors

### LINKED ISSUE #1156: return non-zero exit code upon non-runnable subcommand
```
➜  ~ helm notfoundsubcommand
Error: unknown command "notfoundsubcommand" for "helm"
Run 'helm --help' for usage.
➜  ~ echo $?
1
```

donggangcj:
Close #1156
Signed-off-by: Dong Gang <dong.gang@daocloud.io>

### DIFF
--- a/args.go
+++ b/args.go
@@ -9,16 +9,15 @@ type PositionalArgs func(cmd *Command, args []string) error
 
 // Legacy arg validation has the following behaviour:
 // - root commands with no subcommands can take arbitrary arguments
-// - root commands with subcommands will do subcommand validity checking
-// - subcommands will always accept arbitrary arguments
+// - commands with subcommands cannot take any arguments
 func legacyArgs(cmd *Command, args []string) error {
 	// no subcommand, always take args
 	if !cmd.HasSubCommands() {
 		return nil
 	}
 
-	// root command with subcommands, do subcommand checking.
-	if !cmd.HasParent() && len(args) > 0 {
+	// root command with subcommands, the args should be empty.
+	if cmd.HasSubCommands() && len(args) > 0 {
 		return fmt.Errorf("unknown command %q for %q%s", args[0], cmd.CommandPath(), cmd.findSuggestions(args[0]))
 	}
 	return nil


donggangcj:
@marckhouzam I know you've done a lot of work on helm and could you please this pr if you are free.😀

umarcor:
Ref #842.

marckhouzam:
Thanks @donggangcj.  When I tried this fix with helm, it didn't address the problem described in #1156.

Could you provide a test program that shows the problem before the PR, and that the problem is resolved after the PR?

As for helm, this PR touches the `legacyArgs` function, however, notice that `legacyArgs` is only called when `command.Args == nil`:
https://github.com/spf13/cobra/blob/675ae5f5a98cc705a6d54f4c487ab81230604137/command.go#L627-L630

However, `helm` always uses `command.Args`, except for its root command, so I'm under the impression a different approach will be required to fix this problem.

donggangcj:
@marckhouzam My bad! I've missed important part in Pre commit.  I ran a simple test locally and the result is here:
Before the PR:
```
➜  cobra git:(fix-exit-code-error) helm repo sadd foo https://foo/bar 

This command consists of multiple subcommands to interact with chart repositories.

It can be used to add, remove, list, and index chart repositories.

Usage:
  helm repo [command]

Available Commands:
....

➜  cobra git:(fix-exit-code-error) echo $?
0

```

After the PR:
```
# NOTE: It will print the Command.Args error string instead of help message
➜  bin git:(temp/validate) ✗ ./helm repo sadd foo https://foo/bar
Error: "helm repo" accepts no arguments

Usage:  helm repo add|remove|list|index|update [ARGS] [flags]

➜  bin git:(temp/validate) ✗ echo $?
1
```

The function `legacyArgs` is only called when `Comand.Args == nil` and the pre commit will not work if `Command.Args != nil`.  helm commands that contains subcommands like `get,repo`   
I continue to view the code logic 
```
        // part1
	if !c.Runnable() {
		return flag.ErrHelp
	}

	c.preRun()

       // part2
        argWoFlags := c.Flags().Args()
	if c.DisableFlagParsing {
		argWoFlags = a
	}

	if err := c.ValidateArgs(argWoFlags); err != nil {
		return err
	}
```
The order of part1 and part2 

Finally,I think a command shouldn't accept any args if it has subcommand，or it's easy to confuse `args` and `subcommand` string.😀

donggangcj:
> Ref #842.

Thanks!

marckhouzam:
> I read the source code of cli which is using cobra but it doesn't appear this problem.

Actually, it looks like they hack their way around this problem:
https://github.com/cli/cli/blob/1ddb4d76a750748012c91204740d49991ad2f968/cmd/gh/main.go#L59
and
https://github.com/cli/cli/blob/1ddb4d76a750748012c91204740d49991ad2f968/command/help.go#L69-L73

PR #1159, opened 2020-07-11
Allow specifying licenses using their SPDX identifier

yann-soubeyrand:


jpmcb (approved):
Seems reasonable! Thanks for this one - and thanks for the patience!

PR #1162, opened 2020-07-12
Handle linebreaks in custom completions.

Luap99:
If a command/flag description contains a linebreak than the shell completion script will interpret this as new command/flag.

To fix this we should only use the first line from the description in the output.

e.g. `cmd.Flags().String("flag", "", "Description for flag\nlonger description")`

### DIFF
--- a/custom_completions.go
+++ b/custom_completions.go
@@ -132,6 +132,12 @@ func (c *Command) initCompleteCmd(args []string) {
 					comp = strings.Split(comp, "\t")[0]
 				}
 
+				// Make sure we only write the first line to the output.
+				// This is needed if a description contains a linebreak.
+				// Otherwise the shell scripts will interpret the other lines as new flags
+				// and could therefore provide a wrong completion.
+				comp = strings.Split(comp, "\n")[0]
+
 				// Finally trim the completion.  This is especially important to get rid
 				// of a trailing tab when there are no description following it.
 				// For example, a sub-command without a description should not be completed
--- a/custom_completions_test.go
+++ b/custom_completions_test.go
@@ -501,7 +501,7 @@ func TestFlagNameCompletionInGoWithDesc(t *testing.T) {
 	}
 	rootCmd.AddCommand(childCmd)
 
-	rootCmd.Flags().IntP("first", "f", -1, "first flag")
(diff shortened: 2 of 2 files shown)


Luap99:
@marckhouzam Please take a look.
I noticed this while testing #1146 because this also messed up the formatting of the bash descriptions.

marckhouzam:
@Luap99 Interesting find!  What do you think would be best for a user:
1. only taking the first line of the description or
1. escaping the newline and showing the entire description (if we can)

I'm asking because I wondered if truncating a description would remove useful information.
Do you have concrete examples of descriptions using a newline so we can judge what would be best?

Luap99:
@marckhouzam Here are two examples I found in the podman project:
[tmpdir flag](https://github.com/containers/podman/blob/3d3359086ef0ed0833a8ac8e9d06d9934d835139/cmd/podman/root.go#L268)
[latest flag](https://github.com/containers/podman/blob/3d3359086ef0ed0833a8ac8e9d06d9934d835139/cmd/podman/validate/latest.go#L11)

I would argue that the first line should contain enough information to understand what a command/flag does. If someone wants the full information they should use the man pages or --help.

custom_completions.go
```diff
@@ -132,6 +132,12 @@ func (c *Command) initCompleteCmd(args []string) {
 					comp = strings.Split(comp, "\t")[0]
 				}
 
+				// Make sure we only write the first line to the output.
+				// This is needed if a description contains a linebreak.
+				// Otherwise the shell scripts will interpret the other lines as new flags
+				// and could therefor provide a wrong suggestion.
```
marckhouzam, line 138: Typo: "therefore"

marckhouzam: Could you change "suggestion" to "completion"?  Cobra has a concept of "suggestions" when someone mistypes a command; it is different from "completions".

Luap99: OK, thanks.

marckhouzam:
LGTM 
This will protect both fish and zsh completions from descriptions that have a line break in them.

/cc @jharshman

jpmcb (approved):
Thanks for this fix! LGTM 🎸

Issue #1165, opened 2020-07-15
zsh completion doesn't work


kraymond37:
I create a completion command as [this](https://chromium.googlesource.com/external/github.com/spf13/cobra/+/HEAD/shell_completions.md) said. Bash completion works well, in contrast, zsh doesn't work. 
Did I miss something?  I like cobra very much, and really wish it could work.

Luap99:
How do you load your completion script?
I think you need to do `yourprogram completion zsh > "${fpath[1]}/_yourprogram"`

marckhouzam:
@Luap99 is probably right and it points out that the documentation is mis-leading:
https://github.com/spf13/cobra/blame/master/shell_completions.md#L36
I'll post a PR.

kraymond37:
@Luap99 Thanks. I found `source` didn't work, and `yourprogram completion zsh > "${fpath[1]}/_yourprogram"` didn't work either. Both `yourprogram completion zsh > "${fpath[1]}/_yourprogram"` and `compinit` should be executed, and sometimes also need to relogin shell.

marckhouzam:
> Thanks. I found `source` didn't work, and `yourprogram completion zsh > "${fpath[1]}/_yourprogram"` didn't work either. Both `yourprogram completion zsh > "${fpath[1]}/_yourprogram"` and `compinit` should be executed, and sometimes also need to relogin shell.

Thanks @kraymond37, we have updated #1169 to include that information.
If that is satisfactory, would you mind closing this issue here?

PR #1169, opened 2020-07-16
Improve zsh completion documentation

marckhouzam:
The documentation mistakenly mentions that the user could setup zsh completion by doing:
```
source <(myprogram completion zsh)
```
This has been removed.  It was noticed in #1165.

Also, the documentation to the changes to zsh completion did not include the fact that the `-` prefix is now required to complete flag names (as is done for bash completion).  This was added.  It was noticed in https://github.com/spf13/cobra/pull/899#issuecomment-659392225

### DIFF
--- a/shell_completions.md
+++ b/shell_completions.md
@@ -33,10 +33,15 @@ MacOS:
 
 Zsh:
 
-$ source <(yourprogram completion zsh)
+# If shell completion is not already enabled in your environment you will need
+# to enable it.  You can execute the following once:
+
+$ echo "autoload -U compinit; compinit" >> ~/.zshrc
 
 # To load completions for each session, execute once:
 $ yourprogram completion zsh > "${fpath[1]}/_yourprogram"
+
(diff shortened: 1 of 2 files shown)


jpmcb:
I wonder if it's worth mentioning that the completion system needs to be in place before this will work. For example, someone's `~/.zshrc` should have:
```
autoload -U compinit
compinit
```
otherwise starting a new shell with the completion scripts in place won't work. I believe that's the issue here: https://github.com/spf13/cobra/issues/1165#issuecomment-659892335 . If someone's using something like ohmyzsh, this is usually taken care of for them, but I think being more verbose could help.

Otherwise, this looks good to me!

marckhouzam:
@jpmcb good point. I'll push a tweak.

marckhouzam:
@jpmcb Do you think this is clear enough?

Also, do you know how to source the powershell completion script?  We should add this info as well, but I don't know anything about powershell.

PR #1171, opened 2020-07-17
Complete subcommands when TraverseChildren is set

### LINKED ISSUE #1170: Complete subcommands after local flag
Currently the custom completion does not complete subcommands after a local flag was given.
I think this is wrong since a program can use subcommands even when there were local flags on the root/parent command given.

https://github.com/spf13/cobra/blob/b95db644ed1c0e183a90d8509ef2f9a5d51cc29b/custom_completions.go#L283-L287

### LINKED ISSUE #1172: Sub-commands are sometimes completed in bash after a local non-persistent flag
PR #288 added the feature that sub-commands should not be completed if they follow a local non-persistent flag. This is because the sub-command doesn't know that flag and the sub-command would therefore fail.

However, currently, when using the no-`=` form of a local non-persistent flag with bash, sub-commands are being completed.

Note that when using a boolean flag or the `=` form for a flag (e.g. integer), sub-commands are not completed after a local non-persistent flag as expected.
Here is an example:
```
bash-5.0$ ./testprog --local 1 [tab][tab]
child1      completion  help
# Sub-commands are completed when they shouldn't

bash-5.0$ ./testprog --local=1 [tab][tab]
[file completion is done]
# Sub-commands are NOT completed as expected

bash-5.0$ ./testprog --bool [tab][tab]
[file completion is done]
# Sub-commands are NOT completed as expected
```
My guess is that this bug was introduced along the way as the bash completion script evolved after #288 was implemented.

This only affects bash;  zsh and fish completion behave as expected.

Luap99:
The current custom completion logic does not complete
subcommands when a local flag is set. This is good unless
TraverseChildren is set to true where local flags
can be set on parent commands.

This commit allows subcommands to be completed only
if TraverseChildren is true.

Closes #1170
Closes #1172

### DIFF
--- a/bash_completions.go
+++ b/bash_completions.go
@@ -495,12 +495,14 @@ func writeFlag(buf *bytes.Buffer, flag *pflag.Flag, cmd *Command) {
 
 func writeLocalNonPersistentFlag(buf *bytes.Buffer, flag *pflag.Flag) {
 	name := flag.Name
-	format := "    local_nonpersistent_flags+=(\"--%s"
+	format := "    local_nonpersistent_flags+=(\"--%[1]s\")\n"
 	if len(flag.NoOptDefVal) == 0 {
-		format += "="
+		format += "    local_nonpersistent_flags+=(\"--%[1]s=\")\n"
 	}
-	format += "\")\n"
 	buf.WriteString(fmt.Sprintf(format, name))
+	if len(flag.Shorthand) > 0 {
+		buf.WriteString(fmt.Sprintf("    local_nonpersistent_flags+=(\"-%s\")\n", flag.Shorthand))
+	}
 }
 
 // Setup annotations for go completions for registered flags
@@ -535,7 +537,9 @@ func writeFlags(buf *bytes.Buffer, cmd *Command) {
 		if len(flag.Shorthand) > 0 {
 			writeShortFlag(buf, flag, cmd)
 		}
-		if localNonPersistentFlags.Lookup(flag.Name) != nil {
+		// localNonPersistentFlags are used to stop the completion of subcommands when one is set
+		// if TraverseChildren is true we should allow to complete subcommands
+		if localNonPersistentFlags.Lookup(flag.Name) != nil && !cmd.Root().TraverseChildren {
 			writeLocalNonPersistentFlag(buf, flag)
 		}
 	})
--- a/custom_completions.go
+++ b/custom_completions.go
@@ -175,8 +175,16 @@ func (c *Command) getCompletions(args []string) (*Command, []string, ShellCompDi
 	toComplete := args[len(args)-1]
 	trimmedArgs := args[:len(args)-1]
 
+	var finalCmd *Command
+	var finalArgs []string
+	var err error
 	// Find the real command for which completion must be performed
-	finalCmd, finalArgs, err := c.Root().Find(trimmedArgs)
+	// check if we need to traverse here to parse local flags on parent commands
+	if c.Root().TraverseChildren {
+		finalCmd, finalArgs, err = c.Root().Traverse(trimmedArgs)
+	} else {
+		finalCmd, finalArgs, err = c.Root().Find(trimmedArgs)
+	}
 	if err != nil {
 		// Unable to find the real command. E.g., <program> someInvalidCmd <TAB>
 		return c, []string{}, ShellCompDirectiveDefault, fmt.Errorf("Unable to find a command for arguments: %v", trimmedArgs)
@@ -271,20 +279,24 @@ func (c *Command) getCompletions(args []string) (*Command, []string, ShellCompDi
 	var completions []string
 	directive := ShellCompDirectiveDefault
 	if flag == nil {
-		// Check if there are any local, non-persistent flags on the command-line
 		foundLocalNonPersistentFlag := false
-		localNonPersistentFlags := finalCmd.LocalNonPersistentFlags()
-		finalCmd.NonInheritedFlags().VisitAll(func(flag *pflag.Flag) {
-			if localNonPersistentFlags.Lookup(flag.Name) != nil && flag.Changed {
-				foundLocalNonPersistentFlag = true
-			}
-		})
+		// If TraverseChildren is true on the root command we don't check for
+		// local flags because we can use a local flag on a parent command
+		if !finalCmd.Root().TraverseChildren {
+			// Check if there are any local, non-persistent flags on the command-line
+			localNonPersistentFlags := finalCmd.LocalNonPersistentFlags()
+			finalCmd.NonInheritedFlags().VisitAll(func(flag *pflag.Flag) {
+				if localNonPersistentFlags.Lookup(flag.Name) != nil && flag.Changed {
+					foundLocalNonPersistentFlag = true
+				}
+			})
+		}
 
 		// Complete subcommand names, including the help command
 		if len(finalArgs) == 0 && !foundLocalNonPersistentFlag {
 			// We only complete sub-commands if:
 			// - there are no arguments on the command-line and
-			// - there are no local, non-peristent flag on the command-line
+			// - there are no local, non-peristent flag on the command-line or TraverseChildren is true
 			for _, subCmd := range finalCmd.Commands() {
 				if subCmd.IsAvailableCommand() || subCmd == finalCmd.helpCommand {
 					if strings.HasPrefix(subCmd.Name(), toComplete) {
--- a/custom_completions_test.go
+++ b/custom_completions_test.go
@@ -139,6 +139,8 @@ func TestNoCmdNameCompletionInGo(t *testing.T) {
 		Use: "root",
 		Run: emptyRun,
 	}
+	rootCmd.Flags().String("localroot", "", "local root flag")
+
 	childCmd1 := &Command{
 		Use:   "childCmd1",
 		Short: "First command",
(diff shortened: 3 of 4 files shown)


custom_completions.go
```diff
@@ -271,20 +272,24 @@ func (c *Command) getCompletions(args []string) (*Command, []string, ShellCompDi
...
 	if flag == nil {
-		// Check if there are any local, non-persistent flags on the command-line
 		foundLocalNonPersistentFlag := false
-		localNonPersistentFlags := finalCmd.LocalNonPersistentFlags()
-		finalCmd.NonInheritedFlags().VisitAll(func(flag *pflag.Flag) {
-			if localNonPersistentFlags.Lookup(flag.Name) != nil && flag.Changed {
-				foundLocalNonPersistentFlag = true
-			}
-		})
+		// If TraverseChildren is true we dont check for local flags
```
marckhouzam, line 276: dont -> don't

custom_completions_test.go
```diff
@@ -187,6 +189,49 @@ func TestNoCmdNameCompletionInGo(t *testing.T) {
 		t.Errorf("expected: %q, got: %q", expected, output)
 	}
 
+	// Test that sub-command names are completed if a local non-persistent flag is present and TraverChildren is set to true
+	// set TraverChildren to true on the root cmd
```
marckhouzam, line 193: TraverChildren -> TraverseChildren

custom_completions_test.go
```diff
@@ -187,6 +189,49 @@ func TestNoCmdNameCompletionInGo(t *testing.T) {
 		t.Errorf("expected: %q, got: %q", expected, output)
 	}
 
+	// Test that sub-command names are completed if a local non-persistent flag is present and TraverChildren is set to true
```
marckhouzam, line 192: TraverChildren -> TraverseChildren

custom_completions_test.go
```diff
@@ -187,6 +189,49 @@ func TestNoCmdNameCompletionInGo(t *testing.T) {
...
+		"help",
+		":4",
+		"Completion ended with directive: ShellCompDirectiveNoFileComp", ""}, "\n")
+
+	if output != expected {
+		t.Errorf("expected: %q, got: %q", expected, output)
+	}
+
+	// Test that sub-command names are completed if a local non-persistent flag is present
+	// and TraverChildren is set to true on a child command
```
marckhouzam, line 214: TraverChildren -> TraverseChildren

custom_completions.go
```diff
@@ -271,20 +272,24 @@ func (c *Command) getCompletions(args []string) (*Command, []string, ShellCompDi
...
+		// because we can use a local flag on a parent command
+		if !finalCmd.TraverseChildren {
```
marckhouzam, line 278: From what I understand, for `TraverseChildren` to work (outside of completion), it *must* be set on the Root command.  This is because `TraverseChildren` is only considered here:
https://github.com/spf13/cobra/blob/19e41cf081df9bf9dde5cffa0090e718c3fdc8af/command.go#L922
and that line is *only* run on the root command.
I have confirmed by testing that setting `TraverseChildren` only on a sub-command will still cause an "unknow flag" error for local flags.

Therefore, we should only consider `TraverseChildren` for the root command, so it should probably be:
`if !finalCmd.Root().TraverseChildren {`

custom_completions_test.go
```diff
@@ -187,6 +189,49 @@ func TestNoCmdNameCompletionInGo(t *testing.T) {
...
+	childCmd1.TraverseChildren = true
```
marckhouzam, line 215: Based on the comment above, I believe this test is not valid.  Instead, we should set `rootCmd.TraverseChildren=true` even when testing sub-commands.

custom_completions.go
```diff
@@ -176,7 +176,8 @@ func (c *Command) getCompletions(args []string) (*Command, []string, ShellCompDi
 	trimmedArgs := args[:len(args)-1]
 
 	// Find the real command for which completion must be performed
-	finalCmd, finalArgs, err := c.Root().Find(trimmedArgs)
+	// use Traverse here to parse local flags on parent commands
+	finalCmd, finalArgs, err := c.Root().Traverse(trimmedArgs)
```
marckhouzam, line 180: Could you add a test that fails if we use `Find()` when we shouldn't and when we use`Traverse()` when we shouldn't?

custom_completions.go
```diff
@@ -176,7 +176,8 @@ func (c *Command) getCompletions(args []string) (*Command, []string, ShellCompDi
 	trimmedArgs := args[:len(args)-1]
 
 	// Find the real command for which completion must be performed
-	finalCmd, finalArgs, err := c.Root().Find(trimmedArgs)
+	// use Traverse here to parse local flags on parent commands
```
marckhouzam, line 179: Shouldn't we only use `Traverse()` if `c.Root().TraverseChildren` is true?

marckhouzam:
Also @Luap99, would you mind writing an issue to mention that `TraverseChildren` is not being considered for bash completion?

marckhouzam:
@Luap99 I forgot to ask you to update your commit message, it should have a short title before going into the details.

custom_completions.go
Luap99: Yes you are right, I had a wrong understanding where TraverseChildren should be set. I updated the code according to your notes.

custom_completions.go
Luap99: The test should now fail when we would always use `Find()`.
I'm not sure how to make it fail when we use `Traverse()` though.

Luap99:
> Also @Luap99, would you mind writing an issue to mention that TraverseChildren is not being considered for bash completion?

I added a commit here which should fix this and the bug we discovered (#1172).
I still need to add tests for that one.

custom_completions_test.go
```diff
@@ -187,6 +189,51 @@ func TestNoCmdNameCompletionInGo(t *testing.T) {
...
+		"Completion ended with directive: ShellCompDirectiveNoFileComp", ""}, "\n")
+
+	if output != expected {
+		t.Errorf("expected: %q, got: %q", expected, output)
+	}
+
+	// Test that sub-command names are completed if a local non-persistent flag is present
+	// and TraverseChildren is set to true on a child command
+	rootCmd.TraverseChildren = true
+	childCmd1.TraverseChildren = true
```
marckhouzam, line 216: Let's not set `TraverseChildren` on `childCmd1` since it is not necessary.

custom_completions_test.go
```diff
@@ -187,6 +189,51 @@ func TestNoCmdNameCompletionInGo(t *testing.T) {
...
+
+	output, err = executeCommand(rootCmd, ShellCompNoDescRequestCmd, "--localroot", "value", "childCmd1", "--nonPersistent", "value", "")
+	if err != nil {
+		t.Errorf("Unexpected error: %v", err)
+	}
+	// Reset TraverseChildren for next command
+	rootCmd.TraverseChildren = false
+	childCmd1.TraverseChildren = false
```
marckhouzam, line 224: Can remove this one too.

custom_completions.go
marckhouzam: This test should fail if we use `Traverse` instead of `Find`:
```
executeCommand(rootCmd, ShellCompNoDescRequestCmd, "--localroot", "value", "childCmd1", "")
```
Using `Find` (the passing test) will not return any completions because `--localroot` will be left on the command-line and `ParseFlags` will fail.
But using `Traverse` by mistake (the failing test) will return `childCmd2` as a completion because `--localroot` will have been removed from the command-line, which will allow `ParseFlags` to succeed.

bash_completions.go
```diff
@@ -495,12 +495,15 @@ func writeFlag(buf *bytes.Buffer, flag *pflag.Flag, cmd *Command) {
...
 func writeLocalNonPersistentFlag(buf *bytes.Buffer, flag *pflag.Flag) {
 	name := flag.Name
-	format := "    local_nonpersistent_flags+=(\"--%s"
+	format := "    local_nonpersistent_flags+=(\"--%[1]s"
 	if len(flag.NoOptDefVal) == 0 {
-		format += "="
+		format += "\")\n    local_nonpersistent_flags+=(\"--%[1]s="
 	}
 	format += "\")\n"
 	buf.WriteString(fmt.Sprintf(format, name))
```
marckhouzam, line 503: This works nicely, but I think we can make it more readable with:
```
        name := flag.Name
	format := "    local_nonpersistent_flags+=(\"--%[1]s\")\n"
	if len(flag.NoOptDefVal) == 0 {
		format += "    local_nonpersistent_flags+=(\"--%[1]s=\")\n"
	}
	buf.WriteString(fmt.Sprintf(format, name))
```

What do you think?

Luap99: Yes sure.

custom_completions.go
Luap99: OK, thank you. I added the test.

Luap99:
@marckhouzam I have updated the code according to your suggestions.

jpmcb:
Thanks for this @Luap99!! Would you be able to provide some more context / a good way I can test this? I'll merge this once I've tried this out

Luap99:
@jpmcb I run into this while working on containers/podman#6442. Podman has a local flag called `remote` on the root cmd to indicate that the command should be run against a remote endpoint. Since the root cmd has TraverseChildren set to true, this is a vaild cmd: `podman --remote ps`. I was wondering why the completion didn't offer subcmds if I tried something like this: `podman --remote [TAB]` and found this bug.

Unfortunately podman is not easy to set up. Therefore I can create a simple example program for you if you want to test it.

Luap99:
```
package main

import (
	"fmt"
	"os"
	"path/filepath"
	"strings"

	"github.com/spf13/cobra"
)

var completionNoDesc = false
var remote = false

	Use:   "completion [bash|zsh|fish|powershell]",
	Long: `To load completions:

Bash:

$ source <(yourprogram completion bash)

Linux:
  $ yourprogram completion bash > /etc/bash_completion.d/yourprogram
MacOS:
  $ yourprogram completion bash > /usr/local/etc/bash_completion.d/yourprogram

Zsh:

$ source <(yourprogram completion zsh)

$ yourprogram completion zsh > "${fpath[1]}/_yourprogram"

Fish:

$ yourprogram completion fish | source

$ yourprogram completion fish > ~/.config/fish/completions/yourprogram.fish
`,
	DisableFlagsInUseLine: true,
	ValidArgs:             []string{"bash", "zsh", "fish", "powershell", "bash-old"},
		switch args[0] {
		case "bash":
		case "zsh":
			if !completionNoDesc {
			} else {
			}
		case "fish":
		case "powershell":
		}
	},
}

var rootCmd = &cobra.Command{
	Use:              filepath.Base(os.Args[0]),
	TraverseChildren: true,
		fmt.Println("rootCmd called")
		fmt.Printf("remote %v\n", remote)
	},
}

	Use:       "vargs",
	Short:     "Cmd with vargs and subcmd",
	Args:      cobra.MinimumNArgs(1),
	},
}

	Use:   "childVargs",
	Short: "Child command",
		fmt.Printf("remote %v\n", remote)
	},
}

	Use:   "filterFile",
	Short: "filterFile [*.log,*.txt]",
		if len(args) == 0 {
			//return nil, cobra.ShellCompDirectiveFilterFileExt
		}
	},
		fmt.Printf("remote %v\n", remote)
	},
}

	Use:   "nofile",
	Short: "nofile",
	},
		fmt.Println("nofile called")
		fmt.Printf("remote %v\n", remote)
	},
}

	Use:   "fullFile",
	Short: "fullFile [filename]",
		fmt.Printf("remote %v\n", remote)
	},
}

	Use:   "dirOnly",
	Short: "dirOnly [dirname]",
		if len(args) == 0 {
			return []string{"2"}, cobra.ShellCompDirectiveFilterDirs
		}
	},
		fmt.Println("dirOnly called")
		fmt.Printf("remote %v\n", remote)
	},
}

	Use:   "subdir",
	Short: "subdir [sub-dirname]",
		if len(args) == 0 {
		}
	},
		fmt.Println("subdir called")
		fmt.Printf("remote %v\n", remote)
	},
}

	Use:   "compAndFile",
	Short: "compAndFile [sun|moon|<filename>]",
		var completions []string
		// If we don't filter on "toComplete", zsh and fish will not do file completion
		// even if the prefix typed by the user does not match the returned completions
		// e.g., ./testprog compAndFile f<TAB>
		// will only get file completion for bash, not zsh and fish, if we blindly return
		// the non-matching completions ("sun", "moon")
		for _, comp := range []string{"sun", "moon"} {
			}
		}
	},
		fmt.Printf("remote %v\n", remote)
	},
}

func setFlags() {
	rootCmd.PersistentFlags().String("persistent", "", "persistent flag")

	rootCmd.Flags().BoolVarP(&remote, "remote", "r", false, "bool flag")

}

func main() {
		&completionNoDesc,
		"no-descriptions", false,

	rootCmd.AddCommand(filterFileCmd)
	rootCmd.AddCommand(noFileCmd)
	rootCmd.AddCommand(fullFileCmd)
	rootCmd.AddCommand(dirOnlyCmd)
	rootCmd.AddCommand(subdirCmd)
	rootCmd.AddCommand(compAndFileCmd)

	rootCmd.AddCommand(vargsCmd)

	setFlags()
	rootCmd.Execute()
}

```
@jpmcb Try this with TraverseChildren true and false on the root cmd and do something like this `./testprog --remote [TAB]`

Luap99:
@jpmcb Did you had the time to test this? Is anything else needed?

jpmcb:
Yes, I've tested this out with your example program, thank you very much for the context!! Everything looks good to me. Thanks for the work on this @Luap99 and @marckhouzam

Luap99:
@jpmcb Would you please so kind and take a look at #1162. It is a very small fix.

PR #1173, opened 2020-07-19
feature: Allow short and long descriptions for cobra generator for co…

zloeber:
This adds additional features for the template generation within the `add` command function. This includes;
- Allowing for both short and long description to be passed into the `cobra add` command using `--shortdesc` and `--longdesc` flags respectively. If not defined, these values will be populated with the prior string that was previously hard-coded into the template'
- If rootCmd is the parent for a generated command then the generated command file name will remain as the command name (previous default behaviour). But if any other parent command is specified, then the filename will be generated as `<parentcommand>_<command>` instead. (the 'Cmd' will be stripped from the parent command name if found). This should allow for at least one layer of additional sub-command complexity before file names begin to overlap one another.
- Added a Makefile task for the cobra generator to create a project from scratch in the `_test` folder that creates a default cobra project using combinations of including short and long descriptions, not including them, and generating sub-command file names.


jharshman:
@zloeber Neat functional addition to the `add` command for the generator. Looks like you have some CI failures. Can you address those?

jharshman:
@zloeber this is a feature that I'd like to see added.
Would you mind fixing your CI errors?

zloeber:
> @zloeber this is a feature that I'd like to see added.
> Would you mind fixing your CI errors?

Hello, I'm somewhat new at Golang (novice at best). Please review the additional unit test I added to resolve these CI issues. If needed I could probably hack together further tests if they are deemed necessary. Sorry for the delays with this one as well.

Issue #1178, opened 2020-07-25
Example usage of ValidArgsFunc


infixint943:
Could a simple example for `ValidArgsFunc` be given, as it's not clear what exactly we ought to do in it.
A simple example could be listing all `.go` files in the current directory.

Thanks for this library!

Luap99:
There are examples in the [shell_completions.md](https://github.com/spf13/cobra/blob/master/shell_completions.md) file.

> A simple example could be listing all .go files in the current directory.
```
func AutocompleteGoFiles(cmd *cobra.Command, args []string, toComplete string) ([]string, cobra.ShellCompDirective) {
	return []string{"go"}, ShellCompDirectiveFilterFileExt
}
```

infixint943:
Yes, thanks a lot! Maybe adding it to the readme for better visibility?

wfernandes:
@infixint943 I hope the information in `shell_completions.md` was useful. 
@Luap99 thanks for helping out. 

Since `ValidArgsFunction` is related to shell completion, it makes sense to have it in the `shell_completions.md` file. I believe the original intent was to avoid adding more information to the main README. What do you think?

infixint943:
Yes, that makes sense.
One last request is to state (in the README) that custom completions don't work on zsh, unless you use the master branch. If you could create a new release, it would be much better!

Thanks for the support!

wfernandes:
@infixint943 If you can create a PR adding that note in the README and tag me in it I can take a look at it. Thanks.

Also, we are working on creating a proposal on how we are going to cut releases and support versioning regarding Golang for example. As soon as those guidelines are up, we will plan for cutting future releases.
Until then I think your note/warning in our README would be useful.

wfernandes:
I'll close this issue since it seems to be resolved. Thanks!

PR #1185, opened 2020-08-06
Update readme to reflect support of zsh custom shell completions

infinitepr0:


infixint943:
#1178 @wfernandes

README.md
```diff
@@ -753,7 +753,9 @@ Cobra can generate documentation based on subcommands, flags, etc. Read more abo
 
 ## Generating shell completions
 
-Cobra can generate a shell-completion file for the following shells: Bash, Zsh, Fish, Powershell. If you add more information to your commands, these completions can be amazingly powerful and flexible.  Read more about it in [Shell Completions](shell_completions.md).
+Cobra can generate a shell-completion file for the following shells: Bash, Zsh, Fish, Powershell. Custom, dynamic completions are supported too. If you add more information to your commands, these completions can be amazingly powerful and flexible.  Read more about it in [Shell Completions](shell_completions.md).
+
+*Note: For generating **custom** shell completions for `zsh`, [use the master branch](https://stackoverflow.com/a/56410071/9606036) in your `.mod` file.*
```
marckhouzam, line 758: The intention of helping people with this extra note is good.  However, it would mean that the maintainers would have to remember to remove this line before doing a release.

Personally, I think people have to understand that the documentation on the master branch matches the code on the master branch.  We shouldn't start documenting old code in newer versions.

As an alternative could there be a way to put a link at the top of the readme to say: "For the documentation of the latest release, please see here (link to last release tag)". We'd need a link that wouldn't need to be updated if possible.

infixint943:
Or, we could add in `shell_completions.md` that releases prior to `v1.0.5` don't support zsh dynamic completion.

Would this be better?

marckhouzam:
That sounds better to me.  I'm assuming the next release would be 1.1, but I don't know.

wfernandes:
Thanks for the input @marckhouzam, I think your points makes sense. 

@jharshman We should plan for cutting a minor release that includes the zsh work so that folks can take advantage of it. If we do that, then we can add this warning in the release notes.

jharshman:
@wfernandes agreed.

infixint943:
Any updates regarding this?

marckhouzam:
On October 8th on slack the maintainers mentioned they were going to do a release. There are a couple of PRs to possibly include first, but it should happen soon.

infinitepr0:
Closing since a release has been made. Thanks!

PR #1190, opened 2020-08-15
Remove duplicate main.go example

g1stavo:
There was a duplicate `main.go` example on `README.md`. I removed the first one, since the second seems to make more sense. 

Screen Shot 2020-08-15 at 12 34 20 PM

### DIFF
--- a/README.md
+++ b/README.md
@@ -131,20 +131,6 @@ application will follow the following organizational structure:
       main.go
 ```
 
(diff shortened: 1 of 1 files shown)


jpmcb:
Thanks for contributing! I think both instances are valid and serve a purpose; the first one is good to have for a quick start and getting people off the ground running. The latter one is useful for a more in depth guide. So maybe the last one could be updated to be a bit more verbose and go into more detail of what the the `main.go` file is useful for.

jharshman:
Closing as stale

PR #1192, opened 2020-08-17
Have Cobra create 'completion' command automatically

marckhouzam:
This PR has Cobra automatically provide a `completion` command at the root level if the program does not provide its own, just like for the `help` command.  This lowers the barrier-to-entry and will make shell completion available to all Cobra users.

With this, a program developer does not need to learn or do anything about completion for users of the program to have access to shell completion for all four shells supported by Cobra.

The change is backwards-compatible, since existing handling for completion has not been changed.  However, when no `completion` command is provided by the program, Cobra will now create a standard one.

Cobra will automatically include a `--no-descriptions` flag for `zsh` and `fish`, to allow users to turn off completion descriptions.

The PR also provides three configuration options:
1. Disable the creation of this new default completion command - for programs that:
    * don't want to provide completion, or 
    * want to name the command something else, or
    * want to place the command elsewhere on the command tree.
1. Disable completion descriptions completely - for programs that just don't want to offer completion descriptions.
1. Disable the `--no-descriptions` flag - for programs that want to have completion descriptions enabled all the time.

### DIFF
--- a/command.go
+++ b/command.go
@@ -168,6 +168,9 @@ type Command struct {
 	//FParseErrWhitelist flag parse errors to be ignored
 	FParseErrWhitelist FParseErrWhitelist
 
+	// CompletionOptions is a set of options to control the handling of shell completion
+	CompletionOptions CompletionOptions
+
 	// commandsAreSorted defines, if command slice are sorted or not.
 	commandsAreSorted bool
 	// commandCalledAs is the name or alias value used to call this command.
@@ -914,9 +917,10 @@ func (c *Command) ExecuteC() (cmd *Command, err error) {
 		preExecHookFn(c)
 	}
 
-	// initialize help as the last point possible to allow for user
-	// overriding
+	// initialize help at the last point to allow for user overriding
 	c.InitDefaultHelpCmd()
+	// initialize completion at the last point to allow for user overriding
+	c.initDefaultCompletionCmd()
 
 	args := c.args
 
--- a/completions.go
+++ b/completions.go
@@ -63,6 +63,32 @@ const (
 	ShellCompDirectiveDefault ShellCompDirective = 0
 )
 
+const (
+	// Constants for the completion command
+	compCmdName              = "completion"
+	compCmdNoDescFlagName    = "no-descriptions"
+	compCmdNoDescFlagDesc    = "disable completion descriptions"
+	compCmdNoDescFlagDefault = false
+)
+
+// CompletionOptions are the options to control shell completion
+type CompletionOptions struct {
+	// DisableDefaultCmd prevents Cobra from creating a default 'completion' command
+	DisableDefaultCmd bool
+	// DisableNoDescFlag prevents Cobra from creating the '--no-descriptions' flag
+	// for shells that support completion descriptions
+	DisableNoDescFlag bool
+	// DisableDescriptions turns off all completion descriptions for shells
+	// that support them
+	DisableDescriptions bool
+}
+
+// NoFileCompletions can be used to disable file completion for commands that should
+// not trigger file completions.
+func NoFileCompletions(cmd *Command, args []string, toComplete string) ([]string, ShellCompDirective) {
+	return nil, ShellCompDirectiveNoFileComp
+}
+
 // RegisterFlagCompletionFunc should be called to register a function to provide completion for a flag.
 func (c *Command) RegisterFlagCompletionFunc(flagName string, f func(cmd *Command, args []string, toComplete string) ([]string, ShellCompDirective)) error {
 	flag := c.Flag(flagName)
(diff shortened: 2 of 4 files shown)


custom_completions.go
```diff
@@ -476,6 +502,139 @@ func checkIfFlagCompletion(finalCmd *Command, args []string, lastArg string) (*p
...
+	powershell := &Command{
+		Use:   "powershell",
+		Short: fmt.Sprintf(shortDesc, "powershell"),
+		Long: fmt.Sprintf(`
+Generate the autocompletion script for powershell.
+
+To load completions in your current shell session:
+$ %[1]s completion powershell > comp.ps1
+$ ./comp.ps1
+$ rm ./comp.ps1
```
Luap99, line 626: No need for a temp file.
```suggestion
PS C:\> %[1]s completion powershell | Invoke-Expression
```

Luap99: Also we might want to add: `To load completions for every new session, add the output to your powershell profile`

custom_completions.go
```diff
@@ -476,6 +502,139 @@ func checkIfFlagCompletion(finalCmd *Command, args []string, lastArg string) (*p
...
+
+	out := c.OutOrStdout()
+	noDesc := c.CompletionOptions.DisableDescriptions
+	shortDesc := "generate the autocompletion script for %s"
+	bash := &Command{
+		Use:   "bash",
+		Short: fmt.Sprintf(shortDesc, "bash"),
+		Long: fmt.Sprintf(`
+Generate the autocompletion script for the bash shell.
+
```
Luap99, line 544: Should we add a note here that we require `bash_completion` as dependency?

custom_completions.go
marckhouzam: Yes, I wasn't sure about the PowerShell sequence.  I have installed it on my Mac but haven't tried on Windows (yet?).  I will follow your suggestions.

custom_completions.go
```diff
@@ -476,6 +502,139 @@ func checkIfFlagCompletion(finalCmd *Command, args []string, lastArg string) (*p
...
+			return
+		}
+	}
+
+	haveFlag := !c.CompletionOptions.DisableNoDescFlag && !c.CompletionOptions.DisableDescriptions
+
+	completionCmd := &Command{
+		Use:   compCmdName,
+		Short: "generate the autocompletion script for the specified shell",
+		Long: fmt.Sprintf(`
```
Luap99, line 527: I would argue that making this command hidden makes more sense. If I have a subcommand called `commit` I would be forced to always type more letters before tab completion works. Especially since the completion command is used once or never because the scripts are usually installed with deb/rpm.
Or you add this to the `CompletionOptions` e.g. `HideCommand`

marckhouzam: That's an interesting point.  Hiding the completion command to improve completion 😁

But hiding a command also removes it from the help texts. I'm concerned it would be a little too drastic.  We could definitely put an option though.  

But of the tools I know (kubectl, helm, etc) none of them choose to hide their current `completion` command, so let's leave this point open for discussion with the maintainers.

custom_completions.go
marckhouzam: > This works for me: `./testprog completion powershell | Out-String | Invoke-Expression`

Ok, it seems work but ...

> But powershell completion is lacking quiet a bit behind the other shells. I will look into this and see if I can port this to your custom completions. So we can have an equal user experience for all shells :)

Yes you are right, and it doesn't even work properly for me in its current state, although it may be that I don't understand powershell.

If you can port the powershell completion to Go completions, I think it would be a great improvement for powershell users.

custom_completions.go
marckhouzam: Good idea.  Done.

custom_completions.go
Luap99: I opened #1208 as WIP

marckhouzam:
@jpmcb Do you think the idea of providing a `completion` command automatically is something of value?  I'm not asking to merge but just to know if this suggestion has value in the maintainers' eyes.  If not, we can close this.

jpmcb:
@marckhouzam In my opinion, yes this would be useful. A frequent pattern I've seen in CLI tools is to have a `completion` command that can be injected into a `.rc` type file for completions. 

I think we would want the ability for users to opt out of using this feature (for whatever reason) but otherwise, yes I think it'd be useful!

marckhouzam:
Thanks @jpmcb for the feedback. The current PR does provide a way to opt-out of the completion command as well as a couple of other configurations (see the PR main description). 👍

marckhouzam:
So a colleague wrote a command-line tool using Cobra recently. Looking at it I saw he hasn't implemented shell completion at all.  This reminded me of this PR which would automatically add the `completion` command.

I was thinking it would be nice to include it in the next Cobra release of others agreed.

marckhouzam:
Let's keep this one alive.

jpmcb:
Hi Marc - where is this one at? I can review if it's in a good state! 😄 

On initial look, this is awesome. I think auto-generating completions will be really powerful and remove steps for users to get completions into their CLIs 

Just let me know what this needs and we can get it in!

marckhouzam:
Thanks @jpmcb!  This should be ready for review.  You may want to pay particular attention at how I defined the new completion command options, just to make sure that is a good API.

custom_completions.go
```diff
@@ -494,6 +520,143 @@ func checkIfFlagCompletion(finalCmd *Command, args []string, lastArg string) (*p
...
+PS C:\> %[1]s completion powershell | Out-String | Invoke-Expression
+
+To load completions for every new session, add the output of the above command
+to your powershell profile.
+`, c.Root().Name()),
+		Args:                  NoArgs,
+		DisableFlagsInUseLine: true,
+		ValidArgsFunction:     NoFileCompletions,
+		RunE: func(cmd *Command, args []string) error {
+			return cmd.Root().GenPowerShellCompletion(out)
```
Luap99, line 654: This should be updated to also include the GenPowerShellCompletionWithDesc function

marckhouzam: Thanks for catching that! I will update very soon.

marckhouzam:
At the same time @jpmcb, @Luap99, how about I rename the `custom_completion.go` file to simply `completion.go`.  The original name was from the early functionality what was only providing custom completions; however, when we added support for fish shell, that file learned full completion support but I didn't think about changing the name.  The poor naming has been bothering me for a while.  I'll push that change too, but if you disagree I'll revert.

marckhouzam:
Done and ready.

marckhouzam:
I had missed enabling the `[flags]` string in the help message for `powershell` completion.  The last push fixed it.

marckhouzam:
And I just updated the documentation to mention the default `completion` command and explain its options.

This is ready @jpmcb

jpmcb (approved):
LGTM! Tested this out on a few shells and on a few CLIs and I'm really excited about this one! Great work @marckhouzam and @Luap99

marckhouzam:
Thanks for the review and merge @jpmcb!
Completion for all!!! 🎉

PR #1196, opened 2020-08-20
Modifying "snake-case" to "kebab-case" for clarity.

dhermes:


jpmcb (approved):
Thanks for this! LGTM

PR #1208, opened 2020-08-26
powershell completion with custom comp

### LINKED ISSUE #1229: [Bug] When completion Item description is empty, Powershell completion is not working
https://github.com/spf13/cobra/blob/7f8e83d9366ab8e96959f9c3fbdf3507d4b39c3b/powershell_completions_test.go#L50-L85

From this issue: cli/cli/issues/1775

**I found that 4th argument of `CompletionResult`'s constructor can't use empty string.**

## Steps to reproduce

1. Register completions.

```powershell
 (gh completion -s powershell) -join "`n" | Invoke-Expression
```

2. Completion

```powershell
gh <Ctrl+Space>
```

## Expected behavior

Offer completions.

```none
❯ gh credits
alias        auth         config       environment  help         pr           repo
api          completion   credits      gist         issue        release      version

View credits for this tool
```

## Actual behavior

NOT offer completions.

image

However, `<command> <tab>`**offer** completions.

image

After investigating this bug, I found that if the 4th argument of `CompletionResult`'s constructor is emptystring, it is interpreted as `null` and an `MethodInvocationException` is thrown.
In order to avoid this, it seems to be necessary to set some string, even if it is a whitespace.

```powershell
using namespace System.Management.Automation
using namespace System.Management.Automation.Language

[CompletionResult]::new('version', 'version', [CompletionResultType]::ParameterValue, '')
[CompletionResult]::new('version', 'version', [CompletionResultType]::ParameterValue, ' ')
```

```none
MethodInvocationException:
Line |
   4 |  [CompletionResult]::new('version', 'version', [CompletionResultType]: …
     |  ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
     | Exception calling ".ctor" with "4" argument(s): "Cannot process argument because the value of argument "toolTip" is null. Change the value of argument "toolTip" to a non-null value."

CompletionText ListItemText     ResultType ToolTip
-------------- ------------     ---------- -------
version        version      ParameterValue
```

image

So what should the 4th argument be set to?

### when the 1st and 2nd arguments are the same

[MsDocs said this.](https://docs.microsoft.com/ja-jp/powershell/module/Microsoft.PowerShell.Core/Register-ArgumentCompleter?view=powershell-7#example-3--register-a-custom-native-argument-completer)

>```powershell
>$scriptblock = {
>    param($wordToComplete, $commandAst, $cursorPosition)
>        dotnet complete --position $cursorPosition $commandAst.ToString() | ForEach-Object {
>            [System.Management.Automation.CompletionResult]::new($_, $_, 'ParameterValue', $_)
>        }
>}
>Register-ArgumentCompleter -Native -CommandName dotnet -ScriptBlock $scriptblock
>```

So I tried setting the same value as the 1st and 2nd arguments.

```powershell
$testParams = @{
    Native      = $true
    CommandName = 'gh'
    ScriptBlock = {
        [CompletionResult]::new('repo', 'repo', [CompletionResultType]::ParameterValue, 'Create, clone, fork, and view repositories')
        [CompletionResult]::new('version', 'version', [CompletionResultType]::ParameterValue, 'version') # FIXED '' -> 'version'
    }
}
Register-ArgumentCompleter @testParams
```

image

From this result, it seems that if no details for the completions is shown, it's OK to set the same string as the 1st and 2nd arguments instead of an empty string.

### when the 1st and 2nd arguments are different

This is also the case assumed in `powershell_completions_test.go`. I also checked how it behaves when the 1st argument, which is used as the auto completion result, and the 2nd argument, which is displayed in a list, are different.

```powershell
$testParams = @{
    Native      = $true
    CommandName = 'flags'
    ScriptBlock = {
        [CompletionResult]::new('-a', 'a', [CompletionResultType]::ParameterName, 'a')
        [CompletionResult]::new('--flag1', 'flag1', [CompletionResultType]::ParameterName, 'flag1')
        [CompletionResult]::new('--flag2', 'flag2', [CompletionResultType]::ParameterName, 'flag2')
        [CompletionResult]::new('sub1', 'sub1', [CompletionResultType]::ParameterValue, 'sub1')
    }
}
Register-ArgumentCompleter @testParams
```

image

```powershell
$testParams = @{
    Native      = $true
    CommandName = 'flags'
    ScriptBlock = {
        [CompletionResult]::new('-a', 'a', [CompletionResultType]::ParameterName, '-a')
        [CompletionResult]::new('--flag1', 'flag1', [CompletionResultType]::ParameterName, '--flag1')
        [CompletionResult]::new('--flag2', 'flag2', [CompletionResultType]::ParameterName, '--flag2')
        [CompletionResult]::new('sub1', 'sub1', [CompletionResultType]::ParameterValue, 'sub1')
    }
}
Register-ArgumentCompleter @testParams
```

image

It seems that at the least Console only needs to set one of the arguments and there is no clear difference.

### In VScode

I was somewhat curious about the last one, so I looked into it and found some interesting results.

#### Whitespace and When the 1st and 2nd arguments are the same

image

image

#### In the same case as the 1st argument

image

#### In the same case as the 2nd argument

image

Considering the difference, I wondered if it should be the same string as the 1st argument.
Because in some settings it is possible to see the 1st argument which is not displayed until selected.

## FIX

So my suggestion is to modify the test case as follows.
I think fixing it to pass this modified test will fix the bug.

```diff
diff --git a/powershell_completions_test.go b/powershell_completions_test.go
index 29b609d..4b46fb6 100644
--- a/powershell_completions_test.go
+++ b/powershell_completions_test.go
@@ -47,16 +47,16 @@ func TestPowerShellCompletion(t *testing.T) {
                        }(),
                        expectedExpressions: []string{
                                "'tree'",
-                               "[CompletionResult]::new('sub1', 'sub1', [CompletionResultType]::ParameterValue, '')",
-                               "[CompletionResult]::new('sub2', 'sub2', [CompletionResultType]::ParameterValue, '')",
+                               "[CompletionResult]::new('sub1', 'sub1', [CompletionResultType]::ParameterValue, 'sub1')",
+                               "[CompletionResult]::new('sub2', 'sub2', [CompletionResultType]::ParameterValue, 'sub2')",
                                "'tree;sub1'",
-                               "[CompletionResult]::new('sub11', 'sub11', [CompletionResultType]::ParameterValue, '')",
-                               "[CompletionResult]::new('sub12', 'sub12', [CompletionResultType]::ParameterValue, '')",
+                               "[CompletionResult]::new('sub11', 'sub11', [CompletionResultType]::ParameterValue, 'sub11')",
+                               "[CompletionResult]::new('sub12', 'sub12', [CompletionResultType]::ParameterValue, 'sub12')",
                                "'tree;sub1;sub11'",
                                "'tree;sub1;sub12'",
                                "'tree;sub2'",
-                               "[CompletionResult]::new('sub21', 'sub21', [CompletionResultType]::ParameterValue, '')",
-                               "[CompletionResult]::new('sub22', 'sub22', [CompletionResultType]::ParameterValue, '')",
+                               "[CompletionResult]::new('sub21', 'sub21', [CompletionResultType]::ParameterValue, 'sub21')",
+                               "[CompletionResult]::new('sub22', 'sub22', [CompletionResultType]::ParameterValue, 'sub22')",
                                "'tree;sub2;sub21'",
                                "'tree;sub2;sub22'",
                        },
@@ -76,13 +76,13 @@ func TestPowerShellCompletion(t *testing.T) {
                        }(),
                        expectedExpressions: []string{
                                "'flags'",
-                               "[CompletionResult]::new('-a', 'a', [CompletionResultType]::ParameterName, '')",
-                               "[CompletionResult]::new('--flag1', 'flag1', [CompletionResultType]::ParameterName, '')",
-                               "[CompletionResult]::new('--flag2', 'flag2', [CompletionResultType]::ParameterName, '')",
-                               "[CompletionResult]::new('sub1', 'sub1', [CompletionResultType]::ParameterValue, '')",
+                               "[CompletionResult]::new('-a', 'a', [CompletionResultType]::ParameterName, '-a')",
+                               "[CompletionResult]::new('--flag1', 'flag1', [CompletionResultType]::ParameterName, '--flag1')",
+                               "[CompletionResult]::new('--flag2', 'flag2', [CompletionResultType]::ParameterName, '--flag2')",
+                               "[CompletionResult]::new('sub1', 'sub1', [CompletionResultType]::ParameterValue, 'sub1')",
                                "'flags;sub1'",
-                               "[CompletionResult]::new('-c', 'c', [CompletionResultType]::ParameterName, '')",
-                               "[CompletionResult]::new('--flag3', 'flag3', [CompletionResultType]::ParameterName, '')",
+                               "[CompletionResult]::new('-c', 'c', [CompletionResultType]::ParameterName, '-c')",
+                               "[CompletionResult]::new('--flag3', 'flag3', [CompletionResultType]::ParameterName, '--flag3')",
                        },
                },
                {
```

## Environment data

```none
gh version 1.0.0 (2020-09-16)

VScode 1.49.1 x64
ms-vscode.powershell-preview@2020.9.0

Name                           Value
----                           -----
PSVersion                      7.1.0-preview.7
PSEdition                      Core
GitCommitId                    7.1.0-preview.7
OS                             Microsoft Windows 10.0.19041
Platform                       Win32NT
PSCompatibleVersions           {1.0, 2.0, 3.0, 4.0…}
PSRemotingProtocolVersion      2.3
SerializationVersion           1.1.0.1
WSManStackVersion              3.0
```

Luap99:
The current powershell completion is not very capable.

Let's port it to the go custom completion logic to have a
unified experience accross all shells.

Powershell supports three different completion modes

- TabCompleteNext (default windows style - on each key press the next option is displayed)
- Complete (works like bash)
- MenuComplete (works like zsh)

You set the mode with `Set-PSReadLineKeyHandler -Key Tab -Function <mode>`

Descriptions will only be supported for Complete and MenuComplete.

Fixes #1229


powershell_completions.go
```diff
@@ -11,90 +8,253 @@ import (
...
+    }
+}
+`, name, compCmd,
+		ShellCompDirectiveError, ShellCompDirectiveNoSpace, ShellCompDirectiveNoFileComp,
+		ShellCompDirectiveFilterFileExt, ShellCompDirectiveFilterDirs))
 }
 
 // GenPowerShellCompletion generates PowerShell completion file and writes to the passed writer.
-func (c *Command) GenPowerShellCompletion(w io.Writer) error {
+func (c *Command) GenPowerShellCompletion(w io.Writer, includeDesc bool) error {
```
marckhouzam, line 244: This change is not backwards compatible.  We'll need to add new functions instead.

Luap99: Yes I'm aware. I'm waiting for #1146 to decide how we want to handle this.

Luap99:
@marckhouzam I don't think I can get `ShellCompDirectiveFilterFileExt` and `ShellCompDirectiveFilterDirs` to work, I guess the best way is to provide the standard path completion instead?

marckhouzam:
@Luap99 agreed. That's what I also had to do for fish completion:
https://github.com/spf13/cobra/blob/02a0d2fbc9e61d26f8e5979749f6030964a55a3e/fish_completions.go#L126-L133

powershell_completions.go
```diff
@@ -11,90 +8,253 @@ import (
...
+
+        # Look for the longest completion so that we can format things nicely
+        if ($Longest -lt $Name.Length) {
+            $Longest = $Name.Length
+        }
+
+        # Set the description to a one space string if there is none set.
+        # This is needed because the CompletionResult does not accept an empty string as argument
+        if (-Not $Description) {
+            $Description = " "
```
marckhouzam, line 129: I notice that if there are no descriptions, we see a set of empty parentheses.  It is not so bad, but fish does not put them, but from the comment, I guess you don't have a choice?  For example:

powershell:
```
> ./testprog no<TAB>
nodesc  ( )                        nofile  (No file completion done)
```
fish:
```
> ./testprog no<TAB>
nodesc  nofile  (No file completion done)
```

powershell_completions.go
```diff
@@ -11,90 +8,253 @@ import (
...
+    if (($Directive -band $ShellCompDirectiveFilterFileExt) -ne 0 ) {
 
-func escapeStringForPowerShell(s string) string {
-	return strings.Replace(s, "'", "''", -1)
+        ## get the paths from Get-ChildItem
+        __%[1]s_debug "ShellCompDirectiveFilterFileExt"
+
+        $Values | ForEach-Object {
+            __%[1]s_debug "Filter: $($_.Name)"
+            Get-ChildItem -Recurse -Name -File -Filter "*.$($_.Name)"
```
marckhouzam, line 162: This seems to work pretty decently.  I thought you said `ShellCompDirectiveFilterFileExt` wasn't working?

Also, why to you recurse into sub directories?

powershell_completions.go
```diff
@@ -11,90 +8,253 @@ import (
...
+    # get directive from last line
+    [int]$Directive = $Out[-1].TrimStart(':')
+    if ($Directive -eq "") {
+        # There is not directive specified
+        $Directive = 0
+    }
+    __%[1]s_debug "The completion directive is: $Directive"
+    
+    # remove directive (last element) from out
+    $Out = $Out | Where-Object { $_ –ne $Out[-1] }
```
marckhouzam, line 106: My VSCode tells me the `-` in `-ne` in the above line is not a real `-`.  Was that on purpose?

powershell_completions.go
marckhouzam: This is particularly visible when I use a `--no-descriptions` flag:
```
> ./testprog
cmdColonArgs                     ( )  help                             ( )
cmdLongDesc12345678901234567890  ( )  nodesc                           ( )
compAndFile                      ( )  nofile                           ( )
completion                       ( )  specialChar                      ( )
dirOnly                          ( )  subdir                           ( )
filterFile                       ( )  vargs                            ( )
fullFile                         ( )
```

powershell_completions.go
```diff
@@ -11,90 +8,253 @@ import (
...
+    $Command = "$Command"
+
+    __%[1]s_debug ""
+    __%[1]s_debug "========= starting completion logic =========="
+    __%[1]s_debug "WordToComplete: $WordToComplete Command: $Command CursorPosition: $CursorPosition"
+
+    # The user could have moved the cursor backwards on the command-line.
+    # We need to trigger completion from the $CursorPosition location, so we need
+    # to truncate the command-line ($Command) up to the $CursorPosition location.
+    # Make sure the $Command is longer then the $CursorPosition befor we truncate.
```
marckhouzam, line 44: `befor` missing the final `e`

powershell_completions.go
```diff
@@ -11,90 +8,253 @@ import (
...
+	__%[1]s_debug "Truncated command: $Command"
+	
+    $ShellCompDirectiveError=%[3]d
+    $ShellCompDirectiveNoSpace=%[4]d
+    $ShellCompDirectiveNoFileComp=%[5]d
+    $ShellCompDirectiveFilterFileExt=%[6]d
+    $ShellCompDirectiveFilterDirs=%[7]d
+
+	# Prepare the command to request completions for the program.
+    # Split the command at the first space to seperate the programm and arguments
```
marckhouzam, line 58: `separate` instead of `seperate`

and `programm` with two `m`

powershell_completions.go
```diff
@@ -11,90 +8,253 @@ import (
...
+    $Program,$Arguments = $Command.Split(" ",2)
+    $RequestComp="$Program %[2]s $Arguments"
+    __%[1]s_debug "RequestComp: $RequestComp"
+
+    # we cannont use $WordToComplete because it
```
marckhouzam, line 63: `cannot` instead of `cannont`

powershell_completions.go
```diff
@@ -11,90 +8,253 @@ import (
...
+            "MenuComplete" {
+                # insert space after value
+                # MenuComplete will automatically show the ToolTip of 
+                # the highlighted value at the bottom of the suggestions.
+                [System.Management.Automation.CompletionResult]::new("$($comp.Name)$Space", "$($comp.Name)", 'ParameterValue', "$($comp.Description)")
+            }
+
+            # TabCompleteNext and in case we get something unknown
+            Default {
+                # Like MenuComplete but we dont want to add a space here because
```
marckhouzam, line 229: `don't` instead of `dont`

(it's my VSCode that's reporting me these typos)

powershell_completions.go
```diff
@@ -11,90 +8,253 @@ import (
...
+    $ShellCompDirectiveFilterDirs=%[7]d
+
+	# Prepare the command to request completions for the program.
+    # Split the command at the first space to seperate the programm and arguments
+    $Program,$Arguments = $Command.Split(" ",2)
+    $RequestComp="$Program %[2]s $Arguments"
+    __%[1]s_debug "RequestComp: $RequestComp"
+
+    # we cannont use $WordToComplete because it 
+    # has the wrong values if the curser was moved
```
marckhouzam, line 64: `cursor` instead of `curser`

powershell_completions.go
```diff
@@ -11,90 +8,253 @@ import (
...
+    __%[1]s_debug "Calling $RequestComp"
+    #call the command store the output in $out and redirect stderr and stdout to null
+    # $Out is an array contains each line per element
+    Invoke-Expression -OutVariable out "$RequestComp" 2>&1 | Out-Null
+
+
+    # get directive from last line
+    [int]$Directive = $Out[-1].TrimStart(':')
+    if ($Directive -eq "") {
+        # There is not directive specified
```
marckhouzam, line 100: `no` instead of `not` (that was my typo in the bash script 😄 )

powershell_completions.go
```diff
@@ -11,90 +8,253 @@ import (
...
+
+        # store temporay because switch will overwrite $_
+        $comp = $_
+
+
+        # Powershell supports three different completion modes
+        # - TabCompleteNext (default windows style - on each key press the next option is displayed)
+        # - Complete (works like bash)
+        # - MenuComplete (works like zsh)
+        # You set the mode with Set-PSReadLineKeyHandler -Key Tab -Function <mode>
```
marckhouzam, line 188: Super nice!  Each mode works great!

powershell_completions.go
```diff
@@ -11,90 +8,253 @@ import (
...
-			fmt.Fprintf(out, "\n            [CompletionResult]::new('-%s', '%s', [CompletionResultType]::ParameterName, '%s')", flag.Shorthand, flag.Shorthand, usage)
-		}
-		fmt.Fprintf(out, "\n            [CompletionResult]::new('--%s', '%s', [CompletionResultType]::ParameterName, '%s')", flag.Name, flag.Name, usage)
-	})
-
-	for _, subCmd := range cmd.Commands() {
-		usage := escapeStringForPowerShell(subCmd.Short)
-		fmt.Fprintf(out, "\n            [CompletionResult]::new('%s', '%s', [CompletionResultType]::ParameterValue, '%s')", subCmd.Name(), subCmd.Name(), usage)
+function __%[1]s_debug {
+	if (Get-Variable "POWERSHELL_COMP_DEBUG_FILE" -Scope Global) {
```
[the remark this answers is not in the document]
Luap99: OK, I will change this

powershell_completions.go
```diff
@@ -11,90 +8,253 @@ import (
...
+		"$args" | Out-File -Append -FilePath "${POWERSHELL_COMP_DEBUG_FILE}"
```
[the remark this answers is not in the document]
Luap99: Ah yeah, it's not a env var. You have to set $POWERSHELL_COMP_DEBUG_FILE as normal var in the shell. I will change this to an env var to suit the other shells and in cobra debugging.

powershell_completions.go
```diff
@@ -11,90 +8,253 @@ import (
...
+            "Complete" {
+
+                if ($Values.Length -eq 1) {
+                    __%[1]s_debug "Only one completion left"
+
+                    # insert space after value
+                    [System.Management.Automation.CompletionResult]::new("$($comp.Name)$Space", "$($comp.Name)", 'ParameterValue', "$($comp.Description)")
+
+                } else {
+                    ##TODO Trim Description if to long
```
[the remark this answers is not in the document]
Luap99: Nice, I didn't even consider this for a sec :sweat_smile:

powershell_completions.go
Luap99: That should be no problem to fix

powershell_completions.go
Luap99: No, some copy-paste problem I guess

powershell_completions.go
Luap99: I also use VSCode, is that an extension? It seams like I need that one too.

powershell_completions.go
Luap99: Yeah it works kinda, but I can't make it handle paths correctly. I can only make it complete the full path or only the file name. It won't work with relative paths. It also breaks completely with spaces.

That's why I'm thinking that sticking to standard path completion done by powershell would be better. This is just a ugly hack and doesn't seam to be supported by powershell in any way.

powershell_completions.go
Luap99: Ok that's working now.

Luap99:
@marckhouzam Thanks for the fast review. I ignore `ShellCompDirectiveFilterFileExt` and `ShellCompDirectiveFilterDirs` now because there to many corner cases I can't handle.

powershell_completions.go
marckhouzam: I use the extension "Code Spell Checker"

marckhouzam:
@Luap99 Everything works great!  As you said, the `--flag=` form would be nice, but personally, I don't think it should be a blocker. The documentation is important but after that, I would love to see this merged.

marckhouzam:
Oh, and I think you should remove `powershell_completions_test.go`.
I'm not sure if there are tests you can implement since there is no more generation of code besides the fixed script.
I removed `zsh_completions_test.go` when I did the zsh port to Go completions.

Luap99:
> Luap99 Everything works great! As you said, the `--flag=` form would be nice, but personally, I don't think it should be a blocker. The documentation is important but after that, I would love to see this merged.

I'm still thinking how this could be done. Powershell does not have `COMP_WORDBREAKS`, so I can only imagine something like this.
```
$ ./testprog --flag=[TAB]
--flag=foo
--flag=bar
--flag=foobar
```
Not very nice but it could work.

marckhouzam:
> I'm still thinking how this could be done. Powershell does not have `COMP_WORDBREAKS`, so I can only imagine something like this.
> 
> ```
> $ ./testprog --flag=[TAB]
> --flag=foo
> --flag=bar
> --flag=foobar
> ```
> 
> Not very nice but it could work.

This is what `helm` and `kubectl` currently offer for their zsh completion (not yet migrated to the new Cobra zsh v2):
```
kubectl --context=<TAB>
--context=acc       --context=cor       --context=dev       --context=lab       --context=minikube
--context=bi.dev    --context=default   --context=infra     --context=lab2      --context=prod
```

I think it is just fine.

Luap99:
> This is what helm and kubectl currently offer for their zsh completion (not yet migrated to the new Cobra zsh v2):

I'm curios does that work with file path completion?

I got it working for powershell with arguments, but there is no way to make it work file completion.

Luap99:
Ok, I think the script is in good shape now. I added documentation.

One question remains, how should the API look like?

marckhouzam:
> > This is what helm and kubectl currently offer for their zsh completion (not yet migrated to the new Cobra zsh v2):
> 
> I'm curious does that work with file path completion?

I hadn't realized, but no, it does not work with file completion.  So, currently, for zsh (not using Cobra's zsh support):
```
$ kubectl --kubeconfig /User<TAB>     # Works
$ kubectl --kubeconfig=/User<TAB>     # Does not work
```
> I got it working for powershell with arguments, but there is no way to make it work with file completion.

I think that is fine.

shell_completions.md
```diff
@@ -431,4 +439,38 @@ search  show  status
 
 ## PowerShell completions
 
-Please refer to [PowerShell Completions](powershell_completions.md) for details.
+Cobra supports native PowerShell completions generated from the root `cobra.Command`. You can use the `command.GenPowerShellCompletion()` or `command.GenPowerShellCompletionFile()` functions. You must provide these functions with a parameter indicating if the completions should be annotated with a description. Cobra will provide the description automatically based on usage information. You can choose to make this option configurable by your users.
```
marckhouzam, line 442: This needs to be update to follow the API that will be chosen (which cannot be this one).

shell_completions.md
```diff
@@ -431,4 +439,38 @@ search  show  status
...
+
+The script is designed to support all three Powershell completion modes:
+* TabCompleteNext (default windows style - on each key press the next option is displayed)
+* Complete (works like bash)
+* MenuComplete (works like zsh)
+You set the mode with `Set-PSReadLineKeyHandler -Key Tab -Function <mode>`. Descriptions are only displayed when using the `Complete` or `MenuComplete` mode.
```
marckhouzam, line 448: Need a newline before this line.  GitHub currently puts on the same line as the last bullet, can be seen here:
https://github.com/Luap99/cobra/blob/powershell-custom-comp/shell_completions.md

powershell_completions.go
marckhouzam: Since we know this change cannot be done, I suggest you implement whatever API you prefer and restore the previous methods.  This will at least put the PR in a state that can be merged.

powershell_completions.go
```diff
@@ -11,90 +8,254 @@ import (
...
-	fmt.Fprintf(buf, powerShellCompletionTemplate, c.Name(), c.Name(), subCommandCases.String())
-
+	genPowerShellComp(buf, c.Name(), includeDesc)
 	_, err := buf.WriteTo(w)
 	return err
 }
 
 // GenPowerShellCompletionFile generates PowerShell completion file.
-func (c *Command) GenPowerShellCompletionFile(filename string) error {
+func (c *Command) GenPowerShellCompletionFile(filename string, includeDesc bool) error {
```
marckhouzam, line 253: To make sure we don't forget, this must be reverted as it breaks compatibility.

Luap99:
I just tested this on windows 10 with the default powershell v5.1 and I noticed one problem. `testprog.exe --[TAB]` doesn't work but I think this is not a bug in my script since Powershell doesn't call the completion script at all (there is no debug output). Also ```testprog.exe `--[TAB]``` and `testprog.exe --r[TAB]` are working. Since `testprog.exe --[TAB]` works in Powershell v7, I call this a bug in Powershell v5.1.
I think there is nothing I can do other then adding a note in the docs.

Apart from that, I think it works very well.

jpmcb:
I have no way of testing this on a windows machine. Are there other's in the community that would be able to lend a hand and test this out?

marckhouzam:
> I just tested this on windows 10 with the default powershell v5.1 and I noticed one problem. `testprog.exe --[TAB]` doesn't work but I think this is not a bug in my script since Powershell doesn't call the completion script at all (there is no debug output). Also `` testprog.exe `--[TAB] `` and `testprog.exe --r[TAB]` are working. Since `testprog.exe --[TAB]` works in Powershell v7, I call this a bug in Powershell v5.1.
> I think there is nothing I can do other then adding a note in the docs.

I can confirm I see this same behavior with powershell 5.1.  I think we'll just have to live with it.

I've tested on Mac with `pwsh 7.0.2` and on Windows 10 with `powershell 5.1` by using this PR with `helm`.  
Everything works really well.
Beyond standard completions, I've tested:
* `ShellDirectiveNoSpace`
* `ShellDirectiveNoFileComp`
* flag completion with the `=` form and the normal form
* descriptions
* all three types of completions for powershell

From a functionality perspective, not only do I feel this is complete, but I believe it is dramatically better than what we have right now for powershell.

@jpmcb There is still a need for guidance on what form the API should take for programs to call this.

Luap99:
@marckhouzam Thanks for testing.

I added one small thing to properly escape spacial chars with a backtick. So they can safely be used as input.

I also changed the API to `GenPowerShellCompletion()` and `GenPowerShellCompletionWithDesc()` to keep it backwards compatible.

marckhouzam (approved):
Very nice.
Could you just add `powershell` in `shell_completions.md` in the `Descriptions for completions` section?

rshoemaker:
Hi,

I've been experimenting with your patch and think I've found an issue with it.  If I understand how your patch works, it seems to be generating the completions by interrogating the binary itself via "--help" or usage - I'm not really sure.  The issue with that is that you might have a different command tree that you want to generate completions for.  In my case, we have a complicated Cobra command tree that spans multiple Go binaries.  One binary serves as the core CLI and then we have add-on modules that provide additional sub-commands.  In order to generate completion bindings for Bash and ZSH, we generate a superset of the Cobra CLI commands and then call the GenXXXCompletion() funcs to create a completion file that spans all of our binaries (but they all share a common Cobra root command).  Your patch uses the CLI visible only in the current binary, which may not be the command you called GenXXXCompletion() func on. 

The attached example illustrates the issue.  We have a "root-cmd" with one sub-command "child-cmd".  Then we have an additional Cobra command tree that contains "root-cmd", "child-cmd", and "generated-cmd" (representing a command from another binary).  When we generate Powershell completions on the additional Cobra command tree using your patch, we only get completions for "child-cmd" and NOT "generated-cmd".  This patch seems to break the semantics of the GenXXXCompletion() func and is different from the approach used in the ZSH and Bash versions of the func.

[main.zip.zip](https://github.com/spf13/cobra/files/5271247/main.zip.zip)

rshoemaker:
And by the way, we couldn't get powershell completion working at all without your patch, so it's a big improvement.  We'd like to see this get merged back in, especially if you can tweak it to generate completions for the specified command tree.

Luap99:
@rshoemaker Thanks for your feedback. Your specific problem is not directly related to this PR. To see how the completions are provided see: https://github.com/spf13/cobra/blob/master/shell_completions.md#debugging. This script only parses this output and provides the completions. Currently (master branch) zsh and fish are already using this type of completion script and will face the same problem. This PR only adds support for powershell.

However @marckhouzam (the original author for this type of completions) has a PR which I believe could add support for your use case https://github.com/spf13/cobra/pull/1161#issuecomment-674427951. He can properly clarify.

Luap99:
Also @marckhouzam do you think it would make sense to disable `GenXXXCompletion()` on child commands for custom go completion scripts? This only results in a non working completion script.

rshoemaker:
HI @Luap99 , @marckhouzam ,

Thanks for the quick reply.  I ran some more experiments where I dump all the shell completions for our root command and our generated command tree:

`	_ = RootCmd.GenPowerShellCompletionFile("root.pwr")
	_ = GeneratedCli.GenPowerShellCompletionFile("gen.pwr")

	_ = RootCmd.GenZshCompletionFile("root.zsh")
	_ = GeneratedCli.GenZshCompletionFile("gen.zsh")

	_ = RootCmd.GenBashCompletionFile("root.bsh")
	_ = GeneratedCli.GenBashCompletionFile("gen.bsh")
`

And what I can see by diff'ing the text is that the zsh and bash completion files for "GeneratedCli" contain all the artifacts for the full "GeneratedCli" command tree, while the completion files for "RootCmd" do not.  The two PS completion files are identical and only include artifacts for the "RootCmd".  So, based on experimental data, it seems like your comment about zsh/bash isn't right.

We're running v1.0.0 with your pull request applied.

marckhouzam:
Hi @rshoemaker.

With Cobra 1.0.0, completion for bash, zsh and powershell generate the exhaustive list of completions, as you have seen.  Completion for Fish shell however user a standard generic script that is the same for all commands and binaries.

For the next release (as currently on the master branch), zsh will start using this same approach of having a generic script, and the current powershell PR of @Luap99 also uses that standardized approach.  Bash completion will continue to work the old way, unless #1146 moves forward and also aligns bash completion with the other shells.

I haven't had time to look at your specific case but as pointed out by @Luap99, you may have a standardized way forward, which would be to start using `ValidArgsFunction` for your commands and have the commands call your sub-commands binaries with the `__complete` command.

We do this to handle `helm` plugins, where each plugin is a sub-command of the `helm` rootCmd, but delegates execution and completion to another binary.  Here is the code to delegate the completion to the plugin binary.  You may choose to adapte that code to call you sub-binaries with the `__complete` command:
https://github.com/helm/helm/blob/master/cmd/helm/load_plugins.go#L323-L377

As pointed out by @Luap99 you may need this PR to get flag-name completion to work: https://github.com/spf13/cobra/pull/1161

powershell_completions.go
SilkyFowl: Good!!

powershell_completions.go
```diff
@@ -11,90 +8,278 @@ import (
...
+                    }
+
+                    # Check for empty description and only add parentheses if needed
+                    if ($($comp.Description) -eq " " ) {
+                        $Description = ""
+                    } else {
+                        $Description = "  ($($comp.Description))"
+                    }
+
+                    [System.Management.Automation.CompletionResult]::new("$($comp.Name)$Description", "$($comp.Name)$Description", 'ParameterValue', "$($comp.Description)")
```
SilkyFowl, line 220: Why set `completionText` param to `"$($comp.Name)$Description"`? 
I couldn't figure out when I checked the code, but this seems to be the only constructor that adds a `$Description` to `completionText` parameter, is there any special meaning to it?

```pwsh
 $myParams = @{
  Native=$true
  CommandName='Foo'
}

$comps=@(
  @{n='add';d=':Add Foo'}
  @{n='list';d=':List Foo'}
  @{n='remove';d=':Remove Foo'}
  @{n='v';d=''}
)

Register-ArgumentCompleter @myParams -ScriptBlock {
  param($WordToComplete,$CommandAst,$CursorPosition)
    $comps.where{$_.n -match "^$WordToComplete"}  | % {
      $Mode = (Get-PSReadLineKeyHandler | Where-Object {$_.Key -eq "Tab" }).Function
        
      [CompletionResult]::new("$($_.n) $($_.d)", "$($_.n) $($_.d)", [CompletionResultType]::ParameterValue, " ")
    }
}
```

```console
~
❯ foo afdaf
add :Add Foo        list :List Foo      remove :Remove Foo  v

~
❯ foo add :Add Foo
```

Luap99: As you might know powershell support 3 different completion modes:
- `TabCompleteNext`
- `Complete`
- `MenuComplete`

By default (powershell) only `MenuComplete` will display a description below the suggestions.
Cobra allows descriptions in autocompletions for zsh and fish. PR #1146 adds description support to bash via a little hack. I adopted this hack here to also allow description to be displayed with the `Complete` mode since powershell does not support this natively . The idea is that the descriptions are appended to the real completion values. However if only one matching completion is left the script wont append the description so that the shell can autocomplete the correct value.

Users can choose if they want descriptions to be displayed while generating the scripts with `GenPowerShellCompletion` for no descriptions and  `GenPowerShellCompletionWithDesc` for descriptions.
```
# With descriptions and Mode 'Complete'
$ helm s[tab]
search  (search for a keyword in charts)  show  (show information of a chart)  status  (displays the status of the named release)
# if you type 'se' you will get this autocompleted to 'search'

# With descriptions and Mode 'MenuComplete' The description of the current selected value will be displayed below the suggestions.
$ helm s[tab]
search    show     status  
search for a keyword in charts

# Without descriptions
$ helm s[tab]
search  show  status
```

SilkyFowl: Nice idea!!
I understand the intent of the code.
This is not a problem.

SilkyFowl (approved):

marckhouzam:
@jpmcb what do you think about merging this powershell completion patch?  From my testing @Luap99 did a spectacular job with this.  

Furthermore, it seems no one can get the existing powershell completion to work at all currently, so there is no risk of breaking anything 😄   In fact, I checked-out the commit from the original powershell completion pr #797 and still only got flag completion to work; it seems the feature never properly worked.

We need some good material for the next release and I thought this would be a great one.

jpmcb (approved):
Thanks for the time and patience on this one - I'll go ahead and merge since there seems to be a consensus that this is indeed working on windows / powershell. 

If anyone in the community is trying this out and finds issues, please feel free to open a new issue!!

PR #1217, opened 2020-09-03
Improving the "required flag(s) "%s" not set" error message by not printing "(s)" if a single required flag is missing

nestoroprysk:
Improving the "required flag(s) "%s" not set" error message by not printing "(s)" if a single required flag is missing.

If a single flag is missing, the error message is of the form "required flag "%s" not set". 

Otherwise, the error message remains unchanged.

### DIFF
--- a/command.go
+++ b/command.go
@@ -1004,7 +1004,10 @@ func (c *Command) validateRequiredFlags() error {
 	})
 
 	if len(missingFlagNames) > 0 {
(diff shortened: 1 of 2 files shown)


nestoroprysk:
signed in the agreement

jharshman:
@nestoroprysk please add a description to your PR.
Add details that will help maintainers get on the same page as you about this change.

nestoroprysk:
@jharshman updated the description and resolved conflicts

marckhouzam:
I think this makes sense but needs a proper review.  Re-opening.

Issue #1219, opened 2020-09-05
go get error: unexpected module path "go.etcd.io/bbolt"


2014BDuck:
```
➜  tools go get -u github.com/spf13/cobra/cobra@v1.0.0
go: finding github.com/spf13/cobra/cobra v1.0.0
go: finding github.com/tmc/grpc-websocket-proxy latest
go: finding github.com/armon/consul-api latest
go: finding gopkg.in/check.v1 latest
go: finding github.com/golang/groupcache latest
go: finding golang.org/x/lint latest
go: finding golang.org/x/time latest
go: finding github.com/golang/glog latest
go: finding github.com/xiang90/probing latest
go: finding golang.org/x/net latest
go: finding github.com/coreos/go-systemd latest
go: finding golang.org/x/sync latest
go: finding golang.org/x/oauth2 latest
go: github.com/coreos/bbolt@v1.3.5: parsing go.mod: unexpected module path "go.etcd.io/bbolt"
go: github.com/coreos/go-systemd@v0.0.0-20200630094324-1ac580e0cdfa: go.mod has post-v0 module path "github.com/coreos/go-systemd/v22" at revision 1ac580e0cdfa
go: finding golang.org/x/crypto latest
go get: error loading module requirements
➜  tools
➜  tools go env
GOARCH="amd64"
GOBIN=""
GOCACHE="/Users/jiekunzhu/Library/Caches/go-build"
GOEXE=""
GOFLAGS=""
GOHOSTARCH="amd64"
GOHOSTOS="darwin"
GOOS="darwin"
GOPATH="/Users/jiekunzhu/vhost/go/"
GOPROXY=""
GORACE=""
GOROOT="/usr/local/Cellar/go@1.12/1.12.17/libexec"
GOTMPDIR=""
GOTOOLDIR="/usr/local/Cellar/go@1.12/1.12.17/libexec/pkg/tool/darwin_amd64"
GCCGO="gccgo"
CC="clang"
CXX="clang++"
CGO_ENABLED="1"
GOMOD="/Users/jiekunzhu/tools/go.mod"
CGO_CFLAGS="-g -O2"
CGO_CPPFLAGS=""
CGO_CXXFLAGS="-g -O2"
CGO_FFLAGS="-g -O2"
CGO_LDFLAGS="-g -O2"
PKG_CONFIG="pkg-config"
[...]

Issue #1215, opened 2020-09-03
"ambiguous import"-error when trying to "go get" cobra


BeyondEvil:
When following the [installation instructions](https://github.com/spf13/cobra#installing), I get:

```sh
$ go get -u github.com/spf13/cobra/cobra
go: github.com/spf13/cobra/cobra upgrade => v0.0.0-20200826151851-02a0d2fbc9e6
go get github.com/spf13/cobra/cobra: ambiguous import: found package github.com/spf13/cobra/cobra in multiple modules:
	github.com/spf13/cobra v1.0.0 (/Users/jimbrannlund/go/pkg/mod/github.com/spf13/cobra@v1.0.0/cobra)
	github.com/spf13/cobra/cobra v0.0.0-20200826151851-02a0d2fbc9e6 (/Users/jimbrannlund/go/pkg/mod/github.com/spf13/cobra/cobra@v0.0.0-20200826151851-02a0d2fbc9e6)
```

macOS Mojave 10.14.6
```
Darwin Jims-MacBook-Pro-2.local 18.7.0 Darwin Kernel Version 18.7.0: Thu Jun 20 18:42:21 PDT 2019; root:xnu-4903.270.47~4/RELEASE_X86_64 x86_64
```

```
$ go version
go version go1.14.3 darwin/amd64
```

I'm using the new way with Go Modules, that is no `GOPATH`
```
Jims-MacBook-Pro-2:slack-cli jimbrannlund$ echo $GOPATH

Jims-MacBook-Pro-2:slack-cli jimbrannlund$
````

Directory contents:
```
Jims-MacBook-Pro-2:slack-cli jimbrannlund$ ls -la
total 56
drwxr-xr-x   8 jimbrannlund  staff    256 Sep  3 09:54 .
drwxr-xr-x  49 jimbrannlund  staff   1568 Sep  2 21:32 ..
drwxr-xr-x  13 jimbrannlund  staff    416 Sep  2 21:33 .git
-rw-r--r--   1 jimbrannlund  staff     17 May 25 14:12 README.md
drwxr-xr-x   3 jimbrannlund  staff     96 Sep  3 09:54 cmd
-rw-r--r--   1 jimbrannlund  staff    226 Sep  3 10:09 go.mod
-rw-r--r--   1 jimbrannlund  staff  13321 Sep  3 10:09 go.sum
-rw-r--r--   1 jimbrannlund  staff   1174 Sep  2 20:48 main.go
```

[...]

PR #1139, opened 2020-06-15
modules: add a secondary go.mod to segregate CLI dependencies

MichaelMure:
Fix https://github.com/spf13/cobra/issues/1138

This PR add a secondary `go.mod` specifically for the CLI tool which has the effect of removing its dependencies from the main `go.mod` file imported in projects.

This simple change reduced the number of dependencies of https://github.com/MichaelMure/git-bug from 313 to 148.

PR #1233, opened 2020-09-23
Remove secondary go mod to prevent broken `go get`

jpmcb:
This PR addresses problems with `go get` in issues #1215 and #1219 

The maintainers have discussed the need for a second go module. And despite the CLI package requiring a few additional modules, the pain for users around getting the cobra project as expected does not make this desirable behavior.

Therefore, this effectively reverts PR #1139


jharshman (approved):

umarcor:
According to the changes in this PR, ~299/313=95.5% of the dependencies correspond to the CLI package and are not required by users of the library. While I understand the motivation not to handle the complexity of having two `go.mod` files given the structure of the repo, describing it as "a few additional modules" feels quite unfair.

This PR decided to prioritize the pain of some users having to specify a semver, over others aiming at having an efficient ecosystem. That's ok. Yet, I wish some reference was given about the technical constraints that prevent(ed) `go get github.com/spf13/cobra/cobra` from working. My understanding is that the CLI should not be a child of the library, but a sibling, because the CLI is not part of the library. That would probably avoid the ambiguity, while allowing to have separate dependencies.

marckhouzam:
I too wish we could have an efficient way to avoid those unnecessary imports. 

From my understanding once a project is off the ground, the Cobra cli isn't really that valuable, right?  And yet every project using Cobra keeps downloading it.

I'm not very versed in go packaging but the idea of siblings packages sounds interesting.  In my mind, my project should depend on the library only, and I should install the cli independently if I need it.

umarcor:
@marckhouzam, I would say that we'll all agree on that. The problem is how to deal with it. This repo is in spf13's namespace; hence options are:

- Create spf13/cobra-cli, and move the CLI there.
- Create org cobra and move cobra/lib, cobra/cobra, cobra/cli, whatever.
- Move the current content in the root of this repo (except subdir 'cobra') to a new subdir (say 'lib'). Probably rename 'cobra' to 'cli'.

I prefer the last option, as the result would be:

- `go get github.com/spf13/cobra/lib`
- `go get github.com/spf13/cobra/cli`

each with a separate `go.mod` and `go.sum` file, but both together in the same git history.

Nevertheless, the problem is that any of those are breaking changes. My conviction is that, the later it's fixed, the worse it will be. However, I ain't a maintainer.

PR #1247, opened 2020-10-05
fix: fish output redirection

### LINKED ISSUE #1246: Fish completion error
```sh
Expected a command, but instead found a redirection
complete --do-complete "faas " &> /dev/null
```
I think it should be:
```sh
complete --do-complete "faas " 2> /dev/null
```

matejvasek:
resolves #1246

### DIFF
--- a/fish_completions.go
+++ b/fish_completions.go
@@ -164,7 +164,8 @@ end
 # so we can properly delete any completions provided by another script.
 # The space after the the program name is essential to trigger completion for the program
 # and not completion of the program name itself.
-complete --do-complete "%[2]s " &> /dev/null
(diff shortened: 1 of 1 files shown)


marckhouzam:
Thanks @matejvasek, to make sure we don't break this again, could you put a comment not to use `&>` because Fish 3.0 does not support it?

matejvasek:
@marckhouzam comment added.

marckhouzam (approved):
Thanks.
I've tested this with Fish 3.1 and things continue to work as expected.

@jharshman this looks safe.

jharshman (approved):
thank you @matejvasek for your PR and @marckhouzam for your review!

PR #1213, opened 2020-08-28
Fix zsh completion handling of nospace and file completion

marckhouzam:
Fixes #1211
Fixes #1212

This PR improves the zsh completion script to allow it to properly handle `ShellDirectiveCompletionNoSpace` and file completion all the time.  Specifically, there was an issue where when `ValidArgsFunction` would always return completions even if they don't match what the user types, the script would not behave right.  Please refer to #1211 and #1212 for specific details.

So this PR no longer counts the number of completions it receives, but instead relies on the zsh `_describe` function and its return code to handle things cleanly.

PR #1249, opened 2020-10-07
Fix multiple fish completion issues

### LINKED ISSUE #1214: Fish completion fails with quotes in argument
The fish completion script prints ugly traces if quotes are not balanced in the argument.

To reproduce: `./testprog '[TAB]`

```fish
> ./testprog '- (line 1): Unexpected end of string, quotes are not balanced
./testprog __complete ' 
                      ^
in command substitution
        called on line 35 of file -
in function '__testprog_perform_completion' with arguments './testprog\ \''
        called on line 1 of file -
in command substitution
        called on line 71 of file -
in function '__testprog_prepare_completions'
in command substitution
```

This happens with single and double quotes. Technically the completion still works and offers file completion but the traces makes this unusable.

I also checked this with zsh, bash, #1146 and #1208 but only fish is affected. All other shells offer file completion, which is in my eyes the right thing.

### LINKED ISSUE #1248: Fish completion does not handle "nospace" or file completion properly
This is the same bug that is present for `zsh` and described in #1211 and #1212.

**Bug 1: File completion is not respected**

The `fish` completion script performs file completion only if it receives no completions:
https://github.com/spf13/cobra/blob/b97b5ead31f7d34f764ac8666e40c214bb8e06dc/fish_completions.go#L151

However, it is valid for `ValidArgsFunction` to return completions that don't apply to the current `toComplete` prefix.  This means file completion should be performed, but the script does not realize it as it still believes it received valid completions.

Say I have the following `ValidArgsFunction` for command `withfile`:
```
    // We do not consider the toComplete prefix and always return a completion
    return []string{"why"}, cobra.ShellCompDirectiveDefault
},
``` 
If I have a file called `myfile` and run in `fish`:
```
$ ./testprog withfile my<TAB>
```
I would expect to get file completion and get `myfile` as a completion, but I don't.  This is because
```
$ ./testprog __complete withfile my<ENTER>
why
:0
```
Notice that above, the script would receive a completion even though it is not technically valid.  This is allowed.  But then, the script does not realize it must do file completion.

**Bug 2: NoSpace is not respected all the time**

For the exact same reason, `ShellCompDirectiveNoSpace` also fails when the `__complete` command returns completions that don't match the `toComplete` prefix.

Say I have the following `ValidArgsFunction` for command `nospace`:
```
		if len(args) == 0 {
			return []string{"when\tThe time", "who\tThe person"}, cobra.ShellCompDirectiveNoSpace
		}
		return []string{"why\tThe reason"}, cobra.ShellCompDirectiveNoSpace
	},
``` 

If I run in `fish`:
```
$ ./testprog nospace whe<TAB>
```
a space is added after `./testprog nospace when ` even though the `ShellCompDirectiveNoSpace` was given.

This is because the script only handles `ShellCompDirectiveNoSpace` if a single completion is returned:
https://github.com/spf13/cobra/blob/b97b5ead31f7d34f764ac8666e40c214bb8e06dc/fish_completions.go#L144

But my `ValidArgsFunction` does not filter the completions but instead returns all of them, leaving `fish` to filter them.  In this case:
```
$ ./testprog __complete nospace whe<ENTER>
when	The time
who	The person
:2
Completion ended with directive: ShellCompDirectiveNoSpace
```
Notice that with the result above the completion script would receive two completions even though the `who` completion is not technically valid. This is allowed.

**Bug 3: NoSpace not respected with descriptions**

Finally, the nospace logic for fish does not handle the case when there is a description.  To achieve "nospace" the script adds a second completion with an extra "." to prevent the shell from adding a space. However if there is a description, the extra "." is added to the description instead of the completion, and the nospace directive does not work. 

I will post a PR to handle this better.

### LINKED ISSUE #1303: Fish completion fails when command line starts with a space
If the current command line starts with a space the completion fails. It will only provide file completion.

Note: This is valid syntax used to prevent the shell from writing the command to the history.

BASH_COMP_DEBUG_FILE output:
```
commandLine is:  podman 
Starting __podman_perform_completion with:  podman 
args:  podman 
last arg: 
Setting emptyArg
emptyArg: ""
Cannot find . No completions.
Completion results: 
No completion, probably due to a failure
```

I think this line causes the problem:
https://github.com/spf13/cobra/blob/7df62f7668c7fbeafd6dbb45ad0294f4ea08c0ec/fish_completions.go#L33
We need to trim $argv before we can split on space

This bug was originally reported by @pastalian in https://github.com/containers/podman/issues/8829

### LINKED ISSUE #1306: Fish completion does not work with env vars in the program path
Using an environment variable or a `~` when using a path for the program breaks shell completion for fish.
For instance:
```
$HOME/testprog [TAB]
~/testprog [TAB]
```
currently don't work for fish. This is because of the use of `type -q` in the script.

marckhouzam:
Fixes #1248
Fixes #1303
Fixes #1214
Fixes #1306

This PR fixes multiple bugs in the fish completion script.

Please ses each issue for details.

### DIFF
--- a/fish_completions.go
+++ b/fish_completions.go
@@ -21,7 +21,7 @@ func genFishComp(buf io.StringWriter, name string, includeDesc bool) {
 	WriteStringAndCheck(buf, fmt.Sprintf("# fish completion for %-36s -*- shell-script -*-\n", name))
 	WriteStringAndCheck(buf, fmt.Sprintf(`
 function __%[1]s_debug
-    set file "$BASH_COMP_DEBUG_FILE"
+    set -l file "$BASH_COMP_DEBUG_FILE"
     if test -n "$file"
         echo "$argv" >> $file
     end
@@ -30,26 +30,18 @@ end
 function __%[1]s_perform_completion
     __%[1]s_debug "Starting __%[1]s_perform_completion"
 
-    set args (string split -- " " (string trim -l (commandline -c)))
-    set lastArg "$args[-1]"
+    # Extract all args except the last one
+    set -l args (commandline -opc)
+    # Extract the last arg and escape it in case it is a space
+    set -l lastArg (string escape -- (commandline -ct))
 
     __%[1]s_debug "args: $args"
     __%[1]s_debug "last arg: $lastArg"
 
-    set emptyArg ""
-    if test -z "$lastArg"
-        __%[1]s_debug "Setting emptyArg"
-        set emptyArg \"\"
-    end
-    __%[1]s_debug "emptyArg: $emptyArg"
+    set -l requestComp "$args[1] %[3]s $args[2..-1] $lastArg"
 
-    set requestComp "$args[1] %[3]s $args[2..-1] $emptyArg"
     __%[1]s_debug "Calling $requestComp"
-
-    # Call the command as a sub-shell so that we can redirect any errors
-    # For example, if $requestComp has an unmatched quote
-    # https://github.com/spf13/cobra/issues/1214
-    set results (fish -c "$requestComp" 2> /dev/null)
+    set -l results (eval $requestComp 2> /dev/null)
 
     # Some programs may output extra empty lines after the directive.
     # Let's ignore them or else it will break completion.
@@ -64,12 +56,12 @@ function __%[1]s_perform_completion
         end
     end
 
-    set comps $results[1..-2]
-    set directiveLine $results[-1]
+    set -l comps $results[1..-2]
+    set -l directiveLine $results[-1]
 
     # For Fish, when completing a flag with an = (e.g., <program> -n=<TAB>)
     # completions must be prefixed with the flag
-    set flagPrefix (string match -r -- '-.*=' "$lastArg")
+    set -l flagPrefix (string match -r -- '-.*=' "$lastArg")
 
     __%[1]s_debug "Comps: $comps"
     __%[1]s_debug "DirectiveLine: $directiveLine"
@@ -92,7 +84,7 @@ function __%[1]s_prepare_completions
     # Start fresh
     set --erase __%[1]s_comp_results
 
-    set results (__%[1]s_perform_completion)
+    set -l results (__%[1]s_perform_completion)
     __%[1]s_debug "Completion results: $results"
 
     if test -z "$results"
@@ -101,39 +93,39 @@ function __%[1]s_prepare_completions
         return 1
     end
 
-    set directive (string sub --start 2 $results[-1])
+    set -l directive (string sub --start 2 $results[-1])
     set --global __%[1]s_comp_results $results[1..-2]
 
     __%[1]s_debug "Completions are: $__%[1]s_comp_results"
     __%[1]s_debug "Directive is: $directive"
 
-    set shellCompDirectiveError %[4]d
-    set shellCompDirectiveNoSpace %[5]d
-    set shellCompDirectiveNoFileComp %[6]d
-    set shellCompDirectiveFilterFileExt %[7]d
-    set shellCompDirectiveFilterDirs %[8]d
+    set -l shellCompDirectiveError %[4]d
+    set -l shellCompDirectiveNoSpace %[5]d
+    set -l shellCompDirectiveNoFileComp %[6]d
+    set -l shellCompDirectiveFilterFileExt %[7]d
+    set -l shellCompDirectiveFilterDirs %[8]d
 
     if test -z "$directive"
         set directive 0
     end
 
-    set compErr (math (math --scale 0 $directive / $shellCompDirectiveError) %% 2)
+    set -l compErr (math (math --scale 0 $directive / $shellCompDirectiveError) %% 2)
     if test $compErr -eq 1
         __%[1]s_debug "Received error directive: aborting."
         # Might as well do file completion, in case it helps
         return 1
     end
 
-    set filefilter (math (math --scale 0 $directive / $shellCompDirectiveFilterFileExt) %% 2)
-    set dirfilter (math (math --scale 0 $directive / $shellCompDirectiveFilterDirs) %% 2)
+    set -l filefilter (math (math --scale 0 $directive / $shellCompDirectiveFilterFileExt) %% 2)
+    set -l dirfilter (math (math --scale 0 $directive / $shellCompDirectiveFilterDirs) %% 2)
     if test $filefilter -eq 1; or test $dirfilter -eq 1
         __%[1]s_debug "File extension filtering or directory filtering not supported"
         # Do full file completion instead
         return 1
     end
 
-    set nospace (math (math --scale 0 $directive / $shellCompDirectiveNoSpace) %% 2)
-    set nofiles (math (math --scale 0 $directive / $shellCompDirectiveNoFileComp) %% 2)
+    set -l nospace (math (math --scale 0 $directive / $shellCompDirectiveNoSpace) %% 2)
+    set -l nofiles (math (math --scale 0 $directive / $shellCompDirectiveNoFileComp) %% 2)
 
     __%[1]s_debug "nospace: $nospace, nofiles: $nofiles"
 
@@ -143,34 +135,34 @@ function __%[1]s_prepare_completions
     # may not already be filtered so as to allow fish to match on different
     # criteria than the prefix.
     if test $nospace -ne 0; or test $nofiles -eq 0
-        set prefix (commandline -t)
+        set -l prefix (commandline -t | string escape --style=regex)
         __%[1]s_debug "prefix: $prefix"
 
-        set completions
-        for comp in $__%[1]s_comp_results
-            if test (string match -e -r -- "^$prefix" "$comp")
-                set -a completions $comp
-            end
-        end
+        set -l completions (string match -r -- "^$prefix.*" $__%[1]s_comp_results)
         set --global __%[1]s_comp_results $completions
         __%[1]s_debug "Filtered completions are: $__%[1]s_comp_results"
 
         # Important not to quote the variable for count to work
-        set numComps (count $__%[1]s_comp_results)
+        set -l numComps (count $__%[1]s_comp_results)
         __%[1]s_debug "numComps: $numComps"
 
         if test $numComps -eq 1; and test $nospace -ne 0
-            # To support the "nospace" directive we trick the shell
-            # by outputting an extra, longer completion.
-            # We must first split on \t to get rid of the descriptions because
-            # the extra character we add to the fake second completion must be
-            # before the description.  We don't need descriptions anyway since
-            # there is only a single real completion which the shell will expand
-            # immediately.
-            __%[1]s_debug "Adding second completion to perform nospace directive"
-            set split (string split --max 1 \t $__%[1]s_comp_results[1])
-            set --global __%[1]s_comp_results $split[1] $split[1].
-            __%[1]s_debug "Completions are now: $__%[1]s_comp_results"
+            # We must first split on \t to get rid of the descriptions to be
+            # able to check what the actual completion will be.
+            # We don't need descriptions anyway since there is only a single
+            # real completion which the shell will expand immediately.
+            set -l split (string split --max 1 \t $__%[1]s_comp_results[1])
+        
+            # Fish won't add a space if the completion ends with any
+            # of the following characters: @=/:.,
+            set -l lastChar (string sub -s -1 -- $split)
+            if not string match -r -q "[@=/:.,]" -- "$lastChar"                
+                # In other cases, to support the "nospace" directive we trick the shell
+                # by outputting an extra, longer completion.
+                __%[1]s_debug "Adding second completion to perform nospace directive"
+                set --global __%[1]s_comp_results $split[1] $split[1].
+                __%[1]s_debug "Completions are now: $__%[1]s_comp_results"
+            end
         end
 
         if test $numComps -eq 0; and test $nofiles -eq 0
@@ -186,10 +178,14 @@ end
 
 # Since Fish completions are only loaded once the user triggers them, we trigger them ourselves
 # so we can properly delete any completions provided by another script.
-# The space after the program name is essential to trigger completion for the program
-# and not completion of the program name itself.
-complete --do-complete "%[2]s " > /dev/null 2>&1
-# Using '> /dev/null 2>&1' since '&>' is not supported in older versions of fish.
+# Only do this if the program can be found, or else fish may print some errors; besides,
+# the existing completions will only be loaded if the program can be found.
+if type -q "%[2]s"
+    # The space after the program name is essential to trigger completion for the program
+    # and not completion of the program name itself.
+    # Also, we use '> /dev/null 2>&1' since '&>' is not supported in older versions of fish.
+    complete --do-complete "%[2]s " > /dev/null 2>&1
+end
 
 # Remove any pre-existing completions for the program since we will be handling all of them.
 complete -c %[2]s -e
--- a/fish_completions.go
+++ b/fish_completions.go
@@ -43,16 +43,13 @@ function __%[1]s_perform_completion
     end
     __%[1]s_debug "emptyArg: $emptyArg"
 
-    if not type -q "$args[1]"
-        # This can happen when "complete --do-complete %[2]s" is called when running this script.
-        __%[1]s_debug "Cannot find $args[1]. No completions."
-        return
-    end
-
     set requestComp "$args[1] %[3]s $args[2..-1] $emptyArg"
     __%[1]s_debug "Calling $requestComp"
 
-    set results (eval $requestComp 2> /dev/null)
+    # Call the command as a sub-shell so that we can redirect any errors
+    # For example, if $requestComp has an unmatched quote
+    # https://github.com/spf13/cobra/issues/1214
+    set results (fish -c "$requestComp" 2> /dev/null)
 
     # Some programs may output extra empty lines after the directive.
     # Let's ignore them or else it will break completion.
@@ -66,6 +63,7 @@ function __%[1]s_perform_completion
             break
         end
     end
+
     set comps $results[1..-2]
     set directiveLine $results[-1]
 
--- a/fish_completions.go
+++ b/fish_completions.go
@@ -28,9 +28,9 @@ function __%[1]s_debug
 end
 
 function __%[1]s_perform_completion
-    __%[1]s_debug "Starting __%[1]s_perform_completion with: $argv"
+    __%[1]s_debug "Starting __%[1]s_perform_completion"
 
-    set args (string split -- " " "$argv")
+    set args (string split -- " " (commandline -c))
     set lastArg "$args[-1]"
 
     __%[1]s_debug "args: $args"
@@ -84,31 +84,22 @@ function __%[1]s_perform_completion
     printf "%%s\n" "$directiveLine"
 end
 
-# This function does three things:
-# 1- Obtain the completions and store them in the global __%[1]s_comp_results
-# 2- Set the __%[1]s_comp_do_file_comp flag if file completion should be performed
-#    and unset it otherwise
-# 3- Return true if the completion results are not empty
+# This function does two things:
+# - Obtain the completions and store them in the global __%[1]s_comp_results
+# - Return false if file completion should be performed
 function __%[1]s_prepare_completions
+    __%[1]s_debug ""
+    __%[1]s_debug "========= starting completion logic =========="
+
     # Start fresh
-    set --erase __%[1]s_comp_do_file_comp
     set --erase __%[1]s_comp_results
 
-    # Check if the command-line is already provided.  This is useful for testing.
-    if not set --query __%[1]s_comp_commandLine
-        # Use the -c flag to allow for completion in the middle of the line
-        set __%[1]s_comp_commandLine (commandline -c)
-    end
-    __%[1]s_debug "commandLine is: $__%[1]s_comp_commandLine"
-
-    set results (__%[1]s_perform_completion "$__%[1]s_comp_commandLine")
-    set --erase __%[1]s_comp_commandLine
+    set results (__%[1]s_perform_completion)
     __%[1]s_debug "Completion results: $results"
 
     if test -z "$results"
         __%[1]s_debug "No completion, probably due to a failure"
         # Might as well do file completion, in case it helps
-        set --global __%[1]s_comp_do_file_comp 1
         return 1
     end
 
@@ -132,7 +123,6 @@ function __%[1]s_prepare_completions
     if test $compErr -eq 1
         __%[1]s_debug "Received error directive: aborting."
         # Might as well do file completion, in case it helps
-        set --global __%[1]s_comp_do_file_comp 1
         return 1
     end
 
@@ -141,7 +131,6 @@ function __%[1]s_prepare_completions
     if test $filefilter -eq 1; or test $dirfilter -eq 1
         __%[1]s_debug "File extension filtering or directory filtering not supported"
         # Do full file completion instead
-        set --global __%[1]s_comp_do_file_comp 1
         return 1
     end
 
@@ -150,32 +139,56 @@ function __%[1]s_prepare_completions
 
     __%[1]s_debug "nospace: $nospace, nofiles: $nofiles"
 
-    # Important not to quote the variable for count to work
-    set numComps (count $__%[1]s_comp_results)
-    __%[1]s_debug "numComps: $numComps"
+    # If we want to prevent a space, or if file completion is NOT disabled,
+    # we need to count the number of valid completions.
+    # To do so, we will filter on prefix as the completions we have received
+    # may not already be filtered so as to allow fish to match on different
+    # criteria than the prefix.
+    if test $nospace -ne 0; or test $nofiles -eq 0
+        set prefix (commandline -t)
+        __%[1]s_debug "prefix: $prefix"
 
-    if test $numComps -eq 1; and test $nospace -ne 0
-        # To support the "nospace" directive we trick the shell
-        # by outputting an extra, longer completion.
-        __%[1]s_debug "Adding second completion to perform nospace directive"
-        set --append __%[1]s_comp_results $__%[1]s_comp_results[1].
+        set completions
+        for comp in $__%[1]s_comp_results
+            if test (string match -e -r -- "^$prefix" "$comp")
+                set -a completions $comp
+            end
+        end
+        set --global __%[1]s_comp_results $completions
+        __%[1]s_debug "Filtered completions are: $__%[1]s_comp_results"
+
+        # Important not to quote the variable for count to work
+        set numComps (count $__%[1]s_comp_results)
+        __%[1]s_debug "numComps: $numComps"
+
+        if test $numComps -eq 1; and test $nospace -ne 0
+            # To support the "nospace" directive we trick the shell
+            # by outputting an extra, longer completion.
+            # We must first split on \t to get rid of the descriptions because
+            # the extra character we add to the fake second completion must be
+            # before the description.  We don't need descriptions anyway since
+            # there is only a single real completion which the shell will expand
+            # immediately.
+            __%[1]s_debug "Adding second completion to perform nospace directive"
+            set split (string split --max 1 \t $__%[1]s_comp_results[1])
+            set --global __%[1]s_comp_results $split[1] $split[1].
+            __%[1]s_debug "Completions are now: $__%[1]s_comp_results"
+        end
+
+        if test $numComps -eq 0; and test $nofiles -eq 0
+            # To be consistent with bash and zsh, we only trigger file
+            # completion when there are no other completions
+            __%[1]s_debug "Requesting file completion"
+            return 1
+        end
     end
 
-    if test $numComps -eq 0; and test $nofiles -eq 0
-        __%[1]s_debug "Requesting file completion"
-        set --global __%[1]s_comp_do_file_comp 1
-    end
-
-    # If we don't want file completion, we must return true even if there
-    # are no completions found.  This is because fish will perform the last
-    # completion command, even if its condition is false, if no other
-    # completion command was triggered
-    return (not set --query __%[1]s_comp_do_file_comp)
+    return 0
 end
 
 # Since Fish completions are only loaded once the user triggers them, we trigger them ourselves
 # so we can properly delete any completions provided by another script.
-# The space after the the program name is essential to trigger completion for the program
+# The space after the program name is essential to trigger completion for the program
 # and not completion of the program name itself.
 complete --do-complete "%[2]s " > /dev/null 2>&1
 # Using '> /dev/null 2>&1' since '&>' is not supported in older versions of fish.
@@ -183,16 +196,8 @@ complete --do-complete "%[2]s " > /dev/null 2>&1
 # Remove any pre-existing completions for the program since we will be handling all of them.
 complete -c %[2]s -e
 
-# The order in which the below two lines are defined is very important so that __%[1]s_prepare_completions
-# is called first.  It is __%[1]s_prepare_completions that sets up the __%[1]s_comp_do_file_comp variable.
-#
-# This completion will be run second as complete commands are added FILO.
-# It triggers file completion choices when __%[1]s_comp_do_file_comp is set.
-complete -c %[2]s -n 'set --query __%[1]s_comp_do_file_comp'
-
-# This completion will be run first as complete commands are added FILO.
-# The call to __%[1]s_prepare_completions will setup both __%[1]s_comp_results and __%[1]s_comp_do_file_comp.
-# It provides the program's completion choices.
+# The call to __%[1]s_prepare_completions will setup __%[1]s_comp_results
+# which provides the program's completion choices.
 complete -c %[2]s -n '__%[1]s_prepare_completions' -f -a '$__%[1]s_comp_results'
 
 `, nameForVar, name, compCmd,
--- a/fish_completions.go
+++ b/fish_completions.go
@@ -30,7 +30,7 @@ end
 function __%[1]s_perform_completion
     __%[1]s_debug "Starting __%[1]s_perform_completion"
 
-    set args (string split -- " " (commandline -c))
+    set args (string split -- " " (string trim -l (commandline -c)))
     set lastArg "$args[-1]"
 
     __%[1]s_debug "args: $args"


marckhouzam:
Here is how to test this specific fix.  An extensive test program is included at the bottom.

**Test 1 - ShellCompDirectiveNoSpace not respected**
Before the PR:
```
$ fish
$ ./testprog completion fish | source

$ ./testprog nospace compo<TAB>
# and
$ ./testprog --nospace compo<TAB>
```
will complete to:
```
$ ./testprog nospace compost<SPACE>
# and
$ ./testprog --nospace compost<SPACE>
```
the `<SPACE>` is wrong at the end of the completion because `ValidArgsFunction` requested `ShellCompDirectiveNoSpace`, as can be seen by:
```
$ ./testprog __complete nospace compo
complete	description1
compost	description2
:2
Completion ended with directive: ShellCompDirectiveNoSpace
```
Testing with this PR in the exact same way will not output the `<SPACE>` after the word `compost`.

**Test 2 - ShellCompDirectiveNoSpace not properly handling descriptions**

The nospace logic for fish does not handle the case when there is a description. To achieve "nospace" the script adds a second completion with an extra "." to prevent the shell from adding a space. However if there is a description, the extra "." is added to the description instead of the completion, and the nospace directive does not work.

Before the PR:
```
$ fish
$ ./testprog completion fish | source

$ ./testprog nospace first compo<TAB>
```
will complete to:
```
$ ./testprog nospace first compo<SPACE>
```
the `<SPACE>` is wrong at the end of the completion because `ValidArgsFunction` requested `ShellCompDirectiveNoSpace`, as can be seen by:
```
$ ./testprog __complete nospace compo
compost	description2
:2
Completion ended with directive: ShellCompDirectiveNoSpace
```
**Notice** this is slightly different than the previous case because a single completion is returned by the `__complete` command, but the `<SPACE>` is still added, this time because of the description.

Testing with this PR in the exact same way will not output the `<SPACE>` after the word `compost`.

**Test 3 - File completion not always done when requested**

Before the PR:
```
$ fish
$ ./testprog completion fish | source
$ touch myfile

$ ./testprog withfile myfi<TAB>
```
will **not** perform file completion and will **not** show `myfile` although it should have.

Testing with this PR in the exact same way will trigger file completion and will show `myfile` as expected.

**Extensive test program to execute the above tests, but also any regression tests:**
```
package main

import (
	"fmt"
	"os"
	"strings"

	"github.com/spf13/cobra"
)

var completionNoDesc bool

	Use:   "completion [bashzsh|fish|powershell]",
	Long: `To load completions:

Bash:

Linux:
  $ testprog completion bash > /etc/bash_completion.d/testprog
MacOS:
  $ testprog completion bash > /usr/local/etc/bash_completion.d/testprog

Zsh:

$ compdef _testprog testprog

$ testprog completion zsh > "${fpath[1]}/_testprog"

Fish:

$ testprog completion fish > ~/.config/fish/completions/testprog.fish

PowerShell:

To load completions for every new session, add the output of the above command
to your powershell profile.
`,
	DisableFlagsInUseLine: true,
		return []string{"bash\tBash completion", "zsh", "fish", "powershell"}, cobra.ShellCompDirectiveNoFileComp
	},
	Args: cobra.ExactValidArgs(1),
		switch args[0] {
		case "bash":
		case "zsh":
			if !completionNoDesc {
			} else {
			}
		case "fish":
		case "powershell":
		}
	},
}

var rootCmd = &cobra.Command{
	Use: "testprog",
		fmt.Println("rootCmd called")
	},
}

	Use:        "vargs",
	Short:      "Cmd with ValidArgs AND a subcmd",
	ValidArgs:  []string{"one", "two", "three"},
	ArgAliases: []string{"un", "deux", "trois"},
	Args:       cobra.MinimumNArgs(1),
	},
}

	Use:   "childVargs",
	Short: "Child command",
	},
}

var aliasCmd = &cobra.Command{
	Use:     "cmdWithAlias",
	Short:   "Command with aliases",
	Aliases: []string{"aliasCmd", "cmdAlias"},
		fmt.Println("cmdWithAlias called")
	},
}

// A command with no short description
var nodescriptionCmd = &cobra.Command{
	Use: "nodesc",
		fmt.Println("nodesc called")
	},
}

	Use:   "filterFile",
	Short: "Only list file of type [*.log,*.txt]",
		if len(args) == 0 {
		}
	},
	},
}

	Use:   "nofile",
	Short: "No file completion done",
	},
		fmt.Println("nofile called")
	},
}

	Use:   "fullFile [filename]",
	Short: "Full file completion",
	},
}

	Use:   "dirOnly [dirname]",
	Short: "Only list directories in completion",
		if len(args) == 0 {
			return nil, cobra.ShellCompDirectiveFilterDirs
		}
	},
		fmt.Println("dirOnly called")
	},
}

	Use:   "subdir [sub-dirname]",
	Short: "Only list directories within themes/",
		if len(args) == 0 {
		}
	},
		fmt.Println("subdir called")
	},
}

	Use:   "compAndFile [sun|moon|<filename>]",
	Short: "Provide some completions and file completion when prefix does not match",
		if len(args) != 0 {
			return []string{"nginx", "thanos"}, cobra.ShellCompDirectiveNoFileComp
		}
		var completions []string
		for _, comp := range []string{"install", "uninstall"} {
		}
	},
	},
}

var colonCmd = &cobra.Command{
	Use:       "cmdColonArgs",
	Short:     "Command with colon args",
	ValidArgs: []string{"first:ONE", "first:SECOND"},
		fmt.Println("colonCmd called")
	},
}

var longDescCmd = &cobra.Command{
	Use:       "cmdLongDesc12345678901234567890123456789012345678901234567890",
	Short:     `We have here a quite long description which should go over the length of the shell width.  What should we do with such a long description? zsh and fish handle it nicely, so we should prepare bash to handle it in the same fashion`,
	ValidArgs: []string{"first:ONE", "first:SECOND"},
		fmt.Println("colonCmd called")
	},
}

var specialCharDescCmd = &cobra.Command{
	Use:   "specialChar",
	Short: "Description contains chars like ` and maybe others",
		fmt.Println("specialChar cmd called")
	},
}

var noSpaceCmd = &cobra.Command{
	Use:   "nospace",
	Short: "followed by completion with noSpace, even for file completion",
		comps := []string{"complete\tdescription1", "compost\tdescription2"}
		if len(args) == 0 {
			// Allow file completion with ShellCompDirectiveNoSpace
			return comps, cobra.ShellCompDirectiveNoSpace
		}

		var finalComps []string
		for _, comp := range comps {
				finalComps = append(finalComps, comp)
			}
		}
		return finalComps, cobra.ShellCompDirectiveNoSpace
	},
		fmt.Println("nospace cmd called")
	},
}

var sentenceCmd = &cobra.Command{
	Use:   "sentence",
	Short: "completion has spaces",
		// Allow file completion with ShellCompDirectiveNoSpace
		return []string{"comp is a sentence\tFirst comp", "and another sentence\tsecondComp"}, cobra.ShellCompDirectiveNoFileComp
	},
		fmt.Println("sentence cmd called")
	},
}

var withfileCmd = &cobra.Command{
	Use:   "withfile",
	Short: "completion always returned and has file completion",
		// We do not consider the toComplete prefix and always return a completion
		return []string{"why"}, cobra.ShellCompDirectiveDefault
	},
		fmt.Println("withfile cmd called")
	},
}

func setFlags() {
	rootCmd.Flags().String("theme", "", "theme to use (located in /themes/THEMENAME/)")
	rootCmd.Flags().SetAnnotation("theme", cobra.BashCompSubdirsInDir, []string{"themes"})

	rootCmd.Flags().String("theme2", "", "theme to use (located in /themes/THEMENAME/)")
	rootCmd.RegisterFlagCompletionFunc("theme2", func(cmd *cobra.Command, args []string, toComplete string) ([]string, cobra.ShellCompDirective) {
		return []string{"themes"}, cobra.ShellCompDirectiveFilterDirs
	})

	rootCmd.Flags().String("customComp", "", "test custom comp for flags")
	rootCmd.RegisterFlagCompletionFunc("customComp", func(cmd *cobra.Command, args []string, toComplete string) ([]string, cobra.ShellCompDirective) {
		return []string{"firstComp\tthe first value", "secondComp\tthe second value", "forthComp\tthe forth value"}, cobra.ShellCompDirectiveNoFileComp
	})

	rootCmd.Flags().StringP("file", "f", "", "list files")
	rootCmd.Flags().String("longdesc", "", "before newline\nafter newline")

	rootCmd.Flags().BoolP("bool", "b", false, "bool flag")
	rootCmd.Flags().BoolSliceP("bslice", "B", nil, "bool slice flag")

	rootCmd.PersistentFlags().StringP("persistent", "p", "", "persistent flag")

	// rootCmd.Flags().String("required", "", "required flag")
	// rootCmd.MarkFlagRequired("required")

	vargsCmd.Flags().StringP("nonpersistent", "n", "", "non persistent local flag")

	// If there is only a recommended value before we reach the equal sign or the colon, then the completion works as expected.
	// If the values differ afterwards, completion fails in bash, but works in fish and zsh
	rootCmd.Flags().String("equalSignWorks", "", "test custom comp with equal sign")
	rootCmd.RegisterFlagCompletionFunc("equalSignWorks", func(cmd *cobra.Command, args []string, toComplete string) ([]string, cobra.ShellCompDirective) {
		return []string{"first=Comp\tthe first value", "second=Comp\tthe second value", "forth=Comp\tthe forth value"}, cobra.ShellCompDirectiveNoFileComp
	})

	rootCmd.Flags().String("equalSignWorksNot", "", "test custom comp with equal sign")
	rootCmd.RegisterFlagCompletionFunc("equalSignWorksNot", func(cmd *cobra.Command, args []string, toComplete string) ([]string, cobra.ShellCompDirective) {
		return []string{"first=Comp1\tthe first value", "first=Comp2\tthe second value", "first=Comp3\tthe forth value"}, cobra.ShellCompDirectiveNoFileComp
	})

	rootCmd.Flags().String("colonWorks", "", "test custom comp with colon one value works")
	rootCmd.RegisterFlagCompletionFunc("colonWorks", func(cmd *cobra.Command, args []string, toComplete string) ([]string, cobra.ShellCompDirective) {
		return []string{"first:ONE\tthe first value", "second\tthe second value", "forth\tthe forth value"}, cobra.ShellCompDirectiveNoFileComp
	})

	rootCmd.Flags().String("colonWorksNot", "", "test custom comp with colon doesnt work")
	rootCmd.RegisterFlagCompletionFunc("colonWorksNot", func(cmd *cobra.Command, args []string, toComplete string) ([]string, cobra.ShellCompDirective) {
		return []string{"first:ONE\tthe first value", "first:SECOND\tthe second value", "second:Comp1\tthe forth value"}, cobra.ShellCompDirectiveNoFileComp
	})

	rootCmd.Flags().String("nospace", "", "test custom comp with nospace directive")
	rootCmd.RegisterFlagCompletionFunc("nospace", func(cmd *cobra.Command, args []string, toComplete string) ([]string, cobra.ShellCompDirective) {
		return []string{"complete\tFirst comp", "compost\tsecondComp"}, cobra.ShellCompDirectiveNoSpace
	})

	rootCmd.Flags().StringArray("array", []string{"allo", "toi"}, "string array flag")
}

func main() {
		&completionNoDesc,
		"no-descriptions", false,

	rootCmd.AddCommand(
		filterFileCmd,
		noFileCmd,
		fullFileCmd,
		dirOnlyCmd,
		subdirCmd,
		compAndFileCmd,
		nodescriptionCmd,
		vargsCmd,
		colonCmd,
		aliasCmd,
		longDescCmd,
		specialCharDescCmd,
		noSpaceCmd,
		sentenceCmd,
		withfileCmd,
	)

	setFlags()
	rootCmd.Execute()
}

```

marckhouzam:
Waiting for review

fish_completions.go
```diff
@@ -28,9 +28,9 @@ function __%[1]s_debug
 end
 
 function __%[1]s_perform_completion
-    __%[1]s_debug "Starting __%[1]s_perform_completion with: $argv"
+    __%[1]s_debug "Starting __%[1]s_perform_completion"
 
-    set args (string split -- " " "$argv")
+    set args (string split -- " " (commandline -c))
```
Luap99, line 33: Given that you are already touching this line can you change it to
```suggestion
    set args (string split -- " " (string trim (commandline -c)))
```
Please check #1303 for details. I already checked that both `$argv` or `commandline -c` have to be trimmed

marckhouzam: Thanks @Luap99 !
I will update with your findings.

Luap99: Actually it has to be:
```suggestion
    set args (string split -- " " (string trim -l (commandline -c)))
```
Only trim the left spaces as the right ones are required if complete an empty arg (e.g. ` command [TAB]`).

marckhouzam: Done.
Thanks again!

marckhouzam:
This PR now addresses 5 different fish completion issues.

We have Go tests for the completion logic in Cobra that is coded in Go.  Those tests can be found in `custom_completions_test.go`.  However, we don't have tests to exercise the actual completion script coded in shell script.  This makes it more difficult to make changes to those shell scripts without worrying about causing some regression.

To try to address this concern, I have setup a regression-testing project for the Cobra completion scripts.
It is currently able to test the `fish` and the `bash` scripts.  Specifically for `fish`, it has tests that fail for each issue this PR is fixing and that pass with this PR applied.  This gives me good confidence in this PR.

If anyone is interested it should be simple to run those tests locally:
https://github.com/marckhouzam/cobra-completion-testing

Luap99 (approved):
LGTM

Luap99:
@marckhouzam CI is red. You have to rebase.

fish_completions.go
```diff
@@ -30,7 +30,7 @@ end
 function __%[1]s_perform_completion
     __%[1]s_debug "Starting __%[1]s_perform_completion"
 
-    set args (string split -- " " (commandline -c))
+    set args (string split -- " " (string trim -l (commandline -c)))
```
krobelus, line 33: Should be `string trim -- -l (commandline -c)`

---

Often it's not necessary to pass the current token to fish.
Rather, one can just compute all possible tokens because fish will filter them anyway.
Sometimes it can be useful to do the filtering in the app (for example if it's not feasible to enumerate all possible tokens), so it's fine.

fish_completions.go
```diff
@@ -43,16 +43,13 @@ function __%[1]s_perform_completion
     end
     __%[1]s_debug "emptyArg: $emptyArg"
 
-    if not type -q "$args[1]"
-        # This can happen when "complete --do-complete %[2]s" is called when running this script.
```
krobelus, line 47: Removing this seems sensible. I'm not sure if the comment really matters, because the this fish completion script should only be sourced when the command exists.

fish_completions.go
```diff
@@ -43,16 +43,13 @@ function __%[1]s_perform_completion
...
-    end
-
     set requestComp "$args[1] %[3]s $args[2..-1] $emptyArg"
     __%[1]s_debug "Calling $requestComp"
 
-    set results (eval $requestComp 2> /dev/null)
+    # Call the command as a sub-shell so that we can redirect any errors
+    # For example, if $requestComp has an unmatched quote
+    # https://github.com/spf13/cobra/issues/1214
+    set results (fish -c "$requestComp" 2> /dev/null)
```
krobelus, line 52: Hmm this is interesting. If you didn't include the token at the cursor this wouldn't be a problem.

Usually completions want to use `commandline -opc`, which gives all words left of the cursor, which always tokenize successfully.

It also restricts the tokens to  the current process (`-p`). Observe bind running
```
bind \eg 'commandline -c'
```
and then type `echo foo; echo bar` and <kbd>Ctrl>/kbd>+<kbd>g</kbd>.
So `-p` is usually correct for completions.

The `-o` flag is useful because it already does word splitting, then there is no need for `string split " "`.

fish_completions.go
krobelus: This line should probably be `set args (commandline -opc)`.
If the token at the cursor is really required, `(commandline -ct)` can be appended.

fish_completions.go
krobelus: OK I've played around with this some more, and found a version that should tick all boxes, while not needing a separate fish process. Here's what I would do:

```fish
function __%[1]s_perform_completion
    set -l args (commandline -opc)
    set -l lastArg (commandline -ct)
    set -l requestComp $args[1] %[3]s $args[2..-1] (string escape -- $lastArg)
    if test -d $args[1]
        return 1
    end
    set -l results (eval $requestComp 2> /dev/null)
```

- `set -l` is a minor detail: if a user defines a universal variable with `set -U x foo`, then `set x` inside this function would overwrite the user's variable. This is an ugly edge case in fish which should probably be changed, but for now we can force variables to be local by use `set -l` when they are used for the first time in a function.
- `(commandline -ct)` gives us the raw token at the cursor, including any unclosed quotes. To avoid https://github.com/spf13/cobra/issues/1214, we can simply escape it. This means that variables in the token at the cursor will not be evaluated, but that's a general issue with user-defined fish completions, and I don't think it's an important problem.
- If the command is a directory I got  some errors (trying to complete `./testprog `), so I added a check for that.

fish_completions.go
```diff
@@ -137,49 +126,65 @@ function __%[1]s_prepare_completions
...
+        # Important not to quote the variable for count to work
+        set numComps (count $__%[1]s_comp_results)
+        __%[1]s_debug "numComps: $numComps"
+
+        if test $numComps -eq 1; and test $nospace -ne 0
+            # To support the "nospace" directive we trick the shell
+            # by outputting an extra, longer completion.
+            # We must first split on \t to get rid of the descriptions because
+            # the extra character we add to the fake second completion must be
+            # before the description.  We don't need descriptions anyway since
```
krobelus, line 153: This is clever, and works reliably as far as I can tell.
However, the upcoming fish 3.2.0 will always show the pager on ambiguous completions (https://github.com/fish-shell/fish-shell/issues/6924). This would cause some confusion.

So I'm not sure if this an overall improvement. Who are the users of this option?
Fish already gets this right a lot of times by hardcoding the character set `@=/:.,` - if a completion ends in one of these characters, then no space will be inserted.

fish_completions.go
```diff
@@ -137,49 +126,65 @@ function __%[1]s_prepare_completions
...
+    if test $nospace -ne 0; or test $nofiles -eq 0
+        set prefix (commandline -t)
+        __%[1]s_debug "prefix: $prefix"
+
+        set completions
+        for comp in $__%[1]s_comp_results
+            if test (string match -e -r -- "^$prefix" "$comp")
+                set -a completions $comp
+            end
+        end
```
krobelus, line 140: In case this is relevant: the regex should be escaped, just in case, and this loop can be done in one line because `string match` accepts multiple arguments. I'd also eschew the `-e` option, I think that option was a mistake.

```fish
set prefix (commandline -t | string escape --style=regex)
...
set -l completions (string match -r -- "^$prefix.*" $__%[1]s_comp_results)
```

fish_completions.go
marckhouzam: This scenario is pretty obscure actually.  Let me explain.  

Fish comes with existing completion scripts.  For instance, it has its own `helm` completion script:
https://github.com/fish-shell/fish-shell/blob/master/share/completions/helm.fish
This is script is out of date however and provides some completions that are no longer valid for `helm`.

In the Cobra script this PR updates, we do `complete -c helm -e` to erase all other completions for `helm`.  The problem is that the old `helm` script included in fish may not yet be loaded, which means the user could get completions from both scripts (the old one provided by fish, and the Cobra one).  To fix this, the Cobra script does `complete --do-complete "helm " > /dev/null 2>&1` which forces the loading of the old completion script, so that all these completions can be erased by `complete -c helm -e`.

**The problem happens for programs that are not on PATH**.  In this case, if the Cobra completion script is sourced more than once, calling for example, `complete --do-complete "myprog " > /dev/null 2>&1` would cause fish to print an error because it cannot find `myprog`, and this error printout could not be redirected to `/dev/null`.  To fix that I was catching this special case with `if not type -q "$args[1]"` above.

However, with @Luap99's help, we have found a way to redirect such error to `/dev/null` (see line 52 below).  That is why we no longer need to catch this special case and can remove this code.

fish_completions.go
marckhouzam: Thanks for the improvements @krobelus !

> the regex should be escaped, just in case,

Sounds safer, I'll add this.

> and this loop can be done in one line because `string match` accepts multiple arguments.

Nice!  I admit I was upset at having to do the loop.  Now it can be removed!

> I'd also eschew the `-e` option, I think that option was a mistake.

Actually, according to the following doc, I understood that using `-e` was preferred to adding the suffix `.*`, so I went that route: https://fishshell.com/docs/current/cmds/string-match.html

fish_completions.go
marckhouzam: > This is clever, and works reliably as far as I can tell.
> However, the upcoming fish 3.2.0 will always show the pager on ambiguous completions ([fish-shell/fish-shell#6924](https://github.com/fish-shell/fish-shell/issues/6924)). This would cause some confusion.

Ouch, that will indeed cause confusion for the "nospace" case.

> So I'm not sure if this an overall improvement. Who are the users of this option?

For example, Helm uses this option when completing a chart's full name which is comprised of `<repository>/<chartname>`; the completion first provides all repository names in the form `<repository>/`, and once the slash is present on the command-line, the completion will add the possible chart names, but there should be no space between the two.  This is much more efficient for helm than having to provide all the full `<repository>/<chartname>` from the very start.

`kubectl cp <pod>:<file>` could also use this option to avoid adding a space after the `:`.

> Fish already gets this right a lot of times by hardcoding the character set `@=/:.,` - if a completion ends in one of these characters, then no space will be inserted.

Cool!  I was not aware of this.  This would address both cases I mentioned.  I tested it and it works as you describe.
But I don't know if there are other cases that wouldn't end with `@=/:.,`.  At a minimum, we could turn off this hack for if the completion ended with `@=/:.,`.  I'll have to think what to do for the other (possible) cases for fish 3.2.  Let me know if you have any suggestions.

fish_completions.go
marckhouzam: Thanks!  I'll take the time to look into this as soon as I can and get back to you.

fish_completions.go
marckhouzam: > Should be `string trim -- -l (commandline -cp)`

I don't understand from the documentation what the `-p` flag does.  Could you clarify?

> Often it's not necessary to pass the current token to fish.
> Rather, one can just compute all possible tokens because fish will filter them anyway.
> Sometimes it can be useful to do the filtering in the app (for example if it's not feasible to enumerate all possible tokens), so it's fine. (the `testprog` examples show that the current token is necessary to complete `--flags`)

Right.  So originally, I thought the program should do the filtering, but with experience, I now realize it is actually better to provide `fish` (and `zsh`) with all unfiltered completions so that the shell can do smarter matching than simply on prefix.  For example, if the user types `prog paw<TAB>`, giving shell completions like `bearpaw` will allow it to still match even though it is not on the prefix.

As you saw in this PR however, I didn't find a way to find out if `fish` had found a smart match or not, so, when wanting to do file completion, I was forced to filter on prefix in the script itself, to know if there would be a match or not.  This is because Cobra only wants to do file completion if there are no valid completions, so it can be consistent with `bash` and `zsh`.  This was not ideal, so if you know of a way to figure out in the script if fish will match on a list of completion, that would be helpful.

marckhouzam:
> @marckhouzam CI is red. You have to rebase.

Thanks for the heads up @Luap99 . I rebased on master to get the 2021 date fix which were the cause of the failure.

fish_completions.go
krobelus: One could sidestep the completion engine and directly insert the unique completion if nospace is set

```fish
$ function sneaky-complete
      commandline --replace -t nospace
  end
$ complete -e sneaky-complete; complete -c sneaky-complete -a '(sneaky-complete)' -f
```

According to my tests this works (although it rings the terminal bell, because fish thinks there are no completions). However, I'd advise against it. You are not meant to modifiy the commandline *during* the computation of a completion. That can break other completions, though I didn't find an example here.

fish has not added an option to control the add-a-space behavior yet because AFAIK no real use case has been brought up that doesn't work with the hardcoded list.

I fear that the cost of adding a hack to support this outweighs the benefits.
I think we should drop this until we see a real need. Then an option can be added to fish, avoiding the need for a hacky workaround.

fish_completions.go
krobelus: The `-e` thing is just an opinion, anything works of course. My point is, without it the regex is easier to understand by someone who doesn't know this flag.

fish_completions.go
krobelus: The `-p` is explained in https://github.com/spf13/cobra/pull/1249#discussion_r563333005. Sorry for the wall of text :smile_cat: 

fish does fuzzy substring matching if there is no prefix or exact match. This is currently not exposed in a scriptable way.

I wonder if it's acceptable to do file completions only if there are no results. Doesn't sound too good, so probably the prefix comparison is better (maybe make that substring matching instead of only prefix).

fish_completions.go
krobelus: > This is script is out of date however and provides some completions that are no longer valid for helm.

So the fish script should be changed to a single line `helm completion fish | source`?
Ok this is https://github.com/fish-shell/fish-shell/issues/7357, no one has done that yet. I'll try to do that, I just need to detect the Helm version.

> In the Cobra script this PR updates, we do complete -c helm -e to erase all other completions for helm.

In an ideal world this would not be necessary, because fish's version is installed to, say `/usr/share/fish/completions/helm.fish` but the one that ships with the Helm package is in `/usr/share/fish/vendor_completions.d/helm.fish`, which overrides the former because it's earlier in `$fish_complete_path`.

But for manual (and repeated) sourcing we do need this, so you are right.

I feel like the best solution is to check at the call site if the program is executable:

```fish
if type -q "%[2]s"
    complete --do-complete "%[2]s " > /dev/null 2>&1
end
```

Starting a new fish instance to be able to redirect the error seems a bit overkill.
fish always sources its startup files, so the new instance might cause obscure issues; better to use `eval` (as suggested below).

fish_completions.go
Luap99: I took a look how I use the `ShellCompDirectiveNoSpace` option in podman. Almost all cases ends with one of `=:,` but not all. I found one case where I set this when it ends with a `-`. https://github.com/containers/podman/blob/479fc226044b745ecaafc2d5fa925afe6ca06de0/cmd/podman/common/completion.go#L603-L608
Even more importantly there is a case where I set this together with `ShellCompDirectiveDefault` (path completion). This case does not work with fish currently. It works for bash and zsh.

krobelus: Thanks for investigating. I guess we should just [add `-`](https://i.redd.it/vtttn02gchsz.jpg) to the set of separators in fish.

Regarding the nospace+path, is this what you mean:

```go
// cobra-completion-testing/testprog/testprog.go
var noSpaceCmdPrefix = &cobra.Command{
	Use:   "nospace",
	Short: "Directive: no space",
		return nil, cobra.ShellCompDirectiveNoSpace
	},
	Run: func(cmd *cobra.Command, args []string) {},
}
```

```
testprog/bin/testprog prefix nospace <TAB>
```

this gives me files in both fish and bash (I've merged this branch and the bash v2 one).
If I return some completions instead of `nil`, then I only get those completions in both shells (as expected).

You probably mean something else?

Luap99: Bash does not add a space after the completed path. Fish does.

Luap99: Honestly I don't think adding `-` make much sense. This might solve my specific problem but doesn't solve the problem here.
`cobra` allows me as developer to set `nospace` wherever I want, not only on those characters. I don't see a reason why fish should limit my options as a completion script developer.

fish_completions.go
marckhouzam: > I wonder if it's acceptable to do file completions only if there are no results. 

You mean if the program returns no completions?  This is what the script does right now (before this PR). But that requires that the program do the filtering on itself.  This PR is (in part) trying to improve that situation. By doing the filtering in the fish script, it will allow the program to return all completions, which will then allow the fuzzy matching you mention.  The fuzzy matching will only not work when file completion is requested or when nospace is requested, both of which are the less common cases.

> Doesn't sound too good, so probably the prefix comparison is better (maybe make that substring matching instead of only prefix).

As long as we use an algorithm that is stricter than what fish does, we could indeed improve compared to filtering on prefix.  Do you know where I could see the algorithm used by fish?  Or else I'll just play around and try to see if any pattern match seems to work.

fish_completions.go
marckhouzam: > So the fish script should be changed to a single line `helm completion fish | source`?
> Ok this is [fish-shell/fish-shell#7357](https://github.com/fish-shell/fish-shell/issues/7357), no one has done that yet. I'll try to do that, I just need to detect the Helm version.

You can probably prefix the existing script with something like
```
if helm completion -h | grep -q fish > /dev/null
    helm completion fish | source
else
   # Existing old script
end
```
Note that `helm completion fish` will return a success code even if the command does not exist (#1156), that is why I suggest using the help text to figure out if fish is supported.

> In an ideal world this would not be necessary, because fish's version is installed to, say `/usr/share/fish/completions/helm.fish` but the one that ships with the Helm package is in `/usr/share/fish/vendor_completions.d/helm.fish`, which overrides the former because it's earlier in `$fish_complete_path`.

You are right.  We could choose not to support the explicit sourcing.  In fact, Cobra does not support it for `zsh`.  However, I do find it nicer to support it, as long as the side-effects can be easily controlled.
 
> But for manual (and repeated) sourcing we do need this, so you are right.

👍 

> I feel like the best solution is to check at the call site if the program is executable:
> 
> ```fish
> if type -q "%[2]s"
>     complete --do-complete "%[2]s " > /dev/null 2>&1
> end
> ```

I think that would work, and it is better.  However we thought about adding the extra fish instance to solve another issue where an error was being thrown when a quote was not properly closed, see #1214 

> Starting a new fish instance to be able to redirect the error seems a bit overkill.
> fish always sources its startup files, so the new instance might cause obscure issues; better to use `eval` (as suggested below).

We currently use eval, but it prints error message we cannot catch in the case of #1214

krobelus: > However, I do find explicit sourcing nicer

Yeah, I think it's fine.

> We currently use eval, but it prints error message we cannot catch

The error can be cleanly avoided by escaping `lastArg`, see https://github.com/spf13/cobra/pull/1249#discussion_r563348012.

fish_completions.go
krobelus: Just to clarify for myself; the common cases are
- no files
- only files because the program returns no completions

in those cases fish already does all the filtering. But if we're in between - the program returns some completions but they may not match the token - we want to filter in the completion script. That sounds great.

I agree that this is probably not very common. Most programs should take either custom tokens (like subcommands) or files, not both. Git is a notable exception, but Git completion computes custom file paths based on `git ls-files`.

I'd just ignore nospace for fish for now until, it behaves the right way most of the time already.
I can't think of a good solution right now, it needs support from fish.

> Do you know where I could see the algorithm used by fish?

The two most permissive filters are [here](https://github.com/fish-shell/fish-shell/blob/062f24d91b199780ae7a8493988549a5f651423f/src/wcstringutil.cpp#L206-L214). I think they can be implemented like this:

```fish
$ function fishy_match -a needle haystack
      if string match -irq -- (string escape --style=regex -- $needle) $haystack
          echo substr match
      end
      if string match -rq -- (string split -- '' $needle | string escape --style=regex | string join '.*?') $haystack
          echo subseq match
      end
  end
$ fishy_match String 'case-insensitive substring match'
substr match
$ fishy_match foo follow
subseq match
```

I use `-r` and regex escaping because the needle might contain globs, which would be interpreted by `string match` without `-r`.

fish_completions.go
marckhouzam: @krobelus I finally had a chance to try your suggestion above.  I really love the fact that we no longer have to explicitly check for an empty lastArg thanks to the escaping.  Thanks for all your help!

I will put `-l` everywhere there is a `set` of a new variable.

I've also slightly modified your suggestion to do the escaping right away when calculating `lastArg`; if find it easier to read.

fish_completions.go
marckhouzam: > The error can be cleanly avoided by escaping lastArg, see #1249 (comment).

That works!
So we can do as you suggest:
```
if type -q "%[2]s"
    complete --do-complete "%[2]s " > /dev/null 2>&1
end
```
and avoid having to call a new fish shell to avoid error printouts.

@Luap99 you may be interested in trying the version I will soon push.  But I can confirm that I tested it for unmatched quotes, and completion simply isn't done, with no error message.   Also, the regression tests repo passes as before.

fish_completions.go
marckhouzam: > Most programs should take either custom tokens (like subcommands) or files, not both. 

There are some cases where both are valid.  Helm for example can accept the reference to a remote chart, or the local path to a chart.  What I've done in helm is have the program return the list of files as completions to do both file completion and custom token completion.

I think it is important to keep all shells behaving the same for Cobra because the program returns a list of completions to be used by all shells; we don't want to make the program have to behave differently for each shell.

But we could, as a separate effort, support doing file completion even if there are custom token completion matches.  This was not Cobra's previous behaviour, so I had never considered it myself, but I believe it could be done for all shells.  Maybe a future effort...

> I'd just ignore nospace for fish for now until, it behaves the right way most of the time already.
> I can't think of a good solution right now, it needs support from fish.

Well, our current solution is good until fish 3.2.  And then, when fish shows the choices including the fake one with a trailing `.`,  it may be confusing, but I think it may be not much worse than letting the shell add and unwanted space.

PENDING - I'll turn off adding the extra `.` when fish handles the case itself following the characters you mentioned.

PENDING - And I'll also adapt the filtering based on the algos you've provided.

I feel that after all that, we'll have something really solid for fish completion.

fish_completions.go
marckhouzam: > Even more importantly there is a case where I set this together with ShellCompDirectiveDefault (path completion). 
> Bash does not add a space after the completed path. Fish does.

@Luap99 I never planned for this case.  And now that you mention it, I realize it also does not work for `zsh`.  It is an easy fix though.  I will add it to #1213.  And I see that you made it work for powershell 🎉 .  I don't believe there is anything we can do for `fish` though.  

Out of curiosity, what kind of situation would not want a space after file completion?

marckhouzam: Turns out preventing a space after file completion for `zsh` also prevents the trailing `/` from being added after a directory.  At least in the way I've found to prevent a space.  I think a missing `/` is a bigger problem than the extra space, so I'll leave it alone.

marckhouzam: To get back to the 'nospace' directive for fish, the latest version of this PR keeps the "hack" that adds an extra fake completion but only when the completion does not end with `@=/:.,`.  I think this is the best we can do at this time.

fish_completions.go
krobelus: > What I've done in helm is have the program return the list of files as completions to do both file completion and custom token completion.

Ideally we could tell the shell that we also want file completons, because it can see through variables to tokens like `$PWD/`. Something for the future I guess. The implementation would be a bit hacky:
essentially, you'd need two calls to `complete -c`, one with `-f` and one without.
Each of them has a condition (the `-n` part). If the condition for the one with file completions is true, then fish will also provide file completions (in addition to the provided custom completions).

> we don't want to make the program have to behave differently for each shell.

In terms of what completions to offer, definitely. Of course shells differ in how they display completions.
By default, fish sorts completions, but you can add the `--keep-order` flag to the `complete -c` call, then the program is in control of the ordering.

> I feel that after all that, we'll have something really solid for fish completion.

Yeah, latest changes look great! For some programs  it's hard to compete with manually written completions, but it should work great for conventional programs.

fish_completions.go
Luap99: > Out of curiosity, what kind of situation would not want a space after file completion?

Simple answer `podman/docker run --volume`. This flag accepts a volume name or a path followed by a colon and again a path. This is my current autocompletion function:
```
func AutocompleteVolumeFlag(cmd *cobra.Command, args []string, toComplete string) ([]string, cobra.ShellCompDirective) {
	volumes, _ := getVolumes(cmd, toComplete)
	directive := cobra.ShellCompDirectiveNoSpace | cobra.ShellCompDirectiveDefault
	if strings.Contains(toComplete, ":") {
		// add space after second path
		directive = cobra.ShellCompDirectiveDefault
	}
	return volumes, directive
}
```
After the first path I do not want the space added. The user has to type a colon here. After the colon you get only path completion and then I a space should be added.

> Turns out preventing a space after file completion for zsh also prevents the trailing / from being added after a directory. At least in the way I've found to prevent a space. I think a missing / is a bigger problem than the extra space, so I'll leave it alone.

I am totally fine if it does not work. It is just a small corner case.

jpmcb (approved):
I've tested this with the test program Mark provided and also with his cobra-completion-testing repo. I'm confident this is good to go since all those are passing and work as I'd expect!! Thanks for the hard work and patience on this one! :fish:

jpmcb:
@marckhouzam We'll need to fix the merge conflicts as I've been merging PRs this morning

marckhouzam:
@jpmcb The rebase is done.  The tests in https://github.com/marckhouzam/cobra-completion-testing all pass now except one which is related to the poor handling of `--` by the completion code.  I don't quite recall which conversation made me add this test (surely an exchange with @Luap99), but I am confident it is unrelated to this PR.

Thanks so much for getting so many PRs merged today!  This is very exciting!

Luap99:
@marckhouzam I guess the `--` test is for https://github.com/spf13/cobra/pull/1308

Issue #1198, opened 2020-08-21
Persistent*Run of rootCmd will be skipped if its sub-command has the same Persistent*Run


hyorigo:
I spent some time with https://github.com/carolynvs/stingoftheviper , and I found the strange behavior ---

* If only the *PersistentPreRunE* of *rootCmd* is defined, it will always be triggered.
* If both the *PersistentPreRunE* of *rootCmd* and its sub-command, only the sub-command one will be triggered.

If I need the Persistent*Run of *rootCmd* and its sub-command both being triggered, what should I do? Or maybe the current behavior is a bug?

hyorigo:
@marckhouzam Oh, thanks! I found this logic, so why did you authors make this decision?

https://github.com/spf13/cobra/blob/9ed1d713d619c428c8a8adcd13667d6b3f54b247/command.go#L863-L873

Btw, as per your answer, I tried this code snippet in the sub-cmd, and got stack overflow:

```go
	PersistentPreRunE: func(cmd *cobra.Command, args []string) error {
		if parentPreE := cmd.Parent().PersistentPreRunE; parentPreE != nil {
			if err := parentPreE(cmd, args); err != nil {
				return err
			}
		}

		// blab ... blab ... blah
		return nil
	},
```

It works after I changed `cmd.Parent()` into `rootCmd.Parent()`, but I just couldn't understand why this would cause infinite loop, even I debugged step by step.

marckhouzam:
> @marckhouzam Oh, thanks! I found this logic, so why did you authors make this decision?

I can't say for sure what the original intent was, but based on our current discussion, this approach seems to give the most flexibility.  It allows to either override an ancestor's `PersistentPreRun` function by defining one in a subcmd, and it allows to extend ancestor's `PersistentPreRun` by calling them as you are now doing.


> 
> Btw, as per your answer, I tried this code snippet in the sub-cmd, and got stack overflow:
> 
> ```go
> 	PersistentPreRunE: func(cmd *cobra.Command, args []string) error {
> 		if parentPreE := cmd.Parent().PersistentPreRunE; parentPreE != nil {
> 			if err := parentPreE(cmd, args); err != nil {
> ```

I'm guessing the problem is that you pass `cmd` to `parentPreE` when you should be passing `cmd.Parent()`.

hyorigo:
@marckhouzam Thanks for your help! I think we can close this issue now.

marckhouzam:
[...]

PR #1253, opened 2020-10-14
Adds Persistent{Pre,Post}Run hook chaining

### LINKED ISSUE #252: PersistentPreRun only allowed once

poy:
PersistentPreRun and PersistentPostRun are chained together so that
each child PersistentPreRun is ran, and the PersistentPostRun are ran
in reverse order. For example:

Commands: root -> subcommand-a -> subcommand-b

root - PersistentPreRun
subcommand-a - PersistentPreRun
subcommand-b - PersistentPreRun
subcommand-b - Run
subcommand-b - PersistentPostRun
subcommand-a - PersistentPostRun
root - PersistentPostRun

fixes #252

### DIFF
--- a/command.go
+++ b/command.go
@@ -102,6 +102,21 @@ type Command struct {
 	//   * PersistentPostRun()
 	// All functions get the same args, the arguments after the command name.
 	//
+	// When TraverseChildrenHooks is set, PersistentPreRun and
+	// PersistentPostRun are chained together so that each child
+	// PersistentPreRun is ran, and the PersistentPostRun are ran in reverse
+	// order. For example:
+	//
+	// Commands: root -> subcommand-a -> subcommand-b
+	//
+	// root - PersistentPreRun
+	// subcommand-a - PersistentPreRun
+	// subcommand-b - PersistentPreRun
+	// subcommand-b - Run
+	// subcommand-b - PersistentPostRun
+	// subcommand-a - PersistentPostRun
+	// root - PersistentPostRun
+	//
 	// PersistentPreRun: children of this command will inherit and execute.
 	PersistentPreRun func(cmd *Command, args []string)
 	// PersistentPreRunE: PersistentPreRun but returns an error.
@@ -192,6 +207,11 @@ type Command struct {
 
 	// TraverseChildren parses flags on all parents before executing child command.
 	TraverseChildren bool
+
+	// TraverseChildrenHooks will have each subcommand's PersistentPreRun and
+	// PersistentPostRun instead of overriding. It should be set on the root
+	// command.
+	TraverseChildrenHooks bool
 
 	// Hidden defines, if this command is hidden and should NOT show up in the list of available commands.
 	Hidden bool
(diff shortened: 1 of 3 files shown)


marckhouzam:
Thanks @poy for the contribution.

We recently discussed this topic in #1198 and with my limited experience, I got the impression that the current behaviour allowed for the most flexibility. It allows to either override an ancestor's PersistentPreRun function by defining one in a subcmd, and it allows to extend ancestor's PersistentPreRun by calling them explicitly from a subcmd.

I'm trying to compare the value of this PR's new approach with the current one.  Could you explain a little more how your change will change current behaviour, and why it is better?

I'm also curious if this change is breaking?  Existing programs that have PersistentPreRuns will behave differently with this change?  This was mentioned in https://github.com/spf13/cobra/issues/252#issuecomment-622561414.

poy:
This PR will allow a user to create a hierarchy of `PrePersistentRun` and `PostPersistentRun`. Meaning if the root and sub command both have a hook, then both will be invoked. The pre-hooks get invoked in order of `root` -> `subcommand-1` -> `subcommand-2` -> ... -> `subcommand-N`. This allows the earlier subcommands to setup anything necessary for the laster commands. The post-hooks are invoked in the opposite order: `subcommand-N` -> ... -> `subcommand-2` -> `subcommand-1` -> `root`. This reverse order allows for easier cleanup (similar to how `defer` was implemented).

This is not a breaking change (now :) as this behavior only takes place when `TraverseChildrenHooks` is set to `true` on the root node. It will otherwise utilize the old behavior of subcommands overriding their parent's hooks.

poy:
Is there any chance of this being merged?

infinitepr0:
@marckhouzam A gentle reminder to review (and merge) this PR when free :smile:

marckhouzam:
> @marckhouzam A gentle reminder to review (and merge) this PR when free 😄

I'm not a maintainer 😉 
I just try to help when I can.

marckhouzam:
> This PR will allow a user to create a hierarchy of `PrePersistentRun` and `PostPersistentRun`.

@poy Could you instead call `cmd.Parent().PrePersistentRun()` from the subcommand?

poy:
Yes. However, I think given the name "Persistent", it shouldn't be required. PersistentFlags for example don't have to be added in this way.

poy:
Would still love for this to be considered.

f0m41h4u7:
Hey all.
I am also facing this issue with persistent pre- and post-runs currently, and it leads to some annoying code repeats... So I'm really looking forward to this fix being merged some day.

f0m41h4u7:
Also the PersistentPostRun does not make much sense now, because subcommands are not executing the inherited post-runs. So the hook traversal could fix this.

niamster:
> > @marckhouzam A gentle reminder to review (and merge) this PR when free 😄
> 
> I'm not a maintainer 😉 I just try to help when I can.

@marckhouzam what is required to have this PR merged (now that you are a maintainer 😄) ?

command.go
```diff
@@ -102,6 +102,21 @@ type Command struct {
 	//   * PersistentPostRun()
 	// All functions get the same args, the arguments after the command name.
 	//
+	// When TraverseChildrenHooks is set, PersistentPreRun and
```
niamster, line 105: Maybe worth adding something like `By default, if PersistentPreRun or PersistentPostRun is set, the parent's Pre/Post run hooks are not executed`?

command.go
```diff
@@ -193,6 +208,11 @@ type Command struct {
 	// TraverseChildren parses flags on all parents before executing child command.
 	TraverseChildren bool
 
+	// TraverseChildrenHooks will have each subcommand's PersistentPreRun and
+	// PersistentPostRun instead of overriding. It should be set on the root
+	// command.
+	TraverseChildrenHooks bool
```
niamster, line 214: I think it's worth adding some checks to ensure that this is set in the root command only (because `traverseChildrenHooks` cares only about root command).

niamster:
> > This PR will allow a user to create a hierarchy of `PrePersistentRun` and `PostPersistentRun`.
> 
> @poy Could you instead call `cmd.Parent().PrePersistentRun()` from the subcommand?

It might be very easy to forget adding that if new command creation is not automated and there are many commands added by multiple people, potentially by multiple teams.

polothy:
Think this was fixed by https://github.com/spf13/cobra/pull/2044 and released in 1.8.0

niamster:
> Think this was fixed by #2044 and released in 1.8.0

Great news! I think it's worth closing all issues and non-merged PRs now 🎉

marckhouzam:
Thank you for your efforts @niamster .  I'm going to close this as #2044 addresses this.

Issue #1049, opened 2020-03-03
Man page generation wrong in latest release


dionysius:
There's a difference between cobra 0.0.5 and 0.0.6 negatively affecting the resulting man files:

Example: https://play.golang.org/p/zcjf0eQ35xj (copy it locally to compare with v0.0.5)

```diff
diff --git a/tmp/man-from-005/prog.8 b/tmp/man-from-006/prog.8
index 166d0d1..afda3e3 100644
--- a/tmp/man-from-005/prog.8
+++ b/tmp/man-from-006/prog.8
@@ -1,7 +1,7 @@
-.TH "FOO" "8" "Mar 2020" "The Team" "System Manager's Manual" 
 .nh
-.ad l
-
+.TH FOO(8)Mar 2020
+The Team
+System Manager's Manual
 ```

The resulting visual representation is wrong: see `man -l /tmp/man-from-006/prog.8`.

While we're at it, can we maybe have a golden file/dir for GenMan and GenManTree to detect syntax changes?

I also get the following during execution locally, which might be related (when using GenManTree):
```
WARNING: go-md2man does not handle node type HTMLSpan
```

**Comment by MichaelEischer** [comment]
I've bisected this to 77e4d5aecc4d34e58f72e5a1c4a5a13ef55e6f44 (thanks for the example code), which updates md2man to v2.0.0 .

**Comment by dionysius** [comment]
I'm still using `v0.0.5` because of that...

**Comment by elboulangero** [comment]
Seems like we are facing the same issue when while building the docker package in Debian. Since the go-md2man v2 transition, we can't build the documentation anymore:

```
dh_installman: warning: Section for ./.gopath/src/github.com/docker/cli/man/man1/docker-attach.1 is computed as "2020", which is not a valid section
dh_installman: error: Could not determine section for ./.gopath/src/github.com/docker/cli/man/man1/docker-attach.1
dh_installman: error: Aborting due to earlier error
make: *** [debian/rules:40: binary] Error 25
```

For the full bug report: https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=971789

After some investigation, I observe the same changes in the man pages as the one showed above by @dionysius. And that's why `dh_installman` (the tool that checks the man pages while creating the Debian package) chokes, it's because something is wrong with the man page.

[...]

PR #1255, opened 2020-10-14
fix manpage building with new go-md2man

satta:
This addresses #1049 by changing the format of the generated Markdown input.


jpmcb (approved):
👍 

Working on a patch release - this should be in v1.1.1 shortly

Issue #618, opened 2018-01-17
Traverse doesn't work with grouped short flags


dnephin:
Ref: https://github.com/spf13/cobra/pull/617#discussion_r162150570

Grouping multiple short flags together (ex: `-abs`) does not work when `TraverseChildren` is enabled.

PR #1258, opened 2020-10-15
Custom completion handle multiple shorhand flags together

### LINKED ISSUE #1257: Custom completion logic doesn't support multiple shorthand flags together
> Unlike the flag package, a single dash before an option means something different than a double dash. Single dashes signify a series of shorthand letters for flags. All but the last shorthand letter must be boolean flags or a flag with a default value

from: https://github.com/spf13/pflag#command-line-flag-syntax

The custom completion logic does not respect this and will error in that case. (no completion)

Consider the following simple example:
```go
package main

import (
	"fmt"
	"os"
	"path/filepath"

	"github.com/spf13/cobra"
)

var rootCmd = &cobra.Command{
	Use:              filepath.Base(os.Args[0]),
	TraverseChildren: false,
	ValidArgs:        []string{"foo", "bar"},
		fmt.Println("rootCmd called")
	},
}

func main() {
	f := rootCmd.Flags()
	f.BoolP("short", "s", false, "short flag 1")
	f.BoolP("short2", "d", false, "short flag 2")
	rootCmd.Execute()
}
```
```bash
$ ./testprog -sd
rootCmd called
$ ./testprog __complete -sd ""
[Debug] [Error] Subcommand 'testprog' does not support flag 'sd'
:0
$ ./testprog __complete -s ""
foo
bar
:4
```

Luap99:
Flag definitions like `-asd` are not handled correctly by
the custom completion logic. They should be treated as
multiple flags. For details refer to #1257.

Fixes #1257

### DIFF
--- a/completions.go
+++ b/completions.go
@@ -468,7 +468,16 @@ func checkIfFlagCompletion(finalCmd *Command, args []string, lastArg string) (*p
 	if len(lastArg) > 0 && lastArg[0] == '-' {
 		if index := strings.Index(lastArg, "="); index >= 0 {
 			// Flag with an =
-			flagName = strings.TrimLeft(lastArg[:index], "-")
+			if strings.HasPrefix(lastArg[:index], "--") {
+				// Flag has full name
+				flagName = lastArg[2:index]
+			} else {
+				// Flag is shorthand
+				// We have to get the last shorthand flag name
+				// e.g. `-asd` => d to provide the correct completion
+				// https://github.com/spf13/cobra/issues/1257
+				flagName = lastArg[index-1 : index]
+			}
 			lastArg = lastArg[index+1:]
 			flagWithEqual = true
 		} else {
@@ -485,8 +494,16 @@ func checkIfFlagCompletion(finalCmd *Command, args []string, lastArg string) (*p
 				// If the flag contains an = it means it has already been fully processed,
 				// so we don't need to deal with it here.
 				if index := strings.Index(prevArg, "="); index < 0 {
-					flagName = strings.TrimLeft(prevArg, "-")
-
+					if strings.HasPrefix(prevArg, "--") {
+						// Flag has full name
+						flagName = prevArg[2:]
+					} else {
+						// Flag is shorthand
+						// We have to get the last shorthand flag name
+						// e.g. `-asd` => d to provide the correct completion
+						// https://github.com/spf13/cobra/issues/1257
+						flagName = prevArg[len(prevArg)-1:]
+					}
 					// Remove the uncompleted flag or else there could be an error created
 					// for an invalid value for that flag
 					trimmedArgs = args[:len(args)-1]
--- a/completions_test.go
+++ b/completions_test.go
@@ -2158,3 +2158,96 @@ func TestCompleteCompletion(t *testing.T) {
 		}
 	}
 }
+
+func TestMultipleShorthandFlagCompletion(t *testing.T) {
+	rootCmd := &Command{
+		Use:       "root",
+		ValidArgs: []string{"foo", "bar"},
+		Run:       emptyRun,
+	}
+	f := rootCmd.Flags()
+	f.BoolP("short", "s", false, "short flag 1")
+	f.BoolP("short2", "d", false, "short flag 2")
+	f.StringP("short3", "f", "", "short flag 3")
+	_ = rootCmd.RegisterFlagCompletionFunc("short3", func(*Command, []string, string) ([]string, ShellCompDirective) {
+		return []string{"works"}, ShellCompDirectiveNoFileComp
+	})
+
+	// Test that a single shorthand flag works
+	output, err := executeCommand(rootCmd, ShellCompNoDescRequestCmd, "-s", "")
+	if err != nil {
+		t.Errorf("Unexpected error: %v", err)
+	}
+
+	expected := strings.Join([]string{
+		"foo",
+		"bar",
+		":4",
+		"Completion ended with directive: ShellCompDirectiveNoFileComp", ""}, "\n")
+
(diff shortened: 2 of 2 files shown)


marckhouzam:
Thanks @Luap99, this looks great.  The scenario with `=` does not work perfectly, but it's not critical.

Say I have 2 boolean flags `-b` and a string flag `-s` with completions `string1` and `string2`.

`./testprog -bs=<TAB>`  won't complete properly.

If fixing that makes the code much more complicated, we may want to ignore that case

Luap99:
@marckhouzam Thanks, this should work now.

custom_completions_test.go
```diff
@@ -1958,3 +1958,96 @@ func TestCompleteHelp(t *testing.T) {
...
+		"works",
+		":4",
+		"Completion ended with directive: ShellCompDirectiveNoFileComp", ""}, "\n")
+
+	if output != expected {
+		t.Errorf("expected: %q, got: %q", expected, output)
+	}
+
+	// Test that multiple boolean + string with equal sign with value shorthand flags work
+	output, err = executeCommand(rootCmd, ShellCompNoDescRequestCmd, "-sdf=abc", "")
```
marckhouzam, line 2039: Could you add one more test:
`output, err = executeCommand(rootCmd, ShellCompNoDescRequestCmd, "-sdf", "works", "")`

custom_completions.go
```diff
@@ -442,7 +442,11 @@ func checkIfFlagCompletion(finalCmd *Command, args []string, lastArg string) (*p
 	if len(lastArg) > 0 && lastArg[0] == '-' {
 		if index := strings.Index(lastArg, "="); index >= 0 {
 			// Flag with an =
-			flagName = strings.TrimLeft(lastArg[:index], "-")
+			if strings.HasPrefix(lastArg[:index], "--") {
+				flagName = lastArg[2:index]
+			} else {
```
marckhouzam, line 447: Could you put a comment to mention the handling the scenario of multiple short-hand flag, maybe with a short example?  I feel that in 2 months, I won't remember why we do this 😄

custom_completions.go
```diff
@@ -459,8 +463,11 @@ func checkIfFlagCompletion(finalCmd *Command, args []string, lastArg string) (*p
 				// If the flag contains an = it means it has already been fully processed,
 				// so we don't need to deal with it here.
 				if index := strings.Index(prevArg, "="); index < 0 {
-					flagName = strings.TrimLeft(prevArg, "-")
-
+					if strings.HasPrefix(prevArg, "--") {
+						flagName = prevArg[2:]
+					} else {
```
marckhouzam, line 468: Could you put a comment here too

marckhouzam (approved):
Looks good!  Just some minor nits.

custom_completions_test.go
Luap99: This does not work. :cry:   I think this is a internal cobra problem since the error comes from `c.Root().Find(trimmedArgs)`

marckhouzam: I believe you found a small bug in Cobra.  The `stripFlags()` function does not handle multiple short flags properly.

marckhouzam: It may become important to fix this as it sounds like it would affect some completion scenarios

marckhouzam:
I've just hit this bug using `kubectl`.  

I'm trying out Go completions and native zsh completion in `kubectl` to make sure the PR moving `kubectl` to go completions is good: https://github.com/kubernetes/kubernetes/pull/96087

For the `kubectl exec` command, I always use the form `kubectl exec -it` instead of `kubectl exec -i -t`, which triggers this bug.  It is properly handled for `bash` but would become a problem if/when `kubectl` moved to native zsh completion.

With this in mind, it would be nice to add this PR to the next Cobra release.  Would a maintainer be so kind as to add it to the 2.0.0 milestone? Thank you.

vrothberg:
@eparis, could you take a look?  Podman needs this patch and we don't want to vendor a fork too long. Downstream packaging is impacted by it.

Great work, @Luap99 !

marckhouzam:
@jpmcb This small fix would be nice to include as it causes a regression for bash V2 (#1146) compared to bash V1

Luap99:
> @jpmcb This small fix would be nice to include as it causes a regression for bash V2 (#1146) compared to bash V1

Not only bash v2 it also affects zsh and fish which are already used.

marckhouzam:
> Not only bash v2 it also affects zsh and fish which are already used.

In fact, I now realize it affects all Go completions (`ValidArgsFunction` and `RegisterFlagCompletionFunc()`), including bash v1.  This will cause a regression for `kubectl` when it moves to Go completions (see above comment https://github.com/spf13/cobra/pull/1258#issuecomment-732304019) /cc @jpmcb

Luap99:
I just rebased to resolve the merge conflict.

custom_completions_test.go
jpmcb: I'm assuming fixing `stripFlags()` is out of scope for this PR?

We may want a separate issue to track this and get it fixed, but if it's going to block functionality for shorthand flags working, we may want to tackle it now. Thoughts?

Luap99: This is definitely out of scope for this PR. This PR fixes a very annoying issue, I would love to have this merged.
There is already this issue https://github.com/spf13/cobra/issues/618.

jpmcb: Awesome - I'll approve and merge!

jpmcb (approved):
Thanks for this! LGTM

We also have https://github.com/spf13/cobra/issues/618 to track the found issue around the `stripFlags` function. We'll need to tackle that outside this PR.

PR #1153, opened 2020-07-09
Closes #1152 by upgrading yaml.v2 to v2.3.0

seanprince:
Upgrade to yaml.v2 v2.3.0 to fix vulnerability reported here - https://github.com/kubernetes/kubernetes/issues/89535

PR #1259, opened 2020-10-16
Don't use yaml.v2 2.3.0 which has a breaking change

marckhouzam:
This is to start a discussion to see what would be the best way forward.

PR #1153 moved Cobra to yaml.v2 2.3.0

This change has introduced a breaking change which is preventing `kubectl` and `helm` from using Cobra 1.1 (and probably others).  See https://github.com/kubernetes/kubernetes/pull/95571 and https://github.com/helm/helm/pull/8890#issuecomment-709320952

The breaking change has to do with line wrapping and was introduced by https://github.com/go-yaml/yaml/pull/571

The reason behind Cobra upgrading to yaml.v2 2.3.0 was to address a CVE reported by Kubernetes in https://github.com/kubernetes/kubernetes/issues/89535.  However the fix for Kubernetes was to move to yaml.v2 2.2.8 as seen in https://github.com/kubernetes/kubernetes/pull/87467.  yaml.v2 2.2.8 does not contain the breaking change.

Therefore, Cobra could instead use yaml.v2 2.2.8 and avoid bringing in the breaking change of 2.3.0.  This PR makes the suggested change.

**This would require doing an  quick release of Cobra to address the current problem.**

What are the maintainers thoughts on this?
@jharshman @jpmcb @wfernandes 

The longer term solution will need to be addressed by the yaml.v2 project.


marckhouzam:
@eparis  may have an opinion on this.

umarcor:
Can `kubectl` and `helm` use `1.0.0` until the issue is fixed in `yaml.v2` (or `yaml.v3`)?

marckhouzam:
> Can `kubectl` and `helm` use `1.0.0` until the issue is fixed in `yaml.v2` (or `yaml.v3`)?

Sure, that's what they are doing now.  It's just that the 1.1 release isn't that useful if multiple projects can't use it.  So really, this discussion is about trying to get Cobra improvements in the hands of users.

umarcor:
My only point is that, depending on the reponsiveness of yaml.v2 maintainers, it might be less disruptive to wait until their next release, and then bump cobra. Was it reported in go-yaml already? I didn't find any reference in the linked issues.

marckhouzam:
Someone from kubectl was supposed to reach out to them.  Should be happening soon.

thaJeztah:
would it be possible to do a new release with this PR and https://github.com/spf13/cobra/pull/1255 (which fixes packaging `.deb` packaging failures on the latest Ubuntu release)

jpmcb:
I will work on resolving this today.

I think what we should do is cut a quick patch release and deprecate the version of Yaml to what kubectl and helm need. 

Question: @marckhouzam was this an _unintended_ breaking change? Otherwise, I would have expected to see a yaml 3.x version denoting the breaking change.

jpmcb (approved):
👍

marckhouzam:
@jpmcb Thank you for jumping on this!

> Question: @marckhouzam was this an unintended breaking change? Otherwise, I would have expected to see a yaml 3.x version denoting the breaking change.

From what I understand, the fact that it was a breaking change was overlooked.
Looking at what the author of the change says here https://github.com/kubernetes/kubernetes/pull/95571#issuecomment-709445204, he does realize the change is breaking.  The change was brought in by https://github.com/go-yaml/yaml/pull/571, without much discussion.

jpmcb:
I see - thanks much for the quick response.

> From what I understand, the fact that it was a breaking change was overlooked.

That is unfortunate. We will go ahead and call this a bug in 1.1.x and fix it in 1.1.1 shortly. While I have you @marckhouzam any other bug fixes that should get into this release I'm about to cut?

marckhouzam:
> While I have you @marckhouzam any other bug fixes that should get into this release I'm about to cut?

For a 1.1.1 release, I believe the two PRs you just merged (this one and the manpages one) seem to be it.

jpmcb:
https://github.com/spf13/cobra/releases/tag/v1.1.1 is live!

marckhouzam:
> https://github.com/spf13/cobra/releases/tag/v1.1.1 is live!

Thanks! It's very appreciated!

PR #1270, opened 2020-11-02
Add the ability to specify a filePostpender in GenMarkdownTreeCustom

### LINKED ISSUE #1140: GenMarkdownTreeCustom doesn't have a way to append a footer
[GenMarkdownTreeCustom](https://godoc.org/github.com/spf13/cobra/doc#GenMarkdownTreeCustom) has a way to a header (`filePrepender` arg), however there is not a first class way to add a footer.

sonjoonho:
Closes #1140.

### DIFF
--- a/doc/md_docs.go
+++ b/doc/md_docs.go
@@ -122,17 +122,17 @@ func GenMarkdownCustom(cmd *cobra.Command, w io.Writer, linkHandler func(string)
 func GenMarkdownTree(cmd *cobra.Command, dir string) error {
 	identity := func(s string) string { return s }
 	emptyStr := func(s string) string { return "" }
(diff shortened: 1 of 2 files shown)


jpmcb (approved):
Thanks for this PR! 🚀 and apologies for the long review - been a busy last few months! Cheers

Everything looks good and thanks so much for adding tests!!

marckhouzam:
@jpmcb Helm's compilation is failing with this change.
With this PR, the public function `GenMarkdownTreeCustom()` now requires an extra argument.
Maybe we need a new command instead?

jpmcb:
Thanks for the heads up @marckhouzam - created #1317 to revert and we can revisit this. My apologies - should have caught that!

PR #1282, opened 2020-11-25
fix home directory config not loading

timecode:
### Issue 
The provision to use the convention of an extension-less config file from the user's home directory does not work as expected. The dependency being used to load the config file, `viper`, requires explicit knowledge of the "type" of the file, as it cannot deduce the implicit type from the file extension. Without knowledge of the file's "type", `viper` simply defaults to a "type" of `""`, which results in no attempt being made to load the file.

### Solution 
Specify the config type when specifying the extension-less config file name.

For example:
```go
viper.SetConfigName(".cobra")
viper.SetConfigType("yaml")
```

### Fix
This fix adds an explicit `viper.SetConfigType()` call at points referencing the typically extension-less config file


jharshman:
@timecode , can you provide a failing test that your solution fixes?

timecode:
> @timecode , can you provide a failing test that your solution fixes?

Hi, test added :-)

timecode:
Hi @jharshman, test added :-)

timecode:
😢

jharshman:
@timecode, apologies for letting this go to stale. Additionally, not sure what I was thinking, but you can remove that test. This is just adding an additional method from the viper package. No need to test. But thanks for updating the `.golden` file.

jharshman (approved):

jharshman:
@timecode merged. Thank you for your contribution and patience.

PR #1284, opened 2020-11-27
Fix fish completion for programs that output trailing empty lines

### LINKED ISSUE #1279: Fish completions with `:4` and `:0` as options
We recently enabled completions on goreleaser, and it works pretty well.

One of the issues we found, though, is that on fish it returns `:4` as an option, e.g.:

image

If I choose `:4` and keep pressing TAB, it will keep completing `:0` forever:

image

Same happens for subcommands, e.g.:

image

I think it might be related to the fact that we do some "hacky" stuff to make `release` be the default command instead of root (is there a better way of doing that?): https://github.com/goreleaser/goreleaser/blob/master/cmd/root.go#L30-L32

Anyway, not sure how much of this is a bug, how much is expected and how much is us doing things wrong.

marckhouzam:
Fixes #1279 

Some programs output extra empty lines at the end of their output, which affects the output of the `__complete` command.
`goreleaser` is such an case:
https://github.com/goreleaser/goreleaser/blob/f35534d9d7cbc2d69fcbd4edb78a5d668357fa60/cmd/root.go#L23

Those lines must be ignored for `fish` shell completion to work.
`zsh` and `bash` are not impacted.

To test this fix use the following commands for the program that follows.
```
$ go build -o testprog testprog.go
$ fish
$ ./testprog completion fish | source
$ ./testprog <TAB>

# Without this fix the completion directive :4 will be shown as a valid completion
```
**Test program which adds extra empty lines to the output**:
```
package main

import (
	"fmt"
	"os"

	"github.com/spf13/cobra"
)

var completionNoDesc = false

	Use:   "completion [bashzsh|fish|powershell]",
	Long: `To load completions:

Bash:

Linux:
  $ testprog completion bash > /etc/bash_completion.d/testprog
MacOS:
  $ testprog completion bash > /usr/local/etc/bash_completion.d/testprog

Zsh:

$ compdef _testprog testprog

$ testprog completion zsh > "${fpath[1]}/_testprog"

Fish:

$ testprog completion fish > ~/.config/fish/completions/testprog.fish

PowerShell:

To load completions for every new session, add the output of the above command
to your powershell profile.
`,
	DisableFlagsInUseLine: true,
		return []string{"bash\tBash completion", "zsh", "fish", "powershell"}, cobra.ShellCompDirectiveNoFileComp
	},
	Args: cobra.ExactValidArgs(1),
		switch args[0] {
		case "bash":
		case "zsh":
			if !completionNoDesc {
			} else {
			}
		case "fish":
		}
	},
}

var rootCmd = &cobra.Command{
	Use: "testprog",
		fmt.Println("rootCmd called")
	},
}

func setFlags() {
	rootCmd.PersistentFlags().String("theme", "", "theme to use (located in /themes/THEMENAME/)")
}

func main() {
		&completionNoDesc,
		"no-descriptions", false,

	setFlags()

	fmt.Println("")
	fmt.Println("")
	rootCmd.Execute()
	fmt.Println("     ")
	fmt.Println("")
}
```

### DIFF
--- a/fish_completions.go
+++ b/fish_completions.go
@@ -53,6 +53,19 @@ function __%[1]s_perform_completion
     __%[1]s_debug "Calling $requestComp"
 
     set results (eval $requestComp 2> /dev/null)
+
+    # Some programs may output extra empty lines after the directive.
+    # Let's ignore them or else it will break completion.
+    # Ref: https://github.com/spf13/cobra/issues/1279
+    for line in $results[-1..1]
+        if test (string trim -- $line) = ""
+            # Found an empty line, remove it
+            set results $results[1..-2]
+        else
+            # Found non-empty line, we have our proper output
+            break
+        end
+    end
     set comps $results[1..-2]
     set directiveLine $results[-1]
 


caarlos0 (approved):
Lgtm

fish_completions.go
```diff
@@ -53,6 +53,19 @@ function __%[1]s_perform_completion
     __%[1]s_debug "Calling $requestComp"
 
     set results (eval $requestComp 2> /dev/null)
+
+    # Some programs may output extra empty lines after the directive.
+    # Let's ignore them or else it will break completion.
+    # Ref: https://github.com/spf13/cobra/issues/1279
+    for line in $results[-1..1]
+        if test (string trim $line) = ""
```
krobelus, line 61:
```suggestion
        if test (string trim -- $line) = ""
```

in case a line starts with a dash (unlikely but you never know)

marckhouzam:
Thanks @krobelus for the suggestion.  I have included your suggested change, squashed the two commits and added you as co-author.

marckhouzam:
Waiting for review

krobelus:
Looks great FWIW.

jpmcb (approved):
LGTM - thanks for this PR and thanks for the patience! Glad to see we are getting some fish adoption after all the work that went into adding support for it!

PR #1292, opened 2020-12-04
Cobra User Contract

### LINKED ISSUE #1232: Cobra User Contract (Draft)
Below you will find a draft of expectations for the project. This has been put together and reviewed by maintainers.
We are now asking for community feedback.

### Versioning
Cobra will follow a steady release cadence. Users can expect a major release quarterly. Non breaking changes will also be released as minor versions quarterly. Patch bug releases are at the discretion of the maintainers. Users can expect security patch fixes to be released within the week of these becoming known. Releases will follow [Semantic Versioning](https://semver.org/). Users tracking the Master branch should expect unpredictable breaking changes as the project continues to move forward. For stability, it is highly recommended to use a release.

### Backward Compatibility
We will maintain two major releases in a moving window. The N-1 release will only receive bug fixes and security updates and will be dropped once N+1 is released.

### Deprecation
Deprecation of Go versions or dependent packages will only occur in major releases. To reduce the change of this taking users by surprise, any large deprecation will be preceded by an announcement in the [#cobra slack channel](https://gophers.slack.com/archives/CD3LP1199) and an Issue on Github.

### Communication
Cobra maintainers will use GitHub issues and the [#cobra slack channel](https://gophers.slack.com/archives/CD3LP1199) as the primary means of communication with the community. This is to foster open communication with all users and contributors.

### Breaking Changes
Breaking changes are generally allowed in the master branch, as this is the branch used to develop the next release of Cobra.

There may be times, however, when master is closed for breaking changes. This is likely to happen as we near the release of a new version.

Breaking changes are not allowed in release branches, as these represent minor versions that have already been released. These version have consumers who expect the APIs, behaviors, etc, to remain stable during the lifetime of the patch stream for the minor release.

Examples of breaking changes include:
- Removing or renaming exported constant, variable, type, or function.
- Updating the version of critical libraries such as `spf13/pflag`, `spf13/viper` etc...
  - Some version updates may be acceptable for picking up bug fixes, but maintainers must exercise caution when reviewing.

There may, at times, need to be exceptions where breaking changes are allowed in release branches. These are at the discretion of the project's maintainers, and must be carefully considered before merging.

### CI Testing
Maintainers will ensure the the Cobra test suite utilizes the current supported versions of Golang.

jharshman:
Establish an understanding between user and maintainer.
Set a goal for releases, security fixes and bug patches.

Closes #1232

### DIFF
--- a/CONDUCT.md
+++ b/CONDUCT.md
@@ -1,0 +1,37 @@
+## Cobra User Contract
+
+### Versioning
+Cobra will follow a steady release cadence. Non breaking changes will be released as minor versions quarterly. Patch bug releases are at the discretion of the maintainers. Users can expect security patch fixes to be released within relatively short order of a CVE becoming known. For more information on security patch fixes see the CVE section below. Releases will follow [Semantic Versioning](https://semver.org/). Users tracking the Master branch should expect unpredictable breaking changes as the project continues to move forward. For stability, it is highly recommended to use a release.
+
+### Backward Compatibility
+We will maintain two major releases in a moving window. The N-1 release will only receive bug fixes and security updates and will be dropped once N+1 is released.
+
+### Deprecation
+Deprecation of Go versions or dependent packages will only occur in major releases. To reduce the change of this taking users by surprise, any large deprecation will be preceded by an announcement in the [#cobra slack channel](https://gophers.slack.com/archives/CD3LP1199) and an Issue on Github.
+
+### CVE
+Maintainers will make every effort to release security patches in the case of a medium to high severity CVE directly impacting the library. The speed in which these patches reach a release is up to the discretion of the maintainers. A low severity CVE may be a lower priority than a high severity one.
+
+### Communication
+Cobra maintainers will use GitHub issues and the [#cobra slack channel](https://gophers.slack.com/archives/CD3LP1199) as the primary means of communication with the community. This is to foster open communication with all users and contributors.
+
+### Breaking Changes
+Breaking changes are generally allowed in the master branch, as this is the branch used to develop the next release of Cobra.
+
+There may be times, however, when master is closed for breaking changes. This is likely to happen as we near the release of a new version.
+
+Breaking changes are not allowed in release branches, as these represent minor versions that have already been released. These version have consumers who expect the APIs, behaviors, etc, to remain stable during the lifetime of the patch stream for the minor release.
+
+Examples of breaking changes include:
+- Removing or renaming exported constant, variable, type, or function.
+- Updating the version of critical libraries such as `spf13/pflag`, `spf13/viper` etc...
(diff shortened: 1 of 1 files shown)


CONDUCT.md
```diff
@@ -0,0 +1,37 @@
...
+Deprecation of Go versions or dependent packages will only occur in major releases. To reduce the change of this taking users by surprise, any large deprecation will be preceded by an announcement in the [#cobra slack channel](https://gophers.slack.com/archives/CD3LP1199) and an Issue on Github.
+
+### CVE
+Maintainers will make every effort to release security patches in the case of a medium to high severity CVE directly impacting the library. The speed in which these patches reach a release is up to the discretion of the maintainers. A low severity CVE may be a lower priority than a high severity one.
+
+### Communication
+Cobra maintainers will use GitHub issues and the [#cobra slack channel](https://gophers.slack.com/archives/CD3LP1199) as the primary means of communication with the community. This is to foster open communication with all users and contributors.
+
+### Breaking Changes
+Breaking changes are generally allowed in the master branch, as this is the branch used to develop the next release of Cobra.
```
marckhouzam, line 19: I'm curious about this. It seems to indicate that once a release is out, the master branch can accept breaking changes such as removing functions.  

But such changes will break clients at the next release, so I wonder if that is really what the sentence aims to convey?

marckhouzam (commented):
I love the predictability proposed here. As a user and contributor, it is really appreciated.

CONDUCT.md
jharshman: @marckhouzam, this sentence does feel a bit off. The point we are trying to convey here is that breaking changes _should_ only show up in major releases.  Where we keep those breaking changes til then is a matter of discussion.

jharshman: although... this might be fine if we consider that we might create release branches that are maintained through their lifecycle.
In that case all changes hit master and are then merged into the currently maintained release branch and published as minor releases until the next major release.

marckhouzam: That would work.   But as you point out every PR would need to come in for the release branch and the master branch (except breaking PRs, but this are rare anyway).

Also, it makes me wonder about patch releases. Would there be any? If so, from what branch would they be built?

If you're interested, here's the way helm does its release strategy.
- a minor release triggers a new release branch.  After the release only bug fixes go in that release branch for future patch releases
- the master branch is then used for the next minor release, so PRs target the master branch (which is simple for new contributors) and only bug fixes need to be cherry-picked into the release branch (this is relatively simple and is done by the maintainer on the patch release day, based on a PR tag)
- breaking changes are not committed anywhere but instead the PR is scheduled for the next major milestone; this works well because breaking changes are rare

I feel if this is a pretty simple process but it could be that I'm used to it 😁

jpmcb (approved):
This looks ready to go to me!

PR #1068, opened 2020-03-26
Partial Revert of #922

jharshman:
Issue Reference: https://github.com/spf13/cobra/issues/1056
Bug Introduced: #922 

#922 Introduced a new error type that emitted when a command was not runnable. This caused all commands w/o a run function set to error w/ that message and a status code of 1.

This change reverts the addition of that new error. Similar
functionality can be accomplished by leveraging RunE.

PR #1305, opened 2020-12-29
Bugfix: Requesting help in case of an error needs exit code > 0

gabyx:
Returning nil, when flag.ErrHelp is returned, should really not exit with '0'.

Thanks for merging.


gabyx:
PR #922 solves this.

gabyx:
As #922 is reverted: Shouln't this PR be feasible?

marckhouzam:
Hi @gabyx. Although this is a good idea there may be backwards-compatibility concerns.  Maybe #1068 which reverted prt of #922 can give some info.  Also the tests seenu to fail which could be a hint of the issue with such a change.

PR #1308, opened 2021-01-02
custom comp: do not complete flags after args when interspersed is false

### LINKED ISSUE #1307: custom completion should respect the flagset interspersed option
The flagset interspersed option indicates that flags which follow after the fist argument are also considered arguments and are not parsed as flag.

Consider this simple example:
```
package main

import (
	"fmt"

	"github.com/spf13/cobra"
)

var rootCmd = &cobra.Command{
	Use: "testprog",
		fmt.Println("rootCmd called")
	},
}

var subCmd = &cobra.Command{
	Use: "sub",
		fmt.Println(args)
	},
}

func main() {
	rootCmd.AddCommand(subCmd)
	flags := subCmd.Flags()
	flags.SetInterspersed(false)
	flags.Bool("test", false, "test flag")

	rootCmd.Execute()
}
```

The --test flag is added to args if an argument exists before:
```
$ ./testprog sub --test a
[a]
$ ./testprog sub a --test 
[a --test]
```

This is the expected behaviour however the custom completion logic does not respects this. It still provides flag completion even after the first arg is set.

```
$ ./testprog __complete sub a --t
--test  test flag
:4
```
This should not happen. Instead it should provide argument completion (e.g. ValidArgsFunction).

I looked at the code but I'm not sure how this could be implemented. The problem is that the interspersed option is not exported from the FlagSet so we cannot retrieve it.

@marckhouzam Any Ideas?

Luap99:
If the interspersed option is set false and one arg is already set all
following arguments are counted as arg and not parsed as flags. Because
of that we should not offer flag completion. The same applies to arguments
followed after `--`.

Fixes #1307

### DIFF
--- a/completions.go
+++ b/completions.go
@@ -21,6 +21,15 @@ const (
 // can be instructed to have once completions have been provided.
 type ShellCompDirective int
 
+type flagCompError struct {
+	subCommand string
+	flagName   string
+}
+
+func (e *flagCompError) Error() string {
+	return "Subcommand '" + e.subCommand + "' does not support flag '" + e.flagName + "'"
+}
+
 const (
 	// ShellCompDirectiveError indicates an error occurred and completions should be ignored.
 	ShellCompDirectiveError ShellCompDirective = 1 << iota
@@ -224,18 +233,35 @@ func (c *Command) getCompletions(args []string) (*Command, []string, ShellCompDi
 	// This is important because if we are completing a flag value, we need to also
 	// remove the flag name argument from the list of finalArgs or else the parsing
 	// could fail due to an invalid value (incomplete) for the flag.
-	flag, finalArgs, toComplete, err := checkIfFlagCompletion(finalCmd, finalArgs, toComplete)
-	if err != nil {
-		// Error while attempting to parse flags
-		return finalCmd, []string{}, ShellCompDirectiveDefault, err
-	}
+	flag, finalArgs, toComplete, flagErr := checkIfFlagCompletion(finalCmd, finalArgs, toComplete)
+
+	// Check if interspersed is false or -- was set on a previous arg.
+	// This works by counting the arguments. Normally -- is not counted as arg but
+	// if -- was already set or interspersed is false and there is already one arg then
+	// the extra added -- is counted as arg.
+	flagCompletion := true
+	_ = finalCmd.ParseFlags(append(finalArgs, "--"))
+	newArgCount := finalCmd.Flags().NArg()
 
 	// Parse the flags early so we can check if required flags are set
 	if err = finalCmd.ParseFlags(finalArgs); err != nil {
 		return finalCmd, []string{}, ShellCompDirectiveDefault, fmt.Errorf("Error while parsing flags from args %v: %s", finalArgs, err.Error())
 	}
 
-	if flag != nil {
+	realArgCount := finalCmd.Flags().NArg()
+	if newArgCount > realArgCount {
+		// don't do flag completion (see above)
+		flagCompletion = false
+	}
+	// Error while attempting to parse flags
+	if flagErr != nil {
+		// If error type is flagCompError and we don't want flagCompletion we should ignore the error
+		if _, ok := flagErr.(*flagCompError); !(ok && !flagCompletion) {
+			return finalCmd, []string{}, ShellCompDirectiveDefault, flagErr
+		}
+	}
+
+	if flag != nil && flagCompletion {
 		// Check if we are completing a flag value subject to annotations
 		if validExts, present := flag.Annotations[BashCompFilenameExt]; present {
 			if len(validExts) != 0 {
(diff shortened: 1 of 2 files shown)


custom_completions_test.go
```diff
@@ -1697,6 +1697,107 @@ func TestValidArgsFuncChildCmdsWithDesc(t *testing.T) {
...
+
+	if output != expected {
+		t.Errorf("expected: %q, got: %q", expected, output)
+	}
+
+	// Test that no flags are completed after the -- arg with a flag set
+	output, err = executeCommand(rootCmd, ShellCompRequestCmd, "--bool", "--", "-")
+	if err != nil {
+		t.Errorf("Unexpected error: %v", err)
+	}
```
marckhouzam, line 1743: I think we need one more test:
```
output, err = executeCommand(rootCmd, ShellCompRequestCmd, "--", "--string", "")
```
to make sure that we don't get flag value completion but get `ValidArgsFunction` instead.

custom_completions.go
```diff
@@ -284,7 +298,7 @@ func (c *Command) getCompletions(args []string) (*Command, []string, ShellCompDi
 
 	var completions []string
 	directive := ShellCompDirectiveDefault
-	if flag == nil {
+	if flag == nil && flagCompletion {
```
marckhouzam, line 301: I'm unsure about this one.  I believe it disables subcommand completion and `ValidArgs` completion for the non-interspersed case, right?

I think `ValidArgs` should still be done.
And I'm thinking subcommand completion should also be done. case like:
```
testprog --bool subcomm<TAB>
```
What do you think?

marckhouzam (commented):
Nicely done!

custom_completions.go
Luap99: You are correct this is wrong. It works fine without it. I added a test for it.

custom_completions_test.go
Luap99: Good catch this one was not working correct.

marckhouzam (commented):
This is really nice.
I have mentioned in the comments two more problematic cases, but they are corner cases, especially the one about the `args` being incomplete.

I don't know if supporting these cases is worth the possible extra complexity to the code, but I haven't tried to fix it myself, so I'm not sure.  I personally would not consider them blockers as this PR already improves this scenario for the vast majority of cases.

Thanks for the continuing great work @Luap99 !

custom_completions_test.go
```diff
@@ -1697,6 +1697,159 @@ func TestValidArgsFuncChildCmdsWithDesc(t *testing.T) {
...
+	if output != expected {
+		t.Errorf("expected: %q, got: %q", expected, output)
+	}
+
+	// Check that args are still completed even if flagname with ValidArgsFunction exists
+	output, err = executeCommand(rootCmd, ShellCompRequestCmd, "child", "--", "--string", "")
+	if err != nil {
+		t.Errorf("Unexpected error: %v", err)
+	}
+
```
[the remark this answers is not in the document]
Luap99: Updated, the logic is definitely more complex now but I think it is still fine.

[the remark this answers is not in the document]
marckhouzam: You did a good job, it is not that complex.

custom_completions_test.go
```diff
@@ -1697,6 +1697,228 @@ func TestValidArgsFuncChildCmdsWithDesc(t *testing.T) {
...
+	if output != expected {
+		t.Errorf("expected: %q, got: %q", expected, output)
+	}
+
+	// Check that --validarg is added to args for the ValidArgsFunction
+	childCmd.ValidArgsFunction = func(cmd *Command, args []string, toComplete string) ([]string, ShellCompDirective) {
+		return args, ShellCompDirectiveDefault
+	}
+	output, err = executeCommand(rootCmd, ShellCompRequestCmd, "child", "--", "--validarg", "")
```
marckhouzam, line 1888: We have to use a valid flag for this test, which will then trigger the problem:
```
output, err = executeCommand(rootCmd, ShellCompRequestCmd, "child", "--", "--string", "")
```

`--validarg` is not found as a valid flag by `checkIfFlagCompletion()` so your new logic handling errors is triggered.  However, with `--string`, there will not be an error as it is a valid flag and the `--string` argument will be trimmed:
https://github.com/spf13/cobra/blob/a4ab3fa09e3d6c4ac3fb7deece97a910957305ab/custom_completions.go#L466

But this is such a rare (and possibly invalid) case that I don't think we should fix it (unless it is a very simple fix).

marckhouzam (approved):
This looks good to me.  It makes our completion support even more complete (no pun intended 😄 ).

custom_completions_test.go
Luap99: You are right catching these cases is more difficult and would require a bigger refactor of the `checkIfFlagCompletion` function.

Luap99:
I just rebased to solve the merge conflict.

marckhouzam:
@jpmcb  If you still have energy and time, this is another fix for completion which I had forgotten to mention.  This will make the last Fish shell test of the `cobra-completion-testing` repo work.

jpmcb (changes requested):
Hi @Luap99 - getting a number of these PRs in to cut a release. Do you mind rebasing and resolving the conflict?

Luap99:
@jpmcb done

jpmcb (approved):

PR #1309, opened 2021-01-04
Bump license year to 2021 in golden files

Bowbaq:

PR #1310, opened 2021-01-04
Clarify documentation for completion

marckhouzam:
It was pointed out on Slack that there is no explanation in the docs of what the `toComplete` parameter represents for `ValidArgsFunction`.

This PR explains with examples the `toComplete` and the `args` parameters of `ValidArgsFunction`.

A section is also added to explain about fuzzy completions.

### DIFF
--- a/site/content/completions/_index.md
+++ b/site/content/completions/_index.md
@@ -174,8 +174,8 @@ cmd := &cobra.Command{
 	Use:   "status RELEASE_NAME",
 	Short: "Display the status of the named release",
 	Long:  status_long,
-	RunE: func(cmd *cobra.Command, args []string) {
-		RunGet(args[0])
+	Run: func(cmd *cobra.Command, args []string) {
+		RunStatus(args[0])
 	},
 	ValidArgsFunction: func(cmd *cobra.Command, args []string, toComplete string) ([]string, cobra.ShellCompDirective) {
 		if len(args) != 0 {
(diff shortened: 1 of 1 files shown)


marckhouzam:
I believe the test failures are because of the `2020` years that need to be updated, see #1309

shell_completions.md
```diff
@@ -159,13 +159,14 @@ cmd := &cobra.Command{
 	},
 }
 ```
-Where `getReleasesFromCluster()` is a Go function that obtains the list of current Helm releases running on the Kubernetes cluster.
-Notice we put the `ValidArgsFunction` on the `status` sub-command. Let's assume the Helm releases on the cluster are: `harbor`, `notary`, `rook` and `thanos` then this dynamic completion will give results like:
+Where `getReleasesFromCluster()` is a Go function that returns the list of current Helm release names running on the Kubernetes cluster.
+Similarly as for `Run`, the `args` parameter represents the arguments present on the command-line, while the `toComplete` parameter represents the final argument which the user is trying to complete (e.g., `helm status th<TAB>` will have `toComplete` be `"th"`); the `toComplete` parameter will be empty when the user has requested completions right after typing a space (e.g., `helm status <TAB>`). Notice we put the `ValidArgsFunction` on the `status` sub-command, as it provides completions for this sub-command specifically. Let's assume the Helm releases on the cluster are: `harbor`, `notary`, `rook` and `thanos` then this dynamic completion will give results like:
```
Luap99, line 163: That explains `toComplete` nicely but `args` is not really explained. I know the `args` parameter is common in cobra functions but I think it's important to mention that you can use `len(args)` to control the completion based on the number of arguments given like in the example above.

marckhouzam:
I forgot about this one.  I will rework it.

shell_completions.md
marckhouzam: Well, it took me almost 3 years, but I finally took time to clean up this PR and add more details.

marckhouzam:
This documentation-only PR is ready for review @jpmcb

site/content/completions/_index.md
```diff
@@ -174,8 +174,8 @@ cmd := &cobra.Command{
 	Use:   "status RELEASE_NAME",
 	Short: "Display the status of the named release",
 	Long:  status_long,
-	RunE: func(cmd *cobra.Command, args []string) {
-		RunGet(args[0])
+	Run: func(cmd *cobra.Command, args []string) {
+		RunStatus(args[0])
```
Luap99, line 178: Not sure why this is changed but it shouldn't matter either way as it is an example only so I am fine with that.

Luap99 (approved):
LGTM better late than never.

site/content/completions/_index.md
marckhouzam: That was just a cleanup. The example is about the `helm status` command but I had left the function call to `RunGet` which was a copy/paste from the `kubectl get` example.

marckhouzam:
Things have evolved, not much need for this

PR #1317, opened 2021-01-23
Revert "Add the ability to specify a filePostpender in GenMarkdownTreeCustom"

jpmcb:
Reverts spf13/cobra#1270

The public method requiring a new arguments is a breaking change. Reverting back for now.


marckhouzam:
Thanks for quick reaction @jpmcb !

PR #1321, opened 2021-01-28
fix: unbound variables in bash completion

edentsai:
when `set -o nounset` in Bash,
the warnings of unbound variables will break the bash completion.

use `kubectl` as example:

```sh
$ set -o nounset
$ kubectl <Tab>-bash: BASH_COMP_DEBUG_FILE: unbound variable
$
```

the warning break bash completion without any completion result,
and cause my cursor move to the newline.

Use `${variable:-}` substitution in Bash ,
that assign an empty string as default for unbound variables to fix the warnings.

### DIFF
--- a/bash_completions.go
+++ b/bash_completions.go
@@ -24,7 +24,7 @@ func writePreamble(buf io.StringWriter, name string) {
 	WriteStringAndCheck(buf, fmt.Sprintf(`
 __%[1]s_debug()
 {
-    if [[ -n ${BASH_COMP_DEBUG_FILE} ]]; then
+    if [[ -n ${BASH_COMP_DEBUG_FILE:-} ]]; then
         echo "$*" >> "${BASH_COMP_DEBUG_FILE}"
     fi
 }
@@ -187,7 +187,7 @@ __%[1]s_handle_reply()
                     PREFIX=""
                     cur="${cur#*=}"
                     ${flags_completion[${index}]}
-                    if [ -n "${ZSH_VERSION}" ]; then
+                    if [ -n "${ZSH_VERSION:-}" ]; then
                         # zsh completion needs --flag= prefix
                         eval "COMPREPLY=( \"\${COMPREPLY[@]/#/${flag}=}\" )"
                     fi
@@ -272,7 +272,7 @@ __%[1]s_handle_flag()
 
     # if a command required a flag, and we found it, unset must_have_one_flag()
     local flagname=${words[c]}
-    local flagvalue
+    local flagvalue=""
     # if the word contained an =
     if [[ ${words[c]} == *"="* ]]; then
         flagvalue=${flagname#*=} # take in as flagvalue after the =
@@ -291,7 +291,7 @@ __%[1]s_handle_flag()
 
     # keep flag value with flagname as flaghash
     # flaghash variable is an associative array which is only supported in bash > 3.
-    if [[ -z "${BASH_VERSION}" || "${BASH_VERSINFO[0]}" -gt 3 ]]; then
+    if [[ -z "${BASH_VERSION:-}" || "${BASH_VERSINFO[0]:-}" -gt 3 ]]; then
         if [ -n "${flagvalue}" ] ; then
             flaghash[${flagname}]=${flagvalue}
         elif [ -n "${words[ $((c+1)) ]}" ] ; then
@@ -363,7 +363,7 @@ __%[1]s_handle_word()
         __%[1]s_handle_command
     elif __%[1]s_contains_word "${words[c]}" "${command_aliases[@]}"; then
         # aliashash variable is an associative array which is only supported in bash > 3.
-        if [[ -z "${BASH_VERSION}" || "${BASH_VERSINFO[0]}" -gt 3 ]]; then
+        if [[ -z "${BASH_VERSION:-}" || "${BASH_VERSINFO[0]:-}" -gt 3 ]]; then
             words[c]=${aliashash[${words[c]}]}
             __%[1]s_handle_command
         else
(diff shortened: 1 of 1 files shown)


marckhouzam:
This seems reasonable.
Why do you need to change `${has_completion_function:-}` and `${last_command:-}`?  Those two variables are declared at the start of the script in `__start_kubectl()`.  Do they still cause unbound variable warnings?

edentsai:
> Do they still cause unbound variable warnings?

these variables cause unbound variable warnings.

edentsai:
I just checked the `__start_kubectl()`:

```sh
__start_kubectl()
{
    # ...
    local has_completion_function
    local last_command
    # ...
}
```

`__start_kubectl()` declared the variables with `local` scope that cannot share with other functions.

eparis:
if you're saying tha tlat_command can never be bound then this isn't the right fix, is it?

edentsai:
> __start_kubectl() declared those variables with local scope that cannot share with other functions.

Sorry, my guess is wrong.

i just make a small script to test the case, 
those local variables can be shared when call other functions in the __start_kubectl().

edentsai:
I didn't dig deep whole flow about the kubectl completion, 
just simply fix the unbound variable warnings:

```sh
$ set -o nounset
$ kubectl <Tab>-bash: BASH_COMP_DEBUG_FILE: unbound variable
$
```

```sh
# After fixed "$BASH_COMP_DEBUG_FILE" warning:
$ set -o nounset 
$ kubectl <Tab>-bash: last_command: unbound variable
```

```sh
# After fixed "$last_command" warning:
$ set -o nounset 
$ kubectl <Tab>-bash: has_completion_function: unbound variable
```

maybe we can take a deep look how `last_command` and `has_completion_function` will cause the unbound variables.

edentsai:
I found the root cause:

- Declare a variable `WITHOUT value` will cause unbound variable warning

Here is a script `demo.sh` to reproduce the issue:

```sh demo.sh
#!/usr/bin/env bash

__start_demo()
{
  printf "[INFO] START\n"

  # declare a variable WITHOUT VALUE will cause unbound variable warning.
  local last_command
  printf "[INFO] last_command: ${last_command}\n"

  printf "[INFO] END\n"
}

complete -o default -F __start_demo demo
```

then use `demo.sh` to reproduce the issue:

```sh
$ set -o nounset
$ source demo.sh
$ demo <Tab>demo [INFO] START
-bash: last_command: unbound variable
```

Just declare the variable with an empty string to fix the issue:

```diff
-   local last_command
+   local last_command=""
```

```sh
$ set -o nounset
$ source demo.sh
$ demo <Tab>[INFO] START
[INFO] last_command:
[INFO] END

# finally, the warning is fixed.
```

edentsai:
I updated the changes to declare `last_command` and `has_completion_function` with an empty string.

edentsai:
> This seems reasonable.
> Why do you need to change `${has_completion_function:-}` and `${last_command:-}`? Those two variables are declared at the start of the script in `__start_kubectl()`. Do they still cause unbound variable warnings?

marckhouzam (approved):
Thanks for the investigation @edentsai.  This make sense to me.

@eparis what do you think?

scop:
I filed #1321 yesterday without being aware of this PR, sorry about that. However I think that one is more complete and subjectively better stylewise than the one here. But I don't really care that deeply how, as long as the issues get fixed. Just pointing that one out in case you wish to update this one with the missing parts (flagvalue, ZSH_VERSION).

edentsai:
Hi @scop, 

I can help to add the missing parts of #1351 into this PR if you don't mind. :)

edentsai:
I updated this PR: 

- rebased on master
- add missing parts of #1351

scop:
>     * add the missing parts of #1351 (BASH_VERSION, ZSH_VERSION, flagvalue)

Cool. BASH_VERSION was not in #1351 though, and I'm not sure why it is being checked for emptiness in the first place (maybe the intent is to support loading the bash completion in some non-bash shell?), but since the check is there, guarding it for undefinedness doesn't hurt much. It does have a bit of a code smell now that BASH_VERSION is being guarded, but BASH_VERSINFO is accessed unguarded.

edentsai:
> maybe the intent is to support loading the bash completion in some non-bash shell?

just treat BASH_VERSION like ZSH_VERSION,
maybe someone make a mistake to load the bash/zsh completion in a wrong shell.
although the case should occur some syntax errors before unbound variable warnings.

> It does have a bit of a code smell now that BASH_VERSION is being guarded, but BASH_VERSINFO is accessed unguarded.

humm. I think i should also guard the BASH_VERSINFO,
i will update it later.

edentsai:
PR updated.

scop:
Bump

scop:
Ping :) More duplicates are starting to crop up, e.g. #1441

marckhouzam:
Apologies for missing this PR for the 1.2 release. I'll try to get it added to the list for the next release.

Also, I'll try to add a test for `set -o nounset` in the repo https://github.com/marckhouzam/cobra-completion-testing to avoid introducing other such issues in the future.

scop:
Cool to hear that such a repo exists :+1: 

Another similar option worth considering in the tests would be `set -o posix`, if the limitations are acceptable. But I see at least command substitution (`<( ... )`) is being used, that'll fail in POSIX mode, so maybe not/JFYI for the time being.

scop:
Hi @marckhouzam, anything I can do to help get this moving forward?

scop:
Also, would be good if someone with powers could remove the stale label (unsure if something eventually closes stale bugs), but this isn't really stale anyway.

marckhouzam:
> Hi @marckhouzam, anything I can do to help get this moving forward?

I expect that once the next release of Cobra is planned, the maintainers will start merging things, so nothing to do right now.  Thanks for your patience @scop

stmcginnis (approved):
This looks good to me.

jpmcb (approved):
Thanks all for the patience here - looks good to me and looks like some good conversation. Let's get this in.

PR #1333, opened 2021-02-03
Required flag with a default value should not return an error

owenthereal:
Fix a bug that required flag with a default value should not return an error.


marckhouzam:
Thanks @owenthereal for the contribution.

I wonder however, why a program would set a flag as required if the program provided a default values for that flag? It sounds to me like a mistake that the program should correct.

What do you think?

PR #1245, opened 2020-10-04
Build Hugo site with GitHub Actions and deploy to GitHub Pages

umarcor:
In this PR, a GitHub Actions workflow is added for building a web site with [Hugo](https://gohugo.io/) and deploying it to GitHub Pages (branch `gh-pages`).

- The content of the README is split to multiple markdown sources.
- Markdown sources from the root and from subdir `doc` are moved to `site/content` .
  - Documentation sources regarding completions are placed in `site/content/completions`.
  - Documentation sources regarding documentation generation are placed in `site/content/docgen`.
- Hugo configuration files are added to `site`. @bep's [docuapi](https://github.com/bep/docuapi) theme is used.
- In a GitHub Actions workflow, Hugo is installed, the theme is installed, the site is built, and the site is uploaded as an artifact as well as pushed to branch `gh-pages`.
- The logo from the README is used in the site (with a different font color).
- Any contributor that forks the repo and pushes some change, will have the site automatically built and uploaded to `username.github.io/cobra`.

See:

- Proposed README: https://github.com/umarcor/cobra/blob/ci-site/README.md
- Proposed site: https://umarcor.github.io/cobra
- Example CI run: https://github.com/umarcor/cobra/actions/runs/288401296 (note the artifact)

Ref #985 #1243 #1244

PR #1338, opened 2021-02-05
Add PR labeler with pull_request_target

### LINKED ISSUE #1092: [Bug][Contributor workflow] Github actions labeler doesn't work for forks

jpmcb:
The github [actions/labeler](https://github.com/actions/labeler) workflow now supports the `pull_request_target` trigger which will enable us to have labels on incoming PRs that originated from forks.

This was originally a permissions issue with action/labeler since they did not support actions being run directly from forks. This was documented [in this issue](https://github.com/actions/labeler/issues/12). This new feature has been documented [here](https://github.com/actions/labeler/blob/main/README.md)

And since this is supported directly by Github, I anticipate this will be a much better collaborator experience vs the periodic-labeler tried in #1112

Closes #1092

Let me know if this needs anything else!

### DIFF
--- a/.github/labeler.yml
+++ b/.github/labeler.yml
@@ -1,10 +1,16 @@
-# changes to documentation
-"area/documentation": doc/**/*
+# changes to documentation generation
+"area/doc-gen": doc/**/*
 
-# changes to the core lib package
+# changes to the core Go cobra lib package
 "area/lib": ./*.go
 
-# changes to the shell completion
+# changes to the Cobra CLI
+"area/cli": cobra/**/*
+
+# changes to Github workflows
+"area/github": .github/**/*
+
+# changes to shell completions
 "area/*sh completion":
   - ./*completions*
 
--- a/.github/labeler.yml
+++ b/.github/labeler.yml
@@ -4,10 +4,7 @@
 # changes to the core lib package
 "area/lib": ./*.go
 
-# changes to the zsh completion
+# changes to the shell completion
 "area/*sh completion":
-  - ./zsh_*
-  - ./shell_*
-  - ./powershell_*
-  - ./bash_*
+  - ./*completions*
 
--- a/.github/labeler.yml
+++ b/.github/labeler.yml
@@ -1,0 +1,13 @@
+# changes to documentation
+"area/documentation": doc/**/*
+
+# changes to the core lib package
+"area/lib": ./*.go
+
+# changes to the zsh completion
+"area/*sh completion":
+  - ./zsh_*
+  - ./shell_*
+  - ./powershell_*
+  - ./bash_*
+
--- a/.github/workflows/labeler.yml
+++ b/.github/workflows/labeler.yml
@@ -1,0 +1,12 @@
+name: "Pull Request Labeler"
+on:
+- pull_request_target
+
+jobs:
+  triage:
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/labeler@main
+      with:
+        repo-token: "${{ secrets.GITHUB_TOKEN }}"
+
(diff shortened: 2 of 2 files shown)


.github/labeler.yml
```diff
@@ -0,0 +1,13 @@
...
+
+# changes to the core lib package
+"area/lib": ./*.go
+
+# changes to the zsh completion
+"area/*sh completion":
+  - ./zsh_*
+  - ./shell_*
+  - ./powershell_*
+  - ./bash_*
```
Luap99, line 12: This should also match `fish_completions*` and `
custom_completions*`. Maybe it is easier to just match on the `completions` substring.

umarcor:
@jpmcb, I'd be glad if this was discussed in #952, which has been waiting for feedback from maintainers since ~18 months ago.

.github/labeler.yml
```diff
@@ -0,0 +1,13 @@
+# changes to documentation
+"area/documentation": doc/**/*
+
+# changes to the core lib package
+"area/lib": ./*.go
+
+# changes to the zsh completion
```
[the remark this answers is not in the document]
jpmcb: Good catch!!

.github/labeler.yml
jpmcb: True - ended up just globing for "completions". Thanks for this!!

.github/labeler.yml
```diff
@@ -0,0 +1,10 @@
+# changes to documentation
+"area/documentation": doc/**/*
```
umarcor, line 2: The `doc` subdir in cobra is *not* for documentation. It's the documentation generation codebase of the cobra lib. My proposal in #1245 is to use `site` for the actual documentation of cobra.

.github/labeler.yml
```diff
@@ -0,0 +1,10 @@
...
+
+# changes to the core lib package
+"area/lib": ./*.go
```
umarcor, line 5: I think that this should include `doc/**`. At the same time an "area/cli" category can be created for `cobra/**`.

.github/workflows/labeler.yml
```diff
@@ -0,0 +1,12 @@
...
+on:
+- pull_request_target
+
+jobs:
+  triage:
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/labeler@main
+      with:
+        repo-token: "${{ secrets.GITHUB_TOKEN }}"
```
umarcor, line 11: `${{ github.token }}` is equivalent, just slightly nicer.

.github/workflows/labeler.yml
```diff
@@ -0,0 +1,12 @@
+name: "Pull Request Labeler"
+on:
+- pull_request_target
+
+jobs:
+  triage:
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/labeler@main
```
umarcor, line 9: Maybe we want to pin it to v3?

.github/labeler.yml
jpmcb: Updated this to reflect the docs generation: `doc-gen`

umarcor (approved):

PR #1323, opened 2021-01-29
Deprecate Go < 1.14

jharshman:
In accordance with our adopted best practices, the main branch and the
next major release of Cobra will deprecate older and un-maintained
versions of Golang.

fix #1322

PR #1143, opened 2020-06-22
Add prow-github-actions automatic PR labeler

jpmcb:
I've created a github actions workflow with a feature to automatically label PRs. A persistent problem with the [github sponsored labeler](https://github.com/actions/labeler) is that it will _not_ work from forks (which is a typical project management flow for most open source projects). To get around this, [prow-github-actions](https://github.com/jpmcb/prow-github-actions) will run from the default branch of the repository within a cron job where the github actions token has appropriate permissions.

This new action will run every hour and label PRs based on the globs found in `labels.yml`.

Here is a working example of this: https://github.com/jpmcb/prow-github-actions/pull/12

Let me know if there are any questions!!!

PR #1339, opened 2021-02-08
ci: add GitHub Actions workflow 'Test'

umarcor:
Travis changed their pricing model. Free resources for open source projects were significantly reduced, and maintainers need to periodically request additional "credits". Since this projects is in spf13's namespace, I guess that only he can see the state of the credits.

In practice, most open source projects are migrating from Travis CI to somewhere else, mostly GitHub Actions.

- https://blog.travis-ci.com/2020-11-02-travis-ci-new-billing
- https://github.com/python-pillow/Pillow/issues/5028
- https://www.jeffgeerling.com/blog/2020/travis-cis-new-pricing-plan-threw-wrench-my-open-source-works

Refs: #952 #968 #1044 #1075 #1074 #1143 #1323 #1338

### DIFF
--- a/.github/workflows/Test.yml
+++ b/.github/workflows/Test.yml
@@ -24,7 +24,7 @@ jobs:
     runs-on: ${{ matrix.platform }}-latest
     steps:
 
-    - uses: actions/setup-go@v1
+    - uses: actions/setup-go@v2
       with:
         go-version: ${{ matrix.go }}
 
--- a/.github/workflows/Test.yml
+++ b/.github/workflows/Test.yml
@@ -30,6 +30,12 @@ jobs:
 
     - uses: actions/checkout@v2
 
+    - uses: actions/cache@v2
+      with:
+        path: ~/go/pkg/mod
+        key: ${{ runner.os }}-${{ matrix.go }}-${{ hashFiles('**/go.sum') }}
+        restore-keys: ${{ runner.os }}-${{ matrix.go }}-
+
     - run: |
         export GOBIN=$HOME/go/bin
         curl -sSfL https://raw.githubusercontent.com/golangci/golangci-lint/master/install.sh | bash -s -- -b $GOBIN latest
@@ -61,6 +67,12 @@ jobs:
           mingw-w64-x86_64-go
 
     - uses: actions/checkout@v2
+
+    - uses: actions/cache@v2
+      with:
+        path: ~/go/pkg/mod
+        key: ${{ runner.os }}-${{ matrix.go }}-${{ hashFiles('**/go.sum') }}
+        restore-keys: ${{ runner.os }}-${{ matrix.go }}-
 
     - run: |
         export GOBIN=$HOME/go/bin
--- a/.github/workflows/Test.yml
+++ b/.github/workflows/Test.yml
@@ -1,0 +1,80 @@
+name: Test
+
+on:
+  push:
+  pull_request:
+
+env:
+  GO111MODULE: on
+
+jobs:
+
+
+  test-unix:
+    strategy:
+      fail-fast: false
+      matrix:
+        platform:
+        - { icon: 🐧, os: ubuntu }
+        - { icon: 🍎, os: macOS }
+        go:
+        - 1.14.x
+        - 1.15.x
+    name: '${{ matrix.platform.icon }} ${{ matrix.go }}'
+    runs-on: ${{ matrix.platform.os }}-latest
+    steps:
+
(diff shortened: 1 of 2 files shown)


.github/workflows/Test.yml
```diff
@@ -0,0 +1,79 @@
...
+      uses: actions/checkout@v2
+
+    - name: '⚙️ Install deps'
+      run: |
+        curl -sSfL https://raw.githubusercontent.com/golangci/golangci-lint/master/install.sh | bash -s -- -b $HOME/go/bin/ latest
+        go install github.com/kyoh86/richgo
+        go install github.com/mitchellh/gox
+
+    - name: '🚧 Run tests'
+      run: PATH=$HOME/go/bin/:$PATH make
```
jpmcb, line 42: Will this run all the make tasks? Like `test`, `fmt` etc.

.github/workflows/Test.yml
```diff
@@ -0,0 +1,79 @@
...
+        shell: msys2 {0}
+    runs-on: windows-latest
+    steps:
+
+    - name: '⚙️ Git config'
+      shell: bash
+      run: git config --global core.autocrlf input
+
+    - name: '🟪 Setup MSYS2'
+      uses: msys2/setup-msys2@v2
```
jpmcb, line 58: We should use the official virtual environments: https://github.com/actions/virtual-environments

jpmcb (changes requested):
Looks like you beat me to it 😆 https://github.com/spf13/cobra/pull/1340

I think we can combine our efforts. And looks like yours is much more robust (and has the latest linting PR that got merged). Although I would like to see some changes:

.github/workflows/Test.yml
```diff
@@ -0,0 +1,79 @@
...
+
+jobs:
+
+
+  test:
+    strategy:
+      fail-fast: false
+      matrix:
+        platform:
+        - { icon: 🐧, os: ubuntu }
```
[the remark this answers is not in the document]
umarcor: Emojis/icons are not fancy, but for making communication better, and allowing developers/contributors to more easily spot the failing jobs/steps. That's specially relevant as the complexity of the workflows is increased. See, for instance, https://github.com/ghdl/ghdl/actions/runs/545530319.

Anyway, I will remove then.

.github/workflows/Test.yml
```diff
@@ -0,0 +1,79 @@
...
+        - { icon: 🍎, os: macOS }
+        go:
+        - 1.14.x
+        - 1.15.x
+    name: '${{ matrix.platform.icon }} ${{ matrix.go }}'
```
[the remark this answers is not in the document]
umarcor: Sure.

.github/workflows/Test.yml
umarcor: Yes:

- https://github.com/spf13/cobra/blob/master/Makefile#L16
- https://github.com/spf13/cobra/blob/master/Makefile#L26

.github/workflows/Test.yml
umarcor: We are using the official virtual environments (`ubuntu-latest`, `macos-latest` and `windows-latest`). MSYS2 is preinstalled in `windows-latest`, and Action setup-msys2 allows to either set it up, or get a clean tarball and extract it. As explained in https://github.com/msys2/setup-msys2#context, the problem with the built-in setup is that up to 10-12min update times might be required. That's why the recommended approach is using the clean tarball (with caching enabled by default).

My perception is that you mean to use cmd/powershell and not to rely on MSYS2. I can understand that wish, but AFAICT cobra was never tested on Windows. Moreover, the build/testing plumbing depends on Unix tooling which is not available on Windows by default. Hence, reworking the testing approach for supporting a regular Windows environment is out of the scope of this PR. It can be added as an additional job/matrix afterwards.

jpmcb: I see - thanks for the detailed explanation! And that makes sense. If this provides a clean, cached tar for us to use in the action, sounds good to me. I also looked over the repo and the Github org and seems 👍

umarcor: In fact, there is feedback between MSYS2 and GitHub: https://github.com/actions/virtual-environments/pull/2395#issuecomment-757546669. At some point, the action might be merged/copied in the default environment.

jpmcb (approved):
Looks great! I'll merge this and we'll run both Travis CI and Github Actions side by side for now. Once we're fully confident, we can remove Travis all together!

Thanks for the changes and the thoughtful work you put into this

PR #1340, opened 2021-02-08
Use github actions as test pipeline

jpmcb:
This PR includes a workflow for Github Actions which is comparable to the current Travis CI testing pipeline.

This workflow only uses official Github supported Actions (found in their `actions` org)

You can see this being run on [my fork here](https://github.com/jpmcb/cobra/runs/1856008364) and it only takes 51 seconds to complete. It also uses a cache for the go mod, so we shouldn't have alot of downloads for each run.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -1,0 +1,38 @@
+name: "Build and test"
+on: [push, pull_request]
+
+jobs:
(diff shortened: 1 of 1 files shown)


jpmcb:
Closing - duplicate of #1339

PR #1342, opened 2021-02-10
Fix trailing whitespace in the powershell completion script

Luap99:


marckhouzam:
Just to confirm, this is just esthetic, there is no actual bug?

Luap99:
> Just to confirm, this is just esthetic, there is no actual bug?

Yes I should have mention that.

I want commit the generated files but the podman CI is failing because of the trailing whitespace

Luap99:
@jpmcb This should be an easy one.

marckhouzam (approved):
I did some quick testing with helm and this looks safe.

jpmcb (approved):
Good catch. Removing the trailing whitespace for CI and git compatibility makes sense. 

I currently don't have a great way to test out Windows or powershell things. So thanks for testing this out Marc! I really need to find a way to test these windows changes out. Maybe I'll buy a cheap windows PC or side-load onto my Mac. What do you do to test powershell changes @marckhouzam? 

Anyways - merged! 🚀

Luap99:
@jpmcb You can install powershell on linux. Maybe I can create a simple dockerfile to test it in a container.
https://docs.microsoft.com/en-gb/powershell/scripting/install/installing-powershell-core-on-linux?view=powershell-7.1

marckhouzam:
@jpmcb I've installed PowerShell directly on my Mac.  There is a brew installation for it: https://docs.microsoft.com/en-us/powershell/scripting/install/installing-powershell-core-on-macos?view=powershell-7.1

jpmcb:
Good to know! I figured that powershell was a windows only thing 😅 I'll make sure to give that a try in the future

PR #1352, opened 2021-02-14
Bash completion variable leak fixes

scop:
`split` is from bash_completion's `_init_completion -s`.


jpmcb:
Hi @scop can you give further detail on what you're seeing and what the problem is? Any further detail would be appreciated!!

scop:
Without `local`ising the affected variables in this PR, invoking a completion overwrites their *global* values in the current shell. This either
1) causes problems of varying extents by overwriting contents of those variables the user has set in their shell for whatever reason, or
2) needlessly pollutes the shell with the variables if they weren't already in it

An example of 1), with imaginary `foo` command having cobra's completion:
```shellsessionsame named 
$ split=something
$ foo <kbd>Tab</kbd>
some completions go here
$ ^C
$ echo $split
false
```
Same thing happens to `command_aliases` and `noun_aliases`.

jpmcb (approved):
Good catch! And thanks for the contribution 🚀

PR #1356, opened 2021-02-25
If the file exists, it should be forbidden to regenerate

knight0zh:
When there are multiple commands in the cmd directory, we cannot clearly see which names have been used. 
Because this will overwrite the original code.This is not safe.
So the best way is to prohibit the creation of existing files.

### DIFF
--- a/cobra/cmd/project.go
+++ b/cobra/cmd/project.go
@@ -83,6 +83,11 @@ func (p *Project) createLicenseFile() error {
 }
 
 func (c *Command) Create() error {
+	// check if Command exists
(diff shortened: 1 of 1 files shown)


marckhouzam:
This looks like a reasonable improvement.  Re-opening.

marckhouzam:
@knight0zh Thanks for the contribution!  You make a very valid point.
The Cobra generator has recently moved to its own repository https://github.com/spf13/cobra-cli
Would you be willing to open a PR on that repo instead.  You can tag me on it and hopefully it won't take us long to review.

I'm closing again, as the code has moved.

PR #1357, opened 2021-02-27
Check if command already exists before attempting to create

kodejuice:
These changes makes sure a command doesn't exist before attempting to create it, if it does the user gets a `command already exists` message.

I lost a lot of uncommited work in a project of mine because i mistakenly ran `$ cobra add ..`, this pull request aims to fix that.


kodejuice:
.....

PR #1359, opened 2021-03-02
command: Allow overriding of flag parse function

vkareh:
To allow setting a user-defined flag parser, we add a new
SetFlagParseFunc to Command. This function, when set, will be called
instead of pflag.Parse.

### DIFF
--- a/command.go
+++ b/command.go
@@ -145,6 +145,9 @@ type Command struct {
 	usageFunc func(*Command) error
 	// usageTemplate is usage template defined by user.
 	usageTemplate string
+	// flagParseFunc is func defined by user and it's called when the command is
+	// parsing the flags.
+	flagParseFunc func(*Command, []string) error
 	// flagErrorFunc is func defined by user and it's called when the parsing of
 	// flags returns an error.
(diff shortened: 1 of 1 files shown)


marckhouzam:
Thanks for the contribution @vkareh, sorry it hasn't been reviewed yet.
This seems like an interesting suggestion but I'm not very knowledgeable on the feature of pflags, so I would need to dig deeper into it.  I'll re-open.

@vkareh could you give a concrete example of what you would be able to do with this new support?

vkareh:
@marckhouzam thanks for re-opening this. The way we would use this is to allow commands to accept flags that come from outside of the code.

As a concrete example https://github.com/openshift/rosa/pull/272 would use this to allow the program to accept arbitrary flags that are sent to an external API, but still allow parsing the flags that are part of our program. Today we have to disable flag parsing, override the `PositionalArgs` with our own function, then inside theh `Run` call we need to manually parse the rest of the flags.

Having the ability to override the flag parsing function can get rid of a lot of that by just calling our function directly, then letting cobra handle the rest.

PR #1360, opened 2021-03-02
Changing http:// to https://

paul-nelson-baker:
Various licenses, both in the file comments and in the
strings used for generating new CLI tools, use the http
protocol. Browsers should redirect to https automatically
but why not cut to the heart and directly go to the secure
protocol.

This was performed via 'find' and 'sed':
`find . -type f -iname '*.go' -iname '*.golden'  -exec sed -i '' -e 's/http:\/\//https:\/\//' \{\} \;`

### DIFF
--- a/cobra.go
+++ b/cobra.go
@@ -3,7 +3,7 @@
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
 // You may obtain a copy of the License at
-// http://www.apache.org/licenses/LICENSE-2.0
+// https://www.apache.org/licenses/LICENSE-2.0
 //
 // Unless required by applicable law or agreed to in writing, software
 // distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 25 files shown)


paul-nelson-baker:
Not sure what's not working, it appears to be an environment issue specific to MINGW64. It's not really obvious what the problem is:

```
Run export GOBIN=$HOME/go/bin
golangci/golangci-lint info checking GitHub for tag 'latest'
golangci/golangci-lint info found version: 1.38.0 for v1.38.0/windows/amd64
Archive:  golangci-lint-1.38.0-windows-amd64.zip
  inflating: golangci-lint-1.38.0-windows-amd64/LICENSE  
  inflating: golangci-lint-1.38.0-windows-amd64/README.md  
  inflating: golangci-lint-1.38.0-windows-amd64/golangci-lint.exe  
golangci/golangci-lint info installed /home/runneradmin/go/bin/golangci-lint.exe
no required module provides package github.com/kyoh86/richgo; to add it:
	go get github.com/kyoh86/richgo
Error: Process completed with exit code 1.
```

marckhouzam:
Thanks for the contribution @paul-nelson-baker.
In the long time it took us to get around to looking at your PR, the Cobra generator has moved to https://github.com/spf13/cobra-cli, so this PR would need to be updated.

However, if I look at the official text of the Apache license, it still uses `http`...
https://www.apache.org/licenses/LICENSE-2.0

PR #1363, opened 2021-03-05
powershell completion fix no file comp directive

### LINKED ISSUE #1362: PowerShell completion does not always honour ShellCompDirectiveNoFileComp
This is the same as part of #1248 (fish), as well as #1212 (zsh).

The `powershell` completion script disables file completion only if it receives no completions from the `__complete` command.  Notice the last line of:
https://github.com/spf13/cobra/blob/eb3b6397b1b5d1b0a2cd66a9afe0520f480c0a87/powershell_completions.go#L143-L146

However, it is valid for `ValidArgsFunction` to return completions that don't apply to the current `toComplete` prefix. This means file completion should still be disabled if requested by the directive, but the script does not realize it as it still believes it received valid completions.

Say I have the following `ValidArgsFunction` for command `nofilecomp`:
```
    // We do not consider the toComplete prefix and always return a completion
    return []string{"why"}, cobra.ShellCompDirectiveNoFileComp
},
```
If I have a file called `myfile` and run in powershell:
```
$ ./testprog nofilecomp my<TAB>
```
I would expect not to get file completion, but I do, and `myfile` gets completed. This is because
```
$ ./testprog __complete nofilecomp my<ENTER>
why
:0
```
Notice that above, the script receives a completion even though it is not technically valid. This is allowed. But then, the script does not realize it must disable file completion.

/cc @Luap99

Luap99:
Make sure to filter the returned completions before we check if there are valid completions left.

Fixes #1362

### DIFF
--- a/powershell_completions.go
+++ b/powershell_completions.go
@@ -140,19 +140,6 @@ Register-ArgumentCompleter -CommandName '%[1]s' -ScriptBlock {
         $Space = ""
     }
 
-    if (($Directive -band $ShellCompDirectiveNoFileComp) -ne 0 ) {
-        __%[1]s_debug "ShellCompDirectiveNoFileComp is called"
-
-        if ($Values.Length -eq 0) {
-            # Just print an empty string here so the
-            # shell does not start to complete paths.
-            # We cannot use CompletionResult here because
-            # it does not accept an empty string as argument.
-            ""
-            return
-        }
-    }
-
     if ((($Directive -band $ShellCompDirectiveFilterFileExt) -ne 0 ) -or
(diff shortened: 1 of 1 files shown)


Luap99:
@umarcor The MINGW64 workflow is failing with a missing go module. Can you take a look?

/cc @jpmcb

marckhouzam (approved):
Thanks for the quick fix @Luap99, works great.

umarcor:
@Luap99, see #1366.

jpmcb (approved):
LGTM - thanks for the patience on this one!

PR #1365, opened 2021-03-10
Get MINGW64 action running

bridgetbarnes:
The `gets` need to be before `golangci-lint` is fetched,  so move them earlier to prevent the action from failing because of missing dependancies

### DIFF
--- a/.github/workflows/Test.yml
+++ b/.github/workflows/Test.yml
@@ -76,8 +76,8 @@ jobs:
 
     - run: |
         export GOBIN=$HOME/go/bin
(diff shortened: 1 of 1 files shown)


umarcor:
@bridgetbarnes thanks for having a look at this! Unfortunately, using `get` is not what we want here. `get` might/will modify the `go.mod`, which describes the dependencies of cobra. However, richgo and gox are not dependencies of cobra, but of the testing plumbing. See "*No automatic changes to go.mod and go.sum*" and "*Installing an executable at a specific version*" in https://blog.golang.org/go116-module-changes. See also #1366.

bridgetbarnes:
Oh awesome! Thanks so much for explaining, and for fixing it in your PR! Closing this now

PR #1366, opened 2021-03-12
ci/MSYS2: go install @latest

umarcor:
Close #1365
Ref #1360 #1363

The behaviour of modules, get and install changed in golang 1.16: https://blog.golang.org/go116-module-changes. This PR fixes the problem on MSYS2 when trying to install richgo and gox.

### DIFF
--- a/.github/workflows/Test.yml
+++ b/.github/workflows/Test.yml
@@ -77,7 +77,7 @@ jobs:
     - run: |
         export GOBIN=$HOME/go/bin
         curl -sSfL https://raw.githubusercontent.com/golangci/golangci-lint/master/install.sh | bash -s -- -b $GOBIN latest
-        go install github.com/kyoh86/richgo
-        go install github.com/mitchellh/gox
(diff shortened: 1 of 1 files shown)


umarcor:
@jpmcb mind reviewing and maybe merging this?

georgettica:
got the same issue, glad it is addressed!
I hope this is merged soon 🤞 

good job @umarcor

jpmcb (approved):
LFTM! :rocket: 
would be nice if these had more of a release schedule, but doesn't look like they really make frequent releases or tags, so using the `@latest` seems to be the right thing here

umarcor:
Thanks!
Agree. However, since these are "plumbing deps" and not something the tool (cobra) needs, I believe it's acceptable to be on the bleeding edge, even if we get hurt sometime.
Nonetheless, the relevant modification (the fix) is specifying the version (any).

jpmcb:
I agree, it's pretty low risk! Thanks again for this!

PR #1392, opened 2021-05-06
feat(completion): add option to create default command alone

scop:
The current implementation of the default completion (and help) commands makes them available only when there are some other commands defined. This rules them out for apps that have no other subcommands, either by design anew, or for historical behavior, even though they would benefit from completions both in the first place, and the convenience of having the default generated.

This adds a new completion option for having the default init happen in those cases. Not that clean in the first place and the variable name is a mouthful, but the functionality would be great to have.

Untested, mainly for discussion purposes, but in absence of something better, I'd be fine with something like this.

### DIFF
--- a/command.go
+++ b/command.go
@@ -927,10 +927,10 @@ func (c *Command) ExecuteC() (cmd *Command, err error) {
 		preExecHookFn(c)
 	}
 
-	// initialize help at the last point to allow for user overriding
-	c.InitDefaultHelpCmd()
 	// initialize completion at the last point to allow for user overriding
 	c.initDefaultCompletionCmd()
+	// initialize help at the last point to allow for user overriding
+	c.InitDefaultHelpCmd()
 
 	args := c.args
 
--- a/completions.go
+++ b/completions.go
@@ -95,6 +95,9 @@ type CompletionOptions struct {
 	DisableDescriptions bool
 	// HiddenDefaultCmd makes the default 'completion' command hidden
 	HiddenDefaultCmd bool
+	// InitDefaultEvenIfHasNoSubCommands causes the default 'completion' command to be created
+	// even if there are no other commands defined
+	InitDefaultEvenIfHasNoSubCommands bool
 }
 
 // NoFileCompletions can be used to disable file completion for commands that should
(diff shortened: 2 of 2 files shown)


marckhouzam:
Thanks for the PR @scop.  I now realize that there is value in a `completion` command even for tools that don't have subcommands.  It would allow flags to be completed, as well as allowing for custom completions of arguments and flag values.

However I am not sure of the right approach for the best user experience.  Allowing the creation of the `completion` command changes the help text because sub-commands now exist. There may be other impacts to the tool when creating a first sub-commands, I'm not sure.

And completion itself isn't as smooth as usual because when doing completion of arguments, the `completion` and `help` commands will also be suggested, which are not "arguments", so may cause confusion. 

But I agree that it probably has more value to have the `completion` command than those small disadvantages.  And with your proposal, it is up to the developper to decide to enable the `completion` command, so they can evaluate the impacts.

 I just wanted to continue your discussion and see if others had suggestions about this.

marckhouzam:
Still valid

scop:
Sorry about the delay in following up. I agree with the concerns flagged by @marckhouzam, and made an alternative implementation in #1450 that doesn't have those problems. It's not that nice that the completion command is completely hidden, but better than the alternatives I believe.

scop:
A middle ground could be worth considering as well: let the default completion generate the help section in the usage output, but avoid them being offered as completions. Guess that would be somewhat more code though.

fabstu:
We could also introduce a command like `ForceCreationDespiteLoneCommand` along the lines of `DisableDefaultCmd`. For me it was counter-intuitive that the completion command wasn't available and that not being configurable using a completion option.

marckhouzam:
I believe this has been replaced by #1450

scop:
As mentioned in a couple of comments here, #1450 is an alternative implementation. It will replace this one if the approach in it is chosen, and vice versa, depends on the decision which one is replaced by which.

scop:
Just to drop an explicit note here, #1559 is s superseding implementation of both this and #1450.

marckhouzam:
Replaced by #1559

Issue #1176, opened 2020-07-21
Update OnInitialize usage in readme


kevingentile:
Initializing a viper config similar to the documentation in the readme will not make config values available for use inside init.

## Fails to get "experimental" true from config
```
// rootCmd.go
func init() {
    cobra.OnInitialize(initConfig)
    if viper.GetBool("experimental")
      addExperimentalCommands()
    }
}

func initConfig(){
    viper.SetDefault("experimental", false)

    // read from config file if exists...
    ...
}

func addExperimentalCommands() {
   rootCmd.AddCommand(...)
}
```

## Succeeds getting "experimental" true from config
```
// rootCmd.go
func init() {
   // CHANGE HERE
    initConfig()
    if viper.GetBool("experimental")
      addExperimentalCommands()
    }
}

func initConfig(){
    viper.SetDefault("experimental", false)

    // read from config file if exists...
    ...
}

func addExperimentalCommands() {
   rootCmd.AddCommand(...)
}
```

jpmcb:
Hi @kevingentile can you provide a little more context to how you are running and configuring your rootCmd? 

When adding an initilizer via the `cobra.OnInitialize` method, each of the the associated functions should get invoked during `PreRun` on command execution. Looks like what you are doing in your second example is subverting this behavior by calling `initConfig` manually. I'm not sure this is desirable and I don't think we would want to recommend this in the readme. I wonder if there's a problem with how you've configured the command that isn't allowing for the initializers to get run. 

Any additional info would be great!

kevingentile:
@jpmcb I've created an [example repo](https://github.com/kevingentile/cobra-1176) with my error.

Following the example pattern of adding commands in `init`, I incorrectly assumed that `initConfig` would be executed and the viper config would be available within `init`. Thus, from the above, `viper.GetBool("experimental")` should be moved to allow the viper config to be loaded in `initConfig`:

```
// rootCmd.go
func init() {
    cobra.OnInitialize(initConfig, maybeAddExperimentalCommands)
}

func initConfig(){
    viper.SetDefault("experimental", false)

    // read from config file if exists...
    ...

      addExperimentalCommands()
    }
}

func maybeAddExperimentalCommands() {
    if viper.GetBool("experimental")
        rootCmd.AddCommand(...)
    }
}
```

jpmcb:
I'm not sure this is the right way to go about this.

[...]

PR #1411, opened 2021-06-03
[#1176] include example of config toggled command

kevingentile:
* Root initialize depends on initConfig for proper config evaluation.
* Calling initConfig in OnInitialize yields invalid help text for experimental command.. Addresses #1176

@jpmcb

### DIFF
--- a/README.md
+++ b/README.md
@@ -220,8 +220,7 @@ func Execute() error {
 }
 
 func init() {
-	cobra.OnInitialize(initConfig)
-
+	initConfig()
 	rootCmd.PersistentFlags().StringVar(&cfgFile, "config", "", "config file (default is $HOME/.cobra.yaml)")
 	rootCmd.PersistentFlags().StringP("author", "a", "YOUR NAME", "author name for copyright attribution")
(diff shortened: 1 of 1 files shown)


marckhouzam:
Re-opening until someone has a chance to properly look at this.

johnSchnake:
Looking at this and the linked issue I'm not certain this is a necessary change and it may confuse others.

The point of the `OnInitialize` is to `set the passed functions to be run when each command's Execute method is called.`. Its not a replacement for vanilla-go init() methods or anything like that.

This PR does show how you could run a method early in the invocation of the binary to add commands based on a config, but this section of the README isn't about that narrow use case.

In that use case, I'd even suggest an alternate approach: always show the experimental commands, show them as experimental, and if `experimental` is not set when they're invoked, error so that they are gated. Otherwise you lose any discoverability. If it is just for dev workflows, then it seems fine to just have it be a random workaround and doesn't require updating the main README.

Closing since you've gotten the behavior you want; it just doesn't utilize the `OnInitialize` method. I don't particularly see cobra lacking here.

As always, feel free to reopen if you feel this requires additional conversation.

PR #1417, opened 2021-06-10
Add install instructions for zsh on Mac OS

kylelemons:
zsh is now the default on Mac OS, but the $_fpath version of the installation instructions are likely to put the completion in a strange location that the user might not expect (e.g. an oh-my-zsh plugin's function directory).  So, since Mac OS seems to (as far as I can tell) provide a stable location, this PR recommends using that path instead.

### DIFF
--- a/completions.go
+++ b/completions.go
@@ -608,7 +608,10 @@ to enable it.  You can execute the following once:
 $ echo "autoload -U compinit; compinit" >> ~/.zshrc
 
 To load completions for every new session, execute once:
+# Linux:
 $ %[1]s completion zsh > "${fpath[1]}/_%[1]s"
+# macOS:
+$ %[1]s completion zsh > /usr/local/share/zsh/site-functions/_%[1]s
 
 You will need to start a new shell for this setup to take effect.
 `, c.Root().Name()),


marckhouzam:
Thanks @kylelemons.  Thanks for trying to make things clearer for the user, that is super important.

First you will need to sign Cobra's CLA before any PR can be accepted.

I agree that adding the completion script in `${fpath[1]}` is quite poor; in my case it is the directory of some random oh-my-zsh plugin.

As for the path `/usr/local/share/zsh/site-functions/`, I saw that homebrew installs completions there, so I think it is a good choice.

Could you make the same change in the documentation file `shell_completions.md`?

Finally, can we find a better path on Linux?

kylelemons:
> First you will need to sign Cobra's CLA before any PR can be accepted.

Sorry for the delay, had to wait on legal for the CLA :).

> Could you make the same change in the documentation file shell_completions.md?

Sure.

> Finally, can we find a better path on Linux?

Unfortunately linux seems a bit more fragmented, but one high-SEO [stack overflow post](https://unix.stackexchange.com/questions/607809/where-should-i-install-the-zsh-completion-file-as-a-package-maintainer) actually recommends the same directory, so maybe that?

marckhouzam (approved):
> > Finally, can we find a better path on Linux?
> 
> Unfortunately linux seems a bit more fragmented, but one high-SEO [stack overflow post](https://unix.stackexchange.com/questions/607809/where-should-i-install-the-zsh-completion-file-as-a-package-maintainer) actually recommends the same directory, so maybe that?

I see you are right that things are fragmented.
I ran a docker zsh image and saw that `/usr/share/zsh/site-functions` existed but not `/usr/local/share/zsh/site-functions`.
Same thing after installing zsh on alpine in docker.
But it is the reverse when installing zsh on ubuntu in docker!

So I don't have a good suggestion for a valid path to use.

Let's at least improve things for MacOS with the the change you suggest.

jpmcb (approved):

Issue #1388, opened 2021-05-03
May 2021 release


jpmcb:
Hi all - it's coming time for another release! 

I will plan to cut a new 1.2.0 release sometime in the next week. We've merged a number of PRs already, but let's use this issue as a release tracker to get any necessary CVEs, critical issues, etc into the release!

jpmcb:
*PRs that have been merged already for this release*

New features:
- Now can pass context to execute: https://github.com/spf13/cobra/pull/1265

Bug fixes:
- Handle multiple short hand flags: https://github.com/spf13/cobra/pull/1258
- Fix zsh completions handling nospace and file completions: https://github.com/spf13/cobra/pull/1213
- Fish shell completions fixes / improvements: https://github.com/spf13/cobra/pull/1249
- Fish shell completions new line: https://github.com/spf13/cobra/pull/1284
- Powershell completions fix for comp directive: https://github.com/spf13/cobra/pull/1363

Maintainer Maintenance:
- Fixing MSYS2 in CI: https://github.com/spf13/cobra/pull/1366

marckhouzam:
There is another fix for completion that I have forgot to mention which properly handle the `--` delimiter and the flag `Interspersed` option, it is #1308.

marckhouzam:
If you want this issue to track the content of the 1.2 release, I guess the following should be added, which were merged earlier but after the 1.1.3 release:

New features:
- Create 'completion' command automatically: #1192 

Bug fixes:
- Fix trailing whitespace in the powershell completion script: #1342
- Bash completion variable leak fixes: #1352

Maintainer Maintenance:
- Add PR labeler with pull_request_target: #1338

Luap99:
What about #1146? I would love to use this. This is a great PR and fully backwards compatible.

marckhouzam:
> What about #1146? I would love to use this. This is a great PR and fully backwards compatible.

There is also value in merging the bash v2 for **this release** and not later: it would allow to use bash v2 by default for the automatic `completion` command without worrying about backwards-compatibility.  Please see https://github.com/spf13/cobra/pull/1146#issuecomment-792071457 for an explanation.

I will add tests to [cobra-completion-testing](https://github.com/marckhouzam/cobra-completion-testing) to verify bash v2.

marckhouzam:
[...]

PR #1423, opened 2021-06-24
fix RegisterFlagCompletionFunc concurrent map writes error

### LINKED ISSUE #1320: fatal error: concurrent map writes with RegisterFlagCompletionFunc
Cobra has a global map [here](https://github.com/spf13/cobra/blob/master/custom_completions.go#L21) to store completion function when those exists.

This global map is not protected in any way, which leads to `fatal error: concurrent map writes` when handling **multiple root command concurrently**. Yes, it's a little odd situation but it's a useful one to generate documentation/completion faster (ie, concurrently). Most of those documentation/completion generator will actually mutate things here and there in the commands, which makes them not thread-safe. Now, this global map for completion functions also make things no, thread-safe, even if you duplicated entirely your root command.

I see the following possible solutions:
- move this completion function (no map anymore) directly into `flag.Flag`. No more global, no more indirection, but that's another package.
- move the map into `cobra.Command`, to at least make be thread-safe when you have a copy of your `Command`
- add a mutex or use `sync.Map`.

Example code to trigger the issue:

```go
package main

import (
	"fmt"
	"sync"

	"github.com/spf13/cobra"
)

func main() {
	cmd := cobra.Command{}

	N := 2
	options := make([]string, N)

	var wg sync.WaitGroup
	for i := 0; i < N; i++ {
		flag := fmt.Sprintf("flag%d", i)
		cmd.Flags().StringVarP(&options[i], flag, "", "value", "usage")
		wg.Add(1)
		go func() {
			defer wg.Done()
			err := cmd.RegisterFlagCompletionFunc(flag, nil)
			if err != nil {
				panic(err)
			}
		}()
	}

	wg.Wait()

}
```

_Originally posted by @krobelus in https://github.com/spf13/cobra/issues/1290#issuecomment-767341378_

silenceshell:
Some CI system(like istio) will call package of istioctl in parallel, if cobra use `flagCompletionFunctions` as a global var, will generate `fatal error: concurrent map writes`.
So it is better to store those functions to `flagCompletionFunctions` as a map of root command(thread-safe), instead of a global var.

ref #1320 (this PR will not fix it, as we only get thread-safe, but no lock for multi threads)

### DIFF
--- a/command.go
+++ b/command.go
@@ -88,6 +88,9 @@ type Command struct {
 	// group commands.
 	Annotations map[string]string
 
+	//flagCompletionFunctions is map of flag completion functions.
+	flagCompletionFunctions map[*flag.Flag]func(cmd *Command, args []string, toComplete string) ([]string, ShellCompDirective)
+
 	// Version defines the version for this command. If this value is non-empty and the command does not
 	// define a "version" flag, a "version" boolean flag will be added to the command and, if specified,
 	// will print content of the "Version" variable. A shorthand "v" flag will also be added if the
--- a/command.go
+++ b/command.go
@@ -88,9 +88,6 @@ type Command struct {
 	// group commands.
 	Annotations map[string]string
 
-	//flagCompletionFunctions is map of flag completion functions.
-	flagCompletionFunctions map[*flag.Flag]func(cmd *Command, args []string, toComplete string) ([]string, ShellCompDirective)
-
 	// Version defines the version for this command. If this value is non-empty and the command does not
 	// define a "version" flag, a "version" boolean flag will be added to the command and, if specified,
 	// will print content of the "Version" variable. A shorthand "v" flag will also be added if the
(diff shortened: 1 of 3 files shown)


silenceshell:
@jpmcb would you like to review this PR? Thanks a lot!

command.go
```diff
@@ -88,6 +88,9 @@ type Command struct {
 	// group commands.
 	Annotations map[string]string
 
+	//flagCompletionFunctions is map of flag completion functions.
+	flagCompletionFunctions map[*flag.Flag]func(cmd *Command, args []string, toComplete string) ([]string, ShellCompDirective)
```
marckhouzam, line 92: Could you move this a little lower to be with other non-public fields?

silenceshell: Thanks for your response! 
Nice advice, I have moved `flagCompletionFunctions` after `globNormFunc`.

marckhouzam (approved):
Thanks @silenceshell, I think this is simple and safe improvement regarding the global map.

/cc @jpmcb

silenceshell:
@marckhouzam Thanks for your review.
Will this PR be merged to the next release? #1388
I am working on adding completion support for `istioctl`, and look forward for the next release to support flag completion for `istioctl`, this will make the command more handy.

marckhouzam:
> @marckhouzam Thanks for your review.
> Will this PR be merged to the next release? 

I hope so, but it's up to the maintainers.

> I am working on adding completion support for `istioctl`, ...

Cool!  I've just started using `istioctl` and noticed it was lacking full completion.  Thanks for your work on that.

jpmcb (approved):
This is great! And thanks for bringing this back to the community. LGTM!

PR #1426, opened 2021-06-29
args_test: add helper functions 

umarcor:
This PR adds `expectSuccess`, `getCommand` and other helper functions to `args_test`, in order to reduce the verbosity of the source.

I recommend reviewing the commits one by one.

### DIFF
--- a/args_test.go
+++ b/args_test.go
@@ -5,16 +5,19 @@ import (
 	"testing"
 )
 
-func TestNoArgs(t *testing.T) {
-	c := &Command{Use: "c", Args: NoArgs, Run: emptyRun}
-
-	output, err := executeCommand(c)
+func expectSuccess(output string, err error, t *testing.T) {
 	if output != "" {
-		t.Errorf("Unexpected string: %v", output)
+		t.Errorf("Unexpected output: %v", output)
 	}
 	if err != nil {
 		t.Fatalf("Unexpected error: %v", err)
 	}
+}
+
+func TestNoArgs(t *testing.T) {
+	c := &Command{Use: "c", Args: NoArgs, Run: emptyRun}
+	output, err := executeCommand(c)
+	expectSuccess(output, err, t)
 }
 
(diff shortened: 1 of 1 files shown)


args_test.go
```diff
@@ -5,6 +5,18 @@ import (
 	"testing"
 )
 
+func getCommand(args PositionalArgs, withValid bool) *Command {
```
jpmcb, line 8: Hmmm not sure if I'm a huge fan of using helper methods to abstract away creating a command struct. What benefit do we get adding this complexity when using literal structs inline gets us the same thing?

jpmcb (commented):
Thanks so much for the work that went into this!

I'm not sure it makes sense to abstract so much away from the tests themselves into helper functions. Is there a real need to do this? For me personally, it only makes the tests harder to read since now I have to jump to a different function to understand what it does

args_test.go
umarcor: There are two main reasons: verbosity and human abstraction.

- `get_command` is used 16 times, and it contains 9 lines. That's 144 lines of duplicated code. It's 50% of the file.
- Using this abstraction makes it easier to understand that two orthogonal features are tested: different types of argument constraints and usage of ValidArgs. The orthogonality of those features is fixed in #841. In fact, this PR is a subset of that one.

As you can see in https://github.com/spf13/cobra/pull/841/files#diff-a48034865676903aed3da7be846f56efab79615c941b4b0089d0eb17584d391a, the complexity of this file would explode in #841 if we didn't have this (and other) helper function.

umarcor: See also these two dialogues with @jharshman where he did some suggestions about renaming tests and using sub-tests, which I applied:

- https://github.com/spf13/cobra/pull/841#discussion_r267967922
- https://github.com/spf13/cobra/pull/841#discussion_r267881847

jpmcb: Fair enough! I say let's get it in and if we need to refactor, we can go from there

jpmcb (approved):

PR #1430, opened 2021-06-29
refactor: move cobra/cmd/* to cobra/

umarcor:
Currently, subdir `cobra` contains two sources only (`main.go` and `Makefile`). Most of the actual sources fo the CLI are located in `cobra/cmd`. IMHO, that can be misleading because `cobra` is not meant to be extended in-place. Having most of the sources as package `cmd` implies that `main.go` might use other packages, which is not the case. I think this is a remainder of the initial layout of the repo, when cobra CLI was the only available example about how to use cobra (the library). Nowadays, there are plenty of examples. Nonethless, that belongs to the documentation.

This PR moves `cobra/cmd/*` to `cobra/*` and merges `root.go` into `main.go`.


jpmcb:
Thanks for the thoughtful PR!

This is being superseded by: https://github.com/spf13/cobra/pull/1604 as `cobra/` is [being moved to `spf13/cobra-cli`](https://github.com/spf13/cobra-cli)

For further reference: https://github.com/spf13/cobra/issues/1597

umarcor:
See spf13/cobra-cli#8.

Issue #1240, opened 2020-10-01
Reorganise the cobra library and the CLI as siblings


umarcor:
As discussed in #1233, having the library and the CLI reorganised as siblings in this repo would allow having separated `go.mod` files, while avoiding the ambiguity. However, that's a very disruptive breaking change, as the import path would probably change for all users. Hence, this issue is just a placeholder for now.

Ref #1215 #1219

jpmcb:
I believe the problem is deeper than simply having people upgrade import paths.

Since this repo was an early adopter of go mod, the `go.mod` file itself has fallen a bit behind the current go mod V2 standard. Specifically:

```
The recommended strategy is to develop v2+ modules in a directory named after the major version suffix.

github.com/googleapis/gax-go @ master branch
/go.mod    → module github.com/googleapis/gax-go
/v2/go.mod → module github.com/googleapis/gax-go/v2

This approach is compatible with tools that aren't aware of modules: file paths within the repository match the paths expected by go get in GOPATH mode. This strategy also allows all major versions to be developed together in different directories.
```
[Reference](https://blog.golang.org/v2-go-modules)

Unless we completely overhaul the structure of the repo (which I think could be warranted given how far the go mod has diverted), I think we will continue to see issues related to go paths and this project.

umarcor:
@jpmcb, IMHO, the concern you raised is legit, but unrelated to this issue. ATM, the cobra CLI tool is a child of the cobra library. That is, subdir `cobra` is not explicitly ignored when users import the library. That is unfortunate, because the library does not depend on that subdir at all; but it comes with a non-negligible list of extra dependencies. My understanding is that the most idiomatic solution is to make the library be the child (that is, invert the locations), or to maintain both subprojects as siblings (which I think is desirable).

Whether the `go.mod` file(s) are updated to the standard, that's a related but independent question. As such, it might be addressed either before or after resolving this issue.

umarcor:
Still relevant.

jxsl13:
bump.

jpmcb:
We'll plan on doing this in the next release. But we should think deeply about _how_ we want to go about this to prevent another critical regression the last time this was attempted: https://github.com/spf13/cobra/pull/1233

umarcor:
From https://github.com/spf13/cobra/pull/1233#issuecomment-698890248:

> 1. Create spf13/cobra-cli, and move the CLI there.
[...]

PR #1436, opened 2021-07-01
Module aware generator

spf13:
The `cobra` tool needed some extra updating to make it work seamlessly within the world of Go modules. 

I've dramatically simplified it to just work within a Go module. Now within a module you can just `cobra init` and it works. No paths or module names required. 

I've also made two additional behavioral changes:
1. Viper is no longer included by default. I've heard too much feedback from users that are confused by this and don't know where Cobra ends and Viper begins. Let's simplify that experience. 
2. There is no default license. There's questionable legal ramifications with choosing a license for users without their election. Better to have them take an action to set the license.

### DIFF
--- a/cobra/README.md
+++ b/cobra/README.md
@@ -3,41 +3,78 @@
 Cobra provides its own program that will create your application and add any
 commands you want. It's the easiest way to incorporate Cobra into your application.
 
-In order to use the cobra command, compile it using the following command:
+Install the cobra generator with the command `go install github.com/spf13/cobra/cobra`. 
+Go will automatically install it in your `$GOPATH/bin` directory which should be in your $PATH. 
 
-    go get github.com/spf13/cobra/cobra
+Once installed you should have the `cobra` command available. Confirm by typing `cobra` at a 
+command line. 
 
-This will create the cobra executable under your `$GOPATH/bin` directory.
+There are only two operations currently supported by Cobra generator: 
 
 ### cobra init
 
 The `cobra init [app]` command will create your initial application code
 for you. It is a very powerful application that will populate your program with
-the right structure so you can immediately enjoy all the benefits of Cobra. It
-will also automatically apply the license you specify to your application.
+the right structure so you can immediately enjoy all the benefits of Cobra. 
+It can also apply the license you specify to your application.
 
-Cobra init is pretty smart. You can either run it in your current application directory
-or you can specify a relative path to an existing project. If the directory does not exist, it will be created for you.
+With the introduction of Go modules, the Cobra generator has been simplified to
+take advantage of modules. The Cobra generator works from within a Go module. 
 
-Updates to the Cobra generator have now decoupled it from the GOPATH.
-As such `--pkg-name` is required.
+#### Initalizing a module
 
-**Note:** init will no longer fail on non-empty directories.
+__If you already have a module, skip this step.__
 
+If you want to initialize a new Go module: 
+
+ 1. Create a new directory 
+ 2. `cd` into that directory
+ 3. run `go mod init <MODNAME>`
+
+e.g. 
 ```
-mkdir -p newApp && cd newApp
-cobra init --pkg-name github.com/spf13/newApp
+cd $HOME/code 
+mkdir myapp
+cd myapp
+go mod init github.com/spf13/myapp
 ```
 
-or
+#### Initalizing an Cobra CLI application
 
+From within a Go module run `cobra init`. This will create a new barebones project
+for you to edit. 
+
+You should be able to run you new application immediately. Try it with 
+`go run main.go`. 
+
+You will want to open up and edit 'cmd/root.go' and provide your own description and logic. 
+
+e.g.
 ```
-cobra init --pkg-name github.com/spf13/newApp path/to/newApp
+cd $HOME/code/myapp
+cobra init
+go run main.go
 ```
 
-### cobra add
+Cobra init can also be run from a subdirectory such as how the [cobra generator itself is organized](https://github.com/spf13/cobra).
+This is useful if you want to keep your application code separate from your library code.
 
-Once an application is initialized, Cobra can create additional commands for you.
+#### Optional flags:
+You can provide it your author name with the `--author` flag. 
+e.g. `cobra init --author "Steve Francia spf@spf13.com"`
+
+You can provide a license to use with `--license` 
+e.g. `cobra init --license apache`
+
+Use the `--viper` flag to automatically setup [viper](https://github.com/spf13/viper)
+
+Viper is a companion to Cobra intended to provide easy handling of environment variables and config files and seamlessly connecting them to the application flags.
+
+### Add commands to a project
+
+Once a cobra application is initialized you can continue to use cobra generator to 
+add additional commands to your application. The command to do this is `cobra add`. 
+
 Let's say you created an app and you wanted the following commands for it:
 
 * app serve
@@ -52,6 +89,14 @@ cobra add config
 cobra add create -p 'configCmd'
 ```
 
+`cobra add` supports all the same optional flags as `cobra init` does mentioned above.
+
+You'll notice that this final command has a `-p` flag. This is used to assign a
+parent command to the newly added command. In this case, we want to assign the
+"create" command to the "config" command. All commands have a default parent of rootCmd if not specified.  
+
+By default `cobra` will append `Cmd` to the name provided and uses this to name for the internal variable name. When specifying a parent, be sure to match the variable name used in the code. 
+
 *Note: Use camelCase (not snake_case/kebab-case) for command names.
 Otherwise, you will encounter errors.
 For example, `cobra add add-user` is incorrect, but `cobra add addUser` is valid.*
@@ -62,9 +107,10 @@ the following:
 ```
   ▾ app/
     ▾ cmd/
-        serve.go
         config.go
         create.go
+        serve.go
+        root.go
       main.go
 ```
 
@@ -72,8 +118,11 @@ At this point you can run `go run main.go` and it would run your app. `go run
 main.go serve`, `go run main.go config`, `go run main.go config create` along
 with `go run main.go help serve`, etc. would all work.
 
-Obviously you haven't added your own code to these yet. The commands are ready
-for you to give them their tasks. Have fun!
+You now have a basic Cobra-based application up and running. Next step is to edit the files in cmd and customize them for your application.
+
+For complete details on using the Cobra library, please read the [The Cobra User Guide](https://github.com/spf13/cobra/blob/master/user_guide.md#using-the-cobra-library).
+
+Have fun!
 
 ### Configuring the cobra generator
 
@@ -86,6 +135,7 @@ An example ~/.cobra.yaml file:
 ```yaml
 author: Steve Francia <spf@spf13.com>
 license: MIT
+viper: true
 ```
 
 You can also use built-in licenses. For example, **GPLv2**, **GPLv3**, **LGPL**,
--- a/cobra/README.md
+++ b/cobra/README.md
@@ -9,7 +9,7 @@ Go will automatically install it in your `$GOPATH/bin` directory which should be
 Once installed you should have the `cobra` command available. Confirm by typing `cobra` at a 
 command line. 
 
-There are only two operations currently supported by Cobra generator: 
+There are only two operations currently supported by the Cobra generator: 
 
 ### cobra init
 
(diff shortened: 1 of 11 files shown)


cobra/cmd/init.go
```diff
@@ -77,3 +78,52 @@ func initializeProject(args []string) (string, error) {
 
 	return project.AbsolutePath, nil
 }
+
+func getModImportPath() string {
+	mod, cd := parseModInfo()
+	return path.Join(mod.Path, fileToURL(strings.TrimPrefix(cd.Dir, mod.Dir)))
+}
+
+func fileToURL(in string) string {
```
umarcor, line 87: Is it worth having this function which is called once only? Maybe it can be written inside getModImportPath?

cobra/cmd/init.go
```diff
@@ -77,3 +78,52 @@ func initializeProject(args []string) (string, error) {
...
+type Mod struct {
+	Path, Dir, GoMod string
+}
+
+type CurDir struct {
+	Dir string
+}
+
+func goGet(mod string) error {
+	cmd := exec.Command("go", "get", mod)
```
umarcor, line 119: Might be a oneliner?

```golang
return exec.Command("go", "get", mod).Run()
```

user_guide.md
```diff
@@ -32,7 +32,67 @@ func main() {
 Cobra provides its own program that will create your application and add any
 commands you want. It's the easiest way to incorporate Cobra into your application.
 
-[Here](https://github.com/spf13/cobra/blob/master/cobra/README.md) you can find more information about it.
+Install the cobra generator with the command `go install github.com/spf13/cobra/cobra`.
```
umarcor, line 35: The documentation of the CLI itself is located in `cobra/README.md`. Hence, this feels duplicated and now not matching. Was your purpose to actually remove `cobra/README.md`? Maybe this content should be written there?

cobra/cmd/init.go
```diff
@@ -14,42 +14,41 @@
...
+			if viper.GetBool("useViper") {
+				cobra.CheckErr(goGet("github.com/spf13/viper"))
+			}
 			fmt.Printf("Your Cobra application is ready at\n%s\n", projectPath)
 		},
 	}
 )
 
-func init() {
-	initCmd.Flags().StringVar(&pkgName, "pkg-name", "", "fully qualified pkg name")
```
jharshman, line 49: Because `pkg-name` has been an available flag for a few releases now, i think we should deprecate it instead of removing it entirely.

cobra/cmd/init.go
```diff
@@ -77,3 +78,52 @@ func initializeProject(args []string) (string, error) {
...
+		cobra.CheckErr("Please run `go mod init <MODNAME>` before `cobra init`")
+	}
+
+	e := modInfoJSON("-e")
+	cobra.CheckErr(json.Unmarshal(e, &dir))
+
+	return mod, dir
+}
+
+type Mod struct {
```
jharshman, line 110: nit: can you put each of these variables on their own line?

cobra/cmd/init.go
```diff
@@ -14,42 +14,41 @@
...
-  * If a name is provided, a directory with that name will be created in the current directory;
-  * If no name is provided, the current directory will be assumed;
+Cobra init must be run inside of a go module (please run "go mod init <MODNAME>" first)
 `,
 
 		Run: func(_ *cobra.Command, args []string) {
-
 			projectPath, err := initializeProject(args)
 			cobra.CheckErr(err)
+			cobra.CheckErr(goGet("github.com/spf13/cobra"))
```
jharshman, line 43: I'm not sure how I feel about this utility fetching dependencies for the user. @wfernandes @jpmcb thoughts here?

jharshman (commented):
@spf13 since the documentation is changing here could you also open a PR on https://github.com/spf13/cobra.dev if you haven't made the changes there already?

cobra/cmd/init.go
umarcor: I guess the point is having them explicitly added to the `go.mod` file of the user? Not so much about fetching them.

cobra/cmd/init.go
spf13: The problem is that it doesn't really make sense anymore with modules and can actually create a pretty broken experience for users who follow old documentation without knowing what it is doing (especially if the pkgname doesn't align with the module). Since this is a tool and not a dependency anyone depends on I'd rather let them know it's changed as early as possible and let them view the updated help.

cobra/cmd/init.go
spf13: Initially I thought I'd use it more than once, though I don't see an issue with leaving it as a separate function. To me it's easier to read this way and performs about the same (not that performance is a big factor here).

cobra/cmd/init.go
spf13: Yup, that would be cleaner.

user_guide.md
spf13: Oh, I totally missed that this was where it was. I was surprised their wasn't more documentation. I only noticed the /README.md and this user_guide.md file. This is especially embarrassing because I'm pretty sure I wrote it originally. I'll merge the updated docs into this one and add links to make this more clear again.

cobra/cmd/init.go
spf13: I can. What is the motivation for doing so? (mostly just curious).

cobra/cmd/init.go
spf13: Yes, that is the intent. Russ Cox suggested we do this extra step and I agree. It streamlines the experience for the user as well as simplifying the instructions.

spf13:
> @spf13 since the documentation is changing here could you also open a PR on https://github.com/spf13/cobra.dev if you haven't made the changes there already?

I'll do that as soon as this is accepted. For that, do you want a PR or for me to just make the changes directly? Happy to do either.

user_guide.md
spf13: Addressed in b4ab6634be8a507c3047efdaa89a789d3b6504c0

cobra/cmd/init.go
spf13: Addressed in 31a1f21b87f1f09e7f1d53a7f757f1b113d31034

cobra/cmd/init.go
spf13: One of the motivations for this update is that we are writing a tutorial on building CLIs with Go for [golang.org](https://golang.org) / [go.dev](https://go.dev) which uses Cobra. As we wrote the tutorial it became clear that there were lots of places in the onboarding experience that could be streamlined. This set of changes comes from those learnings.

cobra/cmd/init.go
spf13: In other words, I can't think of a single reason (but there could be one I'm not thinking of) that leaving this option does any good. We can already get the right name from the module. The module needs to be created independently anyway. If we require a module to be initialized first then we can guarantee to use the right name here and keep them in sync without asking the user to do an extra step. 

I debated keeping this and then doing the `go mod init` step if no module is present and concluded that that was a terrible idea and rife with unexpected behaviors and edge cases. 

I think having the tool work simply and consistently is best for everyone.

cobra/cmd/init.go
umarcor: @spf13, in the context of writing a tutorial and streamlining the experience for newcommers, you might want to have a look at #1240. I do think that might be a very disruptive change, but at the same time it would be very valuable to have your feedback.

cobra/README.md
```diff
@@ -3,41 +3,78 @@
 Cobra provides its own program that will create your application and add any
 commands you want. It's the easiest way to incorporate Cobra into your application.
 
-In order to use the cobra command, compile it using the following command:
+Install the cobra generator with the command `go install github.com/spf13/cobra/cobra`.
```
umarcor, line 6: In CI, we are currently using GOBIN (see https://github.com/spf13/cobra/blob/master/.github/workflows/Test.yml). Might be worth adding a reference to https://golang.org/cmd/go/#hdr-Compile_and_install_packages_and_dependencies here?

spf13: good idea.

cobra/cmd/init.go
spf13: See my note on #1240. I think we're likely to resolve this in the next few months by removing the need for cobra generator entirely.

cobra/cmd/init.go
jharshman: Ah I see. I'm okay with this approach :+1:

cobra/cmd/init.go
jharshman: > As we wrote the tutorial it became clear that there were lots of places in the onboarding experience that could be streamlined.

I'm inclined to agree on this point but am always wary of tools that try to do too many things.

jharshman: We can move forward with this change though.

cobra/cmd/init.go
jharshman: For me it's readability. I find it easier to read the structure fields when they are not condensed on one line.

jharshman (approved):
@spf13 are you ready for this to be merged?

spf13:
I still need to write up the PR for the doc site. I've been swamped with
other stuff. I think that's all this is waiting on. It will be a couple
weeks before I'll be able to get to it. I'm fine if anyone else wants to do
that and merge this, but realistically speaking I'll be a couple weeks
before I can.

cobra/cmd/init.go
jpmcb: If this is just in the cobra CLI tool, sounds good to me. I'd be more cautious of anything in the API that would automatically fetch dependencies. But it makes sense to try and streamline the onboarding experience and get people up and going with `cobra init` as quick as possible :+1:

jpmcb (approved):
Sorry for being late to this one - this looks good to me and I think will be a welcome change since most everyone should be on go modules by now! :rocket:

spf13:
I'll try to get to this in the next week or two. Bumping so it's not stale anymore.

cobra/README.md
```diff
@@ -52,6 +89,14 @@ cobra add config
 cobra add create -p 'configCmd'
 ```
 
+`cobra add` supports all the same optional flags as `cobra init` does mentioned above.
```
marckhouzam, line 92: Typo: "... does mentioned above."

cobra/README.md
```diff
@@ -52,6 +89,14 @@ cobra add config
...
+
+You'll notice that this final command has a `-p` flag. This is used to assign a
+parent command to the newly added command. In this case, we want to assign the
+"create" command to the "config" command. All commands have a default parent of rootCmd if not specified.  
+
+By default `cobra` will append `Cmd` to the name provided and uses this to name for the internal variable name. When specifying a parent, be sure to match the variable name used in the code.
```
marckhouzam, line 98: Typo: "... uses this to name for the internal variable name."

cobra/README.md
```diff
@@ -3,41 +3,78 @@
...
-In order to use the cobra command, compile it using the following command:
+Install the cobra generator with the command `go install github.com/spf13/cobra/cobra`. 
+Go will automatically install it in your `$GOPATH/bin` directory which should be in your $PATH. 
 
-    go get github.com/spf13/cobra/cobra
+Once installed you should have the `cobra` command available. Confirm by typing `cobra` at a 
+command line. 
 
-This will create the cobra executable under your `$GOPATH/bin` directory.
+There are only two operations currently supported by Cobra generator:
```
marckhouzam, line 12: Missing "the" before "Cobra generator"

cobra/README.md
```diff
@@ -3,41 +3,78 @@
...
+go mod init github.com/spf13/myapp
 ```
 
-or
+#### Initalizing an Cobra CLI application
+
+From within a Go module run `cobra init`. This will create a new barebones project
+for you to edit. 
+
+You should be able to run you new application immediately. Try it with
```
marckhouzam, line 47: "you" -> "your"

cobra/README.md
```diff
@@ -3,41 +3,78 @@
...
+You can provide a license to use with `--license` 
+e.g. `cobra init --license apache`
+
+Use the `--viper` flag to automatically setup [viper](https://github.com/spf13/viper)
+
+Viper is a companion to Cobra intended to provide easy handling of environment variables and config files and seamlessly connecting them to the application flags.
+
+### Add commands to a project
+
+Once a cobra application is initialized you can continue to use cobra generator to
```
marckhouzam, line 75: Missing "the" before "Cobra generator"

PR #1438, opened 2021-07-02
Fix flag completion

### LINKED ISSUE #1320: fatal error: concurrent map writes with RegisterFlagCompletionFunc

### LINKED ISSUE #1437: 1.2.0 regression: flag completion issues
PR #1423 changed the location where flag completion functions are stored. It always stores them under the root cmd. This breaks Podman because we setup each cobra cmd struct individually with their flags before we join them to the root command.
So when we call  `cmd.RegisterFlagCompletionFunc()` the root cmd will not be the actual root cmd. When the `__complete` command tries to look up flag completions it will not find anything in the real root cmd and thus no completion is provided.

ref https://github.com/containers/podman/pull/10841

Luap99:
The flag completion functions should not be stored in the root cmd.
There is no requirement that the root cmd should be the same when
`RegisterFlagCompletionFunc` was called. Storing the flags there does
not work when you add the the flags to your cmd struct before you add the
cmd to the parent/root cmd. The flags can no longer be found in the rigth
place when the completion command is called and thus the flag completion
does not work.

Also #1423 claims that this would be thread safe but we still have a map
which will fail when accessed concurrently. To truly fix this issue use a
RWMutex.

Fixes #1437
Fixes #1320

### DIFF
--- a/completions.go
+++ b/completions.go
@@ -4,6 +4,7 @@ import (
 	"fmt"
 	"os"
 	"strings"
+	"sync"
 
 	"github.com/spf13/pflag"
 )
@@ -16,6 +17,12 @@ const (
 	// completion results without their description.  It is used by the shell completion scripts.
 	ShellCompNoDescRequestCmd = "__completeNoDesc"
 )
+
+// Global map of flag completion functions. Make sure to use flagCompletionMutex before you try to read and write from it.
+var flagCompletionFunctions = map[*pflag.Flag]func(cmd *Command, args []string, toComplete string) ([]string, ShellCompDirective){}
+
+// lock for reading and writing from flagCompletionFunctions
+var flagCompletionMutex = &sync.RWMutex{}
 
 // ShellCompDirective is a bit map representing the different behaviors the shell
 // can be instructed to have once completions have been provided.
@@ -100,15 +107,13 @@ func (c *Command) RegisterFlagCompletionFunc(flagName string, f func(cmd *Comman
 	if flag == nil {
 		return fmt.Errorf("RegisterFlagCompletionFunc: flag '%s' does not exist", flagName)
 	}
+	flagCompletionMutex.Lock()
+	defer flagCompletionMutex.Unlock()
 
-	root := c.Root()
-	if _, exists := root.flagCompletionFunctions[flag]; exists {
+	if _, exists := flagCompletionFunctions[flag]; exists {
 		return fmt.Errorf("RegisterFlagCompletionFunc: flag '%s' already registered", flagName)
 	}
-	if root.flagCompletionFunctions == nil {
-		root.flagCompletionFunctions = map[*pflag.Flag]func(cmd *Command, args []string, toComplete string) ([]string, ShellCompDirective){}
-	}
-	root.flagCompletionFunctions[flag] = f
+	flagCompletionFunctions[flag] = f
 	return nil
 }
 
@@ -402,7 +407,9 @@ func (c *Command) getCompletions(args []string) (*Command, []string, ShellCompDi
 	// Find the completion function for the flag or command
 	var completionFn func(cmd *Command, args []string, toComplete string) ([]string, ShellCompDirective)
 	if flag != nil && flagCompletion {
-		completionFn = c.Root().flagCompletionFunctions[flag]
+		flagCompletionMutex.RLock()
+		completionFn = flagCompletionFunctions[flag]
+		flagCompletionMutex.RUnlock()
 	} else {
 		completionFn = finalCmd.ValidArgsFunction
 	}
--- a/completions_test.go
+++ b/completions_test.go
@@ -1971,6 +1971,41 @@ func TestFlagCompletionWithNotInterspersedArgs(t *testing.T) {
 	}
 }
 
+func TestFlagCompletionWorksRootCommandAddedAfterFlags(t *testing.T) {
+	rootCmd := &Command{Use: "root", Run: emptyRun}
+	childCmd := &Command{
+		Use: "child",
+		Run: emptyRun,
+		ValidArgsFunction: func(cmd *Command, args []string, toComplete string) ([]string, ShellCompDirective) {
+			return []string{"--validarg", "test"}, ShellCompDirectiveDefault
+		},
+	}
+	childCmd.Flags().Bool("bool", false, "test bool flag")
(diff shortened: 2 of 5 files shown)


Luap99:
@jpmcb @marckhouzam PTAL

Every program who calls cmd.RegisterFlagCompletionFunc() before joining the cmd to the root command will no longer have functional flag completion. I think it will break many applications since it is common to add flags to the cmd before adding it the root/parent cmd.
It would be great if we can get this fixed soon.

Luap99:
OK this patch doesn't work either. Storing them in the correct cmd doesn't work for persistent flags.

I also do not understand what #1423 is supposed to fix. The map is just stored in the root cmd now, writing concurrently will still fail.

Luap99:
OK, this should work now and truly fix the concurrency problem.

completions.go
```diff
@@ -100,15 +107,16 @@ func (c *Command) RegisterFlagCompletionFunc(flagName string, f func(cmd *Comman
...
 
-	root := c.Root()
-	if _, exists := root.flagCompletionFunctions[flag]; exists {
+	if _, exists := flagCompletionFunctions[flag]; exists {
 		return fmt.Errorf("RegisterFlagCompletionFunc: flag '%s' already registered", flagName)
 	}
-	if root.flagCompletionFunctions == nil {
-		root.flagCompletionFunctions = map[*pflag.Flag]func(cmd *Command, args []string, toComplete string) ([]string, ShellCompDirective){}
+	if flagCompletionFunctions == nil {
+		flagCompletionFunctions = map[*pflag.Flag]func(cmd *Command, args []string, toComplete string) ([]string, ShellCompDirective){}
```
marckhouzam, line 117: You no longer need this nil-check, since the map is already initialized at the start.

completions_test.go
```diff
@@ -1763,13 +1763,15 @@ func TestFlagCompletionWithNotInterspersedArgs(t *testing.T) {
...
 	}
-	rootCmd.AddCommand(childCmd, childCmd2)
 	childCmd.Flags().Bool("bool", false, "test bool flag")
 	childCmd.Flags().String("string", "", "test string flag")
 	_ = childCmd.RegisterFlagCompletionFunc("string", func(cmd *Command, args []string, toComplete string) ([]string, ShellCompDirective) {
 		return []string{"myval"}, ShellCompDirectiveDefault
 	})
 
+	// Important: only add the commands after RegisterFlagCompletionFunc was called
+	rootCmd.AddCommand(childCmd, childCmd2)
```
marckhouzam, line 1773: I realize we are trying to fix this quickly, so sorry for the nit, but I feel this should be a separate test, and it should mention in a comment issue #1437 to explain why the order is important

completions_test.go
```diff
@@ -1969,6 +1971,21 @@ func TestFlagCompletionWithNotInterspersedArgs(t *testing.T) {
 	if output != expected {
 		t.Errorf("expected: %q, got: %q", expected, output)
 	}
+
+	// Test that no flag completion works on a subcmd
```
marckhouzam, line 1975: Is this comment right? It seems the "no" is superfluous.

marckhouzam (commented):
Thanks for the quick reaction @Luap99 !

This look good with a couple of nits.  

However, I haven't tested if it really fixes the concern of istioctl mentioned in https://github.com/spf13/cobra/pull/1423#issue-676809165, but I expect it would.

marckhouzam:
@silenceshell are you able to test if this fix addresses the concern you were aiming to fix for istioctl in https://github.com/spf13/cobra/pull/1423#issue-676809165?

silenceshell:
> @silenceshell are you able to test if this fix addresses the concern you were aiming to fix for istioctl in [#1423 (comment)](https://github.com/spf13/cobra/pull/1423#issue-676809165)?

This PR could fix the istioctl test cases, RWMutex is safe enough.

Luap99:
@marckhouzam Updated.

silenceshell:
> OK this patch doesn't work either. Storing them in the correct cmd doesn't work for persistent flags.
> 
> I also do not understand what #1423 is supposed to fix. The map is just stored in the root cmd now, writing concurrently will still fail.

Sorry for the incomplete consideration. 
Istio has some tests that call rootCmd reg func concurrently, #1423 wants to store the map to rootCmd to avoid concurrent map writes errors. This could fix that problem but breaks the current implement. sorry for this again, and thanks for the correction.

marckhouzam (approved):
This looks good.
I was able to reproduce the regression with Helm, and can confirm this PR fixes things.
I had not seen the regression because most of helm flags that use completion are on the root command.

Thanks @Luap99 

@jpmcb what do you think for a 1.2.1 release?

jpmcb (approved):
Merging to get a quick fix out for this regression. Thanks all for the fast find and quick testing efforts on this!

Luap99:
@jpmcb Thanks for the quick review and the new release tag.

PR #1441, opened 2021-07-04
Fix error messages of unbound variables in completions with `set -u`

akinomyoga:
This fixes the issue reported at https://github.com/scop/bash-completion/issues/551


scop:
See https://github.com/spf13/cobra/pull/1351 and https://github.com/spf13/cobra/pull/1321

akinomyoga:
Ah, OK! Thank you for the information. This is a duplicate of #1321, so I close the issue.

PR #1427, opened 2021-06-29
github: add dependabot configuration file

umarcor:
Close #907

Dependabot was an external service for keeping dependencies up to date. It was bought by GitHub, and it's now a built-in service which integrates with security updates. See https://docs.github.com/en/code-security/supply-chain-security/keeping-your-dependencies-updated-automatically.

In April, GitHub announced that the external service (*Dependabot Preview*) is to be shut down in August: https://github.blog/2021-04-29-goodbye-dependabot-preview-hello-dependabot/.

This PR adds a dependabot configuration file for checking go dependencies weekly.

Note that this features does NOT auto-merge. It will open PRs, but maintainers will need to merge them, or bump the dependencies otherwise.

Refs: #1420 #1406 #1404 #1364

PR #1453, opened 2021-07-11
Set up Dependabot

mmorel-35:


umarcor:
This is a duplicate of #1427.

mmorel-35:
Yes, but this one also handles github-actions and and is daily checked, not weekly

umarcor:
> Yes, but this one also handles github-actions and and is daily checked, not weekly

That is correct. The netiquette in this case is to reply in #1427 in order to discuss the motivation for those differences.

mmorel-35:
Ok, I'm closing this one then

PR #1459, opened 2021-07-18
fix: typo in {bash,zsh}_completions.go

favonia:
`subDir` seems to be a typo. It is deep inside `bash_completions.go` used to generate bash completion scripts. I am not sure how to write a test for this.


marckhouzam:
Thanks @favonia!
Could you please also fix the same problem in `zsh_completions.go`?

I think writing a test would be more complicated then it's value, so I think the fix itself is sufficient.

favonia:
Done!

marckhouzam (approved):
This looks good

stmcginnis (approved):
Looks good to me!

jpmcb (approved):

PR #1468, opened 2021-08-06
Enable Go module on install, otherwise install fails with missing HCL dependency

jdb:


jdb:
Nope, sorry for the noise, that is not the right yet.

PR #1473, opened 2021-08-15
completions: fix mixed tab/spaces indentation

thaJeztah:
These templates use 4 spaces for indentation, but some lines used tabs.

### DIFF
--- a/bash_completions.go
+++ b/bash_completions.go
@@ -232,13 +232,13 @@ __%[1]s_handle_reply()
     fi
 
     if [[ ${#COMPREPLY[@]} -eq 0 ]]; then
-		if declare -F __%[1]s_custom_func >/dev/null; then
-			# try command name qualified custom func
-			__%[1]s_custom_func
-		else
-			# otherwise fall back to unqualified for compatibility
-			declare -F __custom_func >/dev/null && __custom_func
-		fi
+        if declare -F __%[1]s_custom_func >/dev/null; then
+            # try command name qualified custom func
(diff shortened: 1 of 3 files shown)


thaJeztah:
@jpmcb  @Luap99 ptal

Luap99 (approved):
LGTM

Luap99:
@marckhouzam PTAL
I think we should setup some linter for the generated files.

thaJeztah:
> I think we should setup some linter for the generated files.

Once support for Go 1.15 is dropped, these templates could potentially stored as separate files (go template?), and included using `//go:embed`, which would (potentially) make validation/linting easier.

marckhouzam (approved):

stmcginnis (approved):
I wonder if this is going to cause problems with go editors that default to using tabs. Was there an actual problem with these, or is this just cosmetic?

Regardless, I think it's fine to change it - I prefer consistency and makes sense for shell code.

thaJeztah:
> I wonder if this is going to cause problems with go editors that default to using tabs.

I don't think editors should format content inside string-literals, but even in that case, this patch shouldn't change that (as there's already content using the same indentation)

> Was there an actual problem with these, or is this just cosmetic?

No real problem as far as I'm aware, but it looked sloppy to have a generated file that doesn't stick to its own coding style 😅

jpmcb (approved):
Looks good - but let's keep a close eye on this in case people report issues with tabs vs. spaces in their generated

PR #1477, opened 2021-08-26
chore(ci): use golangci-lint-action

mmorel-35:
This PR is to use golangci-lint-action instead of direct download of it's executable.
It allows also to see the tool's reviews outside of the logs in github.
I also formated and fixed the actual reviews.
Some linters were deprecated, so I used there actual versions

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -9,6 +9,20 @@ env:
 
 jobs:
 
+  golangci-lint:
+    runs-on: ubuntu-latest
+    steps:
+
+      - uses: actions/setup-go@v2
+        with:
+          go-version: '1.17'
+
+      - uses: actions/checkout@v2
+
+      - uses: golangci/golangci-lint-action@v2
+        with:
+          version: latest
+          args: --verbose
 
   test-unix:
     strategy:
@@ -40,7 +54,6 @@ jobs:
 
     - run: |
         export GOBIN=$HOME/go/bin
-        curl -sSfL https://raw.githubusercontent.com/golangci/golangci-lint/master/install.sh | bash -s -- -b $GOBIN latest
         case "${{ matrix.go }}" in
           16|17) _version='@latest';;
           *) _version='';;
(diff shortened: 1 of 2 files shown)


.github/workflows/test.yml
```diff
@@ -8,8 +8,21 @@ env:
   GO111MODULE: on
 
 jobs:
-
-
+  golangci:
```
umarcor, line 11: Please, keep the style consistent in the YAML file:

- Blank line before and after the job, and before and after steps.
- No need to add a `name` field to the `uses` steps, which are self-explaining by the action name.

.golangci.yml
```diff
@@ -2,7 +2,6 @@ run:
   deadline: 5m
 
 linters:
-  disable-all: true
   enable:
```
umarcor: Please, split the modifications to `.golangci.yml` to a separated commit (maybe PR), since it is unrelated to how golangci-lint is installed.

bash_completions.go
```diff
@@ -617,6 +617,7 @@ func writeCmdAliases(buf io.StringWriter, cmd *Command) {
 	WriteStringAndCheck(buf, `    fi`)
 	WriteStringAndCheck(buf, "\n")
 }
+
```
umarcor, line 620: It seems that some of these changes to `.go` sources are motivated by changing the golangci-lint configuration, which is unrelated to the topic of this PR. Please, keep these changes in a separated commit and/or PR.

umarcor (approved):
LGTM!

umarcor:
Added to #1496 and #1525.

stmcginnis (approved):
I know there are some linters that don't work well with go 1.17 that get pulled in with golangci-lint prior to the 1.43.0 release. I wonder if we should add some sort of check to the makefile for that.

The action config is pulling in the latest golangci-lint, so that should work fine. And the change mentioned above isn't really in the scope of this PR, just thought I would mention it here in case someone has an old version on their system and runs into a problem because of it.

This change looks good to me.

jpmcb (approved):

PR #1481, opened 2021-09-03
Upgrade gopkg.in/yaml.v2 to gopkg.in/yaml.v3

inteon:
Upgrade the yaml dep to version 3.

### DIFF
--- a/doc/yaml_docs.go
+++ b/doc/yaml_docs.go
@@ -23,7 +23,7 @@ import (
 
 	"github.com/spf13/cobra"
 	"github.com/spf13/pflag"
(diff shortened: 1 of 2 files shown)


jpmcb:
We probably still want to do this. Re-opening and marking myself as reviewer to get this in

Jacalz:
It would be great if this could be merged and if a new release could be crated after that. It seems like versions before 3.0.0 contain a security issue. See https://github.com/advisories/GHSA-hp87-p4gw-j4gq for more information.

Jacalz:
It seems like v3.0.1 has been released as well. Perhaps it should be upgraded to that while at it?

PR #1482, opened 2021-09-04
Add Active Help support

marckhouzam:
This PR is a proposal to enhance the shell completion logic with a framework for what I call "Active Help".

Active Help are messages (hints, warnings, etc) that are printed during the use of the program using the shell completion system.  It aims to help teach how to use the program in an "active" fashion.

For example, 
```
bash-5.1$ helm repo add [tab]
You must choose a name for the repo you are adding

bash-5.1$ bin/helm package [tab]
Please specify the path to the chart to package

# This last one is normal completion but shows that it works with the previous active help trigger
bash-5.1$ bin/helm package [tab][tab]
bin/    internal/    scripts/    pkg/     testdata/
```
Active Help messages are defined by the program by adding them as custom completions using `ValidArgsFunction` and `RegisterFlagCompletionFunc` and a new function: `cobra.AppendActiveHelp(completionArray, activeHelpMessage)`.

Here is a 3 minute video illustrating how the feature can be used:

<a href="https://asciinema.org/a/421868" target="_blank"></a>

The PR contains [documentation](https://github.com/VilledeMontreal/cobra/blob/feat/ActiveHelp/active_help.md) describing how to use the feature.

The PR adds the feature to both `bash` and `zsh`.  It is probably do-able for `powershell` but I haven't tried.  It is not possible to implement for `fish` currently.

I'd be interested to know if this is something of interest to the community.

**Note**:
For bash, version >= 4.4 of bash provides a smoother experience.  In older versions of bash, when printing Active Help, the prompt will not be shown before what the user has typed, but will re-appear when normal completion is triggered.  For example:
```
bash-4.3# bin/helm c<TAB>
HINT: The most common actions are: search, pull, install, list

# Notice the lack of prompt in the next line.  This only happens when there is an Active Help message
bin/helm c<TAB>
HINT: The most common actions are: search, pull, install, list

bin/helm c<TAB>
completion    create
bash-4.3# bin/helm c
```

_Should Active Help be disabled completely for bash < 4.4, or is the lack of prompt acceptable?_

**Missing:**
- Updating the [cobra-completion-testing](https://github.com/marckhouzam/cobra-completion-testing) repo

**This can be seen in action using https://github.com/spf13/cobra-cli/pull/24**

**Here is a test program that adds some active help:**
<details>
<summary>Test program</summary>

```go
package main

import (
	"fmt"
	"os"
	"strconv"
	"strings"

	"github.com/spf13/cobra"
)

var (
	completionNoDesc bool
	noMoreArgs       = "This command takes no more arguments"
)

	Use:   "completion [bash|bashv1|zsh|fish|powershell]",
	Long: `Bash:

Zsh:
$ compdef _testprog testprog

Fish:

PowerShell:
`,
	DisableFlagsInUseLine: true,
		switch args[0] {
		case "bash":
			cmd.Root().GenBashCompletionV2(os.Stdout, !completionNoDesc)
		case "bashv1":
		case "zsh":
			if !completionNoDesc {
			} else {
			}
		case "fish":
		case "powershell":
			if completionNoDesc {
			} else {
				cmd.Root().GenPowerShellCompletionWithDesc(os.Stdout)
			}
		}
	},
		return []string{
			"bashv1\tBash completion V1",
			"bash\tBash completion V2",
			"zsh\tZsh completion",
			"fish\tFish completion",
			"pwsh\tPowershell completions"}, cobra.ShellCompDirectiveNoFileComp
	},
}

var rootCmd = &cobra.Command{
	Use: "testprog",
		fmt.Println("rootCmd called")
	},
		return cobra.AppendActiveHelp(nil, "Choose a command to see different Active Help examples"),
			cobra.ShellCompDirectiveNoFileComp
	},
}

var onlyAH = &cobra.Command{
	Use:   "onlyAH",
	Short: "Command with Active Help but no args",
		fmt.Println("onlyAH called")
	},
		return cobra.AppendActiveHelp(nil, noMoreArgs), cobra.ShellCompDirectiveNoFileComp
	},
}

var argsAndAH = &cobra.Command{
	Use:   "argsAndAH",
	Short: "Command with Active Help and one arg",
		fmt.Println("argsAndAH called")
	},
		if len(args) == 0 {
			comps := []string{"firstArg", "secondArg"}
			comps = cobra.AppendActiveHelp(comps, "Please specify a single argument")
			comps = cobra.AppendActiveHelp(comps, "The arguments should be shown below if they match what you typed")
			return comps, cobra.ShellCompDirectiveNoFileComp
		}
		return cobra.AppendActiveHelp(nil, noMoreArgs), cobra.ShellCompDirectiveNoFileComp
	},
}

var filesAndAH = &cobra.Command{
	Use:   "filesAndAH",
	Short: "Command with Active Help and a file path",
		fmt.Println("filesAndAH called")
	},
		if len(args) == 0 {
			return cobra.AppendActiveHelp(nil, "Please specify the path to a file"),
				cobra.ShellCompDirectiveDefault
		}
		return cobra.AppendActiveHelp(nil, noMoreArgs), cobra.ShellCompDirectiveNoFileComp
	},
}

var goFilesAndAH = &cobra.Command{
	Use:   "goFilesAndAH",
	Short: "Command with Active Help and a go file argument",
		fmt.Println("goFilesAndAH called")
	},
		if len(args) == 0 {
			comps := cobra.AppendActiveHelp(nil, "Please specify the path to a file")
			comps = append(comps, "go")
			comps = cobra.AppendActiveHelp(comps, "On file ending with .go are allowed")
			return comps, cobra.ShellCompDirectiveFilterFileExt
		}
		return cobra.AppendActiveHelp(nil, noMoreArgs), cobra.ShellCompDirectiveNoFileComp

	},
}

var noSpaceAndAH = &cobra.Command{
	Use:   "noSpaceAndAH",
	Short: "Command with Active Help using the nospace directive",
		fmt.Println("noSpaceAndAH called")
	},
		var comps []string
		if len(args) == 0 {
			if !strings.Contains(toComplete, "/") {
				comps = []string{"firstRepo/", "secondRepo/"}
				comps = cobra.AppendActiveHelp(comps, "Please choose a repository name")
				return comps, cobra.ShellCompDirectiveNoSpace | cobra.ShellCompDirectiveNoFileComp
			}

			repo := toComplete[:strings.Index(toComplete, "/")+1]
			comps = []string{repo + "firstChart", repo + "secondChart"}
			comps = cobra.AppendActiveHelp(comps, "Please choose a chart name")
			return comps, cobra.ShellCompDirectiveNoFileComp
		}
		return cobra.AppendActiveHelp(comps, "This command takes no more arguments"), cobra.ShellCompDirectiveNoFileComp
	},
}

var levelsForAH = &cobra.Command{
	Use:   "levelsForAH",
	Short: "Command that uses different levels of active help",
		fmt.Println("levelsForAH called")
	},
		var comps []string
		if cobra.GetActiveHelpConfig(cmd) != "off" {
			comps = cobra.AppendActiveHelp(comps, "This msg displays as long as active help is not disabled")
			if cobra.GetActiveHelpConfig(cmd) == "on" {
				comps = cobra.AppendActiveHelp(comps, "This msg should not display if active help level is 'local'")
			}
		}

		comps = cobra.AppendActiveHelp(comps, noMoreArgs)
		return comps, cobra.ShellCompDirectiveNoFileComp

	},
}

func setFlags() {
	rootCmd.PersistentFlags().String("theme", "", "theme to use (located in /themes/THEMENAME/)")
	rootCmd.RegisterFlagCompletionFunc("theme", func(cmd *cobra.Command, args []string, toComplete string) ([]string, cobra.ShellCompDirective) {
		comps := []string{"themes"}
		comps = cobra.AppendActiveHelp(comps, "Please choose a theme from the list below")
		return comps, cobra.ShellCompDirectiveFilterDirs
	})

	rootCmd.PersistentFlags().Int("int", 0, "need an int as input")
	rootCmd.RegisterFlagCompletionFunc("int", func(cmd *cobra.Command, args []string, toComplete string) ([]string, cobra.ShellCompDirective) {
		comps := cobra.AppendActiveHelp(nil, "You must specify an integer")
		if len(toComplete) > 0 {
			if _, err := strconv.Atoi(toComplete); err != nil {
				comps = cobra.AppendActiveHelp(comps, "ERROR: What you have typed is not an integer")
			} else {
				comps = []string{toComplete}
			}
		}
		return comps, cobra.ShellCompDirectiveNoFileComp
	})

	rootCmd.PersistentFlags().Bool("bool", false, "boolean flag")
	rootCmd.RegisterFlagCompletionFunc("bool", func(cmd *cobra.Command, args []string, toComplete string) ([]string, cobra.ShellCompDirective) {
		return cobra.AppendActiveHelp(nil, "This is a boolean flag.  To disable it, remove the flag, or use --bool=false (the = is mandatory)"),
			cobra.ShellCompDirectiveNoFileComp
	})

}

func main() {
		&completionNoDesc,
		"no-descriptions", false,

	rootCmd.AddCommand(
		onlyAH,
		argsAndAH,
		filesAndAH,
		goFilesAndAH,
		noSpaceAndAH,
		levelsForAH,
	)

	setFlags()

	rootCmd.Execute()
}
```
</details>

/cc @jpmcb @jharshman @Luap99

### DIFF
--- a/active_help.go
+++ b/active_help.go
@@ -26,27 +26,24 @@ func AppendActiveHelp(compArray []string, activeHelpStr string) []string {
 	return append(compArray, fmt.Sprintf("%s%s", activeHelpMarker, activeHelpStr))
 }
 
-// activeHelpEnvVar returns the name of the program-specific ActiveHelp environment
-// variable.  It has the format <PROGRAM>_ACTIVE_HELP where <PROGRAM> is the name of the
-// root command in upper case, with all - replaced by _.
-// This format should not be changed: users will be using it explicitly.
-func activeHelpEnvVar(name string) string {
-	activeHelpEnvVar := strings.ToUpper(fmt.Sprintf("%s%s", name, activeHelpEnvVarSuffix))
-	return strings.ReplaceAll(activeHelpEnvVar, "-", "_")
-}
-
-// setActiveHelpConfig first checks the global environment variable
-// of ActiveHelp to see if it is disabling active help, and if it is not,
-// it then looks to the program-specific variable.
-// It then sets the ActiveHelpConfig value to make it available when
-// calling the completion function.  We also set it on the root,
-// just in case users try to access it from there.
-func setActiveHelpConfig(cmd *Command) {
+// GetActiveHelpConfig returns the value of the ActiveHelp environment variable
+// <PROGRAM>_ACTIVE_HELP where <PROGRAM> is the name of the root command in upper
+// case, with all - replaced by _.
+// It will always return "0" if the global environment variable COBRA_ACTIVE_HELP
+// is set to "0".
+func GetActiveHelpConfig(cmd *Command) string {
 	activeHelpCfg := os.Getenv(activeHelpGlobalEnvVar)
 	if activeHelpCfg != activeHelpGlobalDisable {
 		activeHelpCfg = os.Getenv(activeHelpEnvVar(cmd.Root().Name()))
 	}
+	return activeHelpCfg
+}
 
-	cmd.ActiveHelpConfig = activeHelpCfg
-	cmd.Root().ActiveHelpConfig = cmd.ActiveHelpConfig
+// activeHelpEnvVar returns the name of the program-specific ActiveHelp environment
+// variable.  It has the format <PROGRAM>_ACTIVE_HELP where <PROGRAM> is the name of the
+// root command in upper case, with all - replaced by _.
+func activeHelpEnvVar(name string) string {
+	// This format should not be changed: users will be using it explicitly.
+	activeHelpEnvVar := strings.ToUpper(fmt.Sprintf("%s%s", name, activeHelpEnvVarSuffix))
+	return strings.ReplaceAll(activeHelpEnvVar, "-", "_")
 }
--- a/active_help.go
+++ b/active_help.go
@@ -1,0 +1,52 @@
+package cobra
+
+import (
+	"fmt"
+	"os"
+	"strings"
+)
+
+const (
+	activeHelpMarker = "_activeHelp_ "
+	// The below values should not be changed: programs will be using them explicitly
+	// in their user documentation, and users will be using them explicitly.
+	activeHelpEnvVarSuffix  = "_ACTIVE_HELP"
+	activeHelpGlobalEnvVar  = "COBRA_ACTIVE_HELP"
+	activeHelpGlobalDisable = "0"
+)
+
+// AppendActiveHelp adds the specified string to the specified array to be used as ActiveHelp.
+// Such strings will be processed by the completion script and will be shown as ActiveHelp
+// to the user.
+// The array parameter should be the array that will contain the completions.
+// This function can be called multiple times before and/or after completions are added to
+// the array.  Each time this function is called with the same array, the new
+// ActiveHelp line will be shown below the previous ones when completion is triggered.
+func AppendActiveHelp(compArray []string, activeHelpStr string) []string {
+	return append(compArray, fmt.Sprintf("%s%s", activeHelpMarker, activeHelpStr))
+}
+
+// activeHelpEnvVar returns the name of the program-specific ActiveHelp environment
+// variable.  It has the format <PROGRAM>_ACTIVE_HELP where <PROGRAM> is the name of the
+// root command in upper case, with all - replaced by _.
+// This format should not be changed: users will be using it explicitly.
+func activeHelpEnvVar(name string) string {
+	activeHelpEnvVar := strings.ToUpper(fmt.Sprintf("%s%s", name, activeHelpEnvVarSuffix))
+	return strings.ReplaceAll(activeHelpEnvVar, "-", "_")
+}
+
+// setActiveHelpConfig first checks the global environment variable
+// of ActiveHelp to see if it is disabling active help, and if it is not,
+// it then looks to the program-specific variable.
+// It then sets the ActiveHelpConfig value to make it available when
+// calling the completion function.  We also set it on the root,
+// just in case users try to access it from there.
+func setActiveHelpConfig(cmd *Command) {
+	activeHelpCfg := os.Getenv(activeHelpGlobalEnvVar)
+	if activeHelpCfg != activeHelpGlobalDisable {
+		activeHelpCfg = os.Getenv(activeHelpEnvVar(cmd.Root().Name()))
+	}
+
+	cmd.ActiveHelpConfig = activeHelpCfg
+	cmd.Root().ActiveHelpConfig = cmd.ActiveHelpConfig
+}
--- a/completions.go
+++ b/completions.go
@@ -178,6 +178,12 @@ func (c *Command) initCompleteCmd(args []string) {
 
 			noDescriptions := (cmd.CalledAs() == ShellCompNoDescRequestCmd)
 			for _, comp := range completions {
+				if finalCmd.ActiveHelpConfig == activeHelpGlobalDisable {
+					// Remove all activeHelp entries in this case
+					if strings.HasPrefix(comp, activeHelpMarker) {
+						continue
+					}
+				}
 				if noDescriptions {
 					// Remove any description that may be included following a tab character.
 					comp = strings.Split(comp, "\t")[0]
@@ -447,6 +453,9 @@ func (c *Command) getCompletions(args []string) (*Command, []string, ShellCompDi
 		// Go custom completion defined for this flag or command.
 		// Call the registered completion function to get the completions.
 		var comps []string
+
+		setActiveHelpConfig(finalCmd)
+
 		comps, directive = completionFn(finalCmd, finalArgs, toComplete)
 		completions = append(completions, comps...)
 	}
--- a/completions.go
+++ b/completions.go
@@ -178,7 +178,7 @@ func (c *Command) initCompleteCmd(args []string) {
 
 			noDescriptions := (cmd.CalledAs() == ShellCompNoDescRequestCmd)
 			for _, comp := range completions {
-				if finalCmd.ActiveHelpConfig == activeHelpGlobalDisable {
+				if GetActiveHelpConfig(finalCmd) == activeHelpGlobalDisable {
 					// Remove all activeHelp entries in this case
 					if strings.HasPrefix(comp, activeHelpMarker) {
 						continue
@@ -453,9 +453,6 @@ func (c *Command) getCompletions(args []string) (*Command, []string, ShellCompDi
 		// Go custom completion defined for this flag or command.
 		// Call the registered completion function to get the completions.
 		var comps []string
-
-		setActiveHelpConfig(finalCmd)
-
 		comps, directive = completionFn(finalCmd, finalArgs, toComplete)
 		completions = append(completions, comps...)
 	}
--- a/active_help.md
+++ b/active_help.md
@@ -13,6 +13,8 @@ Please specify the path to the chart to package
 bash-5.1$ bin/helm package [tab][tab]
 bin/    internal/    scripts/    pkg/     testdata/
 ```
+
+**Hint**: A good place to use Active Help messages is when the normal completion system does not provide any suggestions. In such cases, Active Help nicely supplements the normal shell completions to guide the user in knowing what is expected by the program.
 ## Supported shells
 
 Active Help is currently only supported for the following shells:
@@ -61,6 +63,7 @@ You must specify the URL for the repo you are adding
 bash-5.1$ helm repo add grafana https://grafana.github.io/helm-charts [tab]
 This command does not take any more arguments
 ```
+**Hint**: As can be seen in the above example, a good place to use Active Help messages is when the normal completion system does not provide any suggestions. In such cases, Active Help nicely supplements the normal shell completions.
 
 ### Active Help for flags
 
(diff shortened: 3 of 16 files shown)


completions.go
```diff
@@ -148,16 +166,21 @@ func (d ShellCompDirective) string() string {
...
+		Use:                fmt.Sprintf("%s [command-line]", ShellCompRequestCmd),
+		Aliases:            []string{ShellCompNoDescRequestCmd},
+		Hidden:             true,
+		DisableFlagParsing: true,
+		Args:               MinimumNArgs(1),
+		Short:              "Request shell completion choices for the specified command-line",
 		Long: fmt.Sprintf("%[2]s is a special command that is used by the shell completion logic\n%[1]s",
 			"to request completion choices for the specified command-line.", ShellCompRequestCmd),
 		Run: func(cmd *Command, args []string) {
+			args, _ = cmd.extractCompleteCmdFlags(args)
```
Luap99, line 178: I wonder if it is really necessary to have this implemented as flag. I think it would be much easier to just use a new cmd alias, .e.g. `ShellCompActiveHelpRequestCmd`. I am also not sure why `ActiveHelpConfig` should be a string, I personally think a boolean would be enough. Do you think people would implement several help levels?  I personally either want help or not.

Luap99 (commented):
This is a very cool feature. Thanks for doing this.

completions.go
marckhouzam: Thank you for having had a look already!

You bring up good points, all of which evolved as I tested Active Help.

> I think it would be much easier to just use a new cmd alias, e.g. `ShellCompActiveHelpRequestCmd`

You are right.  The problem happens when you want mix it with descriptions.  If I want descriptions but not active help, or vice versa, or both together, or neither one.  We would need a different command alias for every permutation.  And in the future, if we were to add a third feature, it would compound the problem.

An alternative to using a flag would be to use an environment variable.  That is actually how I originally implemented things.  However, when writing go tests for the program using Cobra, it became apparent that dealing with environment variables was not as easy as dealing with flags. But that point is open for discussion.

A better implementation for the flag solution would be to let Cobra parse the flag.  This could be done by separating the flag of the `__complete` command from the rest of the flags using the `--` delimiter (e.g., `helm __complete --__activfeHelpCfg=1 -- --namespace ''`).  The problem is that it wouldn't be backwards-compatible, as it would require programs to add the `--` to go tests they already have written; and the `--` would also become mandatory when debugging from the command-line.  That is why I passed on that approach.

> I am also not sure why ActiveHelpConfig should be a string, I personally think a boolean would be enough.

Another good point 😄 .  That was also how I first implemented things.  But when using the feature with helm, I hit some scenarios where the Active Help needed to query the kubernetes cluster; such queries were slow (in comparison).  This made me think that some programs may wish to have different levels of active help, e.g., off, on, local.  Since it is hard to predict the need of programs, I went with the full flexibility of a string.

Note however that a program can still choose to present a boolean flag to their users to only support active help being on or off.

Luap99: > You are right. The problem happens when you want mix it with descriptions. If I want descriptions but not active help, or vice versa, or both together, or neither one. We would need a different command alias for every permutation. And in the future, if we were to add a third feature, it would compound the problem.

Yep, that is what I thought.

> An alternative to using a flag would be to use an environment variable. That is actually how I originally implemented things. However, when writing go tests for the program using Cobra, it became apparent that dealing with environment variables was not as easy as dealing with flags. But that point is open for discussion.

I like the idea of environment variables because that will allow users to change the behaviour more easily. I find it tedious to regenerate the completion script, especially since for Podman we automatically install the script via package manager. Using environment variables would make it easy to document this. One global environment variable would allow me to change it for all my cobra applications.

As far as testing is concerned you can do `os.Setenv()` and `defer os.Unsetenv()` inside your go test function.

> A better implementation for the flag solution would be to let Cobra parse the flag. This could be done by separating the flag of the __complete command from the rest of the flags using the -- delimiter (e.g., helm __complete --__activfeHelpCfg=1 -- --namespace ''). The problem is that it wouldn't be backwards-compatible, as it would require programs to add the -- to go tests they already have written; and the -- would also become mandatory when debugging from the command-line. That is why I passed on that approach.

I don't think cobra needs to do the the flag parsing. We can use pflag directly, e.g. something like this should work:
https://github.com/containers/podman/blob/bbd03df3e120998f57393c25c14ca4fe15f5aec3/pkg/systemd/generate/containers.go#L268-L278

Can you provide a link to your helm changes. I would like to see how this looks in a real application.

marckhouzam:
Here is the helm branch that uses this feature: https://github.com/VilledeMontreal/helm/tree/feat/compInfo

Without needing a kubernetes cluster, one can try:
```
source <(bin/helm completion zsh) # or bash
bin/helm TAB
bin/helm repo add TAB
bin/helm repo add name TAB
bin/helm repo add name https TAB
bin/helm package TAB
bin/helm template TAB
```


I will make a smaller test program and post it in this PR.  I should have thought about doing that before.

marckhouzam:
I've just added a test program in the description of the PR which shows different Active Help scenarios and how to use the active help configuration

jharshman:
@marckhouzam This is a great IMO. I like the interactive nature of it and that it is opt-in. :rocket:

marckhouzam:
I've pushed an extra commit which uses the environment variable `COBRA_ACTIVEHELP` to configure Active Help instead of using a flag as mentioned by @Luap99.  Looking at that one commit (and possibly reverting it) allows to compare the two approaches.

A nice thing about using an env var is that a user could set a different active help setting for individual commands by wrapping them in an alias or script which sets the env var (although that would be a pretty advanced usage).

Using a name like `COBRA_ACTIVEHELP` allows to set an active help level for all Cobra programs.  However, I am afraid that contradicts the fact that different programs can choose whatever active help configuration they want to use.  For example, `helm` could use active help levels of `on/off/local` but `kubectl` could use `enabled/disabled/noremote`; in such a case, using a global name such as `COBRA_ACTIVEHELP` makes the configuration not work.

We could instead use the program name in the variable, something like `HELM_ACTIVEHELP` and `KUBECTL_ACTIVEHELP`, and maybe have the global `COBRA_ACTIVEHELP` to only allow to disable active help for every program.  I'm not sure if we need to go that far though.

Opinions on using a flag versus an env var?

completions.go
marckhouzam: I've pushed a commit that switches to using an env var to continue the discussion.

marckhouzam: And you were right @Luap99 that using environment variables within go tests wasn't a problem.

But if we prefer using environment variables for active help, I don't think we should use only a global one as I mentioned in a previous comment.

I'll update the implementation to illustrate.

marckhouzam:
Still valid

active_help.go
```diff
@@ -0,0 +1,48 @@
...
+// AppendActiveHelp adds the specified string to the specified array to be used as ActiveHelp.
+// Such strings will be processed by the completion script and will be shown as ActiveHelp
+// to the user.
+// The array parameter should be the array that will contain the completions.
+// This function can be called multiple times before and/or after completions are added to
+// the array.  Each time this function is called with the same array, the new
+// ActiveHelp line will be shown below the previous ones when completion is triggered.
+func AppendActiveHelp(compArray []string, activeHelpStr string) []string {
+	return append(compArray, fmt.Sprintf("%s%s", activeHelpMarker, activeHelpStr))
+}
```
Luap99, line 27: I am not sure with the API here, why do I give the compArray as parameter and get the same value with the appended entry back.

IMO It would make more sense to have it like this:
```go
func AppendActiveHelp(activeHelpStr string) string {
	return fmt.Sprintf("%s%s", activeHelpMarker, activeHelpStr)
}
``` 
and then in you code call
```
comps = append(comps, cobra.AppendActiveHelp("text"))
```

With this the caller has control over the append. While I do not think we need to worry about performance here I believe it is better when you call append only once.
```
text1 := cobra.AppendActiveHelp("text1")
text2 := cobra.AppendActiveHelp("text2")
comps = append(comps, text1, text2)
```

active_help.go
```diff
@@ -0,0 +1,48 @@
...
+
+// GetActiveHelpConfig reads the ActiveHelp environment variable
+// of the program and returns its value.
```
Luap99, line 30: Maybe add `<PROGRAM>_ACTIVE_HELP` so the programmer sees the name directly when looking at the godoc comment.

Luap99 (commented):
Nice work, I really like the idea behind this.
I think it is cool to display a message when your command does not take args for example.

I tested this with bash and it works as I would expect.

active_help.go
marckhouzam: Good idea, will do

active_help.go
marckhouzam: I struggled to define a good API to add ActiveHelp messages.  Here is how I remember choosing the `AppendActiveHelp(compArray []string, activeHelpStr string)` API.

1. I was worried that it was not obvious that activeHelp messages should be included in the completions array returned by `ValidArgsFunction`; the two concepts could be seen as independent by users (completions and activeHelp messages).  I therefore felt that by requiring the user to specify an array (and saying in the godoc that the array should be the one also containing the completion choices) we would make it much clearer that activeHelp messages should be included in the array of completions.
2. based on that, I then thought we should mimic the behaviour of Go's `append()` which is why I called it `AppendActiveHelp()` and required the user to assign the result to an array (just like `append()`).  All this in the hope to build on existing knowledge of an existing API.

If end up preferring
```
func AppendActiveHelp(activeHelpStr string) string {
	return fmt.Sprintf("%s%s", activeHelpMarker, activeHelpStr)
}
```
we should probably rename the method to avoid using the word "append".  We could create it as `CreateActiveHelpMsg()` for example.  

But then I worry again that a user won't know easily what to do after having created an activeHelp message.  Will they realize they must include it in the completions array?

Some people had trouble understanding what the arguments to `ValidArgsFunction(cmd, args, toComplete)` represented... So I don't want to assume anything is clear 😄

marckhouzam:
> Nice work, I really like the idea behind this. I think it is cool to display a message when your command does not take args for example.

I expect the most common use of ActiveHelp will/should be to display a message when there are no completions suggested.  When there are no completions, the user doesn't really know what to do, so an ActiveHelp message can explain (e.g., "This command does not take any more arguments", or "Please specify a name for the container that will be created").

> I tested this with bash and it works as I would expect.

Thanks for trying it out!

marckhouzam:
I've created a draft PR for the cobra-cli that uses ActiveHelp.  I hope this show very easily how the feature can be used: https://github.com/spf13/cobra-cli/pull/24

@johnSchnake I've spoken to others about this but not to you.  I'd love your impressions on the feature.

marckhouzam:
I've rebased this PR on top of the bash-v2 improvements that were recently merged.  It is ready again.

@scop if you have a moment, maybe you can have a look at the bash code modifications to see if you notice anything wrong?

jpmcb (commented):
Wow this is incredible. I love it.

Overall, I think this is ready for the spring release of Cobra we'll be cutting soon. I'm not super up to date on the zsh/bash completions scripting but it appears to work as expected. I'm comfortable cutting a path release if we need to fix bugs if needed.

> It is not possible to implement for fish currently.

I'm curious! Something due to fish's spec?

active_help.go
jpmcb: Hmmmmm this is a tricky one. I'm not sure on this.

I think I'm a fan of using the API presented here by Marc, but I'm warry of introducing more complexity to how users get their completions into their program.

But from my understanding of this, it's completely opt in? It's **_not_** another required step users must take to get their completions working?

marckhouzam: Correct, the existing API for normal completions is not affected at all.  ActiveHelp is 100% opt-in and can be completely ignored by programs using Cobra.

marckhouzam:
Thanks for trying this out @jpmcb  🎉  

> > It is not possible to implement for fish currently.

> I'm curious! Something due to fish's spec?

From what I remember of fish completion logic, there is no way to output something else than actual completions (can't generically print to stdout like bash allows); so my expectation is that ActiveHelp would be treated as a real completion and get sorted within the list of completions and would even be selectable as a completion through repeated TABing.  So it wouldn't work as we would need.

Maybe @krobelus can confirm or correct this belief.

BTW, since I first opened the PR I took the time to try ActiveHelp with PowerShell and there too it is not possible to output anything else than actual completions (cannot print to stdout like bash allows).  I may have an idea to hack something that may work but it would need more investigation that I haven't had time to dive deeper into.  So for now, we would have to contend with bash and zsh (which until recently were pretty much the only two shells we supported anyway).

jpmcb:
I think this is good to merge and we can iterate as we get community feedback! Thoughts @marckhouzam? 🚀

krobelus:
> From what I remember of fish completion logic, there is no way to output something else than actual completions (can't generically print to stdout like bash allows)

yeah, fish will likely overwrite that with the completion pager. I guess we can an API for adding arbitrary text.

marckhouzam:
> I think this is good to merge and we can iterate as we get community feedback! Thoughts @marckhouzam? 🚀

Let's do it!  Thanks @jpmcb 🚀

scop:
> @scop if you have a moment, maybe you can have a look at the bash code modifications to see if you notice anything wrong?

Skimmed it real quick and didn't notice anything wrong. There's some room for cleanups along the lines of #1702, which I suppose we should amend to cover the things added here. But the feature looks quite interesting indeed :+1:

PR #1455, opened 2021-07-12
Completion: Capitalize short desc and remove extra space from long 

fcjr:
The `help` short desc is capitalized so we should follow its lead for consistency.

Also fixes #1458.

PR #1485, opened 2021-09-10
Uppercase first char in short desc of generate command

samherrmann:
This commit provides consistency of the letter casing in the short
description between commands. Both the short description as well as
documented examples have the first character in the short description in
uppercase. The generate command was the exception by having the first
character of the short description in lowercase, creating an
inconsistent help menu for the user:

``` sh
Available Commands:
  add              Add a command to a Cobra Application
  completion       generate the autocompletion script for the specified shell
  help             Help about any command
  init             Initialize a Cobra Application
```

* User Guide Example: https://github.com/spf13/cobra/blob/v1.2.1/user_guide.md#example
* Help command: https://github.com/spf13/cobra/blob/master/command.go#L1080

### DIFF
--- a/completions.go
+++ b/completions.go
@@ -589,7 +589,7 @@ func (c *Command) initDefaultCompletionCmd() {
 
 	completionCmd := &Command{
 		Use:   compCmdName,
-		Short: "generate the autocompletion script for the specified shell",
+		Short: "Generate the autocompletion script for the specified shell",
 		Long: fmt.Sprintf(`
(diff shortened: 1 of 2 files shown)


marckhouzam:
Thanks @samherrmann.
This is a duplicate of a slightly larger ajustement in #1455.

samherrmann:
Wow, I'm not sure how I missed that PR given that it's near to the top.

PR #1444, opened 2021-07-07
fix: improve completions help formatting

caarlos0:
did some improvements on the completions docs so they look a bit better if exported as markdown (and still look good on plain text):

- indent the example command lines so they are rendered as code blocks
- removed the $ prefix so they are easier to copy/paste
- standarized the Linux and macOS subsections when present

example: 

- https://nfpm-m71s1p3z3-goreleaser.vercel.app/cmd/nfpm_completion_bash/ (without this patch)
- https://nfpm-nmv190dyc-goreleaser.vercel.app/cmd/nfpm_completion_bash/ (with this patch)

PR #1495, opened 2021-09-23
Remove indentation causing RST errors

melissamahoney-mongodb:
We generate our docs in RST from Cobra. I removed the indentations causing an error in the generated RST. This also makes the long descriptions consistent.

### DIFF
--- a/completions.go
+++ b/completions.go
@@ -616,9 +616,9 @@ $ source <(%[1]s completion bash)
 
 To load completions for every new session, execute once:
 Linux:
-  $ %[1]s completion bash > /etc/bash_completion.d/%[1]s
+$ %[1]s completion bash > /etc/bash_completion.d/%[1]s
 MacOS:
-  $ %[1]s completion bash > /usr/local/etc/bash_completion.d/%[1]s
+$ %[1]s completion bash > /usr/local/etc/bash_completion.d/%[1]s
 
 You will need to start a new shell for this setup to take effect.
   `, c.Root().Name()),


completions.go
```diff
@@ -616,9 +616,9 @@ $ source <(%[1]s completion bash)
 
 To load completions for every new session, execute once:
 Linux:
-  $ %[1]s completion bash > /etc/bash_completion.d/%[1]s
+$ %[1]s completion bash > /etc/bash_completion.d/%[1]s
```
umarcor, line 619: Maybe keep the indentation and add `.. code:: sh` before? See https://docutils.sourceforge.io/docs/ref/rst/directives.html#code

marckhouzam: Would this be interpreted correctly for the help message @umarcor ?

umarcor: To be honest, I am genuinely confused. To my understanding, the help message is printed to screen/terminal, so unrelated to RST. My previous reply was considering that line 619 is going to end being part of a restructuredtext file. However, I now see why you are asking, since this is indeed the "Long" field of a command. Is the same text source used for both printing the help in screen and for generating the rst docs?

marckhouzam: It's definitely used for the help message and I'm learning with this PR that it is also used for RST.  So the original idea of removing the indentation is probably the best way forward. What do you think?

umarcor: I think it is technically not correct to remove the indentation, because the resulting RST is not semantically valid. However, given the constraints, I believe it's fair enough. It allows removing the error found by @melissamahoney-mongodb and we don't need to learn the details of doc generation right now. I'll mark this dialogue as resolved, and we'll deal with docgen some other day :wink:.

marckhouzam:
I think this PR is giving problems:
1. it removes indentation while #1444 adds indentation.  When both are joined in #1525 we are not consistent with the indentation between `completion bash -h` and the other shells (e.g., `completion zsh -h`)
2. the indentation allows nice code-block formatting in markdown, so I feel it is better to have it than not

@melissamahoney-mongodb can you specify how you get an error when generating RST doc?  I haven't been able to reproduce it.

@umarcor what do you think?  I personally would pull this out of the release candidate until we can clarify the situation.

umarcor:
@marckhouzam I removed this from #1496 and I pulled it out of #1525.

It seems that supporting multiple output format might require explicitly handling multiple text sources...

Issue #1496, opened 2021-09-24
Fall release 🍁 - 2021


umarcor:
~Last release was planned for May 2021 and released in July. Next week, it will be 3 months since that. In this period, 3 PRs were merged.~

~Is there any plan for merging PRs and releasing a new version in the following weeks?~

# Release candidate (please try out and test!)

- #1525

# To Do

Approved by a maintainer but merge seems to have been forgotten:

- #841

Left pending in the previous release (#1388):

- #1387 

Small completion fixes from the community:

- #1467

Worth considering:

- #1530

Some random cool ones:

- #1231
- #1003

# Done

- #880
- #896
- #949
- #979
- #1009
- #1159
- #1161
- #1321
- #1372
- #1426 
- #1427 
- #1429
- #1444
- #1454. Closes:
  - #987
  - #1371
  - #1377
  - #1380
  - #1407
  - #1446
  - #1451
  - #1497
  - #1506
  - #1522
  - #1528
- #1455
- #1459
- #1463
- #1473
- #1477
- #1509
- #1510
- #1514
- #1520
- #1541
- #1552
- #1557

# Delayed to a later release

- #1500

marckhouzam:
To help the maintainers how about we try to weed through some of the PRs?  
Here's what I came up with.  

My apologies to the community for not listing some of the ones I'm not familiar with, it does not mean they are not just as valuable, but I had to start somewhere.

**Adding projects to projects_using_cobra.md:**

@umarcor has been efficient and rebased 6 contributions into one PR to avoid the merge conflicts: #1454
This will allow to close: #987 #1371 #1380 #1407 #1446 #1451 

**Approved by a maintainer but merge seems to have been forgotten:**

#1161 DisableFlagParsing must trigger custom completion for flag names
#841  Generalize ValidArgs: use it implicitly with any validator

**Small completion fixes from the community:**

#1455 Completion: Capitalize short desc and remove extra space from long
#1321 fix: unbound variables in bash completion
#1450 feat(completion): add default completion hidden if there are no subcmds
#1372 Update shell_completions.md to have automatically rootcmd
#1467 feat: make InitDefaultCompletionCmd public
#1473 completions: fix mixed tab/spaces indentation
#1459 fix: typo in {bash,zsh}_completions.go
#1444 fix: improve completions help formatting
#1495 Remove indentation causing RST errors

**Some random cool ones:**

#1231 add colors to CLI commands
#1003 Add groups for commands in help 

[...]

PR #1500, opened 2021-10-03
Support subcommands checking for suggestions

### LINKED ISSUE #636: RFE: Support OnlySubCommands arg validation to make subcommand get suggestions
### Motivation

Currently, suggestions for "unknown command" are not working for
**sub**commands. For example code[1](wants to accept only when 
subcommand specified), I expected that cobra give us suggestions like:

_Expected_
~~~
  $ go run main.go echo time
  Error: unknown command "time" for "echo"

  Did you mean this?
   	times
~~~

However, cobra does not suggest anything unfortunately.
`SuggestFor:` attribute also does not work currently.

### Workaround

If I use `OnlyValidArgs` for the subcommands, this could be
solve. However, it does not suit when a command has many subcommands 
and makes it makes difficult to maintain.

### Proposal patch

https://github.com/spf13/cobra/pull/635

[1] example code
~~~
package main

import (
        "fmt"
        "strings"

        "github.com/spf13/cobra"
)

func main() {
        var echoTimes int
        var cmdEcho = &cobra.Command{
                Use:   "echo",
                Short: "My root command",
                        fmt.Println("Don't want to call this...")
                },
        }

        var cmdTimes = &cobra.Command{
                Use: "times [# times] [string to echo]",
                        for i := 0; i < echoTimes; i++ {
                                fmt.Println("Echo: " + strings.Join(args, " "))
                        }
                },
        }
        cmdTimes.Flags().IntVarP(&echoTimes, "times", "t", 1, "times to echo the input")
        var rootCmd = &cobra.Command{Use: "echo"}
        rootCmd.AddCommand(cmdEcho)
        cmdEcho.AddCommand(cmdTimes)
        rootCmd.Execute()
}
~~~

### LINKED ISSUE #834: <Suggestions when "unknown command" happens> only works at first level ?
Suggestions when "unknown command" happens

I have a three levels command, something like (presenting by level)  
 1/ mycmd 
 2/ mycmd list
    mycmd modify 

  3/ mycmd list volumes <flags>
     mycmd list disks <flags>
     mycmd list networks <flags>
     mycmd modify volumes <flags>
     mycmd modify disks <flags>
     mycmd modify networks <flags>

The suggestion works well for level 2 : if ever I do a typo on 'list' or 'modify' then I have the msg "Did you mean this?".
But the suggestions does not work at all if I do a typo on the third level : a typo on 'volumes', 'disks', 'networks'  does not generate the "Did you mean this?" message ?

I ve tried to play with such kind of code (see below), without success : 
	volumesCmd.DisableSuggestions = false
	volumesCmd.SuggestionsMinimumDistance = 2
	listCmd.AddCommand(volumesCmd)

Thank you for any help or clue !

YuviGold:
Resolves #636
Resolves #834
Resolves https://github.com/alexellis/arkade/issues/525

Till now, only the root command looked for suggestions.
Now `findSuggestions` works for nested commands as well.

For example

```
Error: unknown command "rewin" for "root times"

Did you mean this?
	rewind

Run 'root times --help' for usage.
```

### DIFF
--- a/args.go
+++ b/args.go
@@ -8,8 +8,8 @@ import (
 type PositionalArgs func(cmd *Command, args []string) error
 
 // Legacy arg validation has the following behaviour:
-// - root commands with no subcommands can take arbitrary arguments
-// - root commands with subcommands will do subcommand validity checking
+// - commands with no subcommands can take arbitrary arguments
+// - commands with subcommands will do subcommand validity checking
 // - subcommands will always accept arbitrary arguments
 func legacyArgs(cmd *Command, args []string) error {
 	// no subcommand, always take args
@@ -17,8 +17,8 @@ func legacyArgs(cmd *Command, args []string) error {
 		return nil
 	}
 
-	// root command with subcommands, do subcommand checking.
-	if !cmd.HasParent() && len(args) > 0 {
+	// do subcommand checking
+	if len(args) > 0 {
 		return fmt.Errorf("unknown command %q for %q%s", args[0], cmd.CommandPath(), cmd.findSuggestions(args[0]))
 	}
 	return nil
--- a/command_test.go
+++ b/command_test.go
@@ -1194,38 +1194,110 @@ func TestSuggestions(t *testing.T) {
 		SuggestFor: []string{"counts"},
 		Run:        emptyRun,
 	}
+	rewindCmd := &Command{
+		Use:        "rewind",
+		SuggestFor: []string{"dejavu"},
+		Run:        emptyRun,
+	}
+	timesCmd.AddCommand(rewindCmd)
 	rootCmd.AddCommand(timesCmd)
 
-	templateWithSuggestions := "Error: unknown command \"%s\" for \"root\"\n\nDid you mean this?\n\t%s\n\nRun 'root --help' for usage.\n"
-	templateWithoutSuggestions := "Error: unknown command \"%s\" for \"root\"\nRun 'root --help' for usage.\n"
+	templateSuggestions := "\nDid you mean this?\n\t%s\n\n"
 
-	tests := map[string]string{
-		"time":     "times",
-		"tiems":    "times",
-		"tims":     "times",
-		"timeS":    "times",
-		"rimes":    "times",
-		"ti":       "times",
-		"t":        "times",
-		"timely":   "times",
-		"ri":       "",
-		"timezone": "",
-		"foo":      "",
-		"counts":   "times",
+	templatePrefix := "Error: unknown command \"%s\" for \"%s\"\n"
+	templateSuffix := "Run '%s --help' for usage.\n"
+
+	tests := []struct {
+		input              string
+		targetCommand      string
+		expectedSuggestion string
+	}{
+		{
+			input:              "time",
+			expectedSuggestion: "times",
+		},
+		{
+			input:              "tiems",
+			expectedSuggestion: "times",
+		},
+		{
+			input:              "tims",
+			expectedSuggestion: "times",
+		},
+		{
+			input:              "timeS",
+			expectedSuggestion: "times",
+		},
+		{
+			input:              "rimes",
+			expectedSuggestion: "times",
+		},
+		{
+			input:              "ti",
+			expectedSuggestion: "times",
+		},
+		{
+			input:              "t",
+			expectedSuggestion: "times",
+		},
+		{
+			input:              "timely",
+			expectedSuggestion: "times",
+		},
+		{
+			input:              "ri",
+			expectedSuggestion: "",
+		},
+		{
+			input:              "timezone",
+			expectedSuggestion: "",
+		},
+		{
+			input:              "foo",
+			expectedSuggestion: "",
+		},
+		{
+			input:              "counts",
+			expectedSuggestion: "times",
+		},
+		{
+			input:              "foo rewind",
+			expectedSuggestion: "",
+		},
+		{
+			input:              "times rewin",
+			targetCommand:      "root times",
+			expectedSuggestion: "rewind",
+		},
+		{
+			input:              "times dejavu",
+			targetCommand:      "root times",
+			expectedSuggestion: "rewind",
+		},
 	}
 
-	for typo, suggestion := range tests {
+	for index := range tests {
 		for _, suggestionsDisabled := range []bool{true, false} {
+			test := tests[index]
+
+			timesCmd.DisableSuggestions = suggestionsDisabled
 			rootCmd.DisableSuggestions = suggestionsDisabled
 
-			var expected string
-			output, _ := executeCommand(rootCmd, typo)
+			args := strings.Split(test.input, " ")
+			output, _ := executeCommand(rootCmd, args...)
 
-			if suggestion == "" || suggestionsDisabled {
-				expected = fmt.Sprintf(templateWithoutSuggestions, typo)
-			} else {
-				expected = fmt.Sprintf(templateWithSuggestions, typo, suggestion)
+			unknownArg := args[len(args)-1]
+			if test.targetCommand == "" {
+				test.targetCommand = rootCmd.Use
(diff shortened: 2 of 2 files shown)


spf13:
Looks good to me. Would be a nice addition to the #1496

spf13 (commented):
LGTM

umarcor:
Now listed in #1496 and merged in #1525.

@spf13, maybe you can merge some of the PRs listed in #1496? Or, alternatively, provide merging permissions to some active contributor such as @marckhouzam.

marckhouzam:
I am concerned that this change is not backwards-compatible and will break a bunch of programs.

Before this PR, commands (except the root command) that don't specify the `Args` field would accept both sub-commands and arguments.  With this PR, only sub-commands will be accepted.  For example, if I have a program that has one sub-command and one sub-sub-command like so (see full program at the end of this comment):
```
$ prog sub subSub
```
Before this PR:
```
$ ./prog sub somearg
sub called with args [somearg]
```
but with this PR:
```
$ ./prog sub somearg
Error: unknown command "somearg" for "prog sub"
Run 'prog sub --help' for usage.
```
I believe this is a valid scenario that we will find in the wild.  I know that the helm project used to take avantage of this possibility (it no longer does to improve its shell completion usage).  For example, using an old version of helm:
```
$ helm2 get -h
[...]
Usage:
  helm get [flags] RELEASE_NAME
  helm get [command]

Available Commands:
  hooks       Download all hooks for a named release
  manifest    Download the manifest for a named release
  notes       Displays the notes of the named release
  values      Download the values file for a named release
[...]
```
Notice how the `helm2 get` command would accept a release name as an argument, or any of the allowed sub-commands.  This would break with this PR.

<details>
<summary>Test program used in the above example:</summary>

```go
package main

import (
	"fmt"

	"github.com/spf13/cobra"
)

var root = &cobra.Command{
	Use: "prog",
		fmt.Printf("prod called with args %v\n", args)
	},
}

var sub = &cobra.Command{
	Use: "sub",
		fmt.Printf("sub called with args %v\n", args)
	},
}

var subSub = &cobra.Command{
	Use: "subSub",
		fmt.Printf("subSub called with args %v\n", args)
	},
}

func main() {
	root.AddCommand(sub)
	sub.AddCommand(subSub)

	root.Execute()
}
```
</details>

umarcor:
@marckhouzam, I agree that this PR introduces a change that might break a bunch of programs. However, my understanding is that those programs are relaying on "undefined behaviour"; hence, it is not a breaking change from a semver perspective.

See https://github.com/spf13/cobra/blob/master/user_guide.md#positional-and-custom-arguments. There is no documented default behaviour if `Args` is not specified. helm and other projects might be expecting the default to be `ArbitraryArgs`. That is the implicit behaviour of the codebase indeed: 

https://github.com/spf13/cobra/blob/9e1d6f1c2aa8df64b6a6ba39e92517f68580d653/command.go#L998-L1003

https://github.com/spf13/cobra/blob/9e1d6f1c2aa8df64b6a6ba39e92517f68580d653/args.go#L55-L57

In order to assume that defaulting to `ArbitraryArgs` is explicit and not arbitrary (unspecified), we should change L999-L1001 of `command.go` to:

```go
	if c.Args == nil {
		return ArbitraryArgs(c, args)
	}
```

and update the documentation accordingly. I believe that might be done on top of this PR.

On the other hand, as far as I understand, @YuviGold is changing the legacy behaviour. That was apparently inconsistent with regard to the root command and subcommands, and he is making it consistent. It was surprising for me, because I would expect the "legacy" to be kept untouched, then deprecated, last removed; never updated. Moreover, I wonder if this update wouldn't be equivalent to actually removing the usage of `legacyArgs` from the codebase. Anyway, since this PR was explicitly reviewed by @spf13, and he requested it to be included in the next release, I guess it is ok. I'd expect @spf13 to be the one who best understands the rationale behind legacyArgs.

I will move this PR to the end of #1525, so we can keep discussing about it while other PRs are merged.

YuviGold:
Thank you @marckhouzam .I honestly didn't think about this use case.
Although as @umarcor pointed out, this seems to be the expected behavior.
So please, let me know what do you decide and if there's something needed to be done. :)

I believe that this multi-level suggestion is useful and necessary for many tools.
For example, [cli/cli](https://github.com/cli/cli) implemented the suggestions themselves while I would expect cobra to support it.
https://github.com/cli/cli/blob/029d49f3b38ebd54cd84b23732bb7efabfce4896/pkg/cmd/root/help.go#L51-L76
```
// Display helpful error message in case subcommand name was mistyped.
// This matches Cobra's behavior for root command, which Cobra
// confusingly doesn't apply to nested commands
```

marckhouzam:
@umarcor 
> Anyway, since this PR was explicitly reviewed by @spf13, and he requested it to be included in the next release, I guess it is ok. I'd expect @spf13 to be the one who best understands the rationale behind legacyArgs.

I of course agree that @spf13 and the maintainers have the final word on this.  But I think it would be quite understandable if @spf13 did not realize that this was a breaking change.  Considering the stated importance of very strong stability for Cobra, my guess is that this change should be avoided.

> I will move this PR to the end of #1525, so we can keep discussing about it while other PRs are merged.

Thanks!

@YuviGold 
> I believe that this multi-level suggestion is useful and necessary for many tools.

I agree.  Maybe a backwards-compatible approach can be defined?  At a minimum an opt-in solution, would be backwards-compatible, but there may be something even better; I haven't looked into it myself.

umarcor:
> I of course agree that @spf13 and the maintainers have the final word on this. But I think it would be quite understandable if @spf13 did not realize that this was a breaking change. Considering the stated importance of very strong stability for Cobra, my guess is that this change should be avoided.

@marckhouzam, I suggest you or anyone else to propose a PR which adds a test that acknowledges the current *de facto* behaviour that other tools might be relying on. It seems that `legacyArgs` is not properly tested. If we merge that one before this PR, then @YuviGold will need to account for those use cases so that CI is green here.

BTW, see the updated tests in #841.

@YuviGold, I think you, cli/cli and other projects are exploiting an unwanted feature in the codebase. If you want to "support subcommands checking for suggestions" (which is the title of this PR), my understanding is you might achieve it by setting `Args` properly. You should not need to modify `legacyArgs` at all.

Overall, projects relying on both the "previous" behaviour and on the "after" behaviour are doing it wrong by relying on `legacyArgs`. They should not care about this PR, but use the library appropriately instead of driving corner cases to their flavours.

spf13:
I hadn't considered the breaking change this would introduce. 

I disagree that changing undocumented behavior is ok. The code implicitly is the documentation. It's hard to say someone has done something wrong when it worked for them.

I agree with @marckhouzam 

> Maybe a backwards-compatible approach can be defined? At a minimum an opt-in solution would be backwards-compatible, but there may be something even better; I haven't looked into it myself.

I think we should consider how large the impact this is. If Helm is expecting the current behavior still and others are as well we can take 3 approaches. 

1. Make this functionality opt in.
2. Make this the default functionality but provide an opt in for folks wanting to use the legacy behavior (a simple path to working with the fix in place)
3. Wait a release, put some notice / warning in this release that this will change in the next release and since it's a bugfix, it's not a "new version" change.

umarcor:
FTR, I removed this PR from #1525 and I moved it to section "Delayed to a later release" in #1496.

jpmcb:
I agree with Steve - this would be an unexpected breaking change for users.

I'm ok putting a "notice" in and upcoming release of breaking changes upcoming in a 2.0.0 cobra release. I'm not a _huge_ fan of letting lingering tech debt exist next to new implementation (a.k.a, the opt in / opt out approach). We've already done this with the shell completions opt in / opt out and it's a bit painful to maintain.

jpmcb:
Hi @YuviGold - I think this is still valid (and worth discussing). Can you recreate this against the `main` branch since `master` has been deleted?

PR #1503, opened 2021-10-11
Support add flag ignoring conflicts, support category

robberphex:
1. Add func TryAddFlag{,Set}, which add flag{Set} and ignore conflicts. And parseFlag in every level command, not merge flag and parse.
2. Support command category

### DIFF
--- a/command.go
+++ b/command.go
@@ -584,6 +584,9 @@ Loop:
 		case s == "--":
 			// "--" terminates the flags
 			break Loop
+		case strings.HasPrefix(s, "--") && strings.Contains(s, "="):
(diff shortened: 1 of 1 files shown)


marckhouzam:
Hi @robberphex thanks for the PR.
To make it easier to understand can you give more details about this change? Maybe give some examples of behavioural changes, and mention why it seems to be important for `nerdctl`.  Thanks

AkihiroSuda:
@robberphex Could you update the PR text and the commit messages to be more descriptive, so that we can get your changes in the upstream cobra ASAP? Thanks

PR #1504, opened 2021-10-13
Handle requiredFlags errors in the same way as errors originating from parsing flags.

nclaeys:
At the moment there is a difference between the error messages + output when a required argument is not filled in compared to when an unexisting flag is passed.

I validate this by passing in a custom flagErrorFunc which prints both the error and the usage.
This error function is correctly used when the flag parsing fails, but it is not taken into account when validating required arguments fails.

### DIFF
--- a/command.go
+++ b/command.go
@@ -850,7 +850,7 @@ func (c *Command) execute(a []string) (err error) {
 	}
 
 	if err := c.validateRequiredFlags(); err != nil {
-		return err
+		return c.FlagErrorFunc()(c, err)
 	}
 	if c.RunE != nil {
 		if err := c.RunE(c, argWoFlags); err != nil {
--- a/command_test.go
+++ b/command_test.go
@@ -3,6 +3,7 @@ package cobra
 import (
 	"bytes"
 	"context"
+	"errors"
 	"fmt"
 	"io/ioutil"
 	"os"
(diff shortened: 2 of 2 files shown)


marckhouzam:
Thanks for the contribution @nclaeys.

I want to make sure we don't have a backwards-compatibility issue here.
Could you provide an example of the bad behavior and of the improved behavior so that I can be sure I understand what you are aiming to fix?

Thanks

nclaeys:
Hi @marckhouzam,

So the goal is that when you specify a custom flagErrorFunc it is always used when flag validation fails.
There are two code paths that result in an error due to flag validation: 
- a flag is used that does not exists (uses the errorFunc to print the error)
- a required flag is not set (does not use the errorFunc)
Example usage at the moment when you specify a custom error function (in our case it contains: usage string + error):
1. When not specifying a required flag
Output:
```
required flag(s) "name" not set
exit status 1
```
2. When specifying a flag that does not exist:
Output:
```
go run main.go environment new --test bla
Usage:
  datafy environment create [flags]

Aliases:
  create, new

Flags:
      --name string              The name of the environment to create

Global Flags:
      --debug                        Show debug output

unknown flag: --test
failed flag parsing
exit status 1
```
This PR should change the first behavior such that it becomes:
Output:
```
go run main.go environment new

Usage:
  datafy environment create [flags]

Aliases:
  create, new

Flags:
      --name string              The name of the environment to create

Global Flags:
      --debug                        Show debug output

required flag(s) "name" not set
failed flag parsing
exit status 1
```
Note: I updated the test to be more explicit, hopefully this helps you in understanding what I am saying.

pierre-borckmans:
Any update on this?

PR #1509, opened 2021-10-19
Support different bash completion options

### LINKED ISSUE #1508: Bash autocompletion adds command description to completed line
This is pretty much #1443 except I think I know how to reproduce it. I'm using `podman` as the program because I was told (containers/podman#12005) it uses cobra.

1. Add `TAB: menu-complete` to `.inputrc`.
2. Open `bash` and type `podman --c<TAB>`
3. Note that `podman --cgroup-manager  (Cgroup manager to use ("cgroupfs"|"systemd"))` is added to the current line.
4. Press enter and get ``bash: syntax error near unexpected token `('``

It is fixed with `--no-descriptions`, which I don't mind using but I just wanted to make sure the bug is chronicled.

marckhouzam:
Fixes #1508 

Bash completion V2 adds descriptions to completions.  However, with certain options, the description must be removed.  For example the `menu-complete/menu-complete-backward` bash option will immediately insert on the command-line the first completion returned, therefore we must remove the description to avoid it being added on the command-line also.

Based on the documentation found here
https://www.gnu.org/software/bash/manual/html_node/Commands-For-Completion.html
we remove descriptions for the cases:
- menu-complete
- menu-complete-backward
- insert-completions

### DIFF
--- a/bash_completionsV2.go
+++ b/bash_completionsV2.go
@@ -138,11 +138,40 @@ __%[1]s_process_completion_results() {
             _filedir -d
         fi
     else
-        __%[1]s_handle_standard_completion_case
+        __%[1]s_handle_completion_types
     fi
 
     __%[1]s_handle_special_char "$cur" :
     __%[1]s_handle_special_char "$cur" =
+}
+
+__%[1]s_handle_completion_types() {
+    __%[1]s_debug "__%[1]s_handle_completion_types: COMP_TYPE is $COMP_TYPE"
+
+    case $COMP_TYPE in
+    37|42)
+        # Type: menu-complete/menu-complete-backward and insert-completions
+        # If the user requested inserting one completion at a time, or all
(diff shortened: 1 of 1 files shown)


Luap99:
Do you think it is possible to add a test in your completion-testing repo?

marckhouzam:
> Do you think it is possible to add a test in your completion-testing repo?

I've done a PR to add tests: https://github.com/marckhouzam/cobra-completion-testing/pull/9

jpmcb (approved):
Seems good! @marckhouzam - I'm guessing in marckhouzam/cobra-completion-testing we need this merged before we test?

marckhouzam:
> Seems good! @marckhouzam - I'm guessing in marckhouzam/cobra-completion-testing we need this merged before we test?

Yes the cobra-completion-testing already had a test for this which was failing until this merge.

Thanks for merging @jpmcb !

PR #1510, opened 2021-10-21
Fix invalid shell completion when used with ~/.cobra.yaml

darklore:
`cobra completion` outputs invalid output "Using config file:" at the top of the completion script. when ~/.cobra.yaml exists.


marckhouzam:
@darklore thanks for the PR.
Could you describe the steps to reproduce the problem so we can better understand?
Also, a test would be nice to avoid this breaking again.

jpmcb (approved):
Good catch! 👍

PR #1513, opened 2021-10-23
'go get' to 'go install'

opxyc:
Installing using `go get -u` is depricated. So, updated it in cobra/README to `go install`


marckhouzam:
Sorry for the late reply @opxyc but thank you for your contribution.
The cobra generator has been moved to github.com/spf13/cobra-cli and the README was adjusted and now includes the `go install` instruction.

PR #1520, opened 2021-10-27
Remove trailing spaces from bash completion command description

meyermarcel:


stmcginnis (approved):
Aligns with opening tick, and I don't see why any extra space in the string would be needed.

jpmcb (approved):

PR #1523, opened 2021-11-03
Make the default completion command accessible

### LINKED ISSUE #1507: Hide auto-generated completion command
Is there a way to hide the  `completion` cmd?  I like having it, but don't want it to show up in --help

Thanks!

marckhouzam:
Fixes #1507 as requested by @jpmcb 

The ask of #1507 was to be able to make the default `completion` command hidden.
I may have gone too far with this PR so I'm looking for opinions 🤷‍♂️ 

This PR goes much beyond the ask and makes the `completion` command and it's four sub-commands fully accessible so the program can set whatever field it wants before calling `rootCmd.Execute()`.  For example, a program could call
```
cobra.CompletionCmd.Hidden = true  // To make the command hidden
```
or/and
```
cobra.CompletionCmd.Use = "shellcomplete" // To rename the command
```
or/and
```
cobra.PwshCompletionCmd.Use = "pwsh"  // To use 'pwsh' instead of 'powershell' as a sub-command
```

How the new logic works:

1- the default completion command and its 4 sub-commands are now public global variables
2- at initialization the fields of these global variables that don't need special info are initialized (cmd.Use, cmd.Short, cmd.Args, cmd.ValidArgsFunction)
3- during rootCmd.Execute() the function cmd.initDefaultCompletionCmd() is called like before and initializes the remaining fields (these fields needed the program name and root command: cmd.Long, cmd.RunE and the flag)

The magic is that between point 2 and point 3, a program can modify the fields of these commands since they are accessible.

If this brings in too much complexity, we could simply offer a new option in `CompletionOptions` to allow hiding the command.

I made the effort of trying to be as flexible as possible because we got PRs that needed to change the descriptions and such which were slightly off.  With this PR consumers should be able to adjust anything themselves.  But it is more complex.  And I had to do a little bit of resetting things so that go tests, which call `rootCmd.Execute()` multiple times would continue work; this is something we already do for the help command however, as we can see here: https://github.com/spf13/cobra/blob/f09e9476376ba2fa9b29ec321e2944f4bc14503c/command.go#L1114

Opinions please. 😄 

P.S. No tests added yet.


marckhouzam:
I should explain how the new logic works:

1- the default `completion` command and its 4 sub-commands are now public global variables
2- at initialization the fields of these global variables that don't need special info are initialized (`cmd.Use`, `cmd.Short`, `cmd.Args`, `cmd.ValidArgsFunction`)
3- during `rootCmd.Execute()` the function `cmd.initDefaultCompletionCmd()` is called like before and initializes the remaining fields (these fields needed the program name and root command: `cmd.Long`, `cmd.RunE` and the flag)

The magic is that between point 2 and point 3, a program can modify the fields of these commands since they are accessible.

marckhouzam:
This is turning out to be more complicated that it's worth.  It is causing problems with tests which would require Cobra to reset everything about the completion command between tests, which is a big hassle.

I'm abandoning this attempt and have replaced it with #1541 which simply adds a user option to allow hiding the default completion command.

PR #1524, opened 2021-11-04
Docs: Fix command about install cobra executable.

hanjm:
Hi,  I try to install `cobra` and find the installing command in `Readme.md` is wrong.


marckhouzam:
Thanks for the contribution @hanjm.  The Cobra generator has recently been moved to https://github.com/spf13/cobra-cli and the README for it was properly updated.

Therefore I'm going to close this PR as it is no longer needed.  But thank you for bringing this to our attention.

PR #1533, opened 2021-11-11
Completion command helptext capitalization

aidan-mundy:
The default completion command's helptext capitalization should match the default help command's helptext capitalization

Yes, this is entirely nitpicky, but it is bothering me

### DIFF
--- a/completions.go
+++ b/completions.go
@@ -599,7 +599,7 @@ func (c *Command) initDefaultCompletionCmd() {
 
 	completionCmd := &Command{
 		Use:   compCmdName,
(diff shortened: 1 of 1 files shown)


marckhouzam:
Duplicate of #1455

aidan-mundy:
:sob: I really hope that gets merged soon. This is way more annoying to me than it should be, had to disable the command for now :rofl:

marckhouzam:
> 😭 I really hope that gets merged soon. This is way more annoying to me than it should be, had to disable the command for now 🤣

I can understand.  It'll be fixed in the next release which should happen in a few weeks 🤞

PR #1547, opened 2021-12-01
Add backwards-compatibility tests for legacyArgs()

marckhouzam:
As suggested by @umarcor, these tests make sure we don't break backwards-compatibility with respect to the current behaviour of `legacyArgs()`.

See #1500 which would break the current behaviour of `legacyArgs()`.

### DIFF
--- a/args_test.go
+++ b/args_test.go
@@ -292,3 +292,32 @@ func TestMatchAll(t *testing.T) {
 		})
 	}
 }
+
+// This test make sure we keep backwards-compatibility with respect
+// to the legacyArgs() function.
+// It makes sure the root command accepts arguments if it does not have
+// sub-commands.
+func TestLegacyArgsRootAcceptsArgs(t *testing.T) {
+	rootCmd := &Command{Use: "root", Args: nil, Run: emptyRun}
+
+	_, err := executeCommand(rootCmd, "somearg")
+	if err != nil {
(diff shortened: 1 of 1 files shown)


umarcor (commented):
I suggest to also replace https://github.com/spf13/cobra/blob/9e1d6f1c2aa8df64b6a6ba39e92517f68580d653/command.go#L1000 with `return ArbitraryArgs(c, args)`, along with adding the following sentence to the user guide (as done in #841): "*If `Args` is undefined or `nil`, it defaults to `ArbitraryArgs`*".

marckhouzam:
> I suggest to also replace
> 
> https://github.com/spf13/cobra/blob/9e1d6f1c2aa8df64b6a6ba39e92517f68580d653/command.go#L1000
> 
> with `return ArbitraryArgs(c, args)`, along with adding the following sentence to the user guide (as done in #841): "_If `Args` is undefined or `nil`, it defaults to `ArbitraryArgs`_".

I suggest doing this in another PR to keep this one well focused.

marckhouzam:
@jpmcb @johnSchnake This simply adds two new tests to protect us against possible regressions.

marckhouzam:
@jpmcb This is just a couple of extra tests to protect us against regressions.

jpmcb (approved):
Awesome! LGTM, let's do it

PR #1550, opened 2021-12-02
Fix unexpected executing subcommand

JaySon-Huang:
I've defined a subcommand with two boolean arguments and found it unexpectedly run into the subcommand.
In specific, a subcommand naming "sub" with two boolean arguments "a" and "b". If I parse the command with arguments " --a sub" or "--b sub", it returns an error "unknown argument" as expected. But when I parse the command with arguments "--a --b sub", it runs into the "sub" command with those two boolean arguments set.

I trace down the codes in cobra and find that `stripFlags` always assume `'--flag arg'` or `-f arg`, ignoring the pattern of boolean flag. So I file a PR trying to fix this unexpected behavior.

### DIFF
--- a/command.go
+++ b/command.go
@@ -585,11 +585,12 @@ Loop:
 		case strings.HasPrefix(s, "-") && !strings.Contains(s, "=") && len(s) == 2 && !shortHasNoOptDefVal(s[1:], flags):
 			// If '-f arg' then
 			// delete 'arg' from args or break the loop if len(args) <= 1.
-			if len(args) <= 1 {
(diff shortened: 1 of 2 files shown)


marckhouzam:
Thanks @JaySon-Huang! I can reproduce the problem. I haven't looked at the fix yet, but we will need to be very careful not to break anything.

JaySon-Huang:
@marckhouzam sure. I add some comments on the code to make my assumptions more clear. Feel free to leave any comments when you are free.

PR #1551, opened 2021-12-06
Add Command.SetContext

joshcarp:
Basically the same as https://github.com/spf13/cobra/pull/1517 but uses the `Set<field>` naming convention instead of `WithContext` 

Context setting without execution is important because it means that more design patterns can be achieved.

Currently I am using functional options in a project and I can add behaviour through functional options as such:
```go
type GrptlOption func(*cobra.Command) error

func WithFileDescriptors(descriptors ...protoreflect.FileDescriptor) GrptlOption {
	return func(cmd *cobra.Command) error {
		err := CommandFromFileDescriptors(cmd, descriptors...)
		if err != nil {
			return err
		}
		return nil
	}
}
```

I've got a lot more options and this pattern allows me to have nice abstracted pieces of logic that interact with the cobra command. 

This Pattern also allows for adding extra information to a call through `PreRun` functions:

```go

cmd := &cobra.Command{
		Use: "Foobar",
		PersistentPreRun: func(cmd *cobra.Command, args []string) {
			err :=cmd.SetContext(metadata.AppendToOutgoingContext(context.Background(), "pre", "run"))
		},
	}
```

This is a veer nice abstraction and allows for these functions to be very modular

The issue I'm facing at the moment is that I can't write a nifty option (something like `WithAuthentication`) because that needs access to reading and setting the context. Currently I can only read the context.

Needing to use `ExecuteContext` breaks this abstraction because I need to run it right at the end.

### DIFF
--- a/command.go
+++ b/command.go
@@ -230,6 +230,12 @@ func (c *Command) Context() context.Context {
 	return c.ctx
 }
 
+// SetContext sets context for the command. It is set to context.Background by default and will be overwritten by
+// Command.ExecuteContext or Command.ExecuteContextC
+func (c *Command) SetContext(ctx context.Context) {
+	c.ctx = ctx
+}
+
 // SetArgs sets arguments for the command. It is set to os.Args[1:] by default, if desired, can be overridden
 // particularly useful when testing.
 func (c *Command) SetArgs(a []string) {
--- a/command_test.go
+++ b/command_test.go
@@ -2058,3 +2058,101 @@ func TestFParseErrWhitelistSiblingCommand(t *testing.T) {
 	}
 	checkStringContains(t, output, "unknown flag: --unknown")
 }
+
+func TestSetContext(t *testing.T) {
+	key, val := "foo", "bar"
+	root := &Command{
+		Use: "root",
+		Run: func(cmd *Command, args []string) {
+			key := cmd.Context().Value(key)
+			got, ok := key.(string)
+			if !ok {
+				t.Error("key not found in context")
+			}
+			if got != val {
+				t.Errorf("Expected value: \n %v\nGot:\n %v\n", val, got)
+			}
+		},
+	}
+	ctx := context.WithValue(context.Background(), key, val)
+	root.SetContext(ctx)
+	err := root.Execute()
+	if err != nil {
+		t.Error(err)
+	}
+}
+
+func TestSetContextPreRun(t *testing.T) {
+	key, val := "foo", "bar"
+	root := &Command{
+		Use: "root",
+		PreRun: func(cmd *Command, args []string) {
+			ctx := context.WithValue(cmd.Context(), key, val)
+			cmd.SetContext(ctx)
+		},
+		Run: func(cmd *Command, args []string) {
+			key := cmd.Context().Value(key)
+			got, ok := key.(string)
+			if !ok {
+				t.Error("key not found in context")
+			}
+			if got != val {
+				t.Errorf("Expected value: \n %v\nGot:\n %v\n", val, got)
+			}
+		},
+	}
+	err := root.Execute()
+	if err != nil {
+		t.Error(err)
+	}
+}
+
+func TestSetContextPreRunOverwrite(t *testing.T) {
+	key, val := "foo", "bar"
+	root := &Command{
+		Use: "root",
+		Run: func(cmd *Command, args []string) {
+			key := cmd.Context().Value(key)
+			_, ok := key.(string)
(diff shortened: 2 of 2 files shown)


jpmcb:
Awesome! Thanks for this contribution. I'd love some community feedback as well.

And is it possible you could write some tests for this?

I think this would be great for a v2.0 - will hold till we get some more feedback

joshcarp:
Hey, added some tests; the main use cases I see are:
- Setting context before a run (as an alternative to cmd.ExecuteContext())
    - This would allow packages to add information to the context using functional options for example
- Setting the context through a PreRun or PersistentPreRun hook

I've put tests for both these cases, also one to ensure that `ExecuteContext` overwrites any `SetContext` that was called before

jackzampolin:
Would love to see this merged soon. Would make parsing config in `PersistentPreRunE` and setting the values on context much easier!!

alexykot:
I'd like to have this merged asap as well. 

I've started designing my CLI tool and intended to use the context to populate the core data needed by all subcommands, only to find out that there is no legit way right now to update the context after it was passed into the root `ExecuteContext()` call. 

I intend to use it exactly as [the second use case](https://github.com/spf13/cobra/pull/1551#issuecomment-989147556) above describes - parse and load additional data in the `PersistentPreRun` hook, add it to the context, and pass this enriched context along for all downstream subcommands to use.

It's not possible to do that today, so for now I'll have to do all loading before handing off the execution to cobra. I hope this will be merged quickly and I'll be able to use a more elegant way to load data when appropriate rather than fully upfront.

marckhouzam:
@johnSchnake WDYT?

johnSchnake:
I like this PR as its so narrowly scoped. Just needs to fix up the golangci-lint issues and I think this should merge.

joshcarp:
To respond to the question of "why is this needed?":

Currently to pass around information from pre run hooks a global variable needs to be declared:

```go
func TestFoo(t *testing.T){
	globalVar := "default"
	root := &Command{
		Use: "root",
		PreRun: func(cmd *Command, args []string) {
			globalVar = "not default"
		},
			fmt.Println(globalVar)
		},
	}
	root.ExecuteContext(context.Background())
}
```

To avoid using global variables, and to allow for cleaner implementation we can use `SetContext` instead:

```go
func TestFoo(t *testing.T){
	globalVar := "default"
	root := &Command{
		Use: "root",
		PreRun: func(cmd *Command, args []string) {
			ctx := context.WithValue(cmd.Context(), key{}, val)
			cmd.SetContext(ctx)
		},
			fmt.Println(cmd.Context().Value(key{}))
		},
	}
        ctx := context.WithValue(cmd.Context(), "key", "default")
	root.ExecuteContext(ctx)
}
```


This greatly simplifies the readability and maintainability of code because everything is stored within the `context.Context` and no reference to external global variables is made.

I have already simulated this by using a mutable pointer that can be found here: https://github.com/joshcarp/grpctl/blob/main/context.go

I use it to set Authorisation headers which would be a nightmare if I needed to make implementations refer to global variables in my package.

marckhouzam (approved):
LGTM

The tests are great and give a good understanding of what will now be possible.

marckhouzam:
Based on the fact that @johnSchnake also thinks this is good (https://github.com/spf13/cobra/pull/1551#issuecomment-1068740745), I will merge.

joshcarp:
@marckhouzam  thanks,
Also I forgot to link the issue so this can be closed now: https://github.com/spf13/cobra/issues/563

marckhouzam:
> @marckhouzam thanks! Also I forgot to link the issue so this can be closed now: #563

Thanks for following up @joshcarp and thanks for this contribution, it seems it will bring some elegance to the usage of Cobra!

ashleydavies:
Sorry, this isn't related to this PR directly, so I worry it's an undesirable comment, but hopefully you don't mind too much..!

I really wanted to say that the way y'all manage milestones in this repository (I can see above: "@marckhouzam added this to the 1.5.0 milestone") is tremendously helpful and appreciated. I'm so used to digging into tags and digging around for a commit hash, figuring out what version I needed for this feature was so refreshing.

Little things like that make it so much easier to consume a project, and I imagine sometimes must feel like toil or busywork, so wanted to just say it's really helpful 😄 thank you all!

marckhouzam:
Well, we never mind such kind comments ❤️ 
Thanks @ashleydavies !
Credit goes to @jpmcb for setting things up and to @umarcor for keeping use disciplined 👏

PR #1463, opened 2021-07-25
fix: Duplicate error message from cobra init boilerplates

ANGkeith:
closes #1415

PR #1552, opened 2021-12-07
fix root.go.golden

umarcor:
It seems that merging #1463 broke CI. This PR hopefully fixes it.

/cc @jpmcb


jpmcb (approved):
Good call - I always forget the golden files 🏅

PR #1559, opened 2021-12-11
Add default completion command even if there are no other sub-commands

marckhouzam:
This supersedes #1450 and #1392

When a program has no sub-commands, its root command can accept arguments.  If we add the default "completion" command to such programs they will now have a sub-command and will no longer accept arguments.

What we do instead for this special case, is only add the "completion" command if it is being called.

We want to have the "completion" command for such programs because it will allow the completion of flags and of arguments (if provided by the program).

@scop can you confirm this is what you were hoping for?

**Testing done**

```
# Here is a program with no sub-cmds
$ cat tests/minimal.go
package main

import (
	"fmt"

	"github.com/spf13/cobra"
)

var rootCmd = &cobra.Command{
	Use: "prog",
		fmt.Println("value:", value)
		fmt.Println("args:", args)
	},
}

var value string

func main() {
	rootCmd.Flags().StringVarP(&value, "value", "e", "", "value")
	rootCmd.Execute()
}

$ go build -o minimal ./tests/minimal.go

$ ./minimal
value:
args: []
$ ./minimal arg1
value:
args: [arg1]
$ ./minimal arg1 arg2
value:
args: [arg1 arg2]
$ ./minimal arg1 arg2 --value 1
value: 1
args: [arg1 arg2]

# Notice no 'completion' command visible
$ ./minimal -h
Usage:
  prog [flags]

Flags:
  -h, --help           help for prog
  -e, --value string   value

# But the completion command does exist
$ ./minimal completion -h
Generate the autocompletion script for prog for the specified shell.
See each sub-command's help for details on how to use the generated script.

Usage:
  prog completion [command]

Available Commands:
  bash        Generate the autocompletion script for bash
  fish        Generate the autocompletion script for fish
  powershell  Generate the autocompletion script for powershell
  zsh         Generate the autocompletion script for zsh

Flags:
  -h, --help   help for completion

Use "prog completion [command] --help" for more information about a command.

# Shell completion for the 'completion' command itself is not triggered,
# but it is triggered for sub-commands of 'completion'
$ ./minimal __complete compl
:0
$ ./minimal __complete completion ''
bash	Generate the autocompletion script for bash
fish	Generate the autocompletion script for fish
powershell	Generate the autocompletion script for powershell
zsh	Generate the autocompletion script for zsh
:4
```

### DIFF
--- a/completions.go
+++ b/completions.go
@@ -711,8 +711,8 @@ func checkIfFlagCompletion(finalCmd *Command, args []string, lastArg string) (*p
 // 1- the feature has been explicitly disabled by the program,
 // 2- c has no subcommands (to avoid creating one),
 // 3- c already has a 'completion' command provided by the program.
-func (c *Command) InitDefaultCompletionCmd() {
-	if c.CompletionOptions.DisableDefaultCmd || !c.HasSubCommands() {
+func (c *Command) InitDefaultCompletionCmd(args ...string) {
+	if c.CompletionOptions.DisableDefaultCmd {
 		return
 	}
 
@@ -724,6 +724,16 @@ func (c *Command) InitDefaultCompletionCmd() {
 	}
 
 	haveNoDescFlag := !c.CompletionOptions.DisableNoDescFlag && !c.CompletionOptions.DisableDescriptions
+
+	// Special case to know if there are sub-commands or not.
+	hasSubCommands := false
+	for _, cmd := range c.commands {
+		if cmd.Name() != ShellCompRequestCmd && cmd.Name() != helpCommandName {
+			// We found a real sub-command (not 'help' or '__complete')
+			hasSubCommands = true
+			break
+		}
+	}
 
 	completionCmd := &Command{
 		Use:   compCmdName,
@@ -737,6 +747,22 @@ See each sub-command's help for details on how to use the generated script.
 		GroupID:           c.completionCommandGroupID,
 	}
 	c.AddCommand(completionCmd)
+
+	if !hasSubCommands {
+		// If the 'completion' command will be the only sub-command,
+		// we only create it if it is actually being called.
+		// This avoids breaking programs that would suddenly find themselves with
+		// a subcommand, which would prevent them from accepting arguments.
+		// We also create the 'completion' command if the user is triggering
+		// shell completion for it (prog __complete completion '')
+		subCmd, cmdArgs, err := c.Find(args)
+		if err != nil || subCmd.Name() != compCmdName &&
+			!(subCmd.Name() == ShellCompRequestCmd && len(cmdArgs) > 1 && cmdArgs[0] == compCmdName) {
+			// The completion command is not being called or being completed so we remove it.
+			c.RemoveCommand(completionCmd)
+			return
+		}
+	}
 
 	out := c.OutOrStdout()
 	noDesc := c.CompletionOptions.DisableDescriptions
--- a/command.go
+++ b/command.go
@@ -1097,12 +1097,6 @@ func (c *Command) ExecuteC() (cmd *Command, err error) {
 
 	// initialize help at the last point to allow for user overriding
 	c.InitDefaultHelpCmd()
-	// initialize completion at the last point to allow for user overriding
-	c.InitDefaultCompletionCmd()
-
-	// Now that all commands have been created, let's make sure all groups
-	// are properly created also
-	c.checkCommandGroups()
 
 	args := c.args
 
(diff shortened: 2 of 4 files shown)


scop:
Yep, seems to work nicely, thanks!

An observation is that as the `completion` command is hidden, no completions are generated for its arguments (i.e. "foo completion <kbd>Tab</kbd>"), but that's such a minor detail and not a result of this change I believe, that let's not get stuck with that :)

marckhouzam:
> An observation is that as the `completion` command is hidden, no completions are generated for its arguments (i.e. "foo completion Tab")

Nice attention to detail! The completions should work for hidden commands.  In this case however, when we call the "__complete" command to get the completions, we don't create the "completion" command so its subcommands don't get completed.

Let me try to see if I can fix it.

marckhouzam:
While working on this I ran into another completion bug for programs that don't have sub-commands.
I've opened issue #1562 and posted a fix in #1563.

I'm still looking at fixing the issue @scop reported above.

caarlos0:
This looks good, was just about to dig into this myself.

Thanks!

marckhouzam:
I've let this rot for too long. Let's plan it for the next release. Come August iI will run some tests on it and make sure it is good.

caarlos0:
anything I can do to help getting this merged?

marckhouzam:
I've finally re-worked this and I believe it is ready.
Please refer to the description of the PR to see the testing that I did, which shows the new behaviour.

marckhouzam:
@jpmcb When you have moment, this is ready for review.  Don't hesitate to ask for clarifications.

completions.go
```diff
@@ -711,8 +711,8 @@ func checkIfFlagCompletion(finalCmd *Command, args []string, lastArg string) (*p
 // 1- the feature has been explicitly disabled by the program,
 // 2- c has no subcommands (to avoid creating one),
 // 3- c already has a 'completion' command provided by the program.
-func (c *Command) InitDefaultCompletionCmd() {
-	if c.CompletionOptions.DisableDefaultCmd || !c.HasSubCommands() {
+func (c *Command) InitDefaultCompletionCmd(args []string) {
```
jpmcb, line 714: Just wondering what's the impact of this now taking a slice of args? Can this be made optional to still be backwards compatible for users who've adopted `InitDefaultCompletionCmd`? If it's fixing a bug / undefined behavior, that should be fine!

jpmcb (commented):
Closes: https://github.com/spf13/cobra/pull/1450 https://github.com/spf13/cobra/pull/1392

---

So looks like this was unintended behavior where the default completions were only available when some other commands were provided?

marckhouzam:
> So looks like this was unintended behavior where the default completions were only available when some other commands were provided?

It’s the default “completion” _command_ that doesn’t get created unless other commands exist. 

Say you write a tool called `tooly`. If `tooly` has any sub commands (e.g., `tooly list`) then cobra will automatically create `tooly completion` which allow users to generate the shell completion scripts. 

This PR makes this happen for tools that don’t have any sub commands

completions.go
marckhouzam: Oh good point. I don’t recall why disfunction was ever exported. But since cobra does not need to be exported, it must mean that we did that for other people to call it. So good call that this is a breaking change.

we could simply make the argument an ellipsis to avoid breaking compilation. However, I feel this would be worse for programs that don’t have sub commands because we will break their behaviour because this completion command will now appear unless they pass the arguments properly.

I could create a new function instead I guess. I’ll give it some thought and post an update

Thanks for catching this

jpmcb: I found a few cases where this would break people's use of `InitDefaultCompletionCmd`. For example, for Charm, looks like they call this explicitly to correct for this exact issue:

https://github.com/charmbracelet/mods/blob/b037acfe73f811839bbda8e7bdd5f711f4df9799/main.go#L334-L342

I think the right approach is to just make the args "variadic": 

```diff

-func (c *Command) InitDefaultCompletionCmd(args []string) {
+func (c *Command) InitDefaultCompletionCmd(args ...string) {
+       if args == nil {
+               args = []string{}
+       }
```

Then, if someone has a slice of args, they can splat them into the function with `...`:

```diff
-       c.InitDefaultCompletionCmd(args)
+       c.InitDefaultCompletionCmd(args...)
```

And calling it with no args still works, like for Charm's case:

```go
rootCmd.InitDefaultCompletionCmd()
```

marckhouzam: Sounds good, thanks for finding the Charms example. I’ll rework this

marckhouzam:
@jpmcb This is ready now.  Thanks again for catching the breaking change.

jpmcb (approved):

PR #1574, opened 2021-12-29
Introduce FixedCompletions

emersion:
Example usage:

    choices := []string{"choice1", "choice2", "choice3"}
    cmd.RegisterFlagCompletionFunc(cobra.FixedCompletions(choices, ShellCompDirectiveNoFileComp))

### DIFF
--- a/completions.go
+++ b/completions.go
@@ -103,6 +103,14 @@ func NoFileCompletions(cmd *Command, args []string, toComplete string) ([]string
 	return nil, ShellCompDirectiveNoFileComp
 }
 
+// FixedCompletions can be used to create a completion function which always
+// returns the same results.
+func FixedCompletions(choices []string, directive ShellCompDirective) func(cmd *Command, args []string, toComplete string) ([]string, ShellCompDirective) {
+	return func(cmd *Command, args []string, toComplete string) ([]string, ShellCompDirective) {
+		return choices, directive
+	}
+}
+
 // RegisterFlagCompletionFunc should be called to register a function to provide completion for a flag.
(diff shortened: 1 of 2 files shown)


marckhouzam:
This does seem like a pattern that will come up for different programs.  Thanks @emersion!

Could you add a simple test?

A note for completeness. We should normally filter `choices` on `toComplete` before returning those choices.  However, I think avoiding such a filter will allow `zsh` and `fish` to do fuzzy matching; this is an improvement I've been thinking about for a while now, so let's get the ball rolling with your contribution and leave it without such a filter.

emersion:
> Could you add a simple test?

Done!

> A note for completeness. We should normally filter choices on toComplete before returning those choices. However, I think avoiding such a filter will allow zsh and fish to do fuzzy matching; this is an improvement I've been thinking about for a while now, so let's get the ball rolling with your contribution and leave it without such a filter.

That's a good point indeed.

marckhouzam (approved):
@Luap99 do you agree this would be useful for consumer?

Luap99 (commented):
LGTM

emersion:
Thanks for the review! Any chance to get this merged?

marckhouzam:
@jpmcb what do you think if this small addition of a convenience function for shell completions?  It is more flexible than using `ValidArgs` because it can also specify the `ShellCompDirective`.

marckhouzam:
@jpmcb are you ok with this small addition?

jpmcb (approved):
LGTM! Thanks again for the re-ping; apologies for the slow responses

PR #1571, opened 2021-12-23
fix: replace viper with useViper in cobra/README.md' example .cobra.yaml

acehinnnqru:
I found the example .cobra.yaml's config "viper" is not working; I think it is useViper after I have read the source code.

PR #1579, opened 2022-01-12
Update README.md for .cobra.yaml

saltbo:
viper is invalid for init, I found the code is useViper, should we update the Readme?


saltbo:
repeated with #1571

PR #1582, opened 2022-01-19
Command.Context() returns nil instead of context.Background()

### LINKED ISSUE #1578: Command.Context() returns nil instead of context.Background()
Documentation of the [Context() function](https://pkg.go.dev/github.com/spf13/cobra#Command.Context) states a `context.Background()` is returned, but the returned value is `nil`.

> Context returns underlying command context. If command wasn't executed with ExecuteContext Context **returns Background context**. 

```go
cmd := &cobra.Command{}
fmt.Printf("context is %v", cmd.Context()) // returns "context is <nil>"
```
[Go Playground](https://go.dev/play/p/C8jaQ-uQgEv)

It would be really nice if this could return a `context.Background()`, so functions that take a `Command` as parameter and use the context from that command could be tested. There is also no possibility to set this context manually right now.

andreaangiolillo:
Closes #1578

### DIFF
--- a/command.go
+++ b/command.go
@@ -227,6 +227,9 @@ type Command struct {
 // Context returns underlying command context. If command wasn't
 // executed with ExecuteContext Context returns Background context.
 func (c *Command) Context() context.Context {
+	if c.ctx == nil {
+		c.ctx = context.Background()
+	}
 	return c.ctx
 }
 
--- a/command_test.go
+++ b/command_test.go
@@ -2058,3 +2058,10 @@ func TestFParseErrWhitelistSiblingCommand(t *testing.T) {
 	}
 	checkStringContains(t, output, "unknown flag: --unknown")
 }
+
+func TestContext(t *testing.T) {
+	root := &Command{}
+	if root.Context() == nil {
+		t.Error("expected root.Context() != nil")
+	}
+}


johnSchnake:
Looks good to me 👍 Could you squash the commits down before I merge though?

marckhouzam:
Out of curiosity, what specific scenario does this address?
I'm asking because the context is set to `context.Background()` once `Execute()` is called:
https://github.com/spf13/cobra/blob/6d2dc43606da38b673801b8f678c7e5143dfad58/command.go#L915-L918

katexochen:
@marckhouzam I often use package injection in a command's Run function `foo`, which then passes the command to another function `bar`. In unit test, I just call func `bar`, pass a new command, and don't call Execute. If function `bar` uses a context from the command, it panics during the test.
```go
func newFooCommand() *cobra.Command {
	return &cobra.Command{ RunE: foo }
}

func foo(cmd *cobra.Command, args []string) error {
	fs := afero.NewOsFs()
	return bar(cmd, args, fs)
}

func bar(cmd *cobra.Command, args []string, fs afero.Fs) error {

	// use cmd.Context

	return nil
}

func TestFoo(t *testing.T) {
	cmd := newFooCommand()
	fs := afero.NewMemMapFs()
	assert.NoError(t, bar(cmd, []string{}, fs))
}
```

katexochen:
> do we still need to set the context in Context()?

From my perspective, #1551 should at least solve the issue I've described above (https://github.com/spf13/cobra/pull/1582#issuecomment-1054240073). However, if this PR isn't merged, the doc comment of the Command.Context  method should be updated and state that nil is returned if no context has been set.

marckhouzam:
After discussion we felt that there was more flexibility and less risk in allowing a `nil` context.

And now that we have a SetContext() function thanks to #1551 users can call `SetContext()` to set the context to `Background()` without having to call `ExecuteWithContext()`.

The GoDoc was adjusted by #1639.

I'm therefore going to close this but if you still feel something is missing, feel free to re-open.

PR #1584, opened 2022-01-21
Fix ambiguous tags: use 'viper' and remove 'useViper'

atompi:
This approach is more appropriate than just modifying the README. Contrast: #1571

### DIFF
--- a/cobra/cmd/add_test.go
+++ b/cobra/cmd/add_test.go
@@ -9,7 +9,7 @@ import (
 )
 
 func TestGoldenAddCmd(t *testing.T) {
-	viper.Set("useViper", true)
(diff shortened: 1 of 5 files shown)


acehinnnqru:
Maybe they have reasons to use 'useViper'?

marckhouzam:
I also like `viper` as a key for the configuration file.  However, from what I can see from the git history, `useViper` was the name that was used from the very beginning.  Therefore, for backwards-compatibility reasons, we should stay with `useViper`.  There may be configuration files out there that use the correct `useViper` key (after having figured out the documentation was wrong).

I will therefore close this in favour of #1571

PR #1608, opened 2022-02-22
feat: Add how to load completions in your current zsh session

ondrejsika:


marckhouzam (approved):
I like this!
Sourcing the script is not the recommended way of doing things for `zsh` so Cobra does not support it out of the box, but documenting how to make it work for a single enabling, sounds like a good compromise.

@jharshman are you comfortable with this?

jpmcb (approved):
This looks good to me!

marckhouzam:
We have received multiple questions about not being able to source the `zsh` completion script.
Let's go ahead with this PR, as it will help the community

Issue #1603, opened 2022-02-17
Labels, Milestones, Triaging ⚡


jpmcb:
# Labels, Milestones, Triaging ⚡

The following is a living issue, pinned to the top of `cobra`'s GitHub issues for wide visibility, that represents the current labeling, triaging, milestone, and maintenance state of `cobra`. Maintainers, contributors, and users may refer to this in order to understand how bugs and features are being triaged and where their feature request may be in the maintenance cycle.

Inspired by projects like [`kubernetes-sigs/cluster-api`](https://github.com/kubernetes-sigs/cluster-api) and [`helm/helm`](https://github.com/helm/helm)

#### `kind/` - The type of issue
- `kind/bug`: A bug in `cobra`; _unintended_ behavior
- `kind/feature`: A feature request for `cobra`; _new_ or _enhanced_ behavior
- `kind/documentation`: Documentation of `cobra` _itself_
- `kind/testing`: CI/CD, testing, etc. for `cobra`. Usually also gets the github label
- `kind/rel-eng`: Related to tagging and releasing `cobra`
- `kind/upstream`: Go modules `cobra` _depends on_
- `kind/downstream`: Related to projects/libraries that depend _on_ `cobra` (like `helm` or `kubernetes`)
- `kind/security`: Related to security concerns in `cobra` itself. Refer to the security policy before opening *_any_* public issue.
- `kind/cleanup`: General cleanup of code, issues, etc.
- `kind/deprecation`: Related to a feature or part of code being deprecated
- `kind/support`: Questions, supporting users, etc.

#### `area/` - The area of work that needs to be executed against
- `area/cobra-command`: Core `cobra.Command` implementations
- `area/shell-completion`: All shell completions
- `area/docs-generation`: Generation of docs
- `area/flags-args`: Changes to functionality around command line flags and args
- `area/go`: General Go, like `go.mod` or `go.sum` changes
- `area/github`: For changes to Github specific things _not_ shipped in the library (like maintainers files, actions, etc)
- `area/lib`: Catch all for changes in the cobra library

#### `triage/` - The state of an issue being triaged
- `triage/needs-triage`: Needs to be placed into a milestone, or be closed by maintainers. This label is removed once placed into a milestone.
- `triage/blocked`: Cannot move forward until some roadblock is lifted
- `triage/needs-info`: An issue that needs more investigation from maintainers or more info from the issue provider
- `triage/duplicate`: A duplicate issue. These issues are usually closed after receiving this label.

[...]

PR #1610, opened 2022-02-23
Adds size labeler GitHub action

jpmcb:
Adds the size labeler GitHub action.

This helps maintainers eyeball how big a PR is and how much effort should go into reviewing it.

Please see https://github.com/spf13/cobra/issues/1603 for more info


marckhouzam:
For helm, it seemed that there was an issue using the GH action because the token didn't have permissions to modify PRs coming from forks.  But I haven't tried it myself.

I've found this issue that seems to confirm the problem although it also hints that using the `pull_request_target` may be the solution: https://github.com/CodelyTV/pr-size-labeler/issues/8

We may need to test it using a fork from someone that does not have write permissions on Cobra.

jpmcb:
Ah yes! Good call - I found an example [that seems to be working](https://github.com/jina-ai/dashboard/blob/326655d43f5b31dd8ae1207bed36db378bc05a17/.github/workflows/label-pr.yml) that uses `pull_request_target`

Updated the PR

marckhouzam:
I've followed-up with the helm community about this.
There were security concerns in giving a github token with write permissions, as a PR could use that token to do things it shouldn't.  Because of those concerns it was chosen to use this bot instead https://github.com/apps/pull-request-size

Opinions?

jpmcb:
> There were security concerns in giving a github token with write permissions

That shouldn't be a concern with the way this PR is written. From the GitHub docs:

> This event runs in the context of the base of the pull request, rather than in the context of the merge commit, as the pull_request event does. This prevents execution of unsafe code from the head of the pull request that could alter your repository or steal any secrets you use in your workflow

> Warning: For workflows that are triggered by the pull_request_target event, the GITHUB_TOKEN is granted read/write repository permission unless the permissions key is specified and the workflow can access secrets, even when it is triggered from a fork. Although the workflow runs in the context of the base of the pull request, you should make sure that you do not check out, build, or run untrusted code from the pull request with this event.

[Reg github docs](https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#pull_request_target)

I believe this is safe since this action would _only modify a label_, not checkout code from the head of the remote fork and attempt to build or execute on it.

marckhouzam (approved):
LGTM
Thanks for the explanation

PR #1612, opened 2022-02-24
undefined or nil Args default to ArbitraryArgs

umarcor:
This is a follow-up of #1547. See https://github.com/spf13/cobra/pull/1547#issuecomment-1050245825.

### DIFF
--- a/command.go
+++ b/command.go
@@ -1012,7 +1012,7 @@ func (c *Command) ExecuteC() (cmd *Command, err error) {
 
 func (c *Command) ValidateArgs(args []string) error {
 	if c.Args == nil {
-		return nil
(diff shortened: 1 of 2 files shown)


marckhouzam (approved):
LGTM

This PR does not change any functionality and therefore is 100% safe.
The value of it is that it formalizes the fact that when `command.Args == nil`, then `ArbitraryArgs` are accepted.

umarcor:
@jpmcb @johnSchnake can we, please, have this merged?

umarcor:
*ping* @spf13 @johnSchnake @jpmcb @marckhouzam

jpmcb (approved):

PR #1621, opened 2022-03-08
Allow user to add completion for powershell alias

marckhouzam:
I ran across a possible problem with using powershell completion with aliases.  I can't recall where it was though, but the situation is pretty simple.

IIUC, when a user has an alias in powershell, she needs to register that alias for completion using:
```
PS> Register-ArgumentCompleter -CommandName 'aliasname' -ScriptBlock <completionBlock>
```
The problem is that the `<completionBlock>` is currently not accessible in our powershell script.

To address this, this commit stores the completion logic into a scriptblock variable which can easily be accessed by the
user to register aliases.

For example, if the user defines an alias for `helm` she will be able to register an alias like this:
```
PS> sal h helm
PS> Register-ArgumentCompleter -CommandName 'h' -ScriptBlock $__helmCompleterBlock
```

@Luap99 does this make sense to you?  I may be misunderstanding something fundamental about aliases and powershell...

### DIFF
--- a/powershell_completions.go
+++ b/powershell_completions.go
@@ -22,9 +22,15 @@ import (
 	"fmt"
 	"io"
 	"os"
+	"strings"
 )
 
 func genPowerShellComp(buf io.StringWriter, name string, includeDesc bool) {
+	// Variables should not contain a '-' or ':' character
+	nameForVar := name
+	nameForVar = strings.Replace(nameForVar, "-", "_", -1)
+	nameForVar = strings.Replace(nameForVar, ":", "_", -1)
+
 	compCmd := ShellCompRequestCmd
 	if !includeDesc {
 		compCmd = ShellCompNoDescRequestCmd
@@ -41,7 +47,7 @@ filter __%[1]s_escapeStringWithSpecialChars {
 `+"    $_ -replace '\\s|#|@|\\$|;|,|''|\\{|\\}|\\(|\\)|\"|`|\\||<|>|&','`$&'"+`
 }
 
-Register-ArgumentCompleter -CommandName '%[1]s' -ScriptBlock {
+[scriptblock]$__%[2]sCompleterBlock = {
     param(
             $WordToComplete,
             $CommandAst,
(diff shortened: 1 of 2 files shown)


Luap99:
How does this work with the other shells at the moment?

I do not use powershell any more but I think this will work. Back when I worked on this there weren't many projects with powershell completion so I am not sure if there is maybe some pseudo standard which can/should be followed.

marckhouzam:
This whole idea came out of a comment from the PR that added Powershell completion to kubectl: https://github.com/kubernetes/kubernetes/pull/103758#issuecomment-1007586166

There was a project that implemented a hard-coded script for powershell completion for kubectl which handled aliases in this way.  This is the specific line that inspired me: https://github.com/mziyabo/PSKubectlCompletion/blob/f6f28153a616e14789a3a7fbcb493ff7f1f1ae85/PSKubectlCompletion.psm1#L20

Luap99:
I personally do not like to pollute the global variable scope in the shell but the variable name seems reasonable unique so it should be fine. Since there is a legitimate use case for this and I cannot offer a better solution I am fine with this PR.

jpmcb:
Hmmmmm anyone out in the community have an easy way to test this on Windows? I'm without my windows box at the moment.

marckhouzam:
Still valid

marckhouzam:
I fixed this PR which is still relevant.

powershell_completions.go
```diff
@@ -106,12 +112,13 @@ Register-ArgumentCompleter -CommandName '%[1]s' -ScriptBlock {
...
     __%[1]s_debug "Calling $RequestComp"
     # First disable ActiveHelp which is not supported for Powershell
-    $env:%[8]s=0
+    $env:%[9]s=0
 
     #call the command store the output in $out and redirect stderr and stdout to null
     # $Out is an array contains each line per element
     Invoke-Expression -OutVariable out "$RequestComp" 2>&1 | Out-Null
 
+
```
Luap99, line 121: nit: remove this added newline

marckhouzam: Thanks for noticing.  Done.

Luap99 (commented):
LGTM

jpmcb (approved):
@marckhouzam @Luap99 - I see this has the `lgtm` label: I am still without a windows box suitable for testing :( is this good to go? Code looks good to me and If you feel ok with this Marc, go ahead and self merge

marckhouzam:
Yeah, it is ready.  I am able to test it on powershell on Mac, and @Luap99 is happy with the change.  Will merge.

cloud-66:
Sorry. But why it is not working for me ?
```
# PowerShell auto-completion
kubectl completion powershell | Out-String | Invoke-Expression
Set-Alias -Name k -Value kubectl
Register-ArgumentCompleter -CommandName k -ScriptBlock $__kubectlCompleterBlock
```

marckhouzam:
What version of kubectl are you using? I believe you need 1.26 or higher.

cloud-66:
> What version of kubectl are you using? I believe you need 1.26 or higher.

Yeah. Sorry to bother you. Kubectl 1.26+ depends on cobra 1.6.0 with this pr. thank you it works .

PR #1623, opened 2022-03-09
add custom error log

ningmingxiao:


ningmingxiao:
i want to  add custom err ,not  like "requires at least 1 arg(s), only received 0 "

marckhouzam:
Thanks @ningmingxiao for the suggestion.  However, I think that if we add a method for a custom error for `MinimumNArgs()` we would need to add one for all the other methods that print a message (`MaximumNArgs()`, `ExactArgs()`, etc).

I suggest you implement the method you need in your program and call it directly, it does not need to be in Cobra itself.

What do you say?

ningmingxiao:
now if we use MinimumNArgs, i want to add custom log when use check args lenth, but it will not work except i don't use  MinimumNArgs

marckhouzam:
What you need to do is implement your new `minimumNArgsCustomError()` function in your own code.  Here is an example of how Helm calls its own version of `MinimumNArgs()` method:
https://github.com/helm/helm/blob/974a6030c8514591ab0b0f0c898d37f816f698f6/cmd/helm/install.go#L120
and this method is implemented in helm itself:
https://github.com/helm/helm/blob/974a6030c8514591ab0b0f0c898d37f816f698f6/cmd/helm/require/args.go#L68-L80

I'm going to go ahead and close this since there is a solution available.
Feel free to re-open if I misunderstood your need.

PR #1624, opened 2022-03-09
labeler: fix syntax

umarcor:
/cc @jpmcb


marckhouzam:
Woops.  Thanks for the fix @umarcor !

PR #1632, opened 2022-03-15
Fix hidden commands changing padding in usage template

### LINKED ISSUE #1272: Hidden Commands Should be Ignored When Calculating Use Indentation

aliscott:
We added a `fig-autocomplete` hidden subcommand to Infracost and it increased the padding in the usage text (https://github.com/infracost/infracost/pull/1457#discussion_r827182625), which was unexpected. This fixes any hidden commands affecting the padding:

Fixes #1272

Example before:

```
AVAILABLE COMMANDS
  breakdown        Show full breakdown of costs
  comment          Post an Infracost comment to GitHub, GitLab, Azure Repos or Bitbucket
  completion       Generate shell completion script
```

Example with fix:

```
AVAILABLE COMMANDS
  breakdown   Show full breakdown of costs
  comment     Post an Infracost comment to GitHub, GitLab, Azure Repos or Bitbucket
  completion  Generate shell completion script
```

I wasn't sure where to add tests for this since I couldn't find any existing ones for this functionality, but if you point me in the right direction I can do.

### DIFF
--- a/command.go
+++ b/command.go
@@ -1158,18 +1158,20 @@ func (c *Command) AddCommand(cmds ...*Command) {
 			panic("Command can't be a child of itself")
 		}
 		cmds[i].parent = c
-		// update max lengths
-		usageLen := len(x.Use)
-		if usageLen > c.commandsMaxUseLen {
-			c.commandsMaxUseLen = usageLen
-		}
-		commandPathLen := len(x.CommandPath())
-		if commandPathLen > c.commandsMaxCommandPathLen {
-			c.commandsMaxCommandPathLen = commandPathLen
-		}
-		nameLen := len(x.Name())
-		if nameLen > c.commandsMaxNameLen {
-			c.commandsMaxNameLen = nameLen
+		if len(x.Deprecated) == 0 && !x.Hidden {
+			// update max lengths
+			usageLen := len(x.Use)
+			if usageLen > c.commandsMaxUseLen {
+				c.commandsMaxUseLen = usageLen
+			}
+			commandPathLen := len(x.CommandPath())
+			if commandPathLen > c.commandsMaxCommandPathLen {
+				c.commandsMaxCommandPathLen = commandPathLen
+			}
+			nameLen := len(x.Name())
+			if nameLen > c.commandsMaxNameLen {
+				c.commandsMaxNameLen = nameLen
+			}
 		}
 		// If global normalization function exists, update all children
 		if c.globNormFunc != nil {
@@ -1199,6 +1201,9 @@ main:
 	c.commandsMaxCommandPathLen = 0
 	c.commandsMaxNameLen = 0
 	for _, command := range c.commands {
+		if len(command.Deprecated) != 0 || command.Hidden {
+			continue
+		}
 		usageLen := len(command.Use)
 		if usageLen > c.commandsMaxUseLen {
 			c.commandsMaxUseLen = usageLen
--- a/command_test.go
+++ b/command_test.go
@@ -2161,3 +2161,42 @@ func TestSetContextPersistentPreRun(t *testing.T) {
 		t.Error(err)
 	}
 }
+
+func TestPadding(t *testing.T) {
+	rootCmd := &Command{Use: "root", Run: emptyRun}
+	childCmd := &Command{Use: "child", Run: emptyRun}
+	longChildCmd := &Command{Use: "long-name-child-abcdefghijklmnopqrstuvwxyz", Run: emptyRun}
+	// For this test to be useful the hiddenChildCmd and deprecatedChildCmd commands need to have a longer `Use` field than the other commands.
+	hiddenChildCmd := &Command{Use: longChildCmd.Use + "-hidden", Hidden: true, Run: emptyRun}
+	deprecatedChildCmd := &Command{Use: longChildCmd.Use + "-deprecated", Deprecated: "deprecated", Run: emptyRun}
+
+	rootCmd.AddCommand(childCmd)
+	rootCmd.AddCommand(longChildCmd)
+	rootCmd.AddCommand(hiddenChildCmd)
+	rootCmd.AddCommand(deprecatedChildCmd)
+
+	expectedUsePad := len(longChildCmd.Use)
+	expectedPathPad := len(longChildCmd.CommandPath())
+	expectedNamePad := len(longChildCmd.Name())
+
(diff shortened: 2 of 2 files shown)


marckhouzam:
This makes sense.  I'm surprised it wasn't fixed earlier 😄 .  Thanks for the PR!
In fact I think the same problem applies to "deprecated" commands.  I think that instead of checking for `Hidden`, we could instead use `IsAvailableCommand()` to know if the command should be considered for padding.

As for tests, I didn't find any related to padding (but I may have missed them), but you can add them in the `command_test.go` file.

aliscott:
Thanks @marckhouzam have updated to use `IsAvailableCommand()` and added some tests.

command_test.go
```diff
@@ -2058,3 +2058,40 @@ func TestFParseErrWhitelistSiblingCommand(t *testing.T) {
 	}
 	checkStringContains(t, output, "unknown flag: --unknown")
 }
+
+func TestPadding(t *testing.T) {
+	rootCmd := &Command{Use: "root", Run: emptyRun}
+	childCmd := &Command{Use: "child", Run: emptyRun}
+	longChildCmd := &Command{Use: "long-name-child-abcdefghijklmnopqrstuvwxyz", Run: emptyRun}
+	hiddenChildCmd := &Command{Use: "hidden-long-name-child-abcdefghijklmnopqrstuvwxyz", Hidden: true, Run: emptyRun}
+	deprecatedChildCmd := &Command{Use: "deprecated-long-name-child-abcdefghijklmnopqrstuvwxyz", Deprecated: "deprecated", Run: emptyRun}
```
marckhouzam, line 2067: I believe that for the test to be useful, it is important that the deprecated command and hidden command have a longer `Use` field than the other ones.  This is the case now, but seems brittle.
Could you put a comment to say that it is needed?

Also, how about using `longChildCmd.Use + "-hidden"` for the `Use` of the hidden command, and similarly for the deprecated command?  That way we are sure they will be longer.

Luap99 (commented):
I think [IsAvailableCommand()](https://github.com/spf13/cobra/blob/f848943afd7212766aadc19256cf9e7384980281/command.go#L1385) only works when the command has subcommands or has a run function.
I think it is possible to have doc only commands which do not have a run function so this would break, see https://github.com/spf13/cobra/blob/f848943afd7212766aadc19256cf9e7384980281/command.go#L1396
https://github.com/spf13/cobra/blob/f848943afd7212766aadc19256cf9e7384980281/command.go#L516-L517

marckhouzam:
> I think [IsAvailableCommand()](https://github.com/spf13/cobra/blob/f848943afd7212766aadc19256cf9e7384980281/command.go#L1385) only works when the command has subcommands or has a run function. I think it is possible to have doc only commands which do not have a run function so this would break

Good point @Luap99.  I hadn't thought of that.

I haven't used an `AdditionalHelpTopic` before (what you call "doc only command"), do you have a use case that would help understand when to use such a thing?

marckhouzam:
@aliscott As pointed out by @Luap99, we should consider the case of `AdditionalHelpTopic`, so I think we should check that the command is neither hidden nor deprecated explicitly instead of checking `IsAvailableCommand()`.

Luap99:
I don't use it and I am not aware of cobra examples but one simple example is `go help`, e.g. `go help go.mod`

Luap99:
Also see the linked issue in the function comment: https://github.com/spf13/cobra/issues/393#issuecomment-282741924

marckhouzam:
> Also see the linked issue in the function comment: [#393 (comment)](https://github.com/spf13/cobra/issues/393#issuecomment-282741924)

Brilliant!  I had not noticed the link in the comment 🤦‍♂️

marckhouzam (approved):
LGTM
Thanks @aliscott

marckhouzam:
@jpmcb This is ready for a second review.

command.go
```diff
@@ -1158,18 +1158,20 @@ func (c *Command) AddCommand(cmds ...*Command) {
...
-			c.commandsMaxUseLen = usageLen
-		}
-		commandPathLen := len(x.CommandPath())
-		if commandPathLen > c.commandsMaxCommandPathLen {
-			c.commandsMaxCommandPathLen = commandPathLen
-		}
-		nameLen := len(x.Name())
-		if nameLen > c.commandsMaxNameLen {
-			c.commandsMaxNameLen = nameLen
+		if len(x.Deprecated) == 0 && !x.Hidden {
```
jpmcb, line 1161: Could we add a quick comment here? I'd imagine other contributors might see this and be confused why we need this block

command.go
```diff
@@ -1199,6 +1201,9 @@ main:
 	c.commandsMaxCommandPathLen = 0
 	c.commandsMaxNameLen = 0
 	for _, command := range c.commands {
+		if len(command.Deprecated) != 0 || command.Hidden {
```
jpmcb, line 1204: Same here: let's add a quick comment

jpmcb (commented):
This looks good to me - but I'm a _bit_ confused as to the root cause of this. _Why_ are commands that are deprecated or hidden getting different paddings? 

We can go ahead with this, I'm just wondering why this behavior existed in the first place. Aren't commands .. just commands?

marckhouzam:
> This looks good to me - but I'm a _bit_ confused as to the root cause of this. _Why_ are commands that are deprecated or hidden getting different paddings?

The padding, as far as I know is when printing the help text.  Hidden and deprecated commands are not shown in the help text so they should not be used when calculating the padding; you could have a much longer hidden command which would cause a lot of padding, when the help won't even show that command.

But your question is making me wonder if someone overrides the help text and chooses the display deprecated commands for example, would we be causing them an issue?

jpmcb:
> you could have a much longer hidden command which would cause a lot of padding, when the help won't even show that command.

Ahhh that makes sense.

> wonder if someone overrides the help text and chooses the display deprecated commands for example, would we be causing them an issue?

Hmmmm good question. We should investigate a bit further to confirm we won't break users

jpmcb:
Not stale - needs more discussion

marckhouzam:
I put the lifecycle/frozen label on this one.

natsu0529:
Hi! I’d like to help move this issue forward.

I noticed this PR has been inactive for a while and now appears to have conflicts. Would maintainers prefer that I open a fresh PR against the current main branch, based on the approach discussed here?

I’ll make sure to include tests in command_test.go.

PR #1639, opened 2022-03-20
Fix Command.Context comment

### LINKED ISSUE #1578: Command.Context() returns nil instead of context.Background()

katexochen:
Closes #1578, together with #1551.

Is an alternative to #1582. Previous discussion on this topic can be found there.

### DIFF
--- a/command.go
+++ b/command.go
@@ -166,7 +166,7 @@ type Command struct {
 	// errWriter is a writer defined by the user that replaces stderr
 	errWriter io.Writer
 
-	//FParseErrWhitelist flag parse errors to be ignored
+	// FParseErrWhitelist flag parse errors to be ignored
 	FParseErrWhitelist FParseErrWhitelist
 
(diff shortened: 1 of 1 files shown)


wfernandes (approved):
lgtm.
reviewed the relevant functions and the godoc aligns with it.

marckhouzam (approved):
Thanks @katexochen

PR #1642, opened 2022-03-21
Add Kubescape to projects

avinashupadhya99:
[Kubescape](https://github.com/armosec/kubescape) is a Kubernetes security tool with over 5.4k stars on GitHub. Adding it to the list of projects using cobra


marckhouzam (approved):
Thanks @avinashupadhya99.
It's always nice to see successful projects that use Cobra. 🎉

PR #896, opened 2019-06-28
enable composing PositionalArgs

nelz9999:
This may end up being helpful related to issues like #838 or #745...

PR #1643, opened 2022-03-22
deprecate ExactValidArgs and test combinations of args validators

### LINKED ISSUE #745: ExactArgs + ValidArgs + OnlyValidArgs
I have a command called `i3` that must have at exactly one argument given to it, and the argument can either be "load" or "disable". By passing the command "ExactArgs(1)" I can make it do exactly one arg, but it doesn't respect "ValidArgs". I can pass "OnlyValidArgs" to make it respect "ValidArgs", but I can't pass it both. Something like `Args: cobra.ExactArgs(1) && cobra.OnlyValidArgs(cobra, args)` would be awesome.

```
var i3Cmd = &cobra.Command{
	Use:       "i3",
	Short:     "A brief description of your command",
	Long:      ``,
	ValidArgs: []string{"load", "disable"},
	Args:      cobra.ExactArgs(1),
		fmt.Println("i3 called")
	},
}
```

PASS `go run main.go i3` --> Error: accepts 1 arg(s), received 0
PASS `go run main.go i3 hello world` --> Error: accepts 1 arg(s), received 2
FAIL `go run main.go i3 chicken` --> i3 called
PASS `go run main.go i3 load ` --> i3 called

ValidArgs is being ignored because `cobra.OnlyValidArgs(cmd, args)` isn't called. When i pass in chicken, I want it to error because 'load' is the only valid arg.

Trying again with a custom Arg filter.
```
var i3Cmd = &cobra.Command{
	Use:       "i3",
	Short:     "A brief description of your command",
	Long:      ``,
	ValidArgs: []string{"load", "disable"},
		if len(args) != 1 {
			return errors.New("Requires exactly 1 arg.")
		}
		return cobra.OnlyValidArgs(cmd, args)
	},
		fmt.Println("i3 called")
	},
}
```

PASS `go run main.go i3 hello` --> Error: invalid argument "hello" for "dot i3"
PASS `go run main.go i3 load`  --> i3 called
PASS `go run main.go i3 hello world` --> Error: Requires exactly 1 arg
PASS `go run main.go i3` -->  Error: Requires exactly 1 arg.

The example above works... but it doesn't use `cobra.ExactArgs(1)`. I don't like re-inventing the wheel. How can I use both `OnlyValidArgs` and `ExactArgs` together? I tried:
```
var i3Cmd = &cobra.Command{
	Use:       "i3",
	Short:     "A brief description of your command",
	Long:      ``,
	ValidArgs: []string{"load"},
		err := cobra.ExactArgs(1)
		if err != nil {
			return err // <--- this is a compile error
		}	
		return cobra.OnlyValidArgs(cmd, args)
	},
		fmt.Println("i3 called")
	},
}
```
But `ExactArgs` returns `PositionalArgs`.. which looks to be a Type.. I'm not very familiar with Go, so this is confusing to me. I don't know how to combine the two. Any insight would be appreciated!

Also.. If I set `ValidArgs` and `OnlyValidArgs`, and an invalid command arg is provided, the user isn't notified of what the valid Args are. Is this the same feature request as #571? That seems like it would be very helpful.

related #284 
related #571 
related #346

### LINKED ISSUE #838: Feature request: support MatchValid(PositionalArgs) OR MinimumValidNArgs, MaximumValidNArgs, RangeValidArgs

umarcor:
This is a subset of #841, which was approved by @jharshman and reviewed by @marckhouzam .

Fix #838 and Fix #745.

The original validators care about the number of arguments that a command accepts: `NoArgs`, `ArbitraryArgs`, `MinimumNArgs(int)`, `MaximumNArgs(int)`, `ExactArgs(int)`, `RangeArgs(min, max)`. We can consider it an enumeration of 6 items.

When `ValidArgs` was introduced, up to six additional validators might have been added, in order to provide functions covering the full combination matrix. Certainly, `ValidArgs` is a boolean condition which is orthogonal to `Args`, because it is focused on the content of the args, not the number of them. Therefore, the full matrix covers 6 x 2 = 12 combinations. However, `NoArgs` + `ValidArgs` == `NoArgs` + `!ValidArgs`, so 11.

Nonetheless, the implementation was partial and a single one was introduced. Later, a second one was added. Still, there are three combinations missing, which is what was pointed out in https://github.com/spf13/cobra/issues/838 at first.

Hence, this PR enforces the orthogonality in the code by moving the validation of `ValidArgs` from `args.go` to `command.go`, so that:

- Any validator can be used along with `ValidArgs`. `ValidArgs` is checked first, and then the defined validator.
- `OnlyValidArgs` and `ExactValidArgs` are deprecated, but not removed.

The README is updated accordingly.

### DIFF
--- a/args.go
+++ b/args.go
@@ -113,5 +113,5 @@ func MatchAll(pargs ...PositionalArgs) PositionalArgs {
 //
 // Deprecated: use MatchAll(OnlyValidArgs, ExactArgs(n)) instead
 func ExactValidArgs(n int) PositionalArgs {
-	return MatchAll(OnlyValidArgs, ExactArgs(n))
+	return MatchAll(ExactArgs(n), OnlyValidArgs)
 }
--- a/args.go
+++ b/args.go
@@ -32,7 +32,8 @@ func NoArgs(cmd *Command, args []string) error {
 	return nil
 }
 
-// OnlyValidArgs returns an error if any args are not in the list of ValidArgs.
+// OnlyValidArgs returns an error if there are any positional args that are not in
+// the `ValidArgs` field of `Command`
 func OnlyValidArgs(cmd *Command, args []string) error {
 	if len(cmd.ValidArgs) > 0 {
 		// Remove any description that may be included in ValidArgs.
@@ -41,7 +42,6 @@ func OnlyValidArgs(cmd *Command, args []string) error {
 		for _, v := range cmd.ValidArgs {
 			validArgs = append(validArgs, strings.Split(v, "\t")[0])
 		}
-
 		for _, v := range args {
 			if !stringInSlice(v, validArgs) {
 				return fmt.Errorf("invalid argument %q for %q%s", v, cmd.CommandPath(), cmd.findSuggestions(args[0]))
--- a/args.go
+++ b/args.go
@@ -111,7 +111,7 @@ func MatchAll(pargs ...PositionalArgs) PositionalArgs {
 // ExactValidArgs returns an error if there are not exactly N positional args OR
 // there are any positional args that are not in the `ValidArgs` field of `Command`
 //
-// Deprecated: use MatchAll(OnlyValidArgs, ExactArgs(n)) instead
+// Deprecated: use MatchAll(ExactArgs(n), OnlyValidArgs) instead
 func ExactValidArgs(n int) PositionalArgs {
 	return MatchAll(ExactArgs(n), OnlyValidArgs)
 }
--- a/args.go
+++ b/args.go
@@ -86,18 +86,6 @@ func ExactArgs(n int) PositionalArgs {
 	}
 }
 
-// ExactValidArgs returns an error if
-// there are not exactly N positional args OR
-// there are any positional args that are not in the `ValidArgs` field of `Command`
-func ExactValidArgs(n int) PositionalArgs {
-	return func(cmd *Command, args []string) error {
-		if err := ExactArgs(n)(cmd, args); err != nil {
-			return err
-		}
-		return OnlyValidArgs(cmd, args)
-	}
-}
-
 // RangeArgs returns an error if the number of args is not within the expected range.
 func RangeArgs(min int, max int) PositionalArgs {
 	return func(cmd *Command, args []string) error {
@@ -118,4 +106,12 @@ func MatchAll(pargs ...PositionalArgs) PositionalArgs {
 		}
 		return nil
 	}
+}
+
+// ExactValidArgs returns an error if there are not exactly N positional args OR
+// there are any positional args that are not in the `ValidArgs` field of `Command`
+//
+// Deprecated: use MatchAll(OnlyValidArgs, ExactArgs(n)) instead
+func ExactValidArgs(n int) PositionalArgs {
+	return MatchAll(OnlyValidArgs, ExactArgs(n))
 }
--- a/args_test.go
+++ b/args_test.go
@@ -488,7 +488,7 @@ func TestExactValidArgs_WithInvalidCount(t *testing.T) {
 func TestExactValidArgs_WithInvalidCount_WithInvalidArgs(t *testing.T) {
 	c := getCommand(ExactValidArgs(3), true)
 	_, err := executeCommand(c, "three", "a", "two")
-	validOnlyWithInvalidArgs(err, t)
+	exactArgsWithInvalidCount(err, t)
 }
 
 func TestExactValidArgs_WithInvalidArgs(t *testing.T) {
--- a/args_test.go
+++ b/args_test.go
@@ -159,24 +159,6 @@ func TestExactArgsWithInvalidCount(t *testing.T) {
 	exactArgsWithInvalidCount(err, t)
 }
 
-func TestExactValidArgs(t *testing.T) {
-	c := getCommand(ExactValidArgs(3), true)
-	output, err := executeCommand(c, "three", "one", "two")
-	expectSuccess(output, err, t)
-}
-
-func TestExactValidArgsWithInvalidCount(t *testing.T) {
-	c := getCommand(ExactValidArgs(2), false)
-	_, err := executeCommand(c, "three", "one", "two")
-	exactArgsWithInvalidCount(err, t)
-}
-
-func TestExactValidArgsWithInvalidArgs(t *testing.T) {
-	c := getCommand(ExactValidArgs(3), true)
-	_, err := executeCommand(c, "three", "a", "two")
-	validWithInvalidArgs(err, t)
-}
-
 func TestRangeArgs(t *testing.T) {
 	c := getCommand(RangeArgs(2, 4), false)
 	output, err := executeCommand(c, "a", "b", "c")
@@ -291,6 +273,26 @@ func TestMatchAll(t *testing.T) {
 			}
 		})
 	}
+}
+
+// DEPRECATED
+
+func TestExactValidArgs(t *testing.T) {
+	c := getCommand(ExactValidArgs(3), true)
+	output, err := executeCommand(c, "three", "one", "two")
+	expectSuccess(output, err, t)
+}
+
+func TestExactValidArgsWithInvalidCount(t *testing.T) {
+	c := getCommand(ExactValidArgs(2), false)
+	_, err := executeCommand(c, "three", "one", "two")
+	exactArgsWithInvalidCount(err, t)
+}
+
+func TestExactValidArgsWithInvalidArgs(t *testing.T) {
+	c := getCommand(ExactValidArgs(3), true)
+	_, err := executeCommand(c, "three", "a", "two")
+	validWithInvalidArgs(err, t)
 }
 
 // This test make sure we keep backwards-compatibility with respect
--- a/args_test.go
+++ b/args_test.go
@@ -27,7 +27,7 @@ func expectSuccess(output string, err error, t *testing.T) {
 	}
 }
 
-func validWithInvalidArgs(err error, t *testing.T) {
+func validOnlyWithInvalidArgs(err error, t *testing.T) {
 	if err == nil {
 		t.Fatal("Expected an error")
 	}
@@ -43,7 +43,7 @@ func noArgsWithArgs(err error, t *testing.T) {
 		t.Fatal("Expected an error")
 	}
 	got := err.Error()
-	expected := `unknown command "illegal" for "c"`
+	expected := `unknown command "one" for "c"`
 	if got != expected {
 		t.Errorf("Expected: %q, got: %q", expected, got)
 	}
@@ -93,17 +93,39 @@ func rangeArgsWithInvalidCount(err error, t *testing.T) {
 	}
 }
 
+// NoArgs
+
 func TestNoArgs(t *testing.T) {
 	c := getCommand(NoArgs, false)
 	output, err := executeCommand(c)
 	expectSuccess(output, err, t)
 }
 
-func TestNoArgsWithArgs(t *testing.T) {
+func TestNoArgs_WithArgs(t *testing.T) {
 	c := getCommand(NoArgs, false)
-	_, err := executeCommand(c, "illegal")
+	_, err := executeCommand(c, "one")
 	noArgsWithArgs(err, t)
 }
+
+func TestNoArgs_WithValid_WithArgs(t *testing.T) {
+	c := getCommand(NoArgs, true)
+	_, err := executeCommand(c, "one")
+	noArgsWithArgs(err, t)
+}
+
+func TestNoArgs_WithValid_WithInvalidArgs(t *testing.T) {
+	c := getCommand(NoArgs, true)
+	_, err := executeCommand(c, "one")
+	noArgsWithArgs(err, t)
+}
+
+func TestNoArgs_WithValidOnly_WithInvalidArgs(t *testing.T) {
+	c := getCommand(MatchAll(OnlyValidArgs, NoArgs), true)
+	_, err := executeCommand(c, "a")
+	validOnlyWithInvalidArgs(err, t)
+}
+
+// OnlyValidArgs
 
 func TestOnlyValidArgs(t *testing.T) {
 	c := getCommand(OnlyValidArgs, true)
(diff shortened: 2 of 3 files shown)


marckhouzam (changes requested):
I hadn't realized before, but I believe this is a breaking change.  I don't know if it was discussed before, so apologies if I missed something.

But note that I believe it would have broken `kubectl` had `kubectl` not recently stopped using `ValidArgs` in favor of `ValidArgsFunction()`.
Let me try to explain using `kubectl`.

Let's take the `kubectl expose` command which works on a subset of kubernetes resource types.
For simplicity let's say it only works on `deployment` and `pod` (it actually applies to 3 other types of resources, but it does not matter for this example).
So, for the first argument, the user must specify one of the two valid resource types.

Originally, `ValidArgs` was only used for shell completion, as can be see in the following comment:
https://github.com/spf13/cobra/blob/5414d3d45d41008846e74d0f7768363014f3f5c0/command.go#L66-L67 (which should technically be changed by this PR).

So, to help the user, `ValidArgs` was set to `{"deployment", "pod"}` as the two possible completions.
So, when using shell completion the user would end up with either `kubectl expose deployment` or `kubectl expose pod`.

Now, and here is the problem, `kubectl` accepts multiple "aliases" for its resource types.
So, instead of using `deployment` I can just as well use: `deploy`, `deployments`, `deploy.app`, `deploy.apps`, `deployment.app`, `deployments.apps`, etc.
This is not a critical point for shell completion which just guides the user, but it is an important feature for the proper functioning of `kubectl` itself.

If we only allow specified `ValidArgs` as arguments, as suggested by this PR, such a scenario would stop working.

I believe `kubectl` is safe now that it uses `ValidArgsFunction()`, but it shows that we could affect other users of Cobra with this change.

I think we would need to make this opt-in if we are to move forward with this change.
But that is pretty much what we have now, where a program can choose to enforce `ValidArgs` or not, by specifying what validators to use.

umarcor:
@marckhouzam thanks for a very interesting analysis! It seems that the "main" purpose of `ValidArgs` changed since it was introduced in 2015:

- https://github.com/spf13/cobra/commit/9b2e6822e5819fdc8d6c80654b40ba4f805c690a (#69): `ValidArgs` and `BashCompletionFunction` were introduced, to be used in shell completion.
  The arguments were not validated:

```go
// List of all valid non-flag arguments, used for bash completions *TODO* actually validate these
ValidArgs []string
```

- https://github.com/spf13/cobra/commit/d89c499964451591c4e8142464f0d558c4a3698d (#284): the management of arguments was reworked, introducing option `TakesArgs`, which supported `Legacy`, `None`, `Arbitrary` or `ValidOnly`.

```markdown
`Legacy` (or default) the rules are as follows:
- root commands with no subcommands can take arbitrary arguments
- root commands with subcommands will do subcommand validity checking
- subcommands will always accept arbitrary arguments and do no subsubcommand validity checking

`None` the command will be rejected if there are any left over arguments after parsing flags.

`Arbitrary` any additional values left after parsing flags will be passed in to your `Run` function.

`ValidOnly` you must define all valid (non-subcommand) arguments to your command. These are defined in a slice name ValidArgs. For example a command which only takes the argument "one" or "two" would be defined as:
```

- https://github.com/spf13/cobra/commit/f20b4e9c32bb3e9d44773ca208db814f24dcd21b (#284): the previous commit was refactored and `TakesArgs` was removed and built-in validators were added:

```markdown
Validation of positional arguments can be specified using the `Args` field.

The follow validators are built in:

- `NoArgs` - the command will report an error if there are any positional args.
- `ArbitraryArgs` - the command will accept any args.
- `OnlyValidArgs` - the command will report an error if there are any positional args that are not in the ValidArgs list.
- `MinimumNArgs(int)` - the command will report an error if there are not at least N positional args.
- `MaximumNArgs(int)` - the command will report an error if there are more than N positional args.
- `ExactArgs(int)` - the command will report an error if there are not exactly N positional args.
- `RangeArgs(min, max)` - the command will report an error if the number of args is not between the minimum and maximum number of expected args.
```

At that point, any of `MinimumNArgs`, `MaximumNArgs`, `ExactArgs` or `RangeArgs` (which are mutually exclusive) could be combined with `OnlyValidArgs`, by writing a custom validation function.
However, the tests did not include those use cases.

- https://github.com/spf13/cobra/commit/f619abc1d7edb30544d1c956719d203ace73e9b1 (#765): `ExactValidArgs` was added:

```go
func ExactValidArgs(n int) PositionalArgs {
	return func(cmd *Command, args []string) error {
		if err := ExactArgs(n)(cmd, args); err != nil {
			return err
		}
		return OnlyValidArgs(cmd, args)
	}
}
```

So, the inconsistency was introduced in #765.
`MinimumNValidArgs`, `MaximumNValidArgs` and `RangeValidArgs` were missing.

It was not obvious back then, that the implementation should have been a generic helper to combine validators.
`MatchAll` was proposed some months later (#896), which provided the generic solution.
`ExactValidArgs` could have been explicitly deprecated in #896 in favour of showcasing a generic solution, but it was overlooked (even though we did discuss about it).

Note that this PR was created in March 2019 (#841), between #765 (October 2018) and #896 (June 2019).

Note also that there was no reference to the completion feature in neither of #284, #765, #841 or #896, despite `ValidArgs` apparently still being used for that purpose as well.
It seems that it was mostly forgotten until you started to work on it 2 years ago.

Overall, I think that enforcing the valid args in the kubectl example you explained was not a supported use case until you reworked and extended the usability of the completion feature.
Nevertheless, I agree there is no need to apply the `OnlyValidArgs` validator implicitly, because the work introduced in the codebase since this PR was proposed makes it sensible to instead document how to do it explicitly.
Therefore, I will update this PR to:

- Remove https://github.com/spf13/cobra/pull/1643/files#diff-4d7d1923693fc5ce892add2ea2907a744e77ea0b50c1939ccc5067cb48a466a3R1013-R1015.
- Not deprecate `OnlyValidArgs`, but `ExactValidArgs` only.
- Convert all the tests added in this PR to use the explicit solution (combine `OnlyValidArgs` with each of the other validators, through `MatchAll`).
- Update the User Guide to explain which validators are mutually exclusive, and which are orthogonal (can be combined).

As a result, the outcomes of this PR will be:

- Deprecate but not remove `ExactValidArgs`.
- Add tests to check the combination of built-in validators which are orthogonal.
- Document how to combine the built-in validators through the built-in combination function.

Then, in a follow-up PR, we can discuss how to enhance `OnlyValidArgs` with `ValidArgsFunction` in order to support the kubectl example with aliases.

user_guide.md
```diff
@@ -302,30 +302,47 @@ rootCmd.MarkPersistentFlagRequired("region")
...
+  - `ExactArgs(int)` - report an error if there are not exactly N positional args.
+  - `RangeArgs(min, max)` - report an error if the number of args is not between `min` and `max`.
+- Content of the arguments:
+  - `OnlyValidArgs` - report an error if there are any positional args not specified in the `ValidArgs` field of `Command`, which can optionally be set to a list of valid values for positional args.
+
+If `Args` is undefined or `nil`, it defaults to `ArbitraryArgs`.
+
+Moreover, `MatchAll(pargs ...PositionalArgs)` enables combining existing checks with arbitrary other checks.
+For instance, you want to check the ExactArgs length along with OnlyValidArgs to report an error if there are not
+exactly N positional args OR if there are any positional args that are not in the `ValidArgs` field of `Command`.
```
marckhouzam, line 322: This sentence is hard to parse for me.  How about reversing it to something like:

"For instance, if you want to report an error if there are not exactly N positional args OR if there are any positional args that are not in the `ValidArgs` field of `Command`, you can call `MatchAll` on `ExactArgs` and `OnlyValidArgs` as shown below.

umarcor: See https://github.com/spf13/cobra/pull/1646/files#diff-b41dbbb1a3fc0b869a97b02eb8437467af81d0d0aefe2a2cf4bb629fdf01b166R321

umarcor:
> I like the formality this PR introduces. I think it will make maintenance easier.
> And the new tests are great.

This is the goal: https://github.com/umarcor/cobra/blob/args-subtests/args_test.go

umarcor:
@marckhouzam can we please merge this? #1646 has been waiting for three months already...

args_test.go
```diff
@@ -293,6 +471,32 @@ func TestMatchAll(t *testing.T) {
...
+}
+
+func TestExactValidArgs_WithInvalidCount_WithInvalidArgs(t *testing.T) {
+	c := getCommand(ExactValidArgs(3), true)
+	_, err := executeCommand(c, "three", "a", "two")
+	validOnlyWithInvalidArgs(err, t)
+}
+
+func TestExactValidArgs_WithInvalidArgs(t *testing.T) {
+	c := getCommand(ExactValidArgs(3), true)
```
marckhouzam, line 495: Based on the test title, we should probably specify 2 args here.

args_test.go
```diff
@@ -293,6 +471,32 @@ func TestMatchAll(t *testing.T) {
...
+}
+
+func TestExactValidArgs_WithInvalidCount(t *testing.T) {
+	c := getCommand(ExactValidArgs(2), false)
+	_, err := executeCommand(c, "three", "one", "two")
+	exactArgsWithInvalidCount(err, t)
+}
+
+func TestExactValidArgs_WithInvalidCount_WithInvalidArgs(t *testing.T) {
+	c := getCommand(ExactValidArgs(3), true)
```
marckhouzam, line 489: Based on the test title, I believe we want to specify the wrong number of args and the wrong values for them. 
 So, we should probably specify 2 args here since we want the next line to be wrong and it specifies 3 args

args_test.go
```diff
@@ -293,6 +471,32 @@ func TestMatchAll(t *testing.T) {
...
+	_, err := executeCommand(c, "three", "a", "two")
+	validOnlyWithInvalidArgs(err, t)
```
marckhouzam, line 491: This should be `exactArgsWithInvalidCount(err, t)` as we should report invalid number first, when both conditions fail.

args_test.go
```diff
@@ -93,102 +93,280 @@ func rangeArgsWithInvalidCount(err error, t *testing.T) {
...
+
+func TestNoArgs_WithValid_WithArgs(t *testing.T) {
+	c := getCommand(NoArgs, true)
+	_, err := executeCommand(c, "one")
 	noArgsWithArgs(err, t)
 }
 
+func TestNoArgs_WithValid_WithInvalidArgs(t *testing.T) {
+	c := getCommand(NoArgs, true)
+	_, err := executeCommand(c, "one")
```
marckhouzam, line 118: Based on the title of the test, the arg here should be "a".  Currently this is the exact same test as the previous one.

marckhouzam (changes requested):
@umarcor I had another pass at this one and there are couple of things I think we need to tweak, but then it will be good to go.

marckhouzam:
I've added this to the 1.6 milestone

args.go
```diff
@@ -119,3 +107,11 @@ func MatchAll(pargs ...PositionalArgs) PositionalArgs {
 		return nil
 	}
 }
+
+// ExactValidArgs returns an error if there are not exactly N positional args OR
+// there are any positional args that are not in the `ValidArgs` field of `Command`
+//
+// Deprecated: use MatchAll(OnlyValidArgs, ExactArgs(n)) instead
+func ExactValidArgs(n int) PositionalArgs {
+	return MatchAll(OnlyValidArgs, ExactArgs(n))
```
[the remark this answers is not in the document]
umarcor: Done in https://github.com/spf13/cobra/pull/1643/commits/b3982ec9bc06d5f8d2fb6b80739e6fd5c6b1755a.

args_test.go
umarcor: Done in https://github.com/spf13/cobra/pull/1643/commits/83365ac14e01b6700bce04b2e73add6654c0b885.

args_test.go
umarcor: Done in https://github.com/spf13/cobra/pull/1643/commits/4e31482378d51181ed0651cd6daf6a6aa2e946ab.

args_test.go
umarcor: Done in https://github.com/spf13/cobra/pull/1643/commits/b3982ec9bc06d5f8d2fb6b80739e6fd5c6b1755a.

args_test.go
umarcor: Done in https://github.com/spf13/cobra/pull/1643/commits/6863083cddaccf6e875723c9facf3e2216a88882.

marckhouzam (approved):
LGTM!

Thanks for your patience and effort @umarcor !

PR #1645, opened 2022-03-23
Changed branch name in the cobra generator link to 'main'

skywalker2909:
I was going through the ```README.md``` file and noticed at the very end when I clicked on ```The COBRA Generator README``` link, it opened the webpage properly but shows a ```Branch not found, redirected to default branch.``` message every time that kind of feels annoying 🙂 

image

Have updated the link in the ```README.md``` with the correct branch name as a ```good first issue``` contribution to this repo 🙂

### DIFF
--- a/README.md
+++ b/README.md
@@ -102,7 +102,7 @@ It can be installed by running:
 go install github.com/spf13/cobra-cli@latest
 ```
 
-For complete details on using the Cobra-CLI generator, please read [The Cobra Generator README](https://github.com/spf13/cobra-cli/blob/master/README.md)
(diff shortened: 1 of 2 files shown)


marckhouzam:
Nice catch @skywalker2909!  We missed that.  While you are at it would you be so kind as to also update the `user_guide.md` file which has the same problem:
https://github.com/spf13/cobra/blob/03c3eb73f8d00531742e232eb311f8cce0cd6225/user_guide.md?plain=1#L35

skywalker2909:
Thanks @marckhouzam. Have made the change in the ```user_guide.md``` file too.

marckhouzam (approved):
Thanks @skywalker2909 !

PR #1656, opened 2022-04-07
Adjustments to documentation

marckhouzam:
I was going through the user-guide which I'm not sure I had read through and through before.
These are small ajustements I noticed were needed.


user_guide.md
```diff
@@ -603,7 +605,7 @@ Did you mean this?
 Run 'hugo --help' for usage.
 ```
 
-Suggestions are automatic based on every subcommand registered and use an implementation of [Levenshtein distance](http://en.wikipedia.org/wiki/Levenshtein_distance). Every registered command that matches a minimum distance of 2 (ignoring case) will be displayed as a suggestion.
+Suggestions are automatic, based on existing subcommands and use an implementation of [Levenshtein distance](http://en.wikipedia.org/wiki/Levenshtein_distance). Every registered command that matches a minimum distance of 2 (ignoring case) will be displayed as a suggestion.
```
johnSchnake, line 608: nit: I'm not an english major and not sure if a comma is appropriate here. It reads best to me if instead of a comma we use an 'and'
`Suggestions are automatic and based on existing commands...`

johnSchnake (approved):
1 nit but find to merge to me.

johnSchnake:
I'll wait 24hr to see if anyone else wants review or modify the nit. Otherwise I'll merge next time I see revisit.

user_guide.md
marckhouzam: Funny because I started out using an `and`, but then when reading the entire sentence, having two `and`s didn't sound right either: "Suggestions are automatic and based... and use..."

How about: 
 "Suggestions are automatically generated based on existing subcommands and use an implementation of Levenshtein distance."

I'll push an update using that formulation and you can let me know if it sounds better.

jpmcb (approved):
Nice! 👍🏼

marckhouzam:
Hmm... Editing online seems to have done a commit that doesn't pass the CLA check. I'll have to clean that up once I'm at a full computer.

jpmcb:
@marckhouzam - Does this need a squash / force push to make the CLA bot happy? I think this is good to go if you want to get this in a good state 👀

marckhouzam:
> @marckhouzam - Does this need a squash / force push to make the CLA bot happy? I think this is good to go if you want to get this in a good state 👀

Thanks, I had forgotten about that.  I forced push and if fixed the CLA bot.

PR #1659, opened 2022-04-09
Add shell completion to flag groups

marckhouzam:
This PR teaches the shell completion logic how to handle the new flag group feature of #1654.

To avoid mucking about in the already complex shell completion logic, I took a different approach.
When we are doing shell completion, we first check flag groups like this:
1. when a flag in a group is present, other flags in the group will be marked required
1. when a flag in a mutually exclusive group is present, other flags in the group will be marked as hidden

Since the completion logic already handles required flags and hidden flags, this approach allows the standard completion logic to automatically behave appropriately for flag groups.

To specific changes are in the last commit of this PR.

@johnSchnake @Luap99

### DIFF
--- a/completions_test.go
+++ b/completions_test.go
@@ -2691,3 +2691,189 @@ func TestFixedCompletions(t *testing.T) {
 		t.Errorf("expected: %q, got: %q", expected, output)
 	}
 }
+
+func TestCompletionForGroupedFlags(t *testing.T) {
+	getCmd := func() *Command {
+		rootCmd := &Command{
+			Use: "root",
+			Run: emptyRun,
+		}
+		childCmd := &Command{
+			Use: "child",
+			ValidArgsFunction: func(cmd *Command, args []string, toComplete string) ([]string, ShellCompDirective) {
+				return []string{"subArg"}, ShellCompDirectiveNoFileComp
+			},
+			Run: emptyRun,
+		}
+		rootCmd.AddCommand(childCmd)
+
+		rootCmd.PersistentFlags().Int("ingroup1", -1, "ingroup1")
+		rootCmd.PersistentFlags().String("ingroup2", "", "ingroup2")
+
+		childCmd.Flags().Bool("ingroup3", false, "ingroup3")
+		childCmd.Flags().Bool("nogroup", false, "nogroup")
+
+		// Add flags to a group
+		childCmd.MarkFlagsRequiredTogether("ingroup1", "ingroup2", "ingroup3")
+
+		return rootCmd
+	}
+
+	// Each test case uses a unique command from the function above.
+	testcases := []struct {
+		desc           string
+		args           []string
+		expectedOutput string
+	}{
+		{
+			desc: "flags in group not suggested without - prefix",
+			args: []string{"child", ""},
+			expectedOutput: strings.Join([]string{
+				"subArg",
+				":4",
+				"Completion ended with directive: ShellCompDirectiveNoFileComp", ""}, "\n"),
+		},
+		{
+			desc: "flags in group suggested with - prefix",
+			args: []string{"child", "-"},
+			expectedOutput: strings.Join([]string{
+				"--ingroup1",
+				"--ingroup2",
+				"--ingroup3",
+				"--nogroup",
+				":4",
+				"Completion ended with directive: ShellCompDirectiveNoFileComp", ""}, "\n"),
+		},
+		{
+			desc: "when flag in group present, other flags in group suggested even without - prefix",
+			args: []string{"child", "--ingroup2", "value", ""},
+			expectedOutput: strings.Join([]string{
+				"--ingroup1",
+				"--ingroup3",
+				"subArg",
+				":4",
+				"Completion ended with directive: ShellCompDirectiveNoFileComp", ""}, "\n"),
+		},
+		{
+			desc: "when all flags in group present, flags not suggested without - prefix",
+			args: []string{"child", "--ingroup1", "8", "--ingroup2", "value2", "--ingroup3", ""},
+			expectedOutput: strings.Join([]string{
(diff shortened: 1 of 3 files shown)


johnSchnake:
I'll take a look at this on Monday; from me a review on shell completions may be more a learning exercise though so I'd like if someone else could speak to the greater design. Seems like the idea of marking hidden/required is consistent with how the other required flag annotation worked.

Does marking it hidden at this stop not cause any sort of problem if the 'help' has to be printed? e.g. will the hidden flags not be shown in a situation where they should be? Or is it only in the course of autocompletion and out of band from normal functionality?

flag_groups.go
```diff
@@ -157,3 +157,51 @@ func sortedKeys(m map[string]map[string]bool) []string {
...
+	})
+
+	// If a flag that is part of a group is present, we make all the other flags
+	// of that group required so that the shell completion suggests them automatically
+	for flagList, flagnameAndStatus := range groupStatus {
+		for _, isSet := range flagnameAndStatus {
+			if isSet {
+				// One of the flags of the group is set, mark the other ones as required
+				for _, fName := range strings.Split(flagList, " ") {
+					_ = c.MarkFlagRequired(fName)
```
johnSchnake, line 184: Same issue regarding persistent/local flags mixed; may end up with an error here.

flag_groups.go
```diff
@@ -157,3 +157,51 @@ func sortedKeys(m map[string]map[string]bool) []string {
...
+// This allows the standard completion logic to behave appropriately for flag groups
+func (c *Command) enforceFlagGroupsForCompletion() {
+	if c.DisableFlagParsing {
+		return
+	}
+
+	groupStatus := map[string]map[string]bool{}
+	mutuallyExclusiveGroupStatus := map[string]map[string]bool{}
+	c.Flags().VisitAll(func(pflag *flag.Flag) {
+		processFlagForGroupAnnotation(pflag, requiredAsGroup, groupStatus)
```
johnSchnake, line 173: So the way I just modified the other PR was that I pass the flagset into this method and the resulting groupStatus map only has the flag groups of interest.

So by incorporating the new code you may avoid the other problem areas I noted.

completions_test.go
```diff
@@ -2691,3 +2691,171 @@ func TestFixedCompletions(t *testing.T) {
...
+	expected = strings.Join([]string{
+		"--group1-1", // Should be repeated since it is a slice
+		"--flag2",
+		":4",
+		"Completion ended with directive: ShellCompDirectiveNoFileComp", ""}, "\n")
+
+	if output != expected {
+		t.Errorf("expected: %q, got: %q", expected, output)
+	}
+}
```
johnSchnake, line 2861: We should probably include a test case for the persistent/local mixing and ensure it behaves on both the parent and subcommand.

On the parent (which is missing the subcommands local flags) the flag groups shouldn't apply at all.

On the subcommand things should behave as normal.

flag_groups.go
```diff
@@ -157,3 +157,51 @@ func sortedKeys(m map[string]map[string]bool) []string {
...
+	// flags of that group so the shell completion does not suggest them
+	for flagList, flagnameAndStatus := range mutuallyExclusiveGroupStatus {
+		for flagName, isSet := range flagnameAndStatus {
+			if isSet {
+				// One of the flags of the mutually exclusive group is set, mark the other ones as hidden
+				// Don't mark the flag that is already set as hidden because it may be an
+				// array or slice flag and therefore must continue being suggested
+				for _, fName := range strings.Split(flagList, " ") {
+					if fName != flagName {
+						flag := c.Flags().Lookup(fName)
```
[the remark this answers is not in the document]
marckhouzam: I think I'm safe in this case because thanks to your new `hasAllFlags()` I will only be getting the list of flags that are all part of the command.  But I guess we might as well be safe and add the check, it won't hurt.

flag_groups.go
marckhouzam: Here, even if there is an error, we can just ignore it since all it will mean is that the non-existent flag couldn't be marked as required.

flag_groups.go
marckhouzam: Ha!  I should have read all your comments first before replying 😄 
Ok then, no need to add the nil check

completions_test.go
marckhouzam: Will do

marckhouzam:
I meant to answer this earlier but got side-tracked.

> Does marking it hidden at this stop not cause any sort of problem if the 'help' has to be printed? e.g. will the hidden flags not be shown in a situation where they should be? Or is it only in the course of autocompletion and out of band from normal functionality?

Right to the latter.  We are only calling `enforceFlagGroupsForCompletion()` when the `__complete` command is being called and is computing completion choices, so it will not affect any other scenario.

marckhouzam:
Ok, this is updated and two new tests have been added as suggested by @johnSchnake.
I've also changed the new tests to be table-based like John did.  It is much nicer to read.
(I hope to eventually move all the completion tests to a table-based format in a separate PR.)

johnSchnake:
Code, tests, and approach LGTM but tbh I've never used completions via cobra. @Luap99 , would you mind taking a look and trying out (just since you mentioned it on the originating PR).

Otherwise is fine with me to merge this.

marckhouzam:
As we are gearing up for the 1.5 release which will contain the eagerly anticipated flag-grouping feature of @johnSchnake, I suggest we merge this PR as well to complete (no pun intended 😆 )  the feature.

@jpmcb are you ok with me merging this?

jpmcb (approved):
Looks good and seems like previous conversations are resolved. I haven't had _too_ many eyes on the flag group stuff, but I'm good getting this in for the 1.5 release.

PR #1664, opened 2022-04-12
Do not generate docs for hidden flags

### LINKED ISSUE #1663: Generated documentation includes hidden flags for commands
I am trying to generate documentation for a CLI tool that has deprecated (and hidden) command flags, and I believe the generated documentation should not contain these flags unless they are specifically set to not be hidden after being deprecated. This is the behaviour that the generation follows for hidden commands, but not hidden flags on a non-hidden command.

kevinbarbour:
fixes #1663

### DIFF
--- a/doc/yaml_docs.go
+++ b/doc/yaml_docs.go
@@ -149,6 +149,10 @@ func genFlagResult(flags *pflag.FlagSet) []cmdOption {
 	var result []cmdOption
 
 	flags.VisitAll(func(flag *pflag.Flag) {
+		if flag.Hidden {
+			return
+		}
+
 		// Todo, when we mark a shorthand is deprecated, but specify an empty message.
 		// The flag.ShorthandDeprecated is empty as the shorthand is deprecated.
 		// Using len(flag.ShorthandDeprecated) > 0 can't handle this, others are ok.


jpmcb (changes requested):
This only contains the change for docs generated in `yaml`. We should do this for all docs generation, including man page docs, markdown docs, and also the rest docs.

kevinbarbour:
@jpmcb The other doc generators already work this way.

man doc generation skips hidden flags:
https://github.com/spf13/cobra/blob/9e88759b19cd358e9af6f0d778b9a23990cf7f51/doc/man_docs.go#L160

markdown and rest docs use `flags.PrintDefaults()` in pflag which skips hidden flags.

kevinbarbour:
@jpmcb Any plans to merge this one or should I go ahead and close it out?

doc/yaml_docs.go
```diff
@@ -149,6 +149,10 @@ func genFlagResult(flags *pflag.FlagSet) []cmdOption {
 	var result []cmdOption
 
 	flags.VisitAll(func(flag *pflag.Flag) {
+		if flag.Hidden {
```
jpmcb, line 152:
```suggestion
		if len(flag.Deprecated) > 0 || flag.Hidden { 
```

Let's do the same in the `man` doc generation

smottt:
We could really use this bugfix. @kevinbarbour any chance you can fix it up? If not, I'm happy to open a new PR.

PR #1665, opened 2022-04-13
Wrap printf tab with quotes

PapaCharlie:
Without this, slightly older versions of zsh fail to correctly parse the output of the __complete
command. Tested that with zsh 5.0.2 and zsh 5.8. Since this is just correctly quoting the output of
a command, it shouldn't cause any compatibility issues.

### DIFF
--- a/zsh_completions.go
+++ b/zsh_completions.go
@@ -171,7 +171,7 @@ _%[1]s()
             # We first need to escape any : as part of the completion itself.
             comp=${comp//:/\\:}
 
-            local tab=$(printf '\t')
+            local tab="$(printf '\t')"
             comp=${comp//$tab/:}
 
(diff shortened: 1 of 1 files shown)


PapaCharlie:
Agh yes sorry looks like I accidentally created the commit with my work account. Hold on

PapaCharlie:
OK that should have fixed it

marckhouzam (approved):
This is great!  The problem was actually brought up for `kubectl` here https://github.com/kubernetes/kubernetes/issues/105587 but no issue or PR was posted to Cobra.
I've confirmed this PR fixes the issue.  Thanks @PapaCharlie !  And you fix is even better.

marckhouzam:
Since this is XS, I'm going to merge right away.

PR #1668, opened 2022-04-14
Fix ZSH completion to correctly parse the directive and remove the extra lines from output

### LINKED ISSUE #1667: ZSH completions parse directive and debug line removal incorrectly
The ZSH completion currently tries to parse the directive "0:" from last line of the output text, however this isn't valid since this data is actually on the second to last line and the last line has more information which directive was output. The output ends like this:

```
:4
```

The failure to parse the directive correctly causes the error code never to be parsed correctly and the output always has these two extra lines that shouldn't be there.

burmanm:
Fixes #1667 

Reads the directive while parsing all the lines and then removes the lastLine + 3 more characters (0:\ for example), thus making the completion output clean of directive information.

### DIFF
--- a/zsh_completions.go
+++ b/zsh_completions.go
@@ -135,22 +135,26 @@ _%[1]s()
     out=$(eval ${requestComp} 2>/dev/null)
     __%[1]s_debug "completion output: ${out}"
 
-    # Extract the directive integer following a : from the last line
+    # Extract the directive integer following a :
+    directive=0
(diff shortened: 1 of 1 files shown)


marckhouzam:
There was a misunderstanding about the problem which was clarified here https://github.com/spf13/cobra/issues/1667#issuecomment-1099165464

So I believe we don't need this PR.

PR #1530, opened 2021-11-08
Prefer ReplaceAll instead of Replace(..., -1)

WhyNotHugo:
Minor change that gopls pointed out. I think it improves readability a bit.

PR #1675, opened 2022-04-23
fix: generate docs for commands containing slashes

### LINKED ISSUE #1616: bug: cobra doc generator fails when command contains path separators
Due to this line: https://github.com/spf13/cobra/blob/178edbb247f35e466578211dcf5f4892dbbd369b/doc/md_docs.go#L140

I'm unable to generate docs for one of my more complex commands. My subcommands directly match a string value known to my users which contains `/`. For instance:

```
cli run job/first [args]
```

cobra doc generation occurs for the `run` command, but fails for the `job/first` subcommand with a message such as:

```
open docs/cli_run_job/first.md: no such file or directory
```

I think a simple fix would be to include replacement of `/` with `_` when determining the filenames.

jimschubert:
Previously, a command such as:

    &cobra.Command{Use: "run/first [OPTIONS] arg1 arg2"}

would fail generation with an error "no such file or directory" because the generating code would expect some directory ending in `run`.

This PR moves from using `strings.Replace` to `strings.Replacer` which supports mapping multiple replacements in a single invocation. Most of the code can use the utility function `cleanCommandName`. However, manpages use `-` rather than `_` for the seperator so the replacer is defined inline.

I also found that the private genMan function doesn't support passing the custom separator defined in GenManTreeOptions, which is possibly another bug as the SEE ALSO output would be inconsistent whenever the user defines a custom separator. I didn't resolve this issue in this PR because I'm unfamiliar with the manpage format.

Please review the outputs in my example repo: https://github.com/jimschubert/cobraslash_example

fixes #1616


umarcor:
See https://github.com/spf13/cobra/pull/1530.

jimschubert:
If #1530 gets merged first, I have no problem fixing the merge conflicts in this PR. But my PR is solving a different issue.

jpmcb:
Thanks fro this! #1530 has merged - can you rebase and get those changes and we can look at this? Thanks much!!

jimschubert:
@jpmcb thanks. I've just rebased and retested. It's good to go.

jimschubert:
Not stale, just waiting on merge.

jimschubert:
it's no problem. I know how it can be :)

The conflicts were actually just a couple of copyright headers. It's all cleaned up now.

Issue #1680, opened 2022-04-27
bash completion v2 is much slower than v1


scop:
I have a work-in-progress completion that returns 788 items. With cobra's bash completion v1, the results are pretty much instant, but with v2 they take seconds.

Results for v1 vs v2 completions generated by the same executable, the only difference being that the v1 completion is output "manually", and the v2 using cobra's default completion emitter, both built with cobra v1.4.0, both shell files manually modified so that the contents of `__start_sshi` is wrapped in `time { ... }` for benchmarking, v1:
```shellsession
$ sshi ssh <tab>
real	0m0,127s
user	0m0,089s
sys	0m0,040s

Display all 788 possibilities? (y or n)^C
```

v2:
```shellsession
$ sshi ssh <tab>
real	0m2,285s
user	0m1,669s
sys	0m0,713s

Display all 788 possibilities? (y or n)^C
```

For v2, that 2.2 seconds is pretty much the best case. Times like 3.3 and 3.8 seconds are not that rare. v1 is about always < 0.2 seconds.

scop:
The project mentioned is https://github.com/aakso/ssh-inscribe, PR implementing (implicit) switch from the old completions to the V2 ones emitted by the default is https://github.com/aakso/ssh-inscribe/pull/21, completion giving those ~800 completions in a test setup of mine is https://github.com/aakso/ssh-inscribe/pull/21 (but the slowness isn't creating the list, but processing it in bash).

Generating the V2 completions with `--no-descriptions` shaves a few tenths of a second, but it's still ~2 seconds.

marckhouzam:
The v1 script does have an avantage over the v2 one in the fact that the static completions are hard-coded into the script.  With v2, as you mention, your binary is explicitly called to get the completions.  Can you run
```
time sshi __complete ssh ''
```
to see how long it takes your `sshi` binary to compute the completions?  If it takes multiple seconds we'll have to look at why.

scop:
```shellsession
$ time sshi __complete ssh '' | wc -l
Completion ended with directive: ShellCompDirectiveNoFileComp
810

real	0m0,048s
user	0m0,030s
sys	0m0,029s
```

scop:
Thanks for looking into the low hanging fruit PR's so quickly. I have a slightly more involved, promising looking one in draft stage locally, will try to get it cleaned up and sliced/diced into smaller easier reviewable commits later today.

marckhouzam:
[...]

PR #1681, opened 2022-04-27
style(bash-v2): out is not an array variable, do not refer to it as such

scop:
Even though this to my surprise works, it doesn't accomplish anything but some confusion. Remove it.


scop:
Just something trivial to get out of the way for other likely improvements related to #1680.

marckhouzam (approved):
Hmmm... I'm not sure why I used `out` as an array variable.  It might have started as an array in the early iterations of the script and then I didn't notice it wasn't one anymore.

But thanks @scop for cleaning that up.
I have tested with the cobra-completion-testing repo and with helm and it looks great.

PR #1682, opened 2022-04-27
perf(bash-v2): use backslash escape string expansion for tab

scop:
Using a command substitution, i.e. a subshell, with `printf` is expensive for this purpose. For example `__*_format_comp_descriptions` is run once for each completion candidate; the expense adds up and shows when there are a lot of them.

This shaves off roughly 25% of the times I receive for my test case in #1680 (~2 seconds before, ~1.5 after).

### DIFF
--- a/bash_completionsV2.go
+++ b/bash_completionsV2.go
@@ -154,8 +154,7 @@ __%[1]s_handle_completion_types() {
         # If the user requested inserting one completion at a time, or all
         # completions at once on the command-line we must remove the descriptions.
         # https://github.com/spf13/cobra/issues/1508
-        local tab comp
-        tab=$(printf '\t')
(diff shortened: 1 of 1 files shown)


marckhouzam:
Thanks @scop!  I hadn't thought about the performance impact of calling the subshell, especially since the tab is created in a loop.  I would have suggested taking it out of the loop, but with your PR, it takes basically no time at all!  I wasn't familiar with this escape sequence.

I'll run a couple of tests and get this merged.

marckhouzam (approved):
I've tested this improvement with https://github.com/marckhouzam/cobra-completion-testing (after disabling the centos tests that no longer build), and also with helm.  Everything looks great.

LGTM

PR #1683, opened 2022-04-28
perf(bash-v2): standard completion optimizations

scop:
Refactor to remove two loops over the entire list of candidates.

Format descriptions only for completions that are actually going to be displayed, instead of for all candidates.

Format descriptions inline in completions array, removing need for a command substitution/subshell and a printf escape per displayed completion.

These changes shave off a second or so from my case at hand in #1680, it was roughly ~1.7s before, is now ~0.7s after.
The diff is largish, but the bulk of it is in `__%[1]s_format_comp_descriptions` due to indentation changes, there aren't that many lines actually changed in it either. `git show -w` helps with the review.

Caveat: I have not actually tested generating completions with this code. I worked on the improvements by tweaking an existing emitted completion shell file, then manually ported the changes to the template in Go code here. Hopefully I didn't introduce bugs while doing that. (The code compiles, and test suite fails just like it did before this change, no regressions noticed.)

### DIFF
--- a/bash_completionsV2.go
+++ b/bash_completionsV2.go
@@ -177,40 +177,28 @@ __%[1]s_handle_standard_completion_case() {
     local tab=$'\t' comp
 
     local longest=0
+    local compline
     # Look for the longest completion so that we can format things nicely
-    while IFS='' read -r comp; do
+    while IFS='' read -r compline; do
         # Strip any description before checking the length
-        comp=${comp%%%%$tab*}
+        comp=${compline%%$tab*}
         # Only consider the completions that match
         comp=$(compgen -W "$comp" -- "$cur")
+        [[ -z $comp ]] && continue
+        COMPREPLY+=("$compline")
         if ((${#comp}>longest)); then
             longest=${#comp}
         fi
     done < <(printf "%%s\n" "${out}")
 
-    local completions=()
-    while IFS='' read -r comp; do
-        if [ -z "$comp" ]; then
-            continue
-        fi
-
-        __%[1]s_debug "Original comp: $comp"
-        comp="$(__%[1]s_format_comp_descriptions "$comp" "$longest")"
-        __%[1]s_debug "Final comp: $comp"
-        completions+=("$comp")
-    done < <(printf "%%s\n" "${out}")
-
-    while IFS='' read -r comp; do
-        COMPREPLY+=("$comp")
-    done < <(compgen -W "${completions[*]}" -- "$cur")
-
     # If there is a single completion left, remove the description text
     if [ ${#COMPREPLY[*]} -eq 1 ]; then
         __%[1]s_debug "COMPREPLY[0]: ${COMPREPLY[0]}"
-        comp="${COMPREPLY[0]%%%% *}"
+        comp="${COMPREPLY[0]%%$tab*}"
         __%[1]s_debug "Removed description from single completion, which is now: ${comp}"
-        COMPREPLY=()
-        COMPREPLY+=("$comp")
+        COMPREPLY[0]=$comp
+    else # Format the descriptions
+        __sshi_format_comp_descriptions $longest
     fi
 }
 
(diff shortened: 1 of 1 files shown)


scop:
(Force-pushed just to generalize the commit message some.)

marckhouzam:
Thanks @scop !  I will have a look over the weekend.

bash_completionsV2.go
```diff
@@ -177,40 +177,28 @@ __%[1]s_handle_standard_completion_case() {
...
     local longest=0
+    local compline
     # Look for the longest completion so that we can format things nicely
-    while IFS='' read -r comp; do
+    while IFS='' read -r compline; do
         # Strip any description before checking the length
-        comp=${comp%%%%$tab*}
+        comp=${compline%%$tab*}
         # Only consider the completions that match
         comp=$(compgen -W "$comp" -- "$cur")
```
scop, line 186: Doing this for every candidate separately would be something to look into getting rid of next, but I suggest we postpone looking into it as far as this PR is concerned and not get delayed because of it here.

bash_completionsV2.go
```diff
@@ -177,40 +177,28 @@ __%[1]s_handle_standard_completion_case() {
     local tab=$'\t' comp
 
     local longest=0
+    local compline
     # Look for the longest completion so that we can format things nicely
-    while IFS='' read -r comp; do
+    while IFS='' read -r compline; do
         # Strip any description before checking the length
-        comp=${comp%%%%$tab*}
+        comp=${compline%%$tab*}
```
marckhouzam, line 184: Because `%` is a special character in Go's `Sprintf()` function, we need to escape it by making it double, so a single `%` must become `%%` but in our case we need a double `%%` in the final bash script so as to remove the description that follows the first tab character (in case the description contains a tab itself).  So we need to put 4 `%%%%` to get the final double `%%`.

bash_completionsV2.go
```diff
@@ -177,40 +177,28 @@ __%[1]s_handle_standard_completion_case() {
...
-
-    while IFS='' read -r comp; do
-        COMPREPLY+=("$comp")
-    done < <(compgen -W "${completions[*]}" -- "$cur")
-
     # If there is a single completion left, remove the description text
     if [ ${#COMPREPLY[*]} -eq 1 ]; then
         __%[1]s_debug "COMPREPLY[0]: ${COMPREPLY[0]}"
-        comp="${COMPREPLY[0]%%%% *}"
+        comp="${COMPREPLY[0]%%$tab*}"
```
marckhouzam, line 197: Here too, I think we should have `%%%%`

bash_completionsV2.go
```diff
@@ -177,40 +177,28 @@ __%[1]s_handle_standard_completion_case() {
...
         __%[1]s_debug "Removed description from single completion, which is now: ${comp}"
-        COMPREPLY=()
-        COMPREPLY+=("$comp")
+        COMPREPLY[0]=$comp
+    else # Format the descriptions
+        __sshi_format_comp_descriptions $longest
```
marckhouzam, line 201: Not `__sshi_*` but `__%[1]s_format_comp_descriptions`.
I did the same mistake a couple of time with `__helm_*` when I first wrote this code 😄

bash_completionsV2.go
```diff
@@ -230,43 +218,47 @@ __%[1]s_handle_special_char()
...
-        fi
+    local comp desc maxdesclength
+    local longest=$1
+
+    local i ci
+    for ci in ${!COMPREPLY[*]}; do
+        comp=${COMPREPLY[ci]}
+        # Properly format the description string which follows a tab character if there is one
+        if [[ "$comp" == *$tab* ]]; then
+            __sshi_debug "Original comp: $comp"
```
marckhouzam, line 229: `__sshi_*` -> `__%[1]s_debug`

bash_completionsV2.go
```diff
@@ -230,43 +218,47 @@ __%[1]s_handle_special_char()
...
+            # truncate the descriptions that are too long for the shell width
+            if [ $maxdesclength -gt 0 ]; then
+                if [ ${#desc} -gt $maxdesclength ]; then
+                    desc=${desc:0:$(( maxdesclength - 1 ))}
+                    desc+="…"
+                fi
+                comp+="  ($desc)"
+            fi
+            COMPREPLY[ci]=$comp
+            __sshi_debug "Final comp: $comp"
```
marckhouzam, line 259: `__sshi_*` -> `__%[1]s_debug`

marckhouzam (changes requested):
Things are getting faster 🎉 

I didn't finish the review yet but the following changes are necessary no matter what, so I'm posting them now.

scop:
Thanks! Addressed. Quite a bit more care would have been required for the method I used to do make the submitted change. Will have a look at the other open PR if it had similar issues.

bash_completionsV2.go
```diff
@@ -230,43 +218,47 @@ __%[1]s_handle_special_char()
...
-            done
-        else
-            # Don't pad the descriptions so we can fit more text after the completion
-            maxdesclength=$(( COLUMNS - ${#comp} - 4 ))
-        fi
+    local comp desc maxdesclength
+    local longest=$1
+
+    local i ci
+    for ci in ${!COMPREPLY[*]}; do
```
marckhouzam, line 225: I wasn't familiar with this syntax.  From what I understand it loops over the array but put the index in $ci.
For my education why is that better than something like `for ((ci = 0 ; ci < ${#COMPREPLY} ; ci++)); do` ?

When trying it I notice that $COMPREPLY has a bunch of empty elements which your syntax ignores automatically.  Is that why you choose that syntax?

bash_completionsV2.go
```diff
@@ -230,43 +218,47 @@ __%[1]s_handle_special_char()
...
+            else
+                # Don't pad the descriptions so we can fit more text after the completion
+                maxdesclength=$(( COLUMNS - ${#comp} - 4 ))
             fi
-            comp+="  ($desc)"
-        fi
-    fi
 
-    # Must use printf to escape all special characters
-    printf "%%q" "${comp}"
```
marckhouzam, line 269: Don't we need to keep this?

marckhouzam: Oh, right, we don't need it as we store the result directly in $COMPREPLY instead of outputting it to be consumed by the calling function as before.  Sorry.

marckhouzam (approved):
This is great!

The https://github.com/marckhouzam/cobra-completion-testing tests pass and we can see a nice improvement in speed using https://github.com/marckhouzam/cobra-completion-testing/pull/15.  On my machine I see :
- 1000 completions without descriptions 2.688s sec on master -> 1.073 sec with the PR
- 1000 completions with descriptions 2.965s sec on master -> 1.245 sec with the PR

The helm completion tests also pass and I did some manual testing too.

Excellent work @scop, thank you!

I will go ahead and merge to continue working on the other speed improvements.

bash_completionsV2.go
scop: The niceties of that syntax boil down to that it excludes unassigned indices. Sparse arrays "just work" (if they work in the context at hand in other respects in the first place), no need to worry about accessing unset data in `nounset` mode or to wonder if all desired items are actually iterated over, for example:

```shellsession
$ shopt -po nounset
set -o nounset
$ a=(1 2 3)
$ unset 'a[1]'
$ for i in ${!a[*]}; do echo "${a[i]}"; done
1
3
$ for ((i = 0; i < ${#a[*]}; i++)); do echo "${a[i]}"; done
1
bash: a[i]: unbound variable
# oops!

$ shopt -po nounset
set +o nounset
$ for ((i = 0; i < ${#a[*]}; i++)); do echo "${a[i]}"; done
1

# oops!
```

I suppose in this particular case it's not that likely that `COMPREPLY` would actually contain unassigned indices, so it doesn't matter that much which approach would be used. But I find the syntax I used here more succinct, and if one can rely on it being used to iterate over arrays, the convenience of being able to remove array elements by simply `unset`ing them like in the example above is a nice bonus.

Also, that's the syntax for iterating over associative array keys, so there's a consistency bonus attached as well.

marckhouzam: Thanks for the explanation. I leaned something new 👍

PR #1684, opened 2022-04-28
style(bash): out is not an array variable, do not refer to it as such

scop:
For legacy bash completions, similarly as commit 4f0facbcee0aadb87179d3fed80e92709de491a9 is for bash completions v2.

As a side effect, fixes test suite with shellcheck 0.8.0 installed; apparently the 0.7.0 that's in GitHub Actions' ubuntu-latest at the
moment does not flag the array quoting related issue that was provoked from 0.8.0 before this change.

### DIFF
--- a/bash_completions.go
+++ b/bash_completions.go
@@ -99,7 +99,7 @@ __%[1]s_handle_go_custom_completion()
         directive=0
     fi
     __%[1]s_debug "${FUNCNAME[0]}: the completion directive is: ${directive}"
-    __%[1]s_debug "${FUNCNAME[0]}: the completions are: ${out[*]}"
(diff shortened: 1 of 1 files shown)


marckhouzam (approved):
LGTM
I don't know why I treated `out` as an array for some places and not others.  It may have been that the variable used to be an array during previous iterations of the Go completion feature and I didn't fix all usage of it.

Finally, the cobra-completion-testing repo passes and the helm completion tests pass as well.

Thanks @scop !

And I can confirm that before this PR the tests started to fail after I upgraded to shellcheck v0.8.0, but they pass with the PR.  Note that I had to run `go clean -testcache` for the test to actually re-run after upgrading shellcheck.

PR #1685, opened 2022-04-29
Check if a command is 'hidden' before adding length to usage

vicmarbev:
This PR is related to issue #1272. When adding or removing a command, only take into account commands that are not hidden when calculating length for the usage.

### DIFF
--- a/command.go
+++ b/command.go
@@ -1162,18 +1162,20 @@ func (c *Command) AddCommand(cmds ...*Command) {
 			panic("Command can't be a child of itself")
 		}
 		cmds[i].parent = c
-		// update max lengths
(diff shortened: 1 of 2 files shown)


marckhouzam:
Thanks for the contribution @vicmarbev !
However we already have #1632  which surprisingly was not mentioned in issue #1272. I'll go mention it in the issue.

We are having some backward-compatibility concerns about the fix however, as stated here https://github.com/spf13/cobra/pull/1632#issuecomment-1082478764

Maybe you can give your opinion?

vicmarbev:
No problem! I'll close the PR.

PR #1686, opened 2022-04-30
perf(bash-v2): short-circuit descriptionless candidate lists

scop:
If the list of candidates has no descriptions, short circuit all the description processing logic, basically just do a `compgen -W` for the whole list and be done with it.

We could conceivably do some optimizations like this and more when generating the completions with `--no-descriptions` in Go code, by omitting some parts we know won't be needed, or doing some things differently. But doing it this way in bash, the improvements are available also to completions generated with descriptions enabled when they are invoked for completion cases that produce no descriptions. The result after this for descriptionless entries seems fast enough so it seems there's no immediate need to look into doing that.

This alone pretty much eliminates all the slowness related to my specific use case in #1680, bringing it down to ~0.07s (yes, there _is_ a zero on the right of the period), so for the time being I'm not inclined to look into further optimizations at least for the code path where there are no descriptions. Related to #1683, but I suggest merging both as they affect different scenarios.

Same caveat as in #1683 about testing and the way the change was made applies here.

### DIFF
--- a/bash_completionsV2.go
+++ b/bash_completionsV2.go
@@ -176,6 +176,14 @@ __%[1]s_handle_completion_types() {
 __%[1]s_handle_standard_completion_case() {
     local tab=$'\t' comp
 
+    # Short circuit to optimize if we don't have descriptions
+    if [[ $out != *$tab* ]]; then
+        while IFS='' read -r comp; do
+            COMPREPLY+=("$comp")
+        done < <(IFS=$'\n' compgen -W "$out" -- "$cur")
+        return 0
+    fi
+
     local longest=0
     # Look for the longest completion so that we can format things nicely
     while IFS='' read -r comp; do


bash_completionsV2.go
```diff
@@ -176,6 +176,14 @@ __%[1]s_handle_completion_types() {
 __%[1]s_handle_standard_completion_case() {
     local tab=$'\t' comp
 
+    # Short circuit to optimize if we don't have descriptions
+    if [[ $out != *$tab* ]]; then
```
marckhouzam, line 180: Nice way to check if there are no descriptions!

bash_completionsV2.go
```diff
@@ -176,6 +176,14 @@ __%[1]s_handle_completion_types() {
...
+        while IFS='' read -r comp; do
+            COMPREPLY+=("$comp")
+        done < <(IFS=$'\n' compgen -W "$out" -- "$cur")
```
marckhouzam, line 183: Why the `IFS=$'\n'` here?  We don't use it for the bash v1 script which uses similar loops.  Do we need to?

marckhouzam (approved):
This looks good and does speed up the bash completion v2 impressively when there are no descriptions.  Having no descriptions will be a common case I think because:
- some programs will not choose to support descriptions
- some users will disable them (if allowed by the program)
- more often then not, custom completions provided by the program will not include descriptions

I just want to see why the `IFS=$'\n'` was used before merging.

marckhouzam:
Oh, and according to https://github.com/marckhouzam/cobra-completion-testing/pull/15
Before this PR, 1000 completions without descriptions took 1.146 secs.
With this PR they take 0.054 secs!  That is an insane speedup!  But it also shows that dealing with descriptions still costs us a lot...

bash_completionsV2.go
scop: It's there to protect spaces in individual completions:
```shellsession
$ compgen -W $'foo bar\nquux baz\n' -- f
foo
$ IFS=$'\n' compgen -W $'foo bar\nquux baz\n' -- f
foo bar
```

marckhouzam: Ok, seems better.  Thanks

PR #1690, opened 2022-05-03
fix: Don't complete _command on zsh

twpayne:
As reported by @VorpalBlade in https://github.com/twpayne/chezmoi/issues/2047:

Previously the generated zsh completion script started with the line

    #compdef _<command> <command>

where \<command\> is the command that the zsh completion script is generated for.

This enabled completions for both \<command> and _\<command>, but _\<command> is the completion function itself and should not be completed. Furthermore, attempting to autocomplete _<command> (e.g. typing "_\<command>\<Space>\<Tab>" in a zsh shell) causes zsh to hang.

This commit fixes the #compdef line to only complete \<command>, not _\<command>.

Co-authored-by: Arvid Norlander <VorpalBlade@users.noreply.github.com>

### DIFF
--- a/zsh_completions.go
+++ b/zsh_completions.go
@@ -75,7 +75,7 @@ func genZshComp(buf io.StringWriter, name string, includeDesc bool) {
 	if !includeDesc {
 		compCmd = ShellCompNoDescRequestCmd
 	}
-	WriteStringAndCheck(buf, fmt.Sprintf(`#compdef _%[1]s %[1]s
+	WriteStringAndCheck(buf, fmt.Sprintf(`#compdef %[1]s
 
 # zsh completion for %-36[1]s -*- shell-script -*-
 


marckhouzam:
Wow 🤯  
Looking at the documentation for `compdef` I see it uses the form: `compdef <function> <command>`, which is why Cobra uses it like we do.  However, the doc says that the `#compdef` instruction is different and does not specify the `<function>` parameter.

In fact the first zsh PR for Cobra, #497, did use `#compdef <command>`, but the next version of #646, moved to the current `#compdef <function> <command>` and then I came along and just kept it as is without investigating.

I'll run some tests to confirm, but this seems to be right.

Thanks for bringing it back to use @twpayne !

twpayne:
Any updates on this? My local testing indicates that this PR fixes a genuine bug.

marckhouzam:
I plan on getting to this one before the release to run a couple of tests before merging.

marckhouzam (approved):
I've tested this with zsh 5.9 and zsh 5.1 (to get a range of versions covered) using a modified completion script for helm and it does work to use `#compdef helm` as the first line.

I've also look at a few completions script in https://github.com/zsh-users/zsh-completions/tree/master/src and do see that they all use the `#compdef <program>` form.

Thanks again @twpayne .

PR #1691, opened 2022-05-04
fix(bash-v2): skip empty completions when filtering descriptions

scop:
`read` gives a last null value following a trailing newline.

Regression from fb8031162c2ffab270774f13c6904bb04cbba5a7.

We could drop the `\n` from the feeding `printf` to accomplish the same end result, but I'm planning to submit some related cleanups a bit later that wouldn't play well with that approach, so I went with explicit filtering here.

### DIFF
--- a/bash_completionsV2.go
+++ b/bash_completionsV2.go
@@ -188,6 +188,7 @@ __%[1]s_handle_standard_completion_case() {
     local compline
     # Look for the longest completion so that we can format things nicely
     while IFS='' read -r compline; do
+        [[ -z $compline ]] && continue
         # Strip any description before checking the length
         comp=${compline%%%%$tab*}
         # Only consider the completions that match


marckhouzam:
Could you give a way to reproduce the bug?

scop:
Invoke any completion that has descriptions by hitting just a tab and no prefix characters for the completions. An empty line (the empty completion) before the list of visible suggestions appears.

marckhouzam:
> Invoke any completion that has descriptions by hitting just a tab and no prefix characters for the completions. An empty line (the empty completion) before the list of visible suggestions appears.

Ah yes.  I'll see if I can't add a test to catch this in `cobra-completion-testing`.
Thanks!

marckhouzam (approved):
I had to fight with the tests repo to be able to detect such a bug but I finally got it to work: https://github.com/marckhouzam/cobra-completion-testing/pull/17

Thanks for noticing the problem and fixing it @scop !

LGTM

Issue #798, opened 2018-12-20
cobra should support initializers that return errors


waynr:
It looks like the `execute` function already propagates errors: https://github.com/spf13/cobra/blob/d2d81d9a96e23f0255397222bb0b4e3165e492dc/command.go#L675

And so does `Command.ExecuteC`, where `execute` is called: https://github.com/spf13/cobra/blob/d2d81d9a96e23f0255397222bb0b4e3165e492dc/command.go#L805

So it seems like cobra could fairly easily support both initializes that propagate errors and those that don't by adding a `OnInitializeE` function and a corresponding top-level slice of error-returning functions.

I'd be happy to implement this myself if any project maintainers would provide some guidance in terms of what kind of unit tests I should add (if any).

PR #1693, opened 2022-05-09
Add OnInitializeE

fmenezes:
Today we have `cobra.OnInitialize` but unfortunately we don't have an API to return initialization errors.

This aims to add `cobra.OnInitializeE` to accept `func() error` and returns error accordingly.


marckhouzam:
Hi @fmenezes.  Could you provide a description of what this PR does and if it aims to fix an existing issue.
Thank you.

fmenezes:
> Hi @fmenezes.  Could you provide a description of what this PR does and if it aims to fix an existing issue.
> 
> Thank you.

Hi @marckhouzam, my mistake, I completely forgot to add it. It is there now.

Let me know if you want more clarifications.

Thanks for the quick reply.

johnSchnake:
Is there an issue for this?

Also, I noticed in the implementation that you call the `preRunE` then `preRun`; this  is a bit different than the other `xyzE` methods since it is assumed you choose either one or the other, not both. Was there a reason for this particular implementation? It may be the right one, I just wanted context on the choice.

fmenezes:
> Is there an issue for this?
> 
> Also, I noticed in the implementation that you call the `preRunE` then `preRun`; this is a bit different than the other `xyzE` methods since it is assumed you choose either one or the other, not both. Was there a reason for this particular implementation? It may be the right one, I just wanted context on the choice.

@johnSchnake thanks for pointing out, I have no strong preference I can keep it consistent with the same behavior and check if `preRunE` has been set otherwise go with `preRun`.

fmenezes:
I _think_ this PR addresses #798

jpmcb (changes requested):
I need to more fully understand why we need this.

I also haven't used the initializers much before so I'm unsure what this is solving or what user workflow this would work well for that other APIs can't fulfill.

I'm all for propagating errors, but I don't want to flood the API just for the sake of have 1 to 1 `abc` and `abcE` methods.

---

Anyone in the community have a user story or issue that this would solve for you?

fmenezes:
@jpmcb my example lies on https://github.com/mongodb/mongodb-atlas-cli/blob/1c845fa25ad07e5c62266a91de8d14c348caca37/cmd/mongocli/mongocli.go#L100 for mongodb-atlas-cli we have to handle early exits ourselves and now that we are adding tracking capabilities we have to figure out which command was "early exited".

fmenezes:
tagging @jpmcb for copy review

PR #1698, opened 2022-05-17
Added asciidoc generation and associated tests.

bovinemagnet:
Added Asciidoc as a generation target for documentation.

Based off the markdown, additional tests are added to ensure the nested file generation.

### DIFF
--- a/doc/asciidoc_docs.go
+++ b/doc/asciidoc_docs.go
@@ -1,16 +1,3 @@
-//Copyright 2015 Red Hat Inc. All rights reserved.
-//
-// Licensed under the Apache License, Version 2.0 (the "License");
-// you may not use this file except in compliance with the License.
-// You may obtain a copy of the License at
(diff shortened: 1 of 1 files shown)


bovinemagnet:
This change makes use of the following AsciiDoc elements (that differ from a straight MD to AsciiDoc type approach).

Also see uses `xref:` to make it compatible with github asciidocs, Antora asciidocs, and `asciidoc-pdf`  generation.

The initial heading is set to level `=` to ensure it will not  cause parsing errors with Antora (as apposed to the example which would have made it `==`).

All options are wrapped in "text" source blocks, and given a title for the source block, such as:

```
.Option Title
[source,text]
----
  -h, --help   help for command
----
````

bovinemagnet:
I removed the licence that I accidentally copied into the header of the file. to address the concerns of @jpmcb

PR #1702, opened 2022-05-17
style(bash-v2): various cleanups

scop:
Here's a bunch of bash code cleanups I made on the side while working on the performance improvements. See individual commits and messages for details. Some of these are also performance improvements in principle in addition to stylistic ones, but I believe they're not on any hot path so that side is negligible.

Some of these might be a bit opinionated/subjective but read the way I'd personally put them, let me know if you'd rather not make some of these changes and I'll follow suit.


bash_completionsV2.go
```diff
@@ -124,13 +124,12 @@ __%[1]s_process_completion_results() {
...
         $filteringCmd
-    elif [ $((directive & shellCompDirectiveFilterDirs)) -ne 0 ]; then
+    elif (((directive & shellCompDirectiveFilterDirs) != 0)); then
         # File completion for directories only
 
-        # Use printf to strip any trailing newline
         local subdir
-        subdir=$(printf "%%s" "${out}")
-        if [ -n "$subdir" ]; then
+        subdir=${out%%$'\n'}
```
marckhouzam, line 131: Here, in the final bash script, do we want a single or a double `%%`?  If we want a double one to remove all trailing new lines, we need to put four of them in go (`%%%%`)

marckhouzam (commented):
Thanks for sharing your bash expertise!
This all makes sense and all the tests pass.

I'll just wait for the one question about the double `%%`

bash_completionsV2.go
scop: It doesn't make a difference; there are no globs involved, so we are removing just one character (the linefeed) no matter if we do `%` or `%%`. Given this, the former seems a bit cleaner to me here, which is why I chose that option.

There is indeed a small difference in behavior here: the former implementation removed all trailing linefeeds, this removes just one. But I assume we don't have cases where there would be more than one, I chose the simplest available approach enabled by that assumption.

One way to remove multiple with a simple parameter expansion would require use of extended globbing, which we currently do not rely on in the script. bash-completion is a requirement and it does enable `extglob` unconditionally, and I don't see that changing, so in that sense it would be safe to use it. In that case the expansion could be written as `${out%%%%+($'\n')}` (in Go code). Another a bit hacky way which doesn't require extended globbing would be to do `${out//$'\n'}` which is semantically different, but would for practical purposes work just fine here I think. Let me know if you prefer one of these two instead and I'll adjust.

marckhouzam: > It doesn't make a difference; there are no globs involved, so we are removing just one character (the linefeed) no matter if we do % or %%.

Ah yes. I missed that. Thanks.

I agree it should be safe to strip a single linefeed.

marckhouzam (approved):
Thanks once more @scop!

I'll just run a bit of extra manual testing later today, especially for the lesser used directive code paths and will merge after.

marckhouzam:
Just an update that I haven't forgot about this, I'm just having trouble with my setup (waiting on a new computer). I will get back to this as soon as I can.

scop:
Sure, no rush at all as far as I'm concerned, take all the time you like :+1:

marckhouzam:
@scop Finally getting back to this but the PR now has conflicts 😞 Could you rebase when you have a moment?

scop:
Reworked and still untested, hopefully I didn't add too many bugs.

marckhouzam:
> Reworked and still untested, hopefully I didn't add too many bugs.

Thanks @scop. After looking at this one again, I suggest changing our strategy.
Since this is a cleanup and not an actual fix, and that there will be a Cobra release soon, let's play it safe and wait on this one until after the release so we can get some more soak time.

scop:
Rebased.

marckhouzam:
I'm going to merge this improvement now to give it a lot of soak time until the next release.

PR #1707, opened 2022-05-24
Fix minor annoyance of missing '--version' flag from some Help output

fnickels:
In the current code, with a version set, `app help` does not display the switch option `--version`, while `app --help` does.

* Fix missing '--version' flag from help command
  *  with unit tests
* Add vscode to .gitignore

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -37,3 +37,5 @@ bin
 
 .idea/
 *.iml
+
+.vscode
--- a/.gitignore
+++ b/.gitignore
@@ -37,5 +37,3 @@ bin
 
 .idea/
 *.iml
-
-.vscode
--- a/command_test.go
+++ b/command_test.go
@@ -2162,8 +2162,8 @@ func TestSetContextPersistentPreRun(t *testing.T) {
 	}
 }
 
-const VERSION_FLAG = "--version"
-const HELP_FLAG = "--help"
+const VersionFlag = "--version"
+const HelpFlag = "--help"
 
 func TestNoRootRunCommandExecutedWithVersionSet(t *testing.T) {
 	rootCmd := &Command{Use: "root", Version: "1.0.0", Long: "Long description"}
@@ -2175,8 +2175,8 @@ func TestNoRootRunCommandExecutedWithVersionSet(t *testing.T) {
 	}
 
 	checkStringContains(t, output, rootCmd.Long)
-	checkStringContains(t, output, HELP_FLAG)
-	checkStringContains(t, output, VERSION_FLAG)
+	checkStringContains(t, output, HelpFlag)
+	checkStringContains(t, output, VersionFlag)
 }
 
 func TestNoRootRunCommandExecutedWithoutVersionSet(t *testing.T) {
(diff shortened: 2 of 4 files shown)


gusega:
great that you fixed it @fnickels , I also just noticed this thing. Hope it gets merged  soon.

.gitignore
```diff
@@ -37,3 +37,5 @@ bin
 
 .idea/
 *.iml
+
+.vscode
```
marckhouzam, line 41: I guess this makes sense, but:
1. it shouldn't be in this PR
2. it is recommended to use your own global gitignore for those things

My guess is that you didn't realize it sneaked in here.  Could you remove it?

marckhouzam (changes requested):
This is great @fnickels !  Thanks for fixing this.

Everything looks good except one nit.

fnickels:
Will do

command_test.go
```diff
@@ -2161,3 +2161,90 @@ func TestSetContextPersistentPreRun(t *testing.T) {
 		t.Error(err)
 	}
 }
+
+const VERSION_FLAG = "--version"
+const HELP_FLAG = "--help"
```
marckhouzam, line 2166: I just noticed the linter is failing on these two names, it says: "don't use ALL_CAPS in Go names; use CamelCase"

.gitignore
fnickels: Done.  Thanks for suggesting this.  I sort of always knew this was an option, but was too lazy to set it up.

marckhouzam: If it helps: `git config --global core.excludesfile ~/.gitignore`

command_test.go
fnickels: fixed

fnickels:
Yeah, That's what I used.

I sort of knew this existed but have always been too lazy to look it up and
configure it.

Should the `.idea/` reference also be removed from `.gitignore`?



francis…


*Francis J. Nickels III*

Riot Games

***@***.*** | C: +1.310.598.8267

Summoner Name: target1

marckhouzam:
> Should the `.idea/` reference also be removed from `.gitignore`?

There is no official agreement to not put those things in the `.gitignore` for this project, but I at least didn't want to "pollute" your nice PR with those changes.

marckhouzam (approved):
LGTM

Thanks @fnickels !

marckhouzam:
@jpmcb This PR fixes the fact that the `--version` flag was not output by the `help` command.  I don't think that changing help output is a backwards-compatible issue, so I fee this should go in.  Ok with you?

jpmcb (dismissed):
Nice! Just a quick touch up:

- Can you squash your commits and write a single commit message that covers this? (if not, I'd be happy to squash / merge and write one)

marckhouzam:
> Nice! Just a quick touch up:
> 
> * Can you squash your commits and write a single commit message that covers this? (if not, I'd be happy to squash / merge and write one)

I've gone ahead and done it myself.
Thanks @fnickels for the fix!

fnickels:
happy to help

Issue #1708, opened 2022-05-24
Unclear how "Out" and "Err" streams should be used


mislav:
Consider the following program:
```golang
c := cobra.Command{...}
c.SetOut(os.Stdout)
c.SetErr(os.Stderr)
```

**Expected behavior:** Cobra prints help text, version information, and shell completion to _Stdout_; and error messages (e.g. "unknown command") and warning messages (e.g. using deprecated commands or flags) are printed to _Stderr_.

**Actual behavior:** Cobra prints some of its error messages and warning messages (e.g. deprecation notices) to _Stdout_.

I might have misunderstood the semantics of Cobra's "Out" stream. Based on the name alone, I thought it was for setting an equivalent of a "standard output" stream, but the method documentation says: `SetOut sets the destination for usage messages`. In Cobra, "usage messages" are help text that is rendered when the user has specified a wrong command name, used the wrong flag, or passed an invalid flag value. However, Cobra prints much more than just usage messages to the "Out" stream; see below.

Usage messages are always printed as a result of an error, and thus should be (as is my opinion) printed to an error stream. However, in Cobra that is only true until the developer explicitly invokes `command.SetOut()` in their app, after which usage messages and deprecation warnings are printed to the "Out" stream, which is the same stream where Cobra prints help text, version information, and completion scripts. Thus, Cobra blends the output of successful commands with the output of error messages in the same stream.

To resolve this issue, Cobra could change the logic of where usage messages and deprecation warnings are printed, but this could be backwards-incompatible for apps that already rely on this behavior. At minimum, if Cobra wanted to avoid making any functional changes, the developers could expand the documentation to clarify:

- It is not clear to me what the semantics of `command.Print*()` set of functions are and when should they be used. Should app developers use those functions in their command implementations to print to stdout? If so, why do they print to Stderr as default? If not, why are they exported functions?

- Is is not clear to me what the semantics of `command.OutOrStderr()` function are, which is currently the destination that `command.Print*()` are printing to. The documentation says "OutOrStderr returns output to stderr".  However, if the "Out" writer was explicitly specified, then it will return that writer. With that in mind, when should app developers be using this method?

Ref. https://github.com/spf13/cobra/pull/822

P.S. Here is the breakdown of the current printing logic in the latest master. Notice that the current default destinations and their fallbacks are of mismatched natures:

functionality | cur. destination | fallback | ideal destination (according to OP)
--|--|--|--
`--help` | **Out** | **Stdout** | Out
[...]

PR #1709, opened 2022-05-24
Normalize printing to Out vs. Err streams

mislav:
- Have `command.Print*()` print to stdout, and print errors & warnings to stderr
- Deprecate `command.OutOrStderr()`
- Handle errors in built-in help command

*Warning:* backwards-incompatible.

Fixes https://github.com/spf13/cobra/issues/1708

### DIFF
--- a/command.go
+++ b/command.go
@@ -1115,15 +1115,16 @@ Simply type ` + c.Name() + ` help [path to command] for full details.`,
 				}
 				return completions, ShellCompDirectiveNoFileComp
 			},
-			Run: func(c *Command, args []string) {
+			RunE: func(c *Command, args []string) error {
 				cmd, _, e := c.Root().Find(args)
-				if cmd == nil || e != nil {
+				if e != nil {
+					return e
+				} else if cmd == nil {
 					c.PrintErrf("Unknown help topic %#q\n", args)
-					CheckErr(c.Root().Usage())
-				} else {
-					cmd.InitDefaultHelpFlag() // make possible 'help' flag to be shown
-					CheckErr(cmd.Help())
+					return c.Root().Usage()
 				}
+				cmd.InitDefaultHelpFlag() // make possible 'help' flag to be shown
+				return cmd.Help()
 			},
 		}
 	}
--- a/command.go
+++ b/command.go
@@ -327,6 +327,7 @@ func (c *Command) OutOrStdout() io.Writer {
 }
 
 // OutOrStderr returns output to stderr
+// Deprecated: Use OutOrStdout or ErrOrStderr instead
 func (c *Command) OutOrStderr() io.Writer {
 	return c.getOut(os.Stderr)
 }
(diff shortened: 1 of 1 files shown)


marckhouzam:
Thanks @mislav.  Cobra does have some inconsistencies about printing output.  However Cobra being so widely used we have to prioritize backward-compatibility at the expense of standardizing things.

If we ever decided to have a Cobra v2, we would surely standardize the output handling.  But moving to a new major version had such a large impact to projects that I would personally prefer avoiding it.

mislav:
@marckhouzam Thanks for chiming in. And I understand if backwards compatibility is a priority, then it's not a good idea to merge this. Feel free to close; this changeset was mostly meant to complement the referenced issue.

marckhouzam:
Thank you for taking the time @mislav.
I'll check with the other maintainers what we should do with the PR.

mislav:
Another idea: taking an approach where backwards compatibility is kept, but introducing a new opt-in method that replaces `command.SetOut()` with a new one that does the right thing. This already happened once in Cobra's history: a method names `SetOutput()` was deprecated in favor of `SetOut()`.

So for example, we could publish a new method e.g. `command.SetOutNonBroken()` (final name TBD) and deprecate `SetOut` in favor of it. What SetOutNonBroken would do is to define the "Out" stream where help text, completion, and version information are written; but keep writing errors and warning messages to the "Err" stream.

jpmcb (commented):
Marking myself for review.

I agree - there are some major inconsitencies across many of Cobra's APIs.

I'm unsure what to do; the number of things we'd like to fix that would be a breaking change in Cobra could have wide reaching impacts across kubernetes, enterprise software, GitHub, etc.

But these are improvements we should make. And the maintainers should to some degree feel empowered to make these changes.

TLDR: I'd like to do this. And I think it should exist in some new 2.0 of Cobra, I'm just unsure how to start moving towards that.

mislav:
Thanks for taking the time to chime in. What is the most conservative change that the maintainers of Cobra would be willing to accept in order to fix the main bug in the referenced issue (deprecation warnings going to the "Out" stream)?

I'm thinking something along these lines: do not change the functionality of `cmd.Print*()`, in case Cobra apps depend on those, but change the internal logic that prints warnings (e.g. deprecation messages) to go straight to `ErrOrStderr()` instead of using `cmd.Print*()`.

mislav:
Per my previous comment, I'm still willing to do this:

> What is the most conservative change that the maintainers of Cobra would be willing to accept in order to fix the main bug in the referenced issue (deprecation warnings going to the "Out" stream)?

But I'd like a confirmation from Cobra core team (or person) before I proceed, because I wouldn't like to continue doing work on this if it's going to just get closed by the stale bot (like the referenced issue I opened).

tcarreira:
I'm interested as well. 
I support implementing non-breaking changes, and I believe it can be accomplished if the maintainers give their approval. 
@jpmcb, what are your thoughts on this?

There is one specific task I'd like to perform in my app: adding unit tests to verify the output of stdout/stderr. 
However, I've noticed that using the SetOut/SetErr methods consistently produces inconsistent results."

mislav:
The referenced issue got re-opened and in the meantime this PR got outdated, so I'm closing it in favor of the discussion in the issue. There still doesn't seem to be consensus in how this should be solved, so until there is, I don't think there needs to exist a PR that proposes code changes.

PR #1719, opened 2022-06-07
Modify brew prefix path in macOS system

imxw:
Using `$(brew --prefix)` is more generic, since the intel version is located in `/usr/local`, but the ARM version is located in `/opt/homebrew`

### DIFF
--- a/completions.go
+++ b/completions.go
@@ -652,7 +652,7 @@ To load completions for every new session, execute once:
 
 #### macOS:
 
-	%[1]s completion bash > /usr/local/etc/bash_completion.d/%[1]s
+	%[1]s completion bash > $(brew --prefix)/etc/bash_completion.d/%[1]s
 
 You will need to start a new shell for this setup to take effect.
 `, c.Root().Name()),
@@ -689,7 +689,7 @@ To load completions for every new session, execute once:
 
 #### macOS:
 
-	%[1]s completion zsh > /usr/local/share/zsh/site-functions/_%[1]s
+	%[1]s completion zsh > $(brew --prefix)/share/zsh/site-functions/_%[1]s
 
 You will need to start a new shell for this setup to take effect.
 `, c.Root().Name()),
--- a/shell_completions.md
+++ b/shell_completions.md
@@ -40,7 +40,7 @@ Bash:
   # Linux:
   $ %[1]s completion bash > /etc/bash_completion.d/%[1]s
   # macOS:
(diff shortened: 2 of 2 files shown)


marckhouzam:
Thanks for the contribution @imxw !

I wasn't aware of this difference.  But I'm coincidentally getting an M1 Mac very soon and will be able to confirm this change 👍

I guess it would be technically possibly for someone not to use brew to install the bash-completion package.  But we still couldn't know for sure where the package resided in that case, so I think we can assume brew is used as it will cover the vast majority of cases.

imxw:
> I guess it would be technically possibly for someone not to use brew to install the bash-completion package. But we still couldn't know for sure where the package resided in that case, so I think we can assume brew is used as it will cover the vast majority of cases.

I agree with that.  Alternatively, we can use the [Homebrew](https://brew.sh/) [installation script](https://raw.githubusercontent.com/Homebrew/install/HEAD/install.sh) to determine its installation path.

image

marckhouzam:
> > I guess it would be technically possibly for someone not to use brew to install the bash-completion package. But we still couldn't know for sure where the package resided in that case, so I think we can assume brew is used as it will cover the vast majority of cases.
> 
> I agree with that. Alternatively, we can use the [Homebrew](https://brew.sh/) [installation script](https://raw.githubusercontent.com/Homebrew/install/HEAD/install.sh) to determine its installation path.

Well, if they don't use homebrew for their installation of bash-completion, then the package won't follow the homebrew script location anyway 😃

marckhouzam:
Ok, I got my Mac M1 and can confirm the location of the bash_completion scripts is different as brought up by @imxw.

Before merging, @imxw could you also update the reference in the documentation to your generic solution?
https://github.com/spf13/cobra/blob/ca8e3c2779b8cc6168ab6ffdf9c1f2ac65652404/shell_completions.md?plain=1#L43

Once we have that last bit, I'll merge.

Thanks 🚀

imxw:
> Ok, I got my Mac M1 and can confirm the location of the bash_completion scripts is different as brought up by @imxw.
> 
> Before merging, @imxw could you also update the reference in the documentation to your generic solution?
> 
> https://github.com/spf13/cobra/blob/ca8e3c2779b8cc6168ab6ffdf9c1f2ac65652404/shell_completions.md?plain=1#L43
> 
> Once we have that last bit, I'll merge.
> 
> Thanks 🚀

😃 I have modified the relevant documentation

marckhouzam (approved):
Thanks @imxw !

PR #1720, opened 2022-06-08
Add Pulumi as a project using cobra

iwahbe:
The pulumi project uses (and likes using) cobra.


marckhouzam (approved):
Great to hear Pulumi is using Cobra!
Thanks for adding it @iwahbe !

PR #1728, opened 2022-06-14
Rename `SuggestionsMinimumDistance` to `SuggestionsMaximumDistance`

mrpinsky:
Since Cobra will provide suggestions that are less than or equal to this value, it is a maximum, not a minimum.

I just noticed this and the cognitive dissonance was bugging me, but I understand if changing the public API without providing a new feature isn't worth it.

### DIFF
--- a/command.go
+++ b/command.go
@@ -219,9 +219,9 @@ type Command struct {
 	// that go along with 'unknown command' messages.
 	DisableSuggestions bool
 
-	// SuggestionsMinimumDistance defines minimum levenshtein distance to display suggestions.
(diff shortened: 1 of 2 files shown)


marckhouzam:
Thanks @mrpinsky.  I believe your argument makes sense, "maximum" makes more sense.

However, as you point out, this is a public API so we are not allowed to change it.  We'll just have to live with this terminology 😞

jpmcb:
> I just noticed this and the cognitive dissonance was bugging me, but I understand if changing the public API without providing a new feature isn't worth it.

True. Changing anything in this public API is really tricky. This could be something we table for a `v2.0.0` of Cobra if/when that ever happens.

PR #1729, opened 2022-06-15
Adds `check-latest` for setup-go action v3

jpmcb:
This is needed for[ v3 of `setup-go` to check the latest patch version of Go](https://github.com/actions/setup-go#check-latest-version).

Related to https://github.com/spf13/cobra/pull/1660

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -16,6 +16,7 @@ jobs:
       - uses: actions/setup-go@v3
         with:
           go-version: '1.17'
+          check-latest: true
 
       - uses: actions/checkout@v3
 


.github/workflows/test.yml
```diff
@@ -16,6 +16,7 @@ jobs:
       - uses: actions/setup-go@v3
         with:
           go-version: '1.17'
```
umarcor, line 18: Is `go-version` required if we use `check-latest`?

jpmcb: Yes; it is "checking for the latest" on Go 1.17. So, in order to find the latest patch version, we need that option. We shouldn't be consuming base Go `1.17`

umarcor: I guess my question is why do we want to use "the latest on Go 1.17", instead of "the latest Go", which is 1.18 ATM.

jpmcb: That's fair - I'm up for that. Let me look into it quick!

marckhouzam:
I just merged #1782 which I realize includes this PR 🤷‍♂️, so I'm going to close this as done.

jpmcb:
No worries! Thanks Marc!

PR #1730, opened 2022-06-16
use errors.Is() to check for errors

Luap99:
Since go 1.13 you can wrap errors. This make it no longer possible to
compare with `==`, instead you have to compare with `errors.Is()`.

I noticed this problem because -h was no longer working after I stared
wrapping the errors in my custom FlagErrorFunc function.
Note that this is only a problem when a custom help flag is defined.

### DIFF
--- a/command.go
+++ b/command.go
@@ -18,6 +18,7 @@ package cobra
 import (
 	"bytes"
 	"context"
+	"errors"
 	"fmt"
 	"io"
 	"os"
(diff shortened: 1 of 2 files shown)


marckhouzam (approved):
Thanks @Luap99.   I've run some tests on this and it looks reasonable and safe to me. 

However, as it is a pretty critical path, I'd like to get @jpmcb to also approve it before we merge it.

PR #1731, opened 2022-06-17
Feat(completion): support no-space complitions for short flags

### LINKED ISSUE #1629: Shell completion does not support the "no space" format of short flags
Cobra supports a "no space" format for short flags.  I had used it before, but I never really paid attention to it in the code base.  For example:
```
kubectl -ninfra-prometheus get cm
```
Notice the lack of a space between the `-n` flag and the `infra-prometheus` namespace name.
In fact, this seems to work when using multiple short flags:
```
kubectl exec -itninfra-prometheus prom2-prometheus-server-58f7858467-wlqkg -- sh
```
Notice how the `-i`, `-t` and `-n` short flags are used together, with only the last one requiring an argument, and that that argument does not need to be preceded by a space.

Granted that the second example seems unlikely to be used, however the first example seems reasonable (in fact I used it a lot in the past).

Shell completion does not support those "no space" scenarios:
```
kubectl -ninfra-pro<TAB>
# Nothing suggested
```

I haven't had time to investigate, but I wanted to at least document the issue.

mazitovt:
Closes #1629

### DIFF
--- a/completions.go
+++ b/completions.go
@@ -513,10 +513,6 @@ func completeRequireFlags(finalCmd *Command, toComplete string) []string {
 	return completions
 }
 
-func flagsNoSpace() {
-
-}
-
 func checkIfFlagCompletion(finalCmd *Command, args []string, lastArg string) (*pflag.Flag, []string, string, error) {
 	if finalCmd.DisableFlagParsing {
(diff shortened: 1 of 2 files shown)


mazitovt:
Create new PR

PR #1732, opened 2022-06-17
feat(completion): support no-space completions for short flags

### LINKED ISSUE #1629: Shell completion does not support the "no space" format of short flags

mazitovt:
Close #1629

### DIFF
--- a/completions.go
+++ b/completions.go
@@ -543,6 +543,25 @@ func checkIfFlagCompletion(finalCmd *Command, args []string, lastArg string) (*p
 			lastArg = lastArg[index+1:]
 			flagWithEqual = true
 		} else {
+
+			// Find first shorthand non-boolean flag
+			// Following characters are treated as an argument
+			// e.g. `-ini` => n is a flag, 'i' - argument to complete
+			// https://github.com/spf13/cobra/issues/1629
+			if len(lastArg) > 1 && !strings.HasPrefix(lastArg, "--") {
+				i := 1
+				for ; i < len(lastArg)-1; i++ {
+					flagName = lastArg[i : i+1]
+					f := findFlag(finalCmd, flagName)
+					if f != nil && len(f.NoOptDefVal) == 0 {
+						flagName = lastArg[i : i+1]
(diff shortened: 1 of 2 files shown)


mazitovt:
@marckhouzam pls review the PR

Luap99:
Duplicate of #1727?

marckhouzam:
Thanks for the contribution @mazitovt !

> Duplicate of #1727?

As @Luap99 noticed we just got a similar contribution in #1727.  I've started reviewing that one and we should be getting it merged soon. I'll still have a look at yours as it may give a different perspective.

mazitovt:
@marckhouzam Thanks!

marckhouzam:
@mazitovt I've had a look at this PR and it is a step in the right direction.  It is very similar to #1727 which came in first, so I'll focus on #1727 for now.

One note for your info is that although the Go logic works with your PR, the actual shell completion does not work as expected because the different shell script need to be updated to support this usecase.  Note that #1727 has the same problem, but we'll work at getting it resolved.

marckhouzam:
Reopening. I plan on reviewing this. The other PR was closed by the author.

marckhouzam:
The CLA is not signed @mazitovt

Issue #1740, opened 2022-06-21
Shell completions with whitespace seem to work differently on bash vs. other supported shells


kangasta:
Other shell completions escape whitespace in completions, but bash completions seems to add completions with whitespace as multiple suggestions or multiple arguments depending on bash version. I would expect also bash completions to escape whitespace in completions from `ValidArgsFunction`.

bash v3:

```txt
$ ./sh-completion-demo whitespace [TAB][TAB]
no-whitespace    with whitespace
$ ./sh-completion-demo whitespace w[TAB]
$ ./sh-completion-demo whitespace with whitespace
```

bash v4 & v5

```txt
$ ./sh-completion-demo whitespace [TAB][TAB]
no-whitespace  whitespace     with 
```

fish:

```txt
> ./sh-completion-demo whitespace [TAB]
no-whitespace  with whitespace
> ./sh-completion-demo whitespace w[TAB]
> ./sh-completion-demo whitespace with\ whitespace
```

zsh:

```txt
% ./sh-completion-demo whitespace [TAB]
no-whitespace    with whitespace
% ./sh-completion-demo whitespace w[TAB]
% ./sh-completion-demo whitespace with\ whitespace
```

powershell

```txt
> .\sh-completion-demo.exe whitespace [TAB]
> .\sh-completion-demo.exe whitespace with` whitespace
```

Minimal code to reproduce above usage with `github.com/spf13/cobra v1.5.0` ([Gist](https://gist.github.com/kangasta/b571122daaf2e71b27849c38165114f3)):

```go
package main

import (
	"fmt"
	"os"

	"github.com/spf13/cobra"
)

var RootCmd = &cobra.Command{
	Use:  "sh-completion-demo",
	Long: "Demo shell completion with whitespace",
}

var WhitespaceCmd = &cobra.Command{
	Use:   "whitespace",
	Short: "Noun with completion that includes whitespace",
	RunE: func(cmd *cobra.Command, args []string) error {
		if args[0] != "with whitespace" && args[0] != "no-whitespace" {
			return fmt.Errorf(`argument must be "with whitespace" or "no-whitespace"`)
		}
		return nil
	},
	SilenceUsage: true,
	ValidArgsFunction: func(cmd *cobra.Command, args []string, toComplete string) ([]string, cobra.ShellCompDirective) {
		return []string{"with whitespace", "no-whitespace"}, cobra.ShellCompDirectiveNoFileComp
	},
}

func init() {
[...]

PR #1743, opened 2022-06-23
Add similar whitespace escape logic to bash v2 completions than in other completions

kangasta:
This PR would escape bash v2 completions when there are only one completion left and use line-break to separate completions when passing them to `compgen`. This should fix issues described in #1740 and only affect completions with special characters, such as whitespace.

### DIFF
--- a/bash_completionsV2.go
+++ b/bash_completionsV2.go
@@ -230,7 +230,7 @@ __%[1]s_handle_completion_types() {
             comp=${comp%%%%$tab*}
             # Only consider the completions that match
             if [[ $comp == "$cur"* ]]; then
-                COMPREPLY+=("$comp")
+                COMPREPLY+=( "$(printf %%q "$comp")" )
             fi
         done < <(printf "%%s\n" "${completions[@]}")
         ;;
@@ -248,6 +248,12 @@ __%[1]s_handle_standard_completion_case() {
     # Short circuit to optimize if we don't have descriptions
     if [[ "${completions[*]}" != *$tab* ]]; then
         IFS=$'\n' read -ra COMPREPLY -d '' < <(compgen -W "${completions[*]}" -- "$cur")
+
+        # If there is a single completion left, escape the completion
+        if ((${#COMPREPLY[*]} == 1)); then
+            COMPREPLY[0]=$(printf %%q "${COMPREPLY[0]}")
+        fi
+
         return 0
     fi
 
@@ -266,12 +272,12 @@ __%[1]s_handle_standard_completion_case() {
         fi
     done < <(printf "%%s\n" "${completions[@]}")
 
-    # If there is a single completion left, remove the description text
+    # If there is a single completion left, remove the description text and escape the completion
     if ((${#COMPREPLY[*]} == 1)); then
         __%[1]s_debug "COMPREPLY[0]: ${COMPREPLY[0]}"
         comp="${COMPREPLY[0]%%%%$tab*}"
         __%[1]s_debug "Removed description from single completion, which is now: ${comp}"
-        COMPREPLY[0]=$comp
+        COMPREPLY[0]=$(printf %%q "${comp}")
     else # Format the descriptions
         __%[1]s_format_comp_descriptions $longest
     fi
--- a/bash_completionsV2.go
+++ b/bash_completionsV2.go
@@ -247,7 +247,8 @@ __%[1]s_handle_standard_completion_case() {
 
     # Short circuit to optimize if we don't have descriptions
     if [[ "${completions[*]}" != *$tab* ]]; then
-        IFS=$'\n' read -ra COMPREPLY -d '' < <(compgen -W "${completions[*]}" -- "$cur")
+        local compgen_words=$(printf "%%s\n" "${completions[@]}")
+        IFS=$'\n' read -ra COMPREPLY -d '' < <(IFS=$'\n' compgen -W "${compgen_words}" -- "$cur")
 
         # If there is a single completion left, escape the completion
         if ((${#COMPREPLY[*]} == 1)); then
--- a/bash_completionsV2.go
+++ b/bash_completionsV2.go
@@ -201,6 +201,8 @@ __%[1]s_extract_activeHelp() {
     local endIndex=${#activeHelpMarker}
 
     while IFS='' read -r comp; do
+        [[ -z $comp ]] && continue
+
         if [[ ${comp:0:endIndex} == $activeHelpMarker ]]; then
             comp=${comp:endIndex}
             __%[1]s_debug "ActiveHelp found: $comp"
@@ -223,16 +225,21 @@ __%[1]s_handle_completion_types() {
         # If the user requested inserting one completion at a time, or all
         # completions at once on the command-line we must remove the descriptions.
         # https://github.com/spf13/cobra/issues/1508
-        local tab=$'\t' comp
-        while IFS='' read -r comp; do
-            [[ -z $comp ]] && continue
-            # Strip any description
-            comp=${comp%%%%$tab*}
-            # Only consider the completions that match
-            if [[ $comp == "$cur"* ]]; then
-                COMPREPLY+=( "$(printf %%q "$comp")" )
-            fi
-        done < <(printf "%%s\n" "${completions[@]}")
+
+        # If there are no completions, we don't need to do anything
+        (( ${#completions[@]} == 0 )) && return 0
+
+        local tab=$'\t'
+
+        # Strip any description and escape the completion to handled special characters
+        IFS=$'\n' read -ra completions -d '' < <(printf "%%q\n" "${completions[@]%%%%$tab*}")
+
+        # Only consider the completions that match
+        IFS=$'\n' read -ra COMPREPLY -d '' < <(IFS=$'\n'; compgen -W "${completions[*]}" -- "${cur}")
+
+        # compgen looses the escaping so we need to escape all completions again since they will
+        # all be inserted on the command-line.
+        IFS=$'\n' read -ra COMPREPLY -d '' < <(printf "%%q\n" "${COMPREPLY[@]}")
         ;;
 
     *)
@@ -243,18 +250,25 @@ __%[1]s_handle_completion_types() {
 }
 
 __%[1]s_handle_standard_completion_case() {
-    local tab=$'\t' comp
+    local tab=$'\t'
+
+    # If there are no completions, we don't need to do anything
+    (( ${#completions[@]} == 0 )) && return 0
 
     # Short circuit to optimize if we don't have descriptions
     if [[ "${completions[*]}" != *$tab* ]]; then
-        local compgen_words=$(printf "%%s\n" "${completions[@]}")
-        IFS=$'\n' read -ra COMPREPLY -d '' < <(IFS=$'\n' compgen -W "${compgen_words}" -- "$cur")
+        # First, escape the completions to handle special characters
+        IFS=$'\n' read -ra completions -d '' < <(printf "%%q\n" "${completions[@]}")
+        # Only consider the completions that match what the user typed
+        IFS=$'\n' read -ra COMPREPLY -d '' < <(IFS=$'\n'; compgen -W "${completions[*]}" -- "${cur}")
 
-        # If there is a single completion left, escape the completion
-        if ((${#COMPREPLY[*]} == 1)); then
-            COMPREPLY[0]=$(printf %%q "${COMPREPLY[0]}")
+        # compgen looses the escaping so, if there is only a single completion, we need to
+        # escape it again because it will be inserted on the command-line.  If there are multiple
+        # completions, we don't want to escape them because they will be printed in a list
+        # and we don't want to show escape characters in that list.
+        if (( ${#COMPREPLY[@]} == 1 )); then
+            COMPREPLY[0]=$(printf "%%q" "${COMPREPLY[0]}")
         fi
-
         return 0
     fi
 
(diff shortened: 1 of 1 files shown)


marckhouzam:
Thanks @kangasta for the contribution 🎉
I will need to take the time to give this proper attention but I will not have time for the next couple of weeks. I'll get back to this as soon as I can. Thanks for your patience.

marckhouzam:
Ok, I've started looking into this and I started by running some tests and the PR seems to work as expected.

Next step was to add a test to the bash v2 testing of cobra-completion-testing: https://github.com/marckhouzam/cobra-completion-testing/pull/22.  So now we can see that if a single completion has a space, the space character is not escaped.  Running the same test with the PR shows that the space character is escaped.  I think I have to add a couple more tests to cover each possible scenario, so I will keep working on it.

One issue is that the changes cause a slow down in the completion logic.  I'm a little surprised because the changes are isolated to when there is a single completion left, so I'll need to dig into it.

More to come...

bash_completionsV2.go
```diff
@@ -220,7 +220,14 @@ __%[1]s_handle_standard_completion_case() {
 
     # Short circuit to optimize if we don't have descriptions
     if [[ "${completions[*]}" != *$tab* ]]; then
-        IFS=$'\n' read -ra COMPREPLY -d '' < <(compgen -W "${completions[*]}" -- "$cur")
+        compgen_words=$(printf "%%s\n" "${completions[@]}")
+        IFS=$'\n' read -ra COMPREPLY -d '' < <(IFS=$'\n' compgen -W "${compgen_words}" -- "$cur")
```
kangasta, line 224: Could this cause the performance loss?

This also feels bit clumsy to me, but I was not able to get completions with whitespace to be separate indices of `COMPREPLY` without `printf` here.

bash_completionsV2.go
```diff
@@ -220,7 +220,14 @@ __%[1]s_handle_standard_completion_case() {
 
     # Short circuit to optimize if we don't have descriptions
     if [[ "${completions[*]}" != *$tab* ]]; then
-        IFS=$'\n' read -ra COMPREPLY -d '' < <(compgen -W "${completions[*]}" -- "$cur")
+        compgen_words=$(printf "%%s\n" "${completions[@]}")
```
scop, line 223: `compgen_words` should be made `local`.

bash_completionsV2.go
```diff
@@ -220,7 +220,14 @@ __%[1]s_handle_standard_completion_case() {
...
+        IFS=$'\n' read -ra COMPREPLY -d '' < <(IFS=$'\n' compgen -W "${compgen_words}" -- "$cur")
+
+        # If there is a single completion left, escape the completion
+        if [ ${#COMPREPLY[*]} -eq 1 ]; then
```
scop, line 227: Style, see PR #1702

```suggestion
        if ((${#COMPREPLY[*]} == 1)); then
```

bash_completionsV2.go
```diff
@@ -220,7 +220,14 @@ __%[1]s_handle_standard_completion_case() {
...
+            COMPREPLY[0]=$(printf %%q "${COMPREPLY[0]}")
```
scop, line 228: I think we can eliminate a subshell here.
```suggestion
            printf -v "COMPREPLY[0]" %%q "${COMPREPLY[0]}"
```

bash_completionsV2.go
```diff
@@ -239,12 +246,12 @@ __%[1]s_handle_standard_completion_case() {
...
     done < <(printf "%%s\n" "${completions[@]}")
 
-    # If there is a single completion left, remove the description text
+    # If there is a single completion left, remove the description text and escape the completion
     if [ ${#COMPREPLY[*]} -eq 1 ]; then
         __%[1]s_debug "COMPREPLY[0]: ${COMPREPLY[0]}"
         comp="${COMPREPLY[0]%%%%$tab*}"
         __%[1]s_debug "Removed description from single completion, which is now: ${comp}"
-        COMPREPLY[0]=$comp
+        COMPREPLY[0]=$(printf %%q "${comp}")
```
scop, line 254: As above
```suggestion
        printf -v "COMPREPLY[0]" %%q "$comp"
```

scop (changes requested):

bash_completionsV2.go
scop: It's almost certainly here. One starting point for debugging would be to just with strategically placed `time`s, e.g. `time compgen_words=...`.

`printf -v "arr[index]" ...` in a loop would avoid a subshell, but the net result could be even slower as this doesn't create subshells in a loop.

bash_completionsV2.go
scop: Hmm, actually `printf -v` to an array index is a bash 4.1+ thing, so as/if older compatibility is needed, then scratch this and similar comments of mine elsewhere.

marckhouzam: Yeah, we try to keep it working with bash 3, as much as possible.

bash_completionsV2.go
kangasta: Done

bash_completionsV2.go
kangasta: Done

kangasta:
Rebased this on top of `main`

@marckhouzam how big of a performance loss this caused? One option to get this forward could be to split the fix so that optimization for completions without `\t`, from where the performance loss likely comes, would be addresses in separate PR.

Or is there something I could help with on the testing side?

marckhouzam:
I'll try to get back to this PR next week. Thanks for your patience.

kangasta:
Adding a comment to keep this from being closed.

marckhouzam:
Wow, I completely lost track of this.  Sorry @kangasta.
The slow down is about 3 to 4 times slower.

I just ran `make bash` in https://github.com/marckhouzam/cobra-completion-testing

kangasta:
Yeah, I have not much ideas on how to optimize this further. There were comments from @scop about unnecessary subshells, but those would require dropping support for older bash versions.

We have a hack in `upctl` to bypass this issue that replaces spaces with non-breaking spaces (https://github.com/UpCloudLtd/upcloud-cli/pull/204), but it would be nice to get rid of that.

bash_completionsV2.go
JeffFaer: I don't think we need `printf` at all. Wouldn't something like this work?

```bash
IFS=$'\n' read -ra COMPREPLY -d '' < <(IFS=$'\n'; compgen -W "${completions[*]}" -- "$cur")
```
`IFS=$'\n' compgen ...` sets `IFS` for just the `compgen` call.

`IFS=$'\n'; compgen ...` sets `IFS` for both the `compgen` call and the `"${completions[*]}"` string

And since that change is in a subshell, it won't affect `IFS` for the rest of the script

----

You can see what this does with this simpler example:

```bash
$ arr=( 1 2 3 )
$ IFS=$'\n' echo "${arr[*]}"
$ IFS=$'\n'; echo "${arr[*]}"
```

bash_completionsV2.go
```diff
@@ -233,7 +233,14 @@ __%[1]s_handle_standard_completion_case() {
 
     # Short circuit to optimize if we don't have descriptions
     if [[ "${completions[*]}" != *$tab* ]]; then
-        IFS=$'\n' read -ra COMPREPLY -d '' < <(compgen -W "${completions[*]}" -- "$cur")
+        local compgen_words=$(printf "%%s\n" "${completions[@]}")
+        IFS=$'\n' read -ra COMPREPLY -d '' < <(IFS=$'\n' compgen -W "${compgen_words}" -- "$cur")
```
JeffFaer, line 237: If we want to support other bash special characters, we need more quoting:

```bash
local compgen_words="$(printf "%%q\n" "${completions[@]}")"
cur="$(printf "%%q" "${cur}")"
IFS=$'\n' read -ra COMPREPLY -d '' < <(IFS=$'\n' compgen -W "${compgen_words}" -- "${cur}")
```

`compgen`'s `-W` honors shell quoting, so a completion like `foo\>bar` would become `foo>bar` to compgen. We also have to quote `${cur}`  because `compgen` appears to honor shell quoting _after_ checking to make sure each word begins with `${cur}`

----

Similarly, we need quoting up above: https://github.com/spf13/cobra/blob/main/bash_completionsV2.go#L62-L84

```bash
$ args=( echo "foo\>bar" )
$ eval "${args[*]}"
foo>bar
```

`eval` also honors shell quoting, so if we want to pass the actual quoted value through to the `__complete` call, we need to quote the args in `requestComp`

```bash
$ args=( echo "foo\>bar" )
$ eval "$(printf "%q " "${args[@]}")"
foo\>bar
```

bash_completionsV2.go
```diff
@@ -216,7 +216,7 @@ __%[1]s_handle_completion_types() {
             comp=${comp%%%%$tab*}
             # Only consider the completions that match
             if [[ $comp == "$cur"* ]]; then
-                COMPREPLY+=("$comp")
+                COMPREPLY+=( "$(printf %%q "$comp")" )
```
JeffFaer, line 219: If I'm reading cobra-completion-testing right, this line is the performance regression. We're creating a subshell to quote each individual completion

We could instead batch all of the completions up into another array, and then quote them all at once kinda like how we're doing for the input to this loop in the first place

marckhouzam (commented):
I worked very hard trying to make sure #2126 this was safe to merge, and in the end I felt there were too many changes I didn't understand well enough so I tried to simplify the change as much as I could. By the time I was done and had something working, it looked a lot like this PR here!  So, I felt like the fair thing to do was to come back to this PR and fix it to make it work.

Working with the new tests of https://github.com/marckhouzam/cobra-completion-testing/pull/25, I was able to get this PR working as I feel we needed.

And with the use of `printf -v` mentioned by @scop , the performance is much better.  We just need to discuss if we need to protect ourselves against `printf -v` not working in all shell versions.

I have pushed a new commit on top of yours; and all together I feel we got a solution.

bash_completionsV2.go
```diff
@@ -201,6 +201,8 @@ __%[1]s_extract_activeHelp() {
     local endIndex=${#activeHelpMarker}
 
     while IFS='' read -r comp; do
+        [[ -z $comp ]] && continue
```
marckhouzam, line 204: I added this because I noticed we sometime had empty completions.

bash_completionsV2.go
```diff
@@ -243,11 +250,25 @@ __%[1]s_handle_completion_types() {
...
+    # If there are no completions, we don't need to do anything
+    (( ${#completions[@]} == 0 )) && return 0
 
     # Short circuit to optimize if we don't have descriptions
     if [[ "${completions[*]}" != *$tab* ]]; then
-        IFS=$'\n' read -ra COMPREPLY -d '' < <(compgen -W "${completions[*]}" -- "$cur")
+        # First, escape the completions to handle special characters
+        IFS=$'\n' read -ra completions -d '' < <(printf "%%q\n" "${completions[@]}")
+        # Only consider the completions that match what the user typed
+        IFS=$'\n' read -ra COMPREPLY -d '' < <(IFS=$'\n'; compgen -W "${completions[*]}" -- "${cur}")
```
marckhouzam, line 263: This is also inspired from @JeffFaer's work in #2126.

bash_completionsV2.go
```diff
@@ -256,23 +277,39 @@ __%[1]s_handle_standard_completion_case() {
...
         [[ -z $compline ]] && continue
-        # Strip any description before checking the length
-        comp=${compline%%%%$tab*}
+
+        # Before checking if the completion matches what the user typed,
+        # we need to strip any description and escape the completion to handle special
+        # characters because those escape characters are part of what the user typed.
+        # Don't call "printf" in a sub-shell because it will be much slower
+        # since we are in a loop.
+        printf -v comp "%%q" "${compline%%%%$tab*}" &>/dev/null || comp=$(printf "%%q" "${compline%%%%$tab*}")
```
marckhouzam, line 286: I could not find a way to do this efficiently without using `printf -v`.  But since I wasn't 100% sure `printf -v` would work everywhere, I attempted to have a fallback to a slower use of `printf`, in case the call to `printf -v` failed.

bash_completionsV2.go
```diff
@@ -256,23 +277,39 @@ __%[1]s_handle_standard_completion_case() {
...
+
         # Only consider the completions that match
         [[ $comp == "$cur"* ]] || continue
```
marckhouzam, line 289: Escaping the completions **before** filtering them out when those completions had descriptions was the piece of logic that was missing from both #1743 and #2126, and that I added.

marckhouzam (approved):
I pushed a final commit on top of this PR which I believe achieves what we were aiming for.

I will wait a day or two before merging in case someone else can give it a try and tell me what they think, and then I will merge.

marckhouzam:
BTW, this PR now passes the latest tests of https://github.com/marckhouzam/cobra-completion-testing

bash_completionsV2.go
scop: If the printf being executed is not a shell builtin, I don't think `-v foo` as in "assign to variable foo" can work -- the printf executable can not set variables in the parent shell.

I suspect the `which` you were using does not look up from your shell's builtin etc. `type printf` is a more reliable way to check.

```shellsession
$ docker run --rm -it centos:5 bash
[root@388a3ad77194 /]# echo $BASH_VERSION
3.2.25(1)-release
[root@2edc18906647 /]# type printf
printf is a shell builtin
[root@388a3ad77194 /]# printf -v foo foo
[root@388a3ad77194 /]# echo $foo
foo
[root@388a3ad77194 /]# /usr/bin/printf -v bar bar
-v/usr/bin/printf: warning: ignoring excess arguments, starting with `bar'
[root@388a3ad77194 /]# exit
```

scop: Anyway `printf -v` is available in bash >= 3.1: https://github.com/bminor/bash/blob/6794b5478f660256a1023712b5fc169196ed0a22/NEWS#L1442-L1443

(Official upstream repo gives 503 at the moment, therefore pointing to a random GH mirror I found.)

PR #1751, opened 2022-07-05
feat: added custom error handler function

alexraileanu:
Sometimes I want to manually define a way that an error is presented to the user (for instance to change color output and such). This PR adds a way to do so via the new `SetErrorHandlerFunc` method

### DIFF
--- a/command.go
+++ b/command.go
@@ -161,6 +161,9 @@ type Command struct {
 	// versionTemplate is the version template defined by user.
 	versionTemplate string
 
+	// errorHandlerFunc allows setting a custom error handler by the user.
+	errorHandlerFunc func(error)
+
 	// inReader is a reader defined by the user that replaces stdin
 	inReader io.Reader
 	// outWriter is a writer defined by the user that replaces stdout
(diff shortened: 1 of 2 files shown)


jpmcb (commented):
Hmmm interesting.

The code looks good, but I want to understand what this is trying to solve.

Can you provide a sample cobra program where something like this would be useful? Seems like you'd want to set this _per command_ to handle different errors from different commands differently?

alexraileanu:
@jpmcb thanks for your feedback. i created a gist where i can show you an example where this feature would be useful: https://gist.github.com/alexraileanu/200a4473653efa8fbda6e555bd3fc428.

in the screenshot below, you can see that my custom error handlers have different colored messages, however the `Error: ` prefix is added to both. my code, provides a way to give the developer full control over what is output to the user in case of an error.

image

please let me know if you need any more clarification as i would really like to make use of this feature :)

alexraileanu:
@jpmcb sorry for the mess, i did some mistakes on my side of github but it looks like it's all good now.

marckhouzam:
As we have very maintainers who have very little time, we've decided to remove the auto-close configuration.
I'm re-opening this PR since it was closed automatically.

PR #1755, opened 2022-07-08
Use consistent layout for Usage template

erdaltsksn:
Other sections like Usage, Aliases, Available Commands, Flags and
Additional help topics have two spaces indent.

### DIFF
--- a/command.go
+++ b/command.go
@@ -508,7 +508,7 @@ Aliases:
   {{.NameAndAliases}}{{end}}{{if .HasExample}}
 
 Examples:
-{{.Example}}{{end}}{{if .HasAvailableSubCommands}}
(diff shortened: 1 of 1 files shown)


marckhouzam:
Thanks @erdaltsksn, I agree this is probably what should have been done from the start.

However, I worry that programs using Cobra have worked around this by adding the extra 2 spaces themselves in their example string.  If we add our own, suddenly those programs would have 4 spaces on the first line of their examples.

Since there is a simple workaround, I suggest we play it safe and leave this as it is to avoid "breaking" anyone.

marckhouzam:
I'll close this but feel free to voice your opinion and we can re-open if it is justified.
Thanks again for your efforts.

PR #1757, opened 2022-07-16
Fix error message for NoArgs

piroux:
If this is about arguments and if this is not a legacy argument validation function, maybe using the word `argument` instead of the word `command` makes more sense, in the error message for `NoArgs` ?

### DIFF
--- a/args.go
+++ b/args.go
@@ -27,7 +27,7 @@ func legacyArgs(cmd *Command, args []string) error {
 // NoArgs returns an error if any args are included.
 func NoArgs(cmd *Command, args []string) error {
 	if len(args) > 0 {
-		return fmt.Errorf("unknown command %q for %q", args[0], cmd.CommandPath())
+		return fmt.Errorf("unknown argument %q for %q", args[0], cmd.CommandPath())
 	}
 	return nil
 }


marckhouzam:
Thanks @piroux for the contribution.  I understand your point of view, and it probably makes more sense than what we have currently.

However, in practice, from the user's perspective, when a command does not accept any arguments but the user still specifies one, it could just as well be a non-existent subcommand than an non-existant argument.  That's probably why the error message mentions a "command" instead of an "argument" and can still be seen as a valid error.
 
In my opinion, since we can make a point both ways I would leave things as is.

What do you think?

piroux:
You are totally right, that's a chicken and egg problem, so let's keep as it is.

However, for the next major version of `cobra`, maybe it could be a good thing to design the lib around the clear difference of argument and command. Meaning that the lib could recognize if a word is a registered subcommand or an argument expected or not.
What do you think ?

Closing the PR then, but happy to discuss the subject further.

marckhouzam:
> ... Meaning that the lib could recognize if a word is a registered subcommand or an argument expected or not. What do you think ?

Well, of the program doesn't expect neither a subcommand nor an argument, then there is no way to tell the difference when the user types something extra.

Issue #1758, opened 2022-07-19
is there a way to add flags dynamically?


skeetwu:
we build a binary with Cobra, the primary function is to collect input and send all arguments to a restful server, this server will call other functions to create a K8S cluster.
when we do it, there are a lot of arguments, and we need to enhance and change them more regularly, so, if there is a way to add flags dynamically, it will be beneficial for us.

I have written some code to implement this feature, but I am not familiar with Go lang and Corba, I'm not sure we are on the right way, if there is a more good way, please help me figure it out.

API parts, which return dynamic data.
```python
@ns.route('/custom-api')
class CustomAPI(Resource):
    def get(self):
        flags_list = {
            "cert": {
                "type": "string",
                "des": "add self-signed certs for private registries. Format: registry-name1:/path/to/cert-file1,registry-name2:/path/to/cert-file2,...",
                "default": ""
            },
            "csf-grocery": {
                "type": "bool",
                "des": "set up CSF Grocery",
                "default": True
            }
        }
        return json_success(flags_list)
```

parse API result to flag
```go
package utils

import (
	"fmt"
	"os"

	resty "github.com/go-resty/resty/v2"
	"github.com/spf13/cobra"
	"github.com/spf13/viper"
	"github.com/tidwall/gjson"
)

// return data map 
var customizedMap = make(map[string]interface{})

func ParseCutomizedFlags(url string, cmd *cobra.Command, args []string) map[string]interface{} {
        if len(args) == 0 {
		cmd.Help()
		fmt.Println()
		os.Exit(0)
	}


	fullURL := viper.GetString("datacenter-server-api") + url
	var customizedData gjson.Result
	client := resty.New()
	resp, err := client.R().ForceContentType("application/json").EnableTrace().Get(fullURL)
	if err == nil {
		if gjson.Get(resp.String(), "status").String() == "OK" {
			customizedData = gjson.Get(resp.String(), "data")
		}
	} else {
		fmt.Println(err.Error())
	}


	if len(customizedData.Map()) > 0 {
		for key_name, valueMap := range customizedData.Map() {
			value_type := valueMap.Get("type").String()
			if value_type == "string" {
				cmd.Flags().String(string(key_name), valueMap.Map()["default"].String(), valueMap.Map()["des"].String())
			} else if value_type == "bool" {
				cmd.Flags().Bool(string(key_name), valueMap.Map()["default"].Bool(), valueMap.Map()["des"].String())
			}
		}
	}

        // set flag back
[...]

PR #1760, opened 2022-07-21
expose ValidateRequiredFlags and ValidateFlagGroups

skeetwu:
Hi, dear reviewers,

In my project, there is one complex command with a lot of flags, we need to set part of them as required, and other flags are optional, all optional flags were defined from an API server. I have implemented this feature to add dynamic flags in issue #1758.
as talked with @marckhouzam in the last issue, it looks like my method is on the right way.

In the next work, I add some flags in `init()` function and call `MarkFlagRequired` to mark them as required flags. but, because I have set `DisableFlagParsing` as `true`, and need to call `func (c *Command) ParseFlags(args []string) error` to parse all flags manually, the function which validates the required flag will not be called. 

for now, I have to set up a map and store all required flags, then check if exist in arguments one bye one before `ParseFlags()`. It works but looks not graceful.

I went through the source code of Cobra, and it looks like I need to call `validateRequiredFlags()` and `validateFlagGroups()` manually after `ParseFlags(args []string)`,  and function `validateRequiredFlags()` was called by `Execute()`, so I created this PR to expose these two functions.

Thx for reviewing this PR.

B&R,
Skeet

### DIFF
--- a/command.go
+++ b/command.go
@@ -861,10 +861,10 @@ func (c *Command) execute(a []string) (err error) {
 		c.PreRun(c, argWoFlags)
 	}
 
-	if err := c.validateRequiredFlags(); err != nil {
+	if err := c.ValidateRequiredFlags(); err != nil {
 		return err
 	}
-	if err := c.validateFlagGroups(); err != nil {
+	if err := c.ValidateFlagGroups(); err != nil {
 		return err
 	}
 
@@ -1018,7 +1018,8 @@ func (c *Command) ValidateArgs(args []string) error {
 	return c.Args(c, args)
 }
 
-func (c *Command) validateRequiredFlags() error {
+// ValidateRequiredFlags validates all required flags are present and returns an error otherwise
+func (c *Command) ValidateRequiredFlags() error {
 	if c.DisableFlagParsing {
 		return nil
 	}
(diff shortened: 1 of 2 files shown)


skeetwu:
Hi, exposing these two functions is more convenient and does not conflict with other parts. I have changed it in my local workspace, it works well.
please help to review.

marckhouzam:
This seems reasonable to give users more flexibility in handling flags.

command.go
```diff
@@ -1018,7 +1018,8 @@ func (c *Command) ValidateArgs(args []string) error {
 	return c.Args(c, args)
 }
 
-func (c *Command) validateRequiredFlags() error {
+// ValidateRequiredFlags validates the required flags logic and returns all unset required flags.
```
marckhouzam, line 1021: Sorry to be picky but can you chance the godoc to:

// ValidateRequiredFlags validates all required flags are present and returns an error otherwise

skeetwu: No, it's not picky at all. It's very professional comments.  😄

marckhouzam (approved):
Thanks @skeetwu !

skeetwu:
@marckhouzam  thx for reviewing and helping to merge the code.
I feel so glad to contribute code to a  so popular project. 😃

PR #1762, opened 2022-07-22
fix(bash-v2): activeHelp length check syntax

scop:
```shell
$ set -u
$ foo=()
$ echo ${#foo}
bash: foo: unbound variable
echo ${#foo[*]}
0
```

### DIFF
--- a/bash_completionsV2.go
+++ b/bash_completionsV2.go
@@ -150,7 +150,7 @@ __%[1]s_process_completion_results() {
     __%[1]s_handle_special_char "$cur" =
 
     # Print the activeHelp statements before we finish
-    if [ ${#activeHelp} -ne 0 ]; then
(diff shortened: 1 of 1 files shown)


scop:
#1702 requires a rebase after this is in, will take care of it.

marckhouzam:
Nice catch!  Thanks @scop.
I'll run some tests on this next week and get it merged.
I assume this was the only unbound variable problem with ActiveHelp?

scop:
The only one I happened to hit with my normal usage patterns, yes. I didn't analyze beyond that.

marckhouzam (approved):
I've tested this with the `cobra-cli` also using its ActiveHelp patch, and I can see the problem when I do `set -u` in bash, but with the patch, all is well.

Thanks @scop

Issue #1739, opened 2022-06-21
Could we have 'MarkFlagsDependentOn' group validator


plastikfan:
I was looking forward to this new release for the new group validators, but I have found that `MarkFlagsRequiredTogether` doesn't quite suit my needs, because it enforces that all members of the group to be present. However, there is a need for a 1 way dependency check. It may be that a flag ___A___ is dependent on flag ___B___, but not vice-versa, so `MarkFlagsRequiredTogether` can't be used in this scenario.

Rather, it would be great if we could get a `MarkFlagsDependentOn` (or something similarly named), where you could specify that 1 or more flags can be dependent on another one.

And another variation on this theme would be to have another validator `MarkDependentOnAnyOf`, where you could mark that a flag ___A___ is dependent on any of flags ___X___, ___Y___ or ___Z___

Thanks.

plastikfan:
PS: I don't mind submitting a pull request to implement this

plastikfan:
Hi @yurishkuro, I have just finished coding up a solution to this with a suite of passing test cases. My fix implements 2 new group rules:

- MarkFlagsDependentOn: given a list of flag names, the first is considered the `dependee` and all subsequent flags are `dependents`. If any dependents are present, then the dependee must be present
- MarkFlagsDependentOnAny: given a list of flag names, the first is considered the `dependent` and all subsequent are `dependees`. At least 1 of the dependees must be present if the dependent is present

I think this should suit your needs.

I just need to spend a bit of time testing this with real applications now. I intend to submit a pull request and hopefully it'll get accepted but, given the long list of other outstanding PR's, I'm not sure how long this will take.

plastikfan:
accompanying decription to be added to ___Flag Groups___ section of user guide:

If you need 1 way dependency groups, as opposed to all flags in a group being required together (like `MarkFlagsRequiredTogether`), then you have a further 2 options.

You can specify that 1 or more flags be dependent upon another using `MarkFlagsDependentOn` eg, let's say you have an app that performs filtering and you want to support regex and glob filter types, but by default the _filter_ is regex. The user should be able to define the _filter_ and leave out the _type_. But _type_ can't be specified without the _filter_ also being present. So _type_ is ___dependent___ on _filter_ (the ___dependee___), but not vice-versa. In general terms this equates to

> cmd.MarkFlagsDependentOn(dependee, dependent-1, dependent-2 ...)

eg:

```go
[...]

PR #1768, opened 2022-07-27
feat: add markflagsdependenton (#1739)

plastikfan:
This PR adds to the group validation capability that came in the Spring release with the advent of `MarkFlagsRequiredTogether` and `MarkFlagsMutuallyExclusive` by adding 2 further group validation options (new methods on the cobra.Command):

- ___MarkFlagsDependsOn___: 1 or more flags are dependentent upon another
- ___MarkFlagDependsOnAny___: a flag depends on at least 1 of a group of other flags

These 2 options differ from `MarkFlagsRequiredTogether` and `MarkFlagsMutuallyExclusive` in a key aspect, that being that the relationship is 1 way, ie:

- for ___MarkFlagsDependsOn___: Flags A and B depends on Flag X, but X is not dependent on A or B. 
- for ___MarkFlagDependsOnAny___: A is dependent at least 1 of X or Y, but X and Y ore not dependent on A.

I have tried to minimise the amount of code changes in this PR. If I had free will, I would have made all the code consistent, but that would have resulted in more code changes. If you would prefer for me to make existing code consistent with new, then please let me know and I will update.

More detailed explaination of changes:

- `setting annotations`: 2 new unexported functions has been defined ___markAnnotation___ and ___setFlagAnnotation___ which mimmicks how annotations are added for ___MarkFlagsRequiredTogether___ and ___MarkFlagsMutuallyExclusive___ (referred to as non special flag validators). For those 2 non special validators, code is duplicated and the only difference between the 2 implemenations is the error message that is defined for the panic, if the flag can't be found. ___markAnnotation___/___setFlagAnnotation___ avoid the duplication, by using a ___format___ parameter. Ideally, the non special validators should be updated to use ___markAnnotation___, but I avoided this to reduce on code churn.
- `special status`: for the non special validators, all parameters in the group have identical status. The new special validators (___MarkFlagsDependsOn___ and ___MarkFlagDependsOnAny___) impose a different status to the first flag in the group. For ___MarkFlagsDependsOn___, the first flag is the `dependee`, that all other flags in the group depend on. For ___MarkFlagDependsOnAny___, the first flag depends on at least one of the other flags being present.

More on `special status`: To support this, a new collection ___specialGroupInfoCollection___ and associated structs ___specialStatusInfoData___ and ___specialGroupInfo___ have been defined.

The non special valdators use ___processFlagForGroupAnnotation___ which internally defines a map of string to bool, indicating whether a flag is present or not. For the new special validators, this is not enough. For a particluar group, we need to distinguish which flag in the group is the special one. So:

- ___specialGroupInfoCollection___: maps the group name to a ___specialGroupInfo___
- ___specialGroupInfo___: contains `special` which indicates which is the special flag, `others` the remaing flags. To maintain similarilty with existing code, we also have a member of type ___specialStatusInfoData___ 
- ___specialStatusInfoData___: simply maps a flag name to a boolean value equivalent to ___isSet___

Defined 2 new group validator functions: ___validateDependsOnFlagGroups___ and ___validateDependsOnAnyFlagGroups___, both of which use the newly defined struct ___specialGroupInfoCollection___ for validation which are equivalent to the valitdators that exist for non special group validators ___validateRequiredFlagGroups___ and ___validateExclusiveFlagGroups___.

TODO: still need completions tests for ___MarkFlagsDependsOn___. There is no completions functionality for ___MarkFlagDependsOnAny___, because if flag `A` is dependent on `X` or `Y`, we can't mark either as required using ___MarkFlagRequired___ and we can't hide either of them by setting the ___Hidden___ property to true.


marckhouzam:
Thanks @plastikfan for the contribution!  Don't worry about the size of the PR if it is justified.
Please be patient with us as we'll do our best to review this as soon as we can.

plastikfan:
Thanks @marckhouzam, I was wondering wether I should continue with this PR because of that message. I am just writing completions tests for this as it is something that slipped my mind until I reviewed my code for the PR.

plastikfan:
Just now waiting the outcome of #1775, as that PR claims to implement the 2 new validation groups I have proposed as part of its own design refactor. Because of this, #1739 may be redundant.

plastikfan:
Closing this PR as it being superceded by #1775

PR #1771, opened 2022-08-04
fix: correct command path in see_also for YAML doc

zregvart:
The `see_also` section for child commands would include only the name of command. This adds the whole path, similar to how it's done for the other documentation formats.

### DIFF
--- a/doc/yaml_docs.go
+++ b/doc/yaml_docs.go
@@ -128,7 +128,7 @@ func GenYamlCustom(cmd *cobra.Command, w io.Writer, linkHandler func(string) str
 			if !child.IsAvailableCommand() || child.IsAdditionalHelpTopicCommand() {
 				continue
 			}
-			result = append(result, child.Name()+" - "+child.Short)
+			result = append(result, child.CommandPath()+" - "+child.Short)
 		}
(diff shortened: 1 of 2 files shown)


stmcginnis (approved):
Looks reasonable.

marckhouzam (approved):
That's much better, thanks @zregvart! (and thanks @stmcginnis for having a look)
I have tested it with helm and if we take the `helm repo` command and generate the yaml tree, the generated yaml before this PR would end with:
```
see_also:
- helm - The Helm package manager for Kubernetes.
- add - add a chart repository
- index - generate an index file given a directory containing packaged charts
- list - list chart repositories
- remove - remove one or more chart repositories
- update - update information of available charts locally from chart repositories
```
but now:
```
see_also:
- helm - The Helm package manager for Kubernetes.
- helm repo add - add a chart repository
- helm repo index - generate an index file given a directory containing packaged charts
- helm repo list - list chart repositories
- helm repo remove - remove one or more chart repositories
- helm repo update - update information of available charts locally from chart repositories
```
It is not only much clearer but it does match what we output for `markdown` and `man` docs.

zregvart:
Thanks @stmcginnis and @marckhouzam!

Issue #1327, opened 2021-02-02
group flags for a subcommand


medyagh:
in minikube our start command has many flags and we would like to group them for better readablility

```
$ minikube start --help
Starts a local Kubernetes cluster

Options:
      --addons=[]: Enable addons. see `minikube addons list` for a list of valid addon names.
      --apiserver-ips=[]: A set of apiserver IP Addresses which are used in the generated certificate for kubernetes.
This can be used if you want to make the apiserver available from outside the machine
      --apiserver-name='minikubeCA': The authoritative apiserver hostname for apiserver certificates and connectivity.
This can be used if you want to make the apiserver available from outside the machine
      --apiserver-names=[]: A set of apiserver names which are used in the generated certificate for kubernetes.  This
can be used if you want to make the apiserver available from outside the machine
      --apiserver-port=8443: The apiserver listening port
      --auto-update-drivers=true: If set, automatically updates drivers to the latest version. Defaults to true.

--base-image='gcr.io/k8s-minikube/kicbase:v0.0.17@sha256:1cd2e039ec9d418e6380b2fa0280503a72e5b282adea674ee67882f59f4f546e':
The base image to use for docker/podman drivers. Intended for local development.
      --cache-images=true: If true, cache docker images for the current bootstrapper and load them into the machine.
Always false with --driver=none.
      --cni='': CNI plug-in to use. Valid options: auto, bridge, calico, cilium, flannel, kindnet, or path to a CNI
manifest (default: auto)
      --container-runtime='docker': The container runtime to be used (docker, cri-o, containerd).
      --cpus=2: Number of CPUs allocated to Kubernetes.
      --cri-socket='': The cri socket path to be used.
      --delete-on-failure=false: If set, delete the current cluster if start fails and try again. Defaults to false.
      --disable-driver-mounts=false: Disables the filesystem mounts provided by the hypervisors
      --disk-size='20000mb': Disk size allocated to the minikube VM (format: <number>[<unit>], where unit = b, k, m or
g).
      --dns-domain='cluster.local': The cluster dns domain name used in the Kubernetes cluster
      --dns-proxy=false: Enable proxy for NAT DNS requests (virtualbox driver only)
      --docker-env=[]: Environment variables to pass to the Docker daemon. (format: key=value)
      --docker-opt=[]: Specify arbitrary flags to pass to the Docker daemon. (format: key=value)
      --download-only=false: If true, only download and cache files for later use - don't install or start anything.
[...]

PR #1778, opened 2022-08-18
Add group flags

knqyf263:
## Description
Add group flags

Close https://github.com/spf13/cobra/issues/1327

## Example

<details>
<summary>main.go</summary>

```
package main

import (
	"fmt"

	"github.com/spf13/cobra"
)

func main() {
	var output string
	var rootCmd = &cobra.Command{
		Use:   "root [sub]",
		Short: "My root command",
			fmt.Printf("Output: %s\n", output)
		},
	}
	rootCmd.Flags().Int("replicas", 0, "replicas")
	rootCmd.NamedFlags("cache").String("dir", "", "cache dir")
	rootCmd.NamedFlags("cache").Duration("max-age", 0, "cache ttl")
	rootCmd.NamedFlags("result").StringVarP(&output, "output", "o", "", "output format")
	rootCmd.NamedFlags("result").StringP("filename", "f", "", "filename")

	rootCmd.SetArgs([]string{"--output", "dummy.txt"})
	rootCmd.Execute()
	fmt.Println()
	rootCmd.SetArgs([]string{"--help"})
	rootCmd.Execute()
}
```

</details>

```
$ go run main.go
Output: dummy.txt

My root command

Usage:
  root [sub] [flags]

Flags:
  -h, --help           help for root
      --replicas int   replicas

Cache Flags:
      --dir string         cache dir
      --max-age duration   cache ttl

Result Flags:
  -f, --filename string   filename
  -o, --output string     output format
```

### DIFF
--- a/command.go
+++ b/command.go
@@ -139,6 +139,8 @@ type Command struct {
 	iflags *flag.FlagSet
 	// parentsPflags is all persistent flags of cmd's parents.
 	parentsPflags *flag.FlagSet
+	// lnamedFlagSets contains local named flags.
+	lnamedFlagSets *NamedFlagSets
 	// globNormFunc is the global normalization function
 	// that we can use on every pflag set and children commands
 	globNormFunc func(f *flag.FlagSet, name string) flag.NormalizedName
@@ -514,7 +516,8 @@ Available Commands:{{range .Commands}}{{if (or .IsAvailableCommand (eq .Name "he
   {{rpad .Name .NamePadding }} {{.Short}}{{end}}{{end}}{{end}}{{if .HasAvailableLocalFlags}}
 
 Flags:
-{{.LocalFlags.FlagUsages | trimTrailingWhitespaces}}{{end}}{{if .HasAvailableInheritedFlags}}
+{{.LocalNonNamedFlags.FlagUsages | trimTrailingWhitespaces}}{{end}}{{if .HasAvailableNamedFlags}}
+{{.NamedFlagSets.FlagUsages | trimTrailingWhitespaces}}{{end}}{{if .HasAvailableInheritedFlags}}
 
 Global Flags:
 {{.InheritedFlags.FlagUsages | trimTrailingWhitespaces}}{{end}}{{if .HasHelpSubCommands}}
@@ -578,6 +581,7 @@ func stripFlags(args []string, c *Command) []string {
 		return args
 	}
 	c.mergePersistentFlags()
+	c.mergeNamedFlags()
 
 	commands := []string{}
 	flags := c.Flags()
@@ -1046,6 +1050,7 @@ func (c *Command) validateRequiredFlags() error {
 // If c already has help flag, it will do nothing.
 func (c *Command) InitDefaultHelpFlag() {
 	c.mergePersistentFlags()
+	c.mergeNamedFlags()
 	if c.Flags().Lookup("help") == nil {
 		usage := "help for "
 		if c.Name() == "" {
@@ -1067,6 +1072,7 @@ func (c *Command) InitDefaultVersionFlag() {
 	}
 
 	c.mergePersistentFlags()
+	c.mergeNamedFlags()
 	if c.Flags().Lookup("version") == nil {
 		usage := "version for "
 		if c.Name() == "" {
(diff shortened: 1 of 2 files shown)


named_flag_sets.go
```diff
@@ -0,0 +1,89 @@
+package cobra
```
knqyf263, line 1: Should this file be in spf13/pflag?

command.go
```diff
@@ -1475,6 +1481,31 @@ func (c *Command) Flags() *flag.FlagSet {
...
+// NamedFlagSets returns all the named FlagSet that applies to this command.
+func (c *Command) NamedFlagSets() *NamedFlagSets {
+	if c.lnamedFlagSets == nil {
+		c.lnamedFlagSets = NewNamedFlagSets(c.Name(), flag.ContinueOnError)
+	}
+	return c.lnamedFlagSets
+}
+
+// NamedFlags returns the specific named FlagSet that applies to this command.
+func (c *Command) NamedFlags(name string) *flag.FlagSet {
```
knqyf263, line 1493: NamedFlags doesn't support persistent flags at the moment.

knqyf263:
@marckhouzam I don't think this is the best implementation, but I hope this is a good starting point for discussion.

I borrowed an idea of NamedFlagSets from Kubernetes.
https://github.com/kubernetes/component-base/blob/b5a495af30a7bb04642ce82f4816b47e75f78dbe/cli/flag/sectioned.go#L33-L41

EraYaN:
This would still be a very valuable feature and should not go stale.

pedromotita:
Hi, is there any updates do this PR status? It's a really nice feature. A lot of projects would benefit from it

knqyf263:
@pedromotita I wanted to kick off the discussion, but I've not received any response from maintainers.

marckhouzam:
Thanks @knqyf263 for your patience (and everyone else interested in this), I will start looking at this PR.

marckhouzam:
There is a PR to support this in `spf13/pflag` directly: https://github.com/spf13/pflag/pull/339
I'm mentioning this as it can be a reference, however, there has been no movement in that project for years, so I think we should continue adding support in Cobra instead.

marckhouzam:
@knqyf263 Although I think your API is very elegant, I'm concerned it complicates the code in Cobra and would make maintenance more difficult.  I'm worried about creating yet another "type" of flags and then have: local flags, inherited flags and named flags.

The grouping of flags in the help does not affect their actual behaviour.  So from a behavioural perspective we still have either local or inherited flags.  I would prefer keeping those two groups and adding an independent concept of "flag group" on top of this.

For better or worse we already have an API for command groups (see doc [here](https://github.com/spf13/cobra/blob/main/site/content/user_guide.md#grouping-commands-in-help)), so it may be good for users of Cobra to have a similar API for flag groups.

With that in mind, what would you think of an API that looked something like:
```
var rootCmd = &cobra.Command{
	Use:   "root [sub]",
	Short: "My root command",
		fmt.Printf("Root command")
	},
}

// Create flags exactly like is currently done
rootCmd.Flags().Int("replicas", 0, "replicas")
rootCmd.Flags.()String("dir", "", "cache dir")
rootCmd.Flags().Duration("max-age", 0, "cache ttl")
rootCmd.PersistentFlags().StringVarP(&output, "output", "o", "", "output format")
rootCmd.PersistentFlags().StringP("filename", "f", "", "filename")

// Add new, separate knowledge for grouping flags in the help output
err = rootCmd.AddFlagHelpGroup(cobra.Group{ID: "group1", "Title": "cache"}, cobra.Group{ID: "group2", "Title": "result"})
err = rootCmd.AddFlagToHelpGroupID("dir", "group1")
err = rootCmd.AddFlagToHelpGroupID("max-age", "group1")
err = rootCmd.AddFlagToHelpGroupID("output", "group2")
err = rootCmd.AddFlagToHelpGroupID("filename", "group2")
```
Notice the use of the word "Help" to qualify the flag groups; I think we need something along those lines because Cobra has a concept of flag group for mutual exclusion and such things.

I admit I like your API better, but I am hoping that something along the lines of the above will be safer to implement.  And it should handle global flags transparently, I'm hoping.

Hopefully this can start a discussion leading to something we all agree is useful.

pedromotita:
@knqyf263 any thoughts on @marckhouzam API proposal? He has a fair point

Let us know if you're still willing to work on this PR. I'm new to the community, but would be very happy to help 😄

pedromotita:
@marckhouzam I've been trying to wrap my head around a solution using your API suggestion. The main problem is that `flag.FlagSet` has no reference to `cobra.Group`.

Whereas dealing with `Command` we can use `{{if (and (eq .GroupID $group.ID) ... }}` in the template, we can't do that for flags... We could create a map do tie a `Group.ID` to a `flag.FlagSet`, but it's not a elegant solution at all...

The more I think about this feature, the more I get convinced part of it should be implemented in `spf13/pflag` package 😢

marckhouzam:
I believe flags have an annotation field. Maybe you can use that to associate the group Id?

pedromotita:
@marckhouzam I had not thought of that. Thanks!

I opened the PR #2117 with your API proposal and some test cases :)

PR #1781, opened 2022-08-19
fix: don't remove flag value that matches subcommand name

### LINKED ISSUE #1777: flag value misparsed as subcommand
it looks like the parser is having trouble with space delimited persistent flags, where a sub command aligns with the value of the first of many flags:

for `./app -z one -o one two`: https://go.dev/play/p/cKaEuTEhWAZ
wanted:
```
zero: one
one: one
args: []
```
got:
```
zero: -o
one: nil
args: [one two]
```

note: this does not trigger for the following cases
```sh
app -z two two          # only one flag
app -z=two -o=one two   # using equals delimited flags
app -z one -o two two   # colliding value is not first flag
app two -z two -one one # sub command comes first
```

brianpursley:
When the command searches args to find the arg matching a particular subcommand name, it needs to ignore flag values, as it is possible that the value for a flag might match the name of the sub command.

This change improves argsMinusFirstX() to ignore flag values when it searches for the X to exclude from the result.

Fixes #1777

### DIFF
--- a/command.go
+++ b/command.go
@@ -655,13 +655,37 @@ Loop:
 
 // argsMinusFirstX removes only the first x from args.  Otherwise, commands that look like
 // openshift admin policy add-role-to-user admin my-user, lose the admin argument (arg[4]).
-func argsMinusFirstX(args []string, x string) []string {
-	for i, y := range args {
-		if x == y {
-			ret := []string{}
-			ret = append(ret, args[:i]...)
-			ret = append(ret, args[i+1:]...)
-			return ret
+// Special care needs to be taken not to remove a flag value.
+func (c *Command) argsMinusFirstX(args []string, x string) []string {
+	if len(args) == 0 {
+		return args
+	}
+	c.mergePersistentFlags()
+	flags := c.Flags()
+
+Loop:
+	for pos := 0; pos < len(args); pos++ {
+		s := args[pos]
+		switch {
+		case s == "--":
+			// -- means we have reached the end of the parseable args. Break out of the loop now.
+			break Loop
+		case strings.HasPrefix(s, "--") && !strings.Contains(s, "=") && !hasNoOptDefVal(s[2:], flags):
+			fallthrough
+		case strings.HasPrefix(s, "-") && !strings.Contains(s, "=") && len(s) == 2 && !shortHasNoOptDefVal(s[1:], flags):
+			// This is a flag without a default value, and an equal sign is not used. Increment pos in order to skip
+			// over the next arg, because that is the value of this flag.
+			pos++
+			continue
+		case !strings.HasPrefix(s, "-"):
+			// This is not a flag or a flag value. Check to see if it matches what we're looking for, and if so,
+			// return the args, excluding the one at this position.
+			if s == x {
+				ret := []string{}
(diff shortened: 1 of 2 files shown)


command.go
```diff
@@ -613,13 +613,37 @@ Loop:
...
-			return ret
+// Special care needs to be taken not to remove a flag value.
+func argsMinusFirstX(args []string, x string, c *Command) []string {
+	if len(args) == 0 {
+		return args
+	}
+	c.mergePersistentFlags()
+	flags := c.Flags()
+
+Loop:
```
brianpursley, line 624: This logic resembles (but not exactly) that of `stripFlags()` in this same package. I tried to take a similar approach here for consistency.

command.go
```diff
@@ -613,13 +613,37 @@ Loop:
...
+	for pos := 0; pos < len(args); pos++ {
```
umarcor, line 625: Is there a particular reason to use `for pos := 0; pos < len(args); pos++` instead of `for i, y := range args`?

jpmcb (commented):
This overall makes sense but since we'd be chaning the API from
```go
func argsMinusFirstX(args []string, x string)
```
to
```go
argsMinusFirstX(args []string, x string, c *Command)
```
we'd have to wait for Cobra 2.0.0.

command.go
brianpursley: Sorry, I didn't see you comment until now.

Yes, there is a reason:

Inside the loop, it can advance the position (`pos++` on line 636) in order to handle when there is a flag with an arg and an equal sign is not used (ie `--foo bar`).  In this case it needs to skip over the next token because it is the value of the flag, not an arg.

yxxhero:
any update?

brianpursley:
> any update?

I don't think there are any outstanding questions or changes requested, so it waiting on further review or approval.

yxxhero:
@brianpursley Thanks for your reply. I'm waiting for this PR.

marckhouzam (changes requested):
This look correct to me and I think it is safe with respect to backwards-compatibility.
It will fix the broken case, which I don't see anyone expecting to work as it does now.

Since this PR requires a small rebase, I allowed myself to add two nits that hopefully @brianpursley won't mind addressing at the same time.

Thanks for your patience.

command_test.go
```diff
@@ -2193,3 +2193,92 @@ func TestSetContextPersistentPreRun(t *testing.T) {
...
+	root.PersistentFlags().StringVarP(&bar, "bar", "b", "something", "")
+
+	child := &Command{
+		Use: "child",
+	}
+	root.AddCommand(child)
+
+	testCases := []struct {
+		args          []string
+		expectedFlags []string
```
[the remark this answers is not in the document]
brianpursley: Good catch.  I don't know why I called it flags, but you're right, it is not just flags.

I renamed them to `expectedFoundArgs` and `foundArgs`.  Let me know if you have a different name in mind.

marckhouzam (approved):
Thanks for the quick fix!
LGTM

umarcor:
@marckhouzam please milestone this!

marckhouzam:
> @marckhouzam please milestone this!

This was already milestoned 🙂

PR #1783, opened 2022-08-21
ci: use action/setup-go's cache

umarcor:
Since v3, Action setup-go has built-in support for caching modules. See https://github.com/actions/setup-go#caching-dependency-files-and-build-outputs.

Ref: #1660

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -36,7 +36,6 @@ jobs:
         - ubuntu
         - macOS
         go:
-        - 14
         - 15
         - 16
         - 17
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -13,10 +13,13 @@ jobs:
     runs-on: ubuntu-latest
     steps:
 
+      - uses: actions/checkout@v3
+
       - uses: actions/setup-go@v3
         with:
           go-version: '^1.19'
           check-latest: true
+          cache: true
 
       - uses: actions/checkout@v3
 
@@ -43,17 +46,12 @@ jobs:
     runs-on: ${{ matrix.platform }}-latest
     steps:
 
+    - uses: actions/checkout@v3
+
     - uses: actions/setup-go@v3
       with:
         go-version: 1.${{ matrix.go }}.x
-
-    - uses: actions/checkout@v3
-
-    - uses: actions/cache@v3
-      with:
-        path: ~/go/pkg/mod
-        key: ${{ runner.os }}-1.${{ matrix.go }}.x-${{ hashFiles('**/go.sum') }}
-        restore-keys: ${{ runner.os }}-1.${{ matrix.go }}.x-
+        cache: true
 
     - run: |
         export GOBIN=$HOME/go/bin


.github/workflows/test.yml
```diff
@@ -33,7 +36,6 @@ jobs:
         - ubuntu
         - macOS
         go:
-        - 14
```
jpmcb, line 36: Why was go 1.14 removed here? I think that deserves it's own PR / proposal to remove support for Go 1.14

.github/workflows/test.yml
```diff
@@ -43,17 +45,12 @@ jobs:
     runs-on: ${{ matrix.platform }}-latest
     steps:
 
-    - uses: actions/setup-go@v3
-      with:
-        go-version: 1.${{ matrix.go }}.x
-
     - uses: actions/checkout@v3
```
jpmcb, line 48: Is this chunk redundant now?

.github/workflows/test.yml
```diff
@@ -13,10 +13,13 @@ jobs:
     runs-on: ubuntu-latest
     steps:
 
+      - uses: actions/checkout@v3
+
       - uses: actions/setup-go@v3
         with:
           go-version: '^1.19'
           check-latest: true
+          cache: true
```
jpmcb, line 22: Nice! Glad this is supported in the setup-go function now!

.github/workflows/test.yml
umarcor: See run 232 in https://github.com/umarcor/cobra/actions?query=branch%3Aumarcor%2Fci%2Fcache. Enabling this feature fails with go 1.14.

.github/workflows/test.yml
umarcor: I don't understand. This PR moves action/setup-go to be run after actions/checkout, and actions/cache is removed. actions/checkout is not modified. It is required in order to clone the repo.

.github/workflows/test.yml
stmcginnis: 1.14 has been [end of life for awhile now](https://endoflife.date/go) so this seems fine from me. Do we have a published support policy somewhere?

stmcginnis (approved):
This looks good to me. I've seen the caching really speed things up in other projects.

.github/workflows/test.yml
umarcor: @stmcginnis see https://github.com/spf13/cobra/blob/main/CONDUCT.md#deprecation.

stmcginnis: Thanks! So I think as long as we follow that deprecation policy, this should be good to do.

Since that will require a major version bump according to the policy, maybe we should also think about dropping some of the other EOL Go releases and see if there are any other major changes that need to be merged before making that jump.

umarcor: Agree. Nonetheless, note that 'major' in the context of cobra is probably equivalent to 'minor' in semver semantics.

marckhouzam: Yeah, we should avoid doing a major version bump.  A new major version would mean a new module path: `github.com/spf13/cobra/v2`.  That means every project would need to change _all_ their Cobra _package_ import paths.  That's a big impact.

jpmcb: From the `CONDUCT.md`:

> Deprecation of Go versions or dependent packages will only occur in major releases.

It's been awhile since I read that deprecation policy but I'm not sure I agree with it anymore: we should in theory be able to drop "supported" versions of our dependencies without having any breaking changes to the core functionality of cobra. People could still use cobra without issue it just wouldn't be _"supported"_.

I'm ok with the change in this PR but i think it'd be worth re-visiting the deprecation policy all together. It'll never be sustainable for us to make major releases with how fast Go deprecates it's versions.

stmcginnis: I have to agree. Removing support for an old, end-of-life'd, version of Go without changing any backwards functionality doesn't seem like it's worth a major version bump. It would be good to update that policy to clarify that, but I wouldn't think it's something that needs to happen before this change is made.

jpmcb (approved):

marckhouzam (commented):
This seems reasonable, but I am not well versed in GH actions.
I leave it up to you guys.

stmcginnis (approved):

jpmcb (approved):

PR #1785, opened 2022-08-22
Update golangci configuration

umarcor:


marckhouzam:
Thanks @umarcor, can you provide details on why you had/chose to make these changes?  Thanks

umarcor:
@marckhouzam I just applied the "defaults" as currently documented in https://golangci-lint.run/usage/linters/. I enabled the ones listed in https://golangci-lint.run/usage/linters/#enabled-by-default, which were not enabled here. By the way, I cleaned up the ones marked as "no longer maintained" or "replaced by".

umarcor:
@marckhouzam also, the makefile was modified because `gofmt` is now enabled in the golangci-lint configuration. Hence, it is redundant to run it independently.

.golangci.yml
```diff
@@ -2,47 +2,105 @@ run:
...
+    - gosimple
+    - govet
+    - ineffassign
+    - staticcheck
+    - typecheck
+    - unused
+    - varcheck
+   # Disabled by Default
+    #- asasalint
+    #- asciicheck
```
jpmcb, line 20: I didn't realize our `.golangci.yml` had abunch of rules commented out. That seems weird. And makes the linter file harder to read: I don't think anyone really needs to know what rules we explicitly aren't using, that's implied by them not being in the file to begin with.

Can we just remove them?

Makefile
```diff
@@ -9,27 +9,21 @@ ifeq (, $(shell which richgo))
...
+.PHONY: lint test clean
 
 default: all
 
-all: fmt test
+all: test
 
-fmt:
-	$(info ******************** checking formatting ********************)
-	@test -z $(shell gofmt -l $(SRC)) || (gofmt -d $(SRC); exit 1)
```
jpmcb, line 20: So we no longer need `gofmt` because our linter handles it? We should change this to be a call out to the `golangci-lint` commandline utility then.

jpmcb (changes requested):

Makefile
umarcor: See the 'lint' target in the same Makefile.

jpmcb: Ah gotcha - but is `golangci-lint` going to format poorly formatted code?

umarcor: No. It will check if the format is correct. That's what the current `fmt` Makefile target does:

https://github.com/spf13/cobra/blob/a281c8b47b315fe428cf31ad5c38b526dc136f55/Makefile#L19-L20

This PR does not introduce any change in that regard.

Makefile
```diff
@@ -9,27 +9,21 @@ ifeq (, $(shell which richgo))
 $(warning "could not find richgo in $(PATH), run: go install github.com/kyoh86/richgo@latest")
 endif
 
-.PHONY: fmt lint test install_deps clean
+.PHONY: lint test clean
 
 default: all
 
-all: fmt test
```
marckhouzam, line 16: Removing `fmt` here means that when people run `make` locally, they will no longer get errors for formatting errors.  Considering the linter takes less tan 2 seconds to run, I suggest running it for `make all`.
If someone does not want to run it, they can do `make test` directly.

What do you think?

umarcor: `fmt` is removed here because the Makefile target `fmt` is removed in this PR. We can add `lint` in this target, but that will require contributors to have golangci-lint installed before running `make`. I think it is better not to require having golangci-lint in order to `make` cobra. At the same time, it does not make sense to have an specific make target for one single linter but not for others. Overall, the users/developers interested in contributing code will always want to `make lint test clean`.

marckhouzam: When I started contributing I would do `make` and think everything was ok.  It was only later that I realized this did not include the linters, so now I run `make lint` manually.

But I agree that making `golangc-lint` a requirement may be annoying.  But it will also be annoying to only notice the bad formatting once CI is run...

I'm not sure what is best...

umarcor: IMHO modifying the sources and building cobra are not directly related to contributing code back to the upstream. That is, users might want to use and modify this open source tool regardless of the linting rules enforced in this repository. From this point of view, the requirement/need to use golangci-lint belongs to contributing guidelines: https://github.com/spf13/cobra/blob/main/CONTRIBUTING.md. Everyone should read the contributing guidelines before submitting a PR to a project.

PR #1795, opened 2022-09-05
Add Parent FlagSet

fahedouch:
support pFlag `AddParentFlagSet` (https://github.com/spf13/pflag/pull/357)
Signed-off-by: fahed dorgaa <fahed.dorgaa@gmail.com>

### DIFF
--- a/command.go
+++ b/command.go
@@ -1672,7 +1672,7 @@ func (c *Command) Parent() *Command {
 func (c *Command) mergePersistentFlags() {
 	c.updateParentsPflags()
 	c.Flags().AddFlagSet(c.PersistentFlags())
-	c.Flags().AddFlagSet(c.parentsPflags)
+	c.Flags().AddParentFlagSet(c.parentsPflags)
 }
 
 // updateParentsPflags updates c.parentsPflags by adding


command.go
```diff
@@ -1672,7 +1672,7 @@ func (c *Command) Parent() *Command {
 func (c *Command) mergePersistentFlags() {
 	c.updateParentsPflags()
 	c.Flags().AddFlagSet(c.PersistentFlags())
-	c.Flags().AddFlagSet(c.parentsPflags)
+	c.Flags().AddParentFlagSet(c.parentsPflags)
```
AkihiroSuda, line 1675: Needs unit testing.

Also, not sure whether changing the default behavior is allowed. Probably this should be configurable.

jpmcb (commented):
Can you give a broader description of this change and what use case it's trying to solve? 

Also, I'm assuming this is a subset of this PR? https://github.com/spf13/pflag/pull/357 We would need to wait for a new `pflag` release and then consume it from the upstream.

marckhouzam:
As we have very maintainers who have very little time, we've decided to remove the auto-close configuration.
I'm re-opening this PR since it was closed automatically.

PR #1800, opened 2022-09-07
Use correct stale action `exempt-*` yaml keys

jpmcb:
Somehow we had the wrong `exempt-` keys in our stale action yaml (or maybe they changed it) causing the stale action to label some `lifecycle/frozen` issues and PR as stale.

This should fix that and make `lifecycle/frozen` labeling do the right thing


marckhouzam (approved):
Thanks for figuring this out!

jpmcb:
Thanks Marc!!!

PR #1802, opened 2022-09-08
Support for case-insensitive command names

### LINKED ISSUE #1382: Add support for case-insensitive command names
Right now users decide on supported command name using `Command::Use` and `Command::Aliases`. There is no way to tell Cobra to use case-insensitive name matching.

Cobra has `SetGlobalNormalizationFunc` which can be used to achieve case-insensitive flag name matching but there is no such option for command names. Generating aliases with all possible spellings isn't an option because users would need to have a list of aliases that is `2**len(commandName)` characters long. Users aren't able to customize the `Command::Find` behavior because `Command::AddCommand` function takes a `struct` and not an `interface`.

@jpmcb  @jharshman  @marckhouzam  are maintainers open to accepting a PR that adds functionality which allows users to create commands with case-insensitive command name matching?

# Example

Imagine I'm writing a CLI program like `kubectl`. I want to have `get` subcommand and have `service` (aliases: `svc`, `svcs`) and `deployment` (aliases: `deployments`) subcommands of `get` subcommand. The sub-subcommand names should be case-insensitive similarly to like how you can do `kubectl get sErvICeS`. `kubectl` opted to have no subcommands after `get` and instead takes the type of resource as first argument. If I wanted to have a case-insensitive `service` sub-subcommand there is currently no supported way to achieve that.

YuviGold:
Add a global `EnableCaseInsensitive` variable to allow
case-insensitive command names.

The variable supports commands names and aliases globally.

Resolves #1382

### DIFF
--- a/cobra.go
+++ b/cobra.go
@@ -40,14 +40,23 @@ var templateFuncs = template.FuncMap{
 
 var initializers []func()
 
+const (
+	defaultPrefixMatching  = false
+	defaultCommandSorting  = true
+	defaultCaseInsensitive = false
+)
+
 // EnablePrefixMatching allows to set automatic prefix matching. Automatic prefix matching can be a dangerous thing
 // to automatically enable in CLI tools.
 // Set this to true to enable it.
-var EnablePrefixMatching = false
+var EnablePrefixMatching = defaultPrefixMatching
 
 // EnableCommandSorting controls sorting of the slice of commands, which is turned on by default.
 // To disable sorting, set it to false.
-var EnableCommandSorting = true
+var EnableCommandSorting = defaultCommandSorting
+
+// EnableCaseInsensitive allows case-insensitive commands names. (case sensitive by default)
+var EnableCaseInsensitive = defaultCaseInsensitive
 
 // MousetrapHelpText enables an information splash screen on Windows
 // if the CLI is started from explorer.exe.
--- a/command_test.go
+++ b/command_test.go
@@ -314,7 +314,7 @@ func TestEnablePrefixMatching(t *testing.T) {
 		t.Errorf("aCmdArgs expected: %q, got: %q", onetwo, got)
 	}
 
-	EnablePrefixMatching = false
+	EnablePrefixMatching = defaultPrefixMatching
 }
 
 func TestAliasPrefixMatching(t *testing.T) {
@@ -349,7 +349,7 @@ func TestAliasPrefixMatching(t *testing.T) {
 		t.Errorf("timesCmdArgs expected: %v, got: %v", onetwo, got)
 	}
 
-	EnablePrefixMatching = false
+	EnablePrefixMatching = defaultPrefixMatching
 }
 
 // TestChildSameName checks the correct behaviour of cobra in cases,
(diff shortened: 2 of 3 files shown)


cobra.go
```diff
@@ -40,14 +40,23 @@ var templateFuncs = template.FuncMap{
 
 var initializers []func()
 
+const (
+	DefaultPrefixMatching  = false
+	DefaultCommandSorting  = true
+	DefaultCaseInsensitive = false
```
marckhouzam, line 46: Are you imagining a scenario where a program would like to access these?
I was thinking a program may want to use the same default as Cobra for case-sensitivity of other things; however, it wouldn't be the constant they should use but the value set in `cobra. EnableCaseInsensitive`.

So, if these constants don't need to be exported, let's make them private.

marckhouzam (changes requested):
Thank you @YuviGold for stepping up and implementing this!

I have a couple of comments but this looks great.

marckhouzam:
@jpmcb what do you think of aiming to have this in the 1.6 release?

command_test.go
```diff
@@ -1263,6 +1263,99 @@ func TestSuggestions(t *testing.T) {
 	}
 }
 
+func TestCaseInsensitive(t *testing.T) {
```
jpmcb, line 1266: Hooray tests! 👏🏼

jpmcb (commented):
Overall, I think this is something we can get in the next release - let's address any nits / comments and go for it!

cobra.go
jpmcb: Very true Marc! +1 - I don't expect users to need these constants since they delineate internal default behavior

YuviGold: Agree. Just wanted to make those constants so they can be used in tests where toggling back to default configuration . Casting to private :)

YuviGold:
@marckhouzam  @jpmcb 
Good feedback guys. I addressed all your comments :)

marckhouzam (commented):
Could you add a test to make sure `defaultCaseInsensitive` is false?  That way if a future PR tries to change it, we'll noticed right away that it's not allowed.

YuviGold:
> Could you add a test to make sure `defaultCaseInsensitive` is false? That way if a future PR tries to change it, we'll noticed right away that it's not allowed.

@marckhouzam 
Test was added

marckhouzam (approved):
Thanks @YuviGold this looks great!

LGTM

command_test.go
```diff
@@ -1263,6 +1263,99 @@ func TestSuggestions(t *testing.T) {
...
+	}
+
+	for _, test := range tests {
+		for _, enableCaseInsensitivity := range []bool{true, false} {
+			EnableCaseInsensitive = enableCaseInsensitivity
+
+			output, err := executeCommand(rootCmd, test.args...)
+			expectedFailure := test.failWithoutEnabling && !enableCaseInsensitivity
+
+			if !expectedFailure && output != "" {
```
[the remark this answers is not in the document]
marckhouzam: > I've learned that it's the thoroughness of our tests that help us accept changes more quickly. So forgive me for being demanding.
> 
> These new tests cover the cases where we set `EnableCaseInsensitive=true`. But they won't detect is `EnableCaseInsensitive=false` but no error message is received. Could that be added?

Forget this comment, it is an old one I forgot to delete.  Your new tests covers us properly.

PR #1813, opened 2022-09-21
Include --help and --version flag in completion

### LINKED ISSUE #1786: [feature request] completion for --help/--version
Hi.

It would be nice to allow the user to auto complete flags like `--help, `--version` etc.

For example, the `git` completion allows `git --v[TAB]` to auto complete to `git --version`.

TIA

marckhouzam:
Fixes #1786

Thanks to @fnickels's nice work on #1707, I realized why the `--help`/`-h` and `--version`/`-v` flags where not included in shell completions.

The `--help`, `-h`, `--version` and `-v` flags are normally added when the `execute()` function is called on a command.  When doing completion we don't call `execute()` so we need to add these flags explicitly to the command being completed.

Also, when doing shell completion following those flags, this PR disables all completions.
To do so, we keep track of if the `--help` and `--version` flags were added by Cobra using a new flag annotation.

To test:
```
./prog -<TAB>
# Notice that --help and -h are listed with this PR but not without it

./prog --help <TAB>
# Nothing is suggested with this PR but file completion is triggered without the PR
```

### DIFF
--- a/command.go
+++ b/command.go
@@ -30,6 +30,8 @@ import (
 	flag "github.com/spf13/pflag"
 )
 
+const FlagSetByCobraAnnotation = "cobra_annotation_flag_set_by_cobra"
+
 // FParseErrWhitelist configures Flag parse errors to be ignored
 type FParseErrWhitelist flag.ParseErrorsWhitelist
 
@@ -1055,6 +1057,7 @@ func (c *Command) InitDefaultHelpFlag() {
 			usage += c.Name()
 		}
 		c.Flags().BoolP("help", "h", false, usage)
+		_ = c.Flags().SetAnnotation("help", FlagSetByCobraAnnotation, []string{"true"})
 	}
 }
 
@@ -1080,6 +1083,7 @@ func (c *Command) InitDefaultVersionFlag() {
 		} else {
 			c.Flags().Bool("version", false, usage)
 		}
+		_ = c.Flags().SetAnnotation("version", FlagSetByCobraAnnotation, []string{"true"})
 	}
 }
 
--- a/completions.go
+++ b/completions.go
@@ -274,6 +274,12 @@ func (c *Command) getCompletions(args []string) (*Command, []string, ShellCompDi
 	}
 	finalCmd.ctx = c.ctx
 
+	// These flags are normally added when `execute()` is called on `finalCmd`,
+	// however, when doing completion, we don't call `finalCmd.execute()`.
+	// Let's add the --help and --version flag ourselves.
+	finalCmd.InitDefaultHelpFlag()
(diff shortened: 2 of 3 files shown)


marckhouzam:
I realized that the completion logic for this improvement is still lacking.  For example: `helm status --help <TAB>` will still suggest a list of helm releases.

I will need to treat the `--help` and `--version` flags as exception in Cobra.  This actually make sense since these flags are added by Cobra. (Bare with my being a bit OCD about making completion work perfectly 😉 )

I will post an updated version soon.

jpmcb:
Nice! This all sounds good - give me a ping once this is ready to review! :D

marckhouzam:
Alright, this is ready @jpmcb.  If you ignore the 234 lines of new tests, the change is fairly straightforward (< 30 lines). 

One important aspect that you may have a different opinion on is that I have added a new flag annotation: `FlagSetByCobraAnnotation`.  It is used to keep track if a flag was added by Cobra (the `help` and `version` flags).  This was to avoid short-circuiting completions if these flags were program flags (not added by Cobra); if it is the program that adds such flags, it is the program's responsibility to handle shell completion adequately. 

What to look for when testing (any generic test program will do):
1.  see the `--help` and `-h` flags being completed for any command (e.g., prog -<TAB>)
2. see the `--version` and `-v` flags being completed for commands that have their `Version` field set (e.g., prog -<TAB>)
3. no further completion AFTER any of `--help`, `-h`, `--version` and `-v`, but only if they were created by Cobra (not created by the program itself).  Note that there are also Go tests added for this case. (e.g., prog -h <TAB>)

Also the tests of the cobra-completion-testing repo all pass.
I have also done some manual testing with helm.

command.go
```diff
@@ -30,6 +30,8 @@ import (
 	flag "github.com/spf13/pflag"
 )
 
+const FlagSetByCobraAnnotation = "cobra_annotation_flag_set_by_cobra"
```
jpmcb, line 33: Just to clarify, this is a public const so that users could inspect the flag annotations themselves if needed?

jpmcb (approved):

command.go
marckhouzam: Honestly I wasn't too sure what was best.  Would it be useful to external programs?  I don't know.  So I ended up following the other annotations that were all public:
https://github.com/spf13/cobra/blob/7039e1fa214cfc1de404ed6540158c8fda64a758/bash_completions.go#L30-L33
But we can make it private if you prefer.

jpmcb: > So I ended up following the other annotations that were all public

That sounds good! Thanks for clarifying!! happy to follow the current convention 👍🏼

PR #1817, opened 2022-09-28
completions: do not detect arguments with dash as 2nd char as flag

### LINKED ISSUE #1816: Completion detects non-flag argument as flag
We have encountered some odd behavior in the auto completion.
It appears that arguments with a dash as the second character are detected as flags by mistake.

The full reproducer with various test cases can be found here:
https://go.dev/play/p/UxqRkMcEX2a

The minimal reproducer is the following:

```go
package main

import "github.com/spf13/cobra"

func main() {
	root := &cobra.Command{
		Use: "root",
	}
	root.SetArgs([]string{"__complete", "1-ff00:0:1", ""})
	root.Execute()
}
```

which results in the following output:

```
[Debug] [Error] Subcommand 'root' does not support flag '1'
:0
```

I tracked down the issue and appears that we enter this branch by mistake:
https://github.com/spf13/cobra/blob/fce8d8aeb08dc6afe413cc0af67a7fbb3cffec4c/completions.go#L577-L595

In particular, it appears that the is `isFlagArg` classifies the argument as a flag byistake:

https://github.com/spf13/cobra/blob/fce8d8aeb08dc6afe413cc0af67a7fbb3cffec4c/command.go#L629-L632

oncilla:
Previously, arguments with a dash as the second character (e.g., 1-ff00:0:1) were detected as a flag by mistake.
This resulted in auto completion misbehaving if such an argument was last in the argument list during invocation.

Fixes #1816

### DIFF
--- a/command.go
+++ b/command.go
@@ -692,7 +692,7 @@ Loop:
 }
 
 func isFlagArg(arg string) bool {
-	return ((len(arg) >= 3 && arg[1] == '-') ||
+	return ((len(arg) >= 3 && arg[0:2] == "--") ||
 		(len(arg) >= 2 && arg[0] == '-' && arg[1] != '-'))
 }
 
--- a/completions_test.go
+++ b/completions_test.go
@@ -3115,3 +3115,67 @@ func TestCompletionCobraFlags(t *testing.T) {
 		})
 	}
 }
+
+func TestArgsNotDetectedAsFlagsCompletionInGo(t *testing.T) {
+	// Regression test that ensures the bug described in
+	// https://github.com/spf13/cobra/issues/1816 does not occur anymore.
+
+	root := Command{
+		Use: "root",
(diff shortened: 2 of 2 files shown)


marckhouzam (commented):
Would you be able to add a test that fails before and passes after your change?

marckhouzam:
@oncilla just checking if you will have time to add tests to this PR?

oncilla:
@marckhouzam 
Sorry, I completely forgot about this PR.
I should have time soon to add a unit test. How extensive do you want it to be?

Would it be enough to just add another test for this, or should I integrate with an existing test?

marckhouzam:
I think another test would be better. Thanks!

oncilla:
@marckhouzam updated with a regression test.

marckhouzam (approved):
Thanks @oncilla

PR #1822, opened 2022-10-06
Improve accessibility of markdown docs generation result

kangasta:
Present available options and commands in tables. Also adds aliases of a command to markdown docs and omits short command description if long description is defined (similarly than in command usage template used for --help output).


kangasta:
Example before/after as a [gist](https://gist.github.com/kangasta/1380a52060041be14adcddacce435898)

doc/md_docs.go
```diff
@@ -25,23 +25,52 @@ import (
...
+	"github.com/spf13/pflag"
 )
 
+// For formatting options and aliases. E.g., ["--help", "-h"] becomes "`--help`, `-h`"
+func formatFlags(input []string) string {
+	return fmt.Sprintf("`%s`", strings.Join(input, "`, `"))
+}
+
+func formatDescription(input string) string {
+	return strings.Replace(input, "\n", "<br/>", -1)
```
kangasta, line 37: Alternative would be to replace line changes with spaces, but that would result to different formatting than in `--help` messages. `<br/>` HTML tag should be quite widely supported by markdown parsers.

kangasta:
Adding a comment to keep this from being closed.

This would make the options listing in documentation generated by the included docs generator more screen-reader accessible by displaying options info in tables instead of long unstructured code blocks.

Issue #1828, opened 2022-10-13
Add a way to bypass the terminal autocomplete matching


raphaelvigee:
This would allow to support fuzzy completions:
Lets say `ValidArgsFunction` return `[]string{"some_arg", "some_arg_thats_long"}`

```bash
$ mycmd arg<TAB>
```

I should be presented with the autocomplete options:

```
some_arg
some_arg_thats_long
```

Currently those would be filtered out because of the use of `compgen` on bash and `_describe` on zsh (haven't checked fish) those will be filtered out as the default behavior is to filter based on prefix.

The addition of another directive `ShellCompDirectiveNoMatching` (something like that) would disable the matching

marckhouzam:
Hi @raphaelvigee.  Fuzzy matching is already possible for `ValidArgsFunction` and `RegisterFlagCompletionFunc`.  All that is required is for you not to filter on prefix.  Helm does this for example: https://github.com/helm/helm/pull/10513.

However, this only works for `zsh` and `fish`.  I was under the impression that `bash` did not handle fuzzy matching so the Cobra bash script does not allow it.  Could you give an example where you say `bash` support fuzzy matching?

Finally, for completeness, be aware Cobra does filter on the prefix when it does completion itself.  For instance, when completion subcommands and flag names.  I have been hesitant to change that as I am not convinced it is a common scenario.

raphaelvigee:
I can confirm that fuzzy matching does not work on zsh, the situation highlighted above returns no completion in the terminal as they get filtered out in `_describe` (which follows the user-defined matchers), but i think the CLI author should be able to say "you defo want fuzzy matching here", hence passing the extra directive.
Debugging it, i can see the completion up until it passes through `_describe`.

From the zsh debug:
```
+_mycmd:68> __mycmd_debug $'completions: some_arg\nsome_arg_thats_long'
...
+_mycmd:160> __mycmd_debug 'Calling _describe'
...
+_mycmd:167> __mycmd_debug '_describe did not find completions.'
```

A similar thing is happening with bash

I don't have an example in mind that does fuzzy matching, but for my current use-case, fuzzy matching is almost necessary, basically: `mycmd` takes a single argument which is a list of urls, these urls are long and complicated, and fuzzy matching would make the user life much easier

marckhouzam:
For fuzzy matching to work for `zsh` you need to set the `matcher-list` in `zstyle`.  I get this automatically with `oh-my-zsh`.

```
[...]

PR #1833, opened 2022-10-19
Implement completion NoMatching

raphaelvigee:
Re #1828 

Was only able to get it to work on zsh & fish. bash not supported, was not able to test powershell


fish_completions.go
```diff
@@ -141,19 +142,24 @@ function __%[1]s_prepare_completions
...
     # To do so, we will filter on prefix as the completions we have received
     # may not already be filtered so as to allow fish to match on different
     # criteria than the prefix.
     if test $nospace -ne 0; or test $nofiles -eq 0
-        set -l prefix (commandline -t | string escape --style=regex)
-        __%[1]s_debug "prefix: $prefix"
-
-        set -l completions (string match -r -- "^$prefix.*" $__%[1]s_comp_results)
+		if test $nomatching -ne 0
+			set -l completions (string match -r -- "^.*" $__%[1]s_comp_results)
```
marckhouzam, line 156: Doesn't that match everything?  I guess we don't need the `string match` operation in this case.

fish_completions.go
```diff
@@ -141,19 +142,24 @@ function __%[1]s_prepare_completions
...
 
-    __%[1]s_debug "nospace: $nospace, nofiles: $nofiles"
+    __%[1]s_debug "nospace: $nospace, nofiles: $nofiles, nomatching: $nomatching"
 
     # If we want to prevent a space, or if file completion is NOT disabled,
     # we need to count the number of valid completions.
     # To do so, we will filter on prefix as the completions we have received
     # may not already be filtered so as to allow fish to match on different
     # criteria than the prefix.
     if test $nospace -ne 0; or test $nofiles -eq 0
```
marckhouzam, line 154: This is a special case which the comment may not have made very clear.

Let's take file completion.  We only want to trigger file completion if there are no other completions available (from the point of view of the user).  This is not normally what fish does, but it is how the other shells behave so we decided to standardize on that.

However, if the program does not filter on prefix (to allow fuzzy matching), it will always return some completions and we will not be able to know if fish will match any of them or not; so, we would not know if there are no real completions and couldn't tell if we should trigger file completion.  Therefore, in this case, we turn off fuzzy matching and explicitly match on prefix; if there are no completions left, we know to turn on file completion.

A similar scenario applies to the `nospace` directive.  You can refer to #1249 for details.

This means that in this case, we cannot do fuzzy completion, or it would break the request from the program to do file completion and/or nospace completion.

I don't know if that was clear, so don't hesitate to ask for clarification.

zsh_completions.go
```diff
@@ -108,8 +108,9 @@ _%[1]s()
     local shellCompDirectiveNoFileComp=%[5]d
     local shellCompDirectiveFilterFileExt=%[6]d
     local shellCompDirectiveFilterDirs=%[7]d
+	local shellCompDirectiveNoMatching=%[8]d
```
marckhouzam, line 111: Please use spaces instead of tab for the script part

fish_completions.go
```diff
@@ -141,19 +142,24 @@ function __%[1]s_prepare_completions
...
-
-        set -l completions (string match -r -- "^$prefix.*" $__%[1]s_comp_results)
+		if test $nomatching -ne 0
+			set -l completions (string match -r -- "^.*" $__%[1]s_comp_results)
+		else
+			set -l prefix (commandline -t | string escape --style=regex)
+			__%[1]s_debug "prefix: $prefix"
+	
+			set -l completions (string match -r -- "^$prefix.*" $__%[1]s_comp_results)
+		end
```
marckhouzam, line 162: Please use spaces, not tabs in the script part

marckhouzam (changes requested):
Thanks @raphaelvigee.
Could you provide a small test program that uses the new directive so I can figure out how the user would use this?

fish_completions.go
raphaelvigee: Hum I see, what do you suggest changing the logic to in that case ?

marckhouzam:
> Thanks @raphaelvigee. Could you provide a small test program that uses the new directive so I can figure out how the user would use this?

To be honest, I still don't understand what the PR is trying to achieve.   Fuzzy matching works for me on both zsh and fish right now.  

Could it be that fuzzy matching does not work for you for fish because FileCompletion or NoSpace are requested?  Which is a current limitation as explained in https://github.com/spf13/cobra/pull/1833#discussion_r1002153126

raphaelvigee:
@marckhouzam Sorry I missed that, i m pretty sure zsh does prefix matching, which is what this PR is trying to change to fuzzy matching, but i ll double check

Issue #1831, opened 2022-10-17
Subcommand Group |  panic: Group id 'Group1' is not defined for subcommand 'app1 sub1'


KINGSABRI:
Hi everyone 
I'm trying to implement the subcommand group feature introduced in v[v1.6.0]. I followed the [test code](https://github.com/spf13/cobra/blob/1424b7b92743e634e20f611e99a495e411fa4732/command_test.go#L1770)

`/app1/cmd/root.go`
```go
func init() {
	cobra.EnableCaseInsensitive = true
	cobra.EnableCommandSorting = false

	rootCmd.CompletionOptions.DisableDefaultCmd = true 
	rootCmd.Flags().SortFlags = false
	rootCmd.PersistentFlags().SortFlags = false

	rootCmd.AddGroup(&cobra.Group{ID: "Group1", Title: "Group1 Subcommands:"})
}
```

 `/app1/cmd/sub1.go`
```go
var sub1Cmd = &cobra.Command{
	Use:     "sub1",
	GroupID: "Group1",
	Aliases: []string{"1"},
	Run: func(cmd *cobra.Command, args []string) {
		fmt.Println("Hiii")
	},
}

func init() {
	rootCmd.AddCommand(sub1Cmd)
	sub1Cmd.Flags().SortFlags = false
	sub1Cmd.Flags().StringP("ipaddress", "i", "0.0.0.0", "IP address to listen to")
}

```

When I run the app I get 
```
panic: Group id 'Group1' is not defined for subcommand 'app1 sub1'
goroutine 1 [running]:
github.com/spf13/cobra.(*Command).AddCommand(0xd024e0, {0xc00024fd08, 0x1, 0x1?})
        /home/USER/go/pkg/mod/github.com/spf13/cobra@v1.6.0/command.go:1223 +0x2d4
app1/cmd.init.3()
        /home/USER/Code/app1/cmd/sub1Cmd.go:47 +0x54
```

Q\ is there any complex example that wrapup all cobra's features?

KINGSABRI:
Hi, 
I'm not sure why the issue is being submitted twice. 

Your advice regarding the issue is appreciated 

Thanks

andig:
Same here. It's not obvious where the Groups need be added in order to be specified on the child command?

aawsome:
The `AddGroup` command must be executed before the `AddCommand(sub1Cmd)` is executed. 

[...]

PR #1835, opened 2022-10-20
Make Groups public field for command

andig:
Fix https://github.com/spf13/cobra/issues/1831. 

This PR is a BC break as it replaces `Groups()` with the `Groups` field. This could be fixed by keeping `CommandGroups` as public member name but felt clearer not duplicating the api.

### DIFF
--- a/command.go
+++ b/command.go
@@ -138,7 +138,7 @@ type Command struct {
 	PersistentPostRunE func(cmd *Command, args []string) error
 
 	// groups for subcommands
-	commandgroups []*Group
+	Groups []*Group
 
 	// args is actual args parsed from flags.
 	args []string
@@ -1244,11 +1244,6 @@ func (c *Command) AddCommand(cmds ...*Command) {
 	}
 }
 
-// Groups returns a slice of child command groups.
-func (c *Command) Groups() []*Group {
-	return c.commandgroups
-}
-
 // AllChildCommandsHaveGroup returns if all subcommands are assigned to a group
 func (c *Command) AllChildCommandsHaveGroup() bool {
 	for _, sub := range c.commands {
@@ -1261,7 +1256,7 @@ func (c *Command) AllChildCommandsHaveGroup() bool {
 
 // ContainGroups return if groupID exists in the list of command groups.
 func (c *Command) ContainsGroup(groupID string) bool {
-	for _, x := range c.commandgroups {
+	for _, x := range c.Groups {
 		if x.ID == groupID {
 			return true
 		}
@@ -1271,7 +1266,7 @@ func (c *Command) ContainsGroup(groupID string) bool {
 
 // AddGroup adds one or more command groups to this parent command.
 func (c *Command) AddGroup(groups ...*Group) {
-	c.commandgroups = append(c.commandgroups, groups...)
+	c.Groups = append(c.Groups, groups...)
 }
 
 // RemoveCommand removes one or more commands from a parent command.
--- a/command.go
+++ b/command.go
@@ -138,7 +138,7 @@ type Command struct {
 	PersistentPostRunE func(cmd *Command, args []string) error
 
 	// groups for subcommands
-	Groups []*Group
+	CommandGroups []*Group
 
 	// args is actual args parsed from flags.
 	args []string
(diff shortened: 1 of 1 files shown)


marckhouzam:
Thanks @andig .
With Cobra we cannot break backwards-compatibility.  It is crucial that a project continue to compile after upgrading to a new release of Cobra.  Could you update the PR to retain backwards-compatibility?

Also could you describe how users would make use of this change to improve the use of groups?

andig:
@marckhouzam the feature has only been introduced 9 days ago and like does not have wide-spread use. It also encourages a close coupling. Happy to rework, but it seems better to make this change cleanly and quickly.

marckhouzam:
> @marckhouzam the feature has only been introduced 9 days ago and like does not have wide-spread use. It also encourages a close coupling. Happy to rework, but it seems better to make this change cleanly and quickly.

I'll let @jpmcb disagree, but in my opinion, once it is out, that's it, projects can be using it.  That is why I was so detailed on getting as good an API as we could from the start.

jpmcb:
I'm not seeing how this is a breaking change?

Looks like we want to add the `commandGroups` as part of the public API and since it wasn't to begin with, I don't believe adding it will break anyone's programs?

I'm also not seeing how this solves https://github.com/spf13/cobra/issues/1831. Can you provide an example using this newly exposed API from this patch to give an explanation / example?

andig:
> I'm not seeing how this is a breaking change?

@jpmcb https://github.com/spf13/cobra/pull/1835/commits/7aa995fd55a5862574d401d7005f330a10276e28 is a breaking change (`Groups()` -> `Groups`). https://github.com/spf13/cobra/pull/1835/commits/cd90af1646b39bf0e07132e17c18c9f8e970c3b4 isn't and will have both `Groups()` and `CommandGroups`.

> I'm also not seeing how this solves https://github.com/spf13/cobra/issues/1831.

The problem with #1831 is that the root commands groups must be created for the sub command to use them. Typically root command it created during `init()` which runs _after_ subcommand is declared and hence panics. This PR fixes that by adding groups to the root command variable at creation time instead of during `init()`.

andig:
> I'll let @jpmcb disagree, but in my opinion, once it is out, that's it, projects can be using it. That is why I was so detailed on getting as good an API as we could from the start.

@marckhouzam I do by no means want to imply you did a bad job. I may really be missing something! Just didn't find a good approach that fits (my) current pattern of using cobra commands.

marckhouzam:
> > I'll let @jpmcb disagree, but in my opinion, once it is out, that's it, projects can be using it. That is why I was so detailed on getting as good an API as we could from the start.
> 
> @marckhouzam I do by no means want to imply you did a bad job. Hope it doesn't come across like that.

Not at all!  I was just explaining why we were so demanding with the original review.  Once an API is published we can never break it 😅

And thanks for continuing to improve this feature. Clearly people want it as we already see multiple users 👍

marckhouzam:
I'd like to think a bit about this `init()` ordering issue.
And we'll need to document the final way we settle on so the users know exactly how to create groups safely.

marckhouzam:
Let's look at the basics of the situation.  
Our current API is:
1. Call `AddGroup()`
2. Call `AddCommand()`

These calls *must* be in the above order because (and only because) we verify if all groups properly exist when calling `AddCommand()`.

The problem is that when using `inti()` functions, users don't have a good control of the above order.

Notice that if we don't verify the correct existence of groups, then the ordering does not matter and everything will work.
So, we could simply remove the check and the `panic` from `AddCommand()`.

But as discussed in the original PR, it would become easy for a typo in a group id to go unnoticed which would make the subcommand with the wrong id not to be shown in the help.

So instead of completely removing the check, how about we move it to a point where the order of initialization no longer matters?  I believe we can do that by adding the check in `ExecuteC(), once the `help` and `completion` commands have been created.

What do you think?

andig:
> What do you think?

I had actually thought about that but kinda forgot. Sounds very good- validate existence of _all_ group/IDs without depending on order.

jpmcb:
> So instead of completely removing the check, how about we move it to a point where the order of initialization no longer matters? 

I think that's a very solid solution here: Let's not re-work the API, let's decouple it so calls don't depend on ordering.

andig:
@marckhouzam would you be so kind to provide the follow-up PR? I'm unsure where to put the validation now.

marckhouzam:
> @marckhouzam would you be so kind to provide the follow-up PR? I'm unsure where to put the validation now.

I've created #1839.  If this solution looks good, we can close the current PR.

PR #1836, opened 2022-10-20
Replace deprecated ExactValidArgs with MatchAll in user_guide and shell_completions example

doniacld:
Update the example from `shell_completions` by replacing `ExactValidArgs` which is deprecated from the documentation by `MatchAll(ExactArgs(1), OnlyValidArgs)`.

### DIFF
--- a/shell_completions.md
+++ b/shell_completions.md
@@ -71,7 +71,7 @@ PowerShell:
 `,cmd.Root().Name()),
 	DisableFlagsInUseLine: true,
 	ValidArgs:             []string{"bash", "zsh", "fish", "powershell"},
-	Args:                  cobra.ExactValidArgs(1),
+	Args:                  cobra.MatchAll(cobra.ExactArgs(1), cobra.OnlyValidArgs),
 	Run: func(cmd *cobra.Command, args []string) {
 		switch args[0] {
 		case "bash":
--- a/user_guide.md
+++ b/user_guide.md
@@ -349,7 +349,7 @@ shown below:
 ```go
 var cmd = &cobra.Command{
   Short: "hello",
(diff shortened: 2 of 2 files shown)


shell_completions.md
```diff
@@ -71,7 +71,7 @@ PowerShell:
 `,cmd.Root().Name()),
 	DisableFlagsInUseLine: true,
 	ValidArgs:             []string{"bash", "zsh", "fish", "powershell"},
-	Args:                  cobra.ExactValidArgs(1),
+	Args:                  MatchAll(ExactArgs(1), OnlyValidArgs),
```
marckhouzam, line 74: This should be `cobra.MatchAll(ExactArgs(1), OnlyValidArgs),`

marckhouzam (changes requested):
Thanks for noticing @doniacld

shell_completions.md
doniacld: Thanks, I fixed it it is actually `cobra.MatchAll(cobra.ExactArgs(1), cobra.OnlyValidArgs)` :)

marckhouzam: ha!  Good point!

marckhouzam: While we're at it, would you mind also fixing the 3 missing `cobra.` prefixes in:

https://github.com/spf13/cobra/blob/badcce14f8e828cda4d8ac404a12448700de1441/user_guide.md?plain=1#L352

doniacld: For sure, fixed and checking if there is any left.

marckhouzam (approved):
LGTM

marckhouzam:
Thanks @doniacld !  Label added.  Let me know if I have to do something else for this contribution to count for your stats.

PR #1839, opened 2022-10-23
Check for group presence after full initialization

### LINKED ISSUE #1831: Subcommand Group |  panic: Group id 'Group1' is not defined for subcommand 'app1 sub1'

marckhouzam:
Fixes #1831
Replaces #1835

By moving the check for help group existence to "ExecuteC()" we no longer need groups to be added before AddCommand() is called.  This provides more flexibility to developers and works better with the use of "init()" for command creation.

cc @aawsome

### DIFF
--- a/command.go
+++ b/command.go
@@ -998,6 +998,10 @@ func (c *Command) ExecuteC() (cmd *Command, err error) {
 	// initialize completion at the last point to allow for user overriding
 	c.InitDefaultCompletionCmd()
 
+	// Now that all commands have been created, let's make sure all groups
+	// are properly created also
+	c.checkCommandGroups()
+
 	args := c.args
 
 	// Workaround FAIL with "go test -v" or "cobra.test -test.v", see #155
@@ -1090,6 +1094,19 @@ func (c *Command) ValidateRequiredFlags() error {
 		return fmt.Errorf(`required flag(s) "%s" not set`, strings.Join(missingFlagNames, `", "`))
 	}
 	return nil
+}
+
+// checkCommandGroups checks if a command has been added to a group that does not exists.
+// If so, we panic because it indicates a coding error that should be corrected.
+func (c *Command) checkCommandGroups() {
+	for _, sub := range c.commands {
+		// if Group is not defined let the developer know right away
+		if sub.GroupID != "" && !c.ContainsGroup(sub.GroupID) {
+			panic(fmt.Sprintf("group id '%s' is not defined for subcommand '%s'", sub.GroupID, sub.CommandPath()))
+		}
+
+		sub.checkCommandGroups()
+	}
 }
 
 // InitDefaultHelpFlag adds default help flag to c.
@@ -1218,10 +1235,6 @@ func (c *Command) AddCommand(cmds ...*Command) {
 			panic("Command can't be a child of itself")
 		}
 		cmds[i].parent = c
-		// if Group is not defined let the developer know right away
-		if x.GroupID != "" && !c.ContainsGroup(x.GroupID) {
-			panic(fmt.Sprintf("Group id '%s' is not defined for subcommand '%s'", x.GroupID, cmds[i].CommandPath()))
-		}
 		// update max lengths
 		usageLen := len(x.Use)
 		if usageLen > c.commandsMaxUseLen {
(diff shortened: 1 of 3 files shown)


command.go
```diff
@@ -1092,6 +1096,19 @@ func (c *Command) ValidateRequiredFlags() error {
...
+// checkCommandGroups checks if a command has been added to a group that does not exists.
+// If so, we panic because it indicates a coding error that should be corrected.
+func (c *Command) checkCommandGroups() {
+	for _, sub := range c.commands {
+		// if Group is not defined let the developer know right away
+		if sub.GroupID != "" && !c.ContainsGroup(sub.GroupID) {
+			panic(fmt.Sprintf("group id '%s' is not defined for subcommand '%s'", sub.GroupID, sub.CommandPath()))
+		}
+
+		sub.checkCommandGroups()
```
aawsome, line 1108: I think this introduces a minimal greater runtime overhead than checking within AddCommand, as this does an extra recursion over the subcommand tree.
But this should be neglectable.

command_test.go
```diff
@@ -1862,6 +1862,84 @@ func TestAddGroup(t *testing.T) {
 	checkStringContains(t, output, "\nTest group\n  cmd")
 }
 
+func TestMissingGroupFirstLevel(t *testing.T) {
```
aawsome, line 1865: I think this should be named "TestWrongGroupFirstLevel". This also applies to the other negative tests below.

aawsome (approved):
Very good idea @marckhouzam to put this check in the execute branch!

I've got only nit-picking comments; -)

command.go
marckhouzam: Since this function is only called once, the recursion will perform the check once for every single command.
I believe this is identical to when we were making the check in `AddCommand()` which is called for each command.
Isn't it?

aawsome: The check itself is identical. We just have another recursive traversion of the subcommands tree, thats all.

command_test.go
marckhouzam: Done for all four new tests.

jpmcb (approved):
This looks great to me. Maybe we can cut a quick 1.6.1 to push this out since it was unintended behavior we'd like to fix?

marckhouzam:
> This looks great to me. Maybe we can cut a quick 1.6.1 to push this out since it was unintended behavior we'd like to fix?

If you're up for it I think it would improve the user experience for this popular feature 👍

umarcor:
@marckhouzam please milestone this!

marckhouzam:
> @marckhouzam please milestone this!

Thanks! I created a 1.6.1 milestone for this one.

PR #1850, opened 2022-11-13
Force ForEach-Object to return array in pwsh completion

### LINKED ISSUE #1847: Single word is not completed in PowerShell
When I use the Powershell completion in oh-my-posh (https://github.com/JanDeDobbeleer/oh-my-posh/issues/3050), it fails to complete when I want to complete the first word, e.g. `con<tab>` or `com<tab>` - should result in config or completion respectively. Instead neither happens.

After a little digging I found that OMP uses this project and that the PowerShell completion might have a bug.

```powershell
$Values = $Out | ForEach-Object {
		#Split the output in name and description
		$Name, $Description = $_.Split("`t", 2)
		__oh-my-posh_debug "Name: $Name Description: $Description"

		# Look for the longest completion so that we can format things nicely
		if ($Longest -lt $Name.Length) {
			$Longest = $Name.Length
		}

		# Set the description to a one space string if there is none set.
		# This is needed because the CompletionResult does not accept an empty string as argument
		if (-Not $Description) {
			$Description = " "
		}
		@{Name = "$Name"; Description = "$Description" }
	}
```

When there is only one completion result, which is oddly the case for `oh-my-posh __complete con` but not for `oh-my-posh __complete config ex`, then the above code makes `$Values` a hashtable and not an array, which results in a hidden error down the line, because a hashtable has no `Length` property.

```powershell
if (($Directive -band $ShellCompDirectiveNoFileComp) -ne 0 ) {
		__oh-my-posh_debug "ShellCompDirectiveNoFileComp is called";

		if ($Values.Length -eq 0) {
			# Just print an empty string here so the
			# shell does not start to complete paths.
			# We cannot use CompletionResult here because
			# it does not accept an empty string as argument.
			""
			return
		}
	}
```

My simple workaround has been forcing `$Values` to become an array by doing this `[Array]$Values` in the first code fragment.

totkeks:
Fixes #1847

### DIFF
--- a/powershell_completions.go
+++ b/powershell_completions.go
@@ -137,7 +137,7 @@ filter __%[1]s_escapeStringWithSpecialChars {
     }
 
     $Longest = 0
-    $Values = $Out | ForEach-Object {
+    [Array]$Values = $Out | ForEach-Object {
         #Split the output in name and description
 `+"        $Name, $Description = $_.Split(\"`t\",2)"+`
         __%[1]s_debug "Name: $Name Description: $Description"


Luap99 (commented):
LGTM
@marckhouzam PTAL

marckhouzam (approved):
Works great!  Thanks @totkeks.
Thanks @Luap99 for the review.

PR #1854, opened 2022-11-15
Create unit test illustrating unknown flag bug

brianpursley:
Created a unit test that tests the unknown flag error message when the unknown flag is located in different arg positions.

I think this bug (the first test case with the `FIXME` comment) has been around for a long time, and I decided to create a test to demonstrate it.

I haven't really looked into how this might be fixed yet, just thought I would start with a test first.

### DIFF
--- a/command_test.go
+++ b/command_test.go
@@ -2692,3 +2692,46 @@ func TestFind(t *testing.T) {
 		})
 	}
 }
+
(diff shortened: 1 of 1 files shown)


brianpursley:
main branch name got me, should be fixed now :crossed_fingers:

marckhouzam (approved):
Thanks @brianpursley.
You probably have it in mind already but my guess is that `stripFlags()` swallows the argument (or flag name) that follows `--unknown`

PR #1856, opened 2022-11-15
complete aliases for subcommands

### LINKED ISSUE #1852: Command aliases are not tab-completed
Say you have some command

```
	cmd := &cobra.Command{
		Use:                   "context",
		Aliases:               []string{"ctx"},
	}
```

Then completing `ct` will not work:

```
prog __complete ct

:4
```

I think the issue is this, though I have not debugged into it, it's just by reading the code:

```
			// Complete subcommand names, including the help command
			if len(finalArgs) == 0 && !foundLocalNonPersistentFlag {
				// We only complete sub-commands if:
				// - there are no arguments on the command-line and
				// - there are no local, non-persistent flags on the command-line or TraverseChildren is true
				for _, subCmd := range finalCmd.Commands() {
					if subCmd.IsAvailableCommand() || subCmd == finalCmd.helpCommand {
						if strings.HasPrefix(subCmd.Name(), toComplete) {
							completions = append(completions, fmt.Sprintf("%s\t%s", subCmd.Name(), subCmd.Short))
						}
						directive = ShellCompDirectiveNoFileComp
					}
				}
			}
```

This code path does not take the command's aliases into account.

This appears to be a duplicate of #1296, which has been closed with a reference to some discussion that seems inconclusive to me.

Airblader:
When completing a subcommand, also take its aliases into consideration instead of only its name.

fixes #1852

### DIFF
--- a/completions.go
+++ b/completions.go
@@ -424,10 +424,20 @@ func (c *Command) getCompletions(args []string) (*Command, []string, ShellCompDi
 				// - there are no local, non-persistent flags on the command-line or TraverseChildren is true
 				for _, subCmd := range finalCmd.Commands() {
 					if subCmd.IsAvailableCommand() || subCmd == finalCmd.helpCommand {
+						directive = ShellCompDirectiveNoFileComp
+
+						// Only ever complete the name OR one of the aliases, no need to offer multiple matching ones
+						// for the same command.
 						if strings.HasPrefix(subCmd.Name(), toComplete) {
 							completions = append(completions, fmt.Sprintf("%s\t%s", subCmd.Name(), subCmd.Short))
+						} else {
+							for _, alias := range subCmd.Aliases {
+								if strings.HasPrefix(alias, toComplete) {
+									completions = append(completions, fmt.Sprintf("%s\t%s", alias, subCmd.Short))
+									break
+								}
+							}
 						}
-						directive = ShellCompDirectiveNoFileComp
 					}
 				}
 			}
--- a/completions_test.go
+++ b/completions_test.go
@@ -2176,6 +2176,69 @@ func TestValidArgsNotValidArgsFunc(t *testing.T) {
 	}
 }
 
+func TestCommandAliasesCompletionInGo(t *testing.T) {
+	rootCmd := &Command{
+		Use: "root",
+		Run: emptyRun,
+	}
+
+	subCmd := &Command{
+		Use:     "sandstone",
+		Aliases: []string{"slate", "pumice", "pegmatite"},
+		Run:     emptyRun,
+	}
+
+	rootCmd.AddCommand(subCmd)
+
+	testcases := []struct {
+		desc               string
+		toComplete         string
+		expectedCompletion string
+	}{
+		{
+			desc:               "command name",
+			toComplete:         "sand",
+			expectedCompletion: "sandstone",
+		},
+		{
+			desc:               "command name if an alias also matches",
+			toComplete:         "s",
+			expectedCompletion: "sandstone",
+		},
+		{
+			desc:               "alias if command name does not match",
+			toComplete:         "sla",
+			expectedCompletion: "slate",
+		},
+		{
+			desc:               "only one alias if multiple match",
+			toComplete:         "p",
+			expectedCompletion: "pumice",
+		},
+	}
+
+	for _, tc := range testcases {
+		t.Run(tc.desc, func(t *testing.T) {
+			args := append([]string{ShellCompRequestCmd}, tc.toComplete)
+			output, err := executeCommand(rootCmd, args...)
+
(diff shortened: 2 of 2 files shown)


marckhouzam:
This look reasonable.  The important thing is to avoid aliases coming in the way of other completions.
I'll have to play around with the resulting behaviour to see how it feels.

Thanks @Airblader for the contribution.

marckhouzam:
Sorry for the long delay @Airblader.  I finally got back to this.
I've been trying it with `kubectl` and I'm hesitant about he impact of this change.  
For example, the command `kubectl create quota` has an alias to `kubectl create resourcequota`.
So let's say a user does:
```
$ kubectl create <TAB>
clusterrole          -- Create a cluster role
clusterrolebinding   -- Create a cluster role binding for a particular cluster role
configmap            -- Create a config map from a local file, directory or literal value
cronjob              -- Create a cron job with the specified name
deployment           -- Create a deployment with the specified name
ingress              -- Create an ingress with the specified name
job                  -- Create a job with the specified name
namespace            -- Create a namespace with the specified name
poddisruptionbudget  -- Create a pod disruption budget with the specified name
priorityclass        -- Create a priority class with the specified name
quota                -- Create a quota with the specified name
role                 -- Create a role with single rule
rolebinding          -- Create a role binding for a particular role or cluster role
secret               -- Create a secret using specified subcommand
service              -- Create a service using a specified subcommand
serviceaccount       -- Create a service account with the specified name
token                -- Request a service account token
```
Notice that the choices starting with the letter `r` are `role` and `rolebinding`.  This is fine.
Then I add the letter `r` to my command-line to help complete those choices:
```
$ kubectl create r<TAB>
resourcequota  -- Create a quota with the specified name
role           -- Create a role with single rule
rolebinding    -- Create a role binding for a particular role or cluster role
``` 
Now the completion does not complete `role`/`rolebinding` but suddenly provides a new third choice `resourcequota`.
This may be surprising and confusing to a user.

An alternative would be to only complete command aliases if no other completions are available.
In the above example the user would get the following:
```
$ kubectl create r<TAB>
$ kubectl create role<TAB>
role           -- Create a role with single rule
rolebinding    -- Create a role binding for a particular role or cluster role

$ kubectl create res<TAB>
$ kubectl create resourcequota
```

marckhouzam:
I sent that last comment too early, so I have edited it to add an example of the suggestion.

Airblader:
Thanks for getting back! I see your point. Mh. This is a tricky one. There are two ways to use completions: as a visual list to look at and explore (as you did here), and for mindless tab-completion when you already know the command you're looking for and just want to save yourself keystrokes. Essentially either solution is good at one, but bad at the other.

That said, the solution you propose does also make it difficult to discover potential aliases through completions, because completing "r<TAB>" would instantly complete to "role", and the user would never learn about resourcequota.

A third solution would be to just always provide all completions, even for all aliases. I'd argue this is the "technically correct" solution, but I imagine you wouldn't like this one very much?

marckhouzam:
> A third solution would be to just always provide all completions, even for all aliases. I'd argue this is the "technically correct" solution, but I imagine you wouldn't like this one very much?

That would be always correct, indeed.  But as you expected I feel it would reduce the user experience. It would possibly add a whole bunch of new completions to some already large list of choices.

I'm a big fan of using the shell completion for discovery. However, I’m not sure we can use shell completion to help discover aliases without also reducing the user experience.

I feel like aliases should not be in the way of the normal shell completion. They should just be considered if they don’t add noise.  We can imagine that there can be many aliases for one command that would create noise if we showed them unilaterally.

That’s why I’m thinking that if we do want to consider aliases, we would need to do them only if there are no other completions found.

But this discussion shows that maybe showing aliases altogether is not something we should do, which is maybe why it was never done before.

What do you think?

Airblader:
My concern (and motivation for this PR) isn't so much showing the aliases, but offering the tab completion for them. I do feel strongly that not offering aliases to be completed is more of a usability issue than showing too many completions (personal opinion): why offer an alias at all if it is very awkward to use them?

Since you brought up kubectl as an example, let me name another one that mirrors more closely where this PR is coming from. It's an internal tool similar to minikube, where there is a "pause" command aliases with "suspend" (and others, actually). Suspend might be the more intuitive verb for some users, but not being able to tab complete "sus<TAB>" unfairly penalizes them with a worse experience.

Yet another option here is to give control over this decision to the users of the library. Largely this may even already be achievable (haven't verified), though be a bit awkward to include in every command. A less flexible but easier to use solution would be some global configuration in Cobra. However I would also understand being hesitant to increase configuration surface.

I do want to go on record and say that to me this isn't a minor usability topic, but a fundamental one. Tab completions is IMO one of the most impactful usability features in CLIs, and Cobra specifically has great support for it in general. 🙂

PR #1857, opened 2022-11-15
Completion support for [Nushell](https://nushell.sh)

ayax79:
Completion support for [Nushell](https://nushell.sh). Besides the tests included, I have also tested against minikube.

### DIFF
--- a/nushell_completions.go
+++ b/nushell_completions.go
@@ -1,0 +1,124 @@
+// Copyright 2013-2022 The Cobra Authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package cobra
+
+import (
+	"bytes"
+	"fmt"
+	"io"
+	"os"
+	"regexp"
+	"strings"
+
+	"github.com/spf13/pflag"
+)
+
+var carrageReturnRE = regexp.MustCompile(`\r?\n`)
+
+func descriptionString(desc string) string {
+	// Remove any carriage returns, this will break the extern
+	desc = carrageReturnRE.ReplaceAllString(desc, " ")
+
+	// Lets keep the descriptions short-ish
+	if len(desc) > 100 {
+		desc = desc[0:97] + "..."
+	}
+	return desc
+}
+
+func GenNushellComp(c *Command, buf io.StringWriter, nameBuilder *strings.Builder, isRoot bool, includeDesc bool) {
+	processFlags := func(flags *pflag.FlagSet) {
+		flags.VisitAll(func(f *pflag.Flag) {
+			WriteStringAndCheck(buf, fmt.Sprintf("\t--%[1]s", f.Name))
+
+			if f.Shorthand != "" {
+				WriteStringAndCheck(buf, fmt.Sprintf("(-%[1]s)", f.Shorthand))
+			}
+
+			if includeDesc && f.Usage != "" {
+				desc := descriptionString(f.Usage)
+				WriteStringAndCheck(buf, fmt.Sprintf("\t# %[1]s", desc))
+			}
+
+			WriteStringAndCheck(buf, "\n")
+
+		})
+	}
+
+	cmdName := c.Name()
+	// commands after root name will be like "git pull"
+	if !isRoot {
+		nameBuilder.WriteString(" ")
+	}
+	nameBuilder.WriteString(cmdName)
+
+	// only create an extern block if there is something to put in it
+	if len(c.ValidArgs) > 0 || c.HasAvailableFlags() {
+		builderString := nameBuilder.String()
+
+		// ensure there is a space before any previous content
+		// otherwise it will break descriptions
+		WriteStringAndCheck(buf, "\n")
+
+		funcName := builderString
+		if !isRoot {
+			funcName = fmt.Sprintf("\"%[1]s\"", builderString)
+		}
+
+		if includeDesc && c.Short != "" {
+			desc := descriptionString(c.Short)
+			WriteStringAndCheck(buf, fmt.Sprintf("# %[1]s\n", desc))
+		}
+		WriteStringAndCheck(buf, fmt.Sprintf("export extern %[1]s [\n", funcName))
+
+		// valid args
+		for _, arg := range c.ValidArgs {
+			WriteStringAndCheck(buf, fmt.Sprintf("\t%[1]s?\n", arg))
+		}
+
+		processFlags(c.InheritedFlags())
+		processFlags(c.LocalFlags())
+
+		// End extern statement
+		WriteStringAndCheck(buf, "]\n")
+	}
+
+	// process sub commands
+	for _, child := range c.Commands() {
+		childBuilder := strings.Builder{}
+		childBuilder.WriteString(nameBuilder.String())
+		GenNushellComp(child, buf, &childBuilder, false, includeDesc)
+	}
+
+}
+
+func (c *Command) GenNushellCompletion(w io.Writer, includeDesc bool) error {
+	var nameBuilder strings.Builder
+	buf := new(bytes.Buffer)
+	GenNushellComp(c, buf, &nameBuilder, true, includeDesc)
+
+	_, err := buf.WriteTo(w)
+	return err
+}
+
+func (c *Command) GenNushellCompletionFile(filename string, includeDesc bool) error {
+	outFile, err := os.Create(filename)
+	if err != nil {
+		return err
+	}
+	defer outFile.Close()
+
+	return c.GenNushellCompletion(outFile, includeDesc)
+}
--- a/nushell_completions_test.go
+++ b/nushell_completions_test.go
@@ -1,0 +1,141 @@
+// Copyright 2013-2022 The Cobra Authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package cobra
+
+import (
+	"bytes"
+	"log"
+	"os"
+	"testing"
+)
+
+func TestGenNushellCompletion(t *testing.T) {
+	rootCmd := &Command{
+		Use: "kubectl",
+		Run: emptyRun,
+	}
+	rootCmd.PersistentFlags().String("server", "s", "The address and port of the Kubernetes API server")
+	rootCmd.PersistentFlags().BoolP("skip-headers", "", false, "The address and port of the Kubernetes API serverIf true, avoid header prefixes in the log messages")
+
+	getCmd := &Command{
+		Use:        "get",
+		Short:      "Display one or many resources",
+		ArgAliases: []string{"pods", "nodes", "services", "replicationcontrollers", "po", "no", "svc", "rc"},
+		ValidArgs:  []string{"pod", "node", "service", "replicationcontroller"},
+		Run:        emptyRun,
+	}
+
+	rootCmd.AddCommand(getCmd)
+
+	buf := new(bytes.Buffer)
+	assertNoErr(t, rootCmd.GenNushellCompletion(buf, true))
+	output := buf.String()
+
+	// root command has no local options, it should not be displayed
+	checkOmit(t, output, "export extern kubectl")
+
+	check(t, output, "export extern \"kubectl get\"")
+	check(t, output, "--server")
+	check(t, output, "--skip-headers")
+	check(t, output, "pod?")
+	check(t, output, "node?")
+	check(t, output, "service?")
+	check(t, output, "replicationcontroller?")
+
+	check(t, output, "The address and port of the Kubernetes API serverIf true, avoid header prefixes in the log messages")
+	check(t, output, "The address and port of the Kubernetes API server")
+	check(t, output, "Display one or many resources")
+}
+
+func TestGenNushellCompletionWithoutDesc(t *testing.T) {
+	rootCmd := &Command{
+		Use: "kubectl",
+		Run: emptyRun,
+	}
+	rootCmd.PersistentFlags().String("server", "s", "The address and port of the Kubernetes API server")
+	rootCmd.PersistentFlags().BoolP("skip-headers", "", false, "The address and port of the Kubernetes API serverIf true, avoid header prefixes in the log messages")
+
+	getCmd := &Command{
+		Use:        "get",
+		Short:      "Display one or many resources",
+		ArgAliases: []string{"pods", "nodes", "services", "replicationcontrollers", "po", "no", "svc", "rc"},
+		ValidArgs:  []string{"pod", "node", "service", "replicationcontroller"},
+		Run:        emptyRun,
+	}
+
+	rootCmd.AddCommand(getCmd)
+
+	buf := new(bytes.Buffer)
+	assertNoErr(t, rootCmd.GenNushellCompletion(buf, false))
+	output := buf.String()
+
+	checkOmit(t, output, "The address and port of the Kubernetes API server")
+	checkOmit(t, output, "The address and port of the Kubernetes API serverIf true, avoid header prefixes in the log messages")
+	checkOmit(t, output, "Display one or many resources")
+}
+
+func TestGenNushellCompletionFile(t *testing.T) {
+	err := os.Mkdir("./tmp", 0755)
+	if err != nil {
+		log.Fatal(err.Error())
+	}
+
+	defer os.RemoveAll("./tmp")
+
+	rootCmd := &Command{Use: "root", Args: NoArgs, Run: emptyRun}
+	child := &Command{
+		Use:               "child",
+		ValidArgsFunction: validArgsFunc,
+		Run:               emptyRun,
+	}
+	rootCmd.AddCommand(child)
+
+	assertNoErr(t, rootCmd.GenNushellCompletionFile("./tmp/test", false))
+}
+
+func TestFailGenNushellCompletionFile(t *testing.T) {
+	err := os.Mkdir("./tmp", 0755)
+	if err != nil {
+		log.Fatal(err.Error())
+	}
+
+	defer os.RemoveAll("./tmp")
+
+	f, _ := os.OpenFile("./tmp/test", os.O_CREATE, 0400)
+	defer f.Close()
+
+	rootCmd := &Command{Use: "root", Args: NoArgs, Run: emptyRun}
+	child := &Command{
+		Use:               "child",
+		ValidArgsFunction: validArgsFunc,
+		Run:               emptyRun,
+	}
+	rootCmd.AddCommand(child)
+
+	got := rootCmd.GenNushellCompletionFile("./tmp/test", false)
+	if got == nil {
+		t.Error("should raise permission denied error")
+	}
+
+	if os.Getenv("MSYSTEM") == "MINGW64" {
+		if got.Error() != "open ./tmp/test: Access is denied." {
+			t.Errorf("got: %s, want: %s", got.Error(), "open ./tmp/test: Access is denied.")
+		}
+	} else {
+		if got.Error() != "open ./tmp/test: permission denied" {
+			t.Errorf("got: %s, want: %s", got.Error(), "open ./tmp/test: permission denied")
+		}
+	}
+}
--- a/shell_completions.md
+++ b/shell_completions.md
@@ -6,6 +6,7 @@ The currently supported shells are:
 - Zsh
 - fish
 - PowerShell
+- Nushell
 
 Cobra will automatically provide your program with a fully functional `completion` command,
 similarly to how it provides the `help` command.
@@ -68,9 +69,18 @@ PowerShell:
   # To load completions for every new session, run:
   PS> %[1]s completion powershell > %[1]s.ps1
   # and source this file from your PowerShell profile.
+
+Nushell:
+  
+  # To generate completions (replace YOUR_COMPLETION_DIR with actual path to save)
+  > %[1]s completion nushell | save /YOUR_COMPLETION_DIR/%[1]s-completions.nu
+
+  # To load completions for each session, execute once (replace YOUR_COMPLETION_DIR with actual path):
+  > echo "use /YOUR_COMPLETION_DIR/%[1]s-completions.nu *" | save --append $nu.config-path
+
 `,cmd.Root().Name()),
 	DisableFlagsInUseLine: true,
-	ValidArgs:             []string{"bash", "zsh", "fish", "powershell"},
+	ValidArgs:             []string{"bash", "zsh", "fish", "powershell", "nushell"},
 	Args:                  cobra.MatchAll(cobra.ExactArgs(1), cobra.OnlyValidArgs),
 	Run: func(cmd *cobra.Command, args []string) {
 		switch args[0] {
@@ -82,6 +92,8 @@ PowerShell:
 			cmd.Root().GenFishCompletion(os.Stdout, true)
 		case "powershell":
 			cmd.Root().GenPowerShellCompletionWithDesc(os.Stdout)
+		case "nushell":
+			cmd.Root().GenNushellCompletion(os.Stdout, true)
 		}
 	},
 }
--- a/nushell_completions.md
+++ b/nushell_completions.md
@@ -1,0 +1,4 @@
+## Generating Nushell Completions For Your cobra.Command
+
+Please refer to [Shell Completions](shell_completions.md) for details.
+


marckhouzam:
Hey @ayax79 this exciting!  I have never used `nushell` but I have now installed it on my Mac.

I see that this PR generates a script containing the completions.  However, this is no longer the approach used by Cobra.  Instead, the program, through Cobra, provides the completions itself through the hidden `__complete` command.  The nushell script should then call `prog __complete <all other args on the command-line>` to obtain the list of completions and it should use them as appropriate for nushell.  You can test out the `__complete` command manually by doing things like `prog __complete ''` or `prog __complete --` and such.

This approach means that we have centralized all the logic to generate completions and coded it in Go (in Cobra's `completions.go` file).  Then Cobra generates a shell script for each supported shell which uses the `__complete` command to get the completions.

Following this approach would be your first step.  You can look at the completion scripts for any of the other shells to see how this is done.  These scripts are in files named for the shell they support.

Eventually, the nushell script should also support the ShellCompDirectives, which you will notice in the other scripts.

Feel free to ask for more info if needed.

marckhouzam:
We had a similar situation with Fish where normal completions are a long list of choices. So in our case we have one generic completion entry which calls a function that calls `prog __complete ...` and processes the results.

The nushell doc talks about custom completions which seem like an avenue to investigate. For example it shows:
```
export extern "git push" [
    remote?: string@"nu-complete git remotes"
...
```
which calls `git remotes` to get the list of completions.

marckhouzam:
An important advantage of Cobras approach is that using the `__complete` command in your nushell script will automatically support dynamic completions provided by the program using Cobra (which your solution doesn't).  Please see details here: https://github.com/spf13/cobra/blob/main/shell_completions.md#dynamic-completion-of-nouns

ayax79:
I have played around with a few options and haven't found anything that works well utilizing __command.

This works for only the first level for commands and breaks flags:
```
def 'nu-complete minikube' [] {
	minikube __command
}

export extern minikube [
	command?: string@'nu-complete minikube'
]
```

Screenshot_minikube_level0

Broken help flag:
Screenshot_minikube_broken_help

To access a flag you have to bypass nushell via:
```
^minikube --help
```

By adding another level:
```
def 'nu-complete minikube start' [] {
    minikube start __command
}

export extern 'minikube start' [
	command?: string@'nu-complete minikube start'
]
```
This overrides the completion list and removes the description (unless it is manually added as comment). It also breaks any flags. 

The other option is to utilize the [external completer](https://www.nushell.sh/blog/2022-09-06-nushell-0_68.html#external-completions-experimental-herlon214-rsteube) support. This however only provides an option for one global fallback completer. This how [carapace-bin](https://github.com/rsteube/carapace-bin) works with nushell. 

I'll poke around more, but it seems like I am may be running out of options.

marckhouzam:
Thanks for continuing the investigation @ayax79.

In the above comment you typed `__command` everywhere, but I assume you meant `__complete`?
Also be aware that the format is `prog __complete <other args>`, so the `__complete` command should *always* be the first argument to the program.

It is also important to pass the current command-line arguments that the user typed.  So, if the user typed `minikube start --<tab>` to complete flags, the way to get the completions is to call `minikube __complete start --`.

Finally, if the last character on the command-line before the user pressed TAB is a space, you must pass an extra empty argument.  So, if the user typed `minikube start <TAB>` (notice the space after `start`) you need to call `minikube __complete start ""`.  This complexity allows cobra to tell the difference between the user typing `minikube start<TAB>` which is request to complete the `start` argument, versus `minikube start <TAB>` which is a request to complete the argument *after* `start`.

You can find documentation about all that here: https://github.com/spf13/cobra/blob/main/shell_completions.md#debugging

rsteube:
Yeah you'll definitely have no success with the `extern` approach, it won't allow unknown flags.
Sorry if it wasn't clear before, but you can do some nu scripting magic to support mutliple completers - it's just bad in regards to "registering" them.

```sh
let external_completer = {|spans| 
  {
    $spans.0: { } # default
    kubectl: { kubectl __complete ($spans | into df | slice 1 1000) | handleOutput }
    minikube: { minikube __complete ($spans | into df | slice 1 1000) | handleOutput }
  } | get $spans.0 | each {|it| do $it}
}
```

ayax79:
I finally got some more time to work on this. I also spent some time digging into the cobra completions and the nushell external completions code .

This seems to work, although it needs some more testing. There might be some weirdness around mixing flags and commands:

```
# Executes a cobra apps __complete cmd with the following span
export def exec_cobra_complete [
  cmd: string,  # The cobra command (e.g. kubectl)
  spans: list,  # The list of spans
 ] {
  # skip the first entry in the span (the command) and join the rest of the span to create __complete args
  let cmd_args = ($spans | skip 1 | str join ' ') 

  # If the last span entry was empty add "" to the end of the command args
  let cmd_args = if ($spans | last | str trim | is-empty) {
    $'($cmd_args) ""'
  } else {
    $cmd_args
  }

  # The full command to be executed
  let full_cmd = $'($cmd) __complete ($cmd_args)'

  # Since nushell doesn't have anything like eval, execute in a subshell
  let result = (do -i { nu -c $"'($full_cmd)'" } | complete)

  # Create a record with all completion related info. 
  # directive and directive_str are for posterity
  let stdout_lines = ($result.stdout | lines)
  {
    completions: ($stdout_lines | drop | lines | parse "{value}\t{description}")
    directive_str: ($result.stderr)
    directive: ($stdout_lines | last)
  }
}

# Execute the __complete command and only return completions
export def get_cobra_completions [
  cmd: string,  # The cobra command (e.g. kubectl)
  spans: list,  # The list of spans
] {
  # return just the completions from the result. 
  (exec_cobra_complete $cmd $spans).completions
}

let external_completer = {|spans|
  {
    $spans.0: { }
    kubectl: { get_cobra_completions 'kubectl' $spans }
    minikube: { get_cobra_completions 'minikube' $spans }
  } | get $spans.0 | each {|it| do $it}
}
```

As @rsteube mentioned, registration will be a little bit odd. I think that can be documented in commends for the completion generation though. It will require adding the that code block to the nushell config file ($nu.config-path) and updating config.external_completer entry:
```
let-env config {
    external_completer: $external_completer
}
```

One nice thing is that for other cobra commands, only the external_completer block will need to be updated.

@marckhouzam, if this approach looks good, I'll update nushell_completer.go with it.

marckhouzam:
Thanks @ayax79.
I was able to test the above script and things look promising.
However, we need to come up with a solution to enable completion for more than one tool using Cobra.  Here is where I think it needs some thought.

I'm assuming that each tool, `myprogram` for example, will generate the above script ending with
```
let external_completer = {|spans|
  {
    $spans.0: { }
    myprogram: { get_cobra_completions 'myprogram' $spans }
  } | get $spans.0 | each {|it| do $it}
}
```
The problem is that if I have multiple tools using this, only the last sourced script will work as each one sets the same `external_completer` variable.  What we do in Cobra is prefix each variable and function in the script by `<programName>_`.  In this case the script would set variables such as `myprogram_external_completer`.

After that, we need to figure out a way to have the one
```
let-env config = {
    external_completer: $external_completer
}
```
manage to include _all_ tools that have cobra completions.

ayax79:
If we took this approach, it would still be possible to wrap the cobra completer and chain completers together -- it would be up to the user to figure how they wanted to handle it.

The configuration itself isn't changeable once the shell is loaded. Nushell works much more like a [compiled language](https://www.nushell.sh/book/thinking_in_nu.html#think-of-nushell-as-a-compiled-language).  Dynamic loading of resources and execution isn't really possible. The two phase execution approach is part of the reason why nushell is so fast, but it does create some awkwardness. You can see how I got around this, by creating a subshell to execute the command I assembled:


```
 # Since nushell doesn't have anything like eval, execute in a subshell
 let result = (do -i { nu -c $"'($full_cmd)'" } | complete)
```

What I really don't like about this approach is that nushell's environment has to load to execute the command (I think, need to verify). It _might_ be feasible to do something kind of like execute completions from a directory, but it could get expensive:

```
let external_completer = {|spans|
    let completion_command = $"($env.COMPLETION_DIR)/($spans.0).nu ($spans | str join ' ')"
    nu -c $completion_file 
}
```

Using something like [carapace-bin](https://github.com/rsteube/carapace-bin) is the best use of external_completers, because carpace itself is plugable.

marckhouzam:
Not knowing anything about nushell, I'm having trouble following the subtleties.  
If you could describe the steps the user would take to get the completions to work for multiple tool, we could see if it is something that we can move forward with.

If we can reach a solution where any one tool can generate some nushell completion script and the user could add something to their configuration file to setup any number of such scripts, we would have something sufficient.

ayax79:
I simplified the nushell external configurator to this:
```
# An external configurator that works with any cobra based
# command line application (e.g. kubectl, minikube)
let cobra_configurator = {|spans| 
  let cmd = $spans.0

  # skip the first entry in the span (the command) and join the rest of the span to create __complete args
  let cmd_args = ($spans | skip 1 | str join ' ') 

  # If the last span entry was empty add "" to the end of the command args
  let cmd_args = if ($spans | last | str trim | is-empty) {
    $'($cmd_args) ""'
  } else {
    $cmd_args
  }

  # The full command to be executed
  let full_cmd = $'($cmd) __complete ($cmd_args)'

  # Since nushell doesn't have anything like eval, execute in a subshell
  let result = (do -i { nu -c $"'($full_cmd)'" } | complete)

  # Create a record with all completion related info. 
  # directive and directive_str are for posterity
  let stdout_lines = ($result.stdout | lines)
  let result = ({
    completions: ($stdout_lines | drop | lines | parse "{value}\t{description}")
    directive_str: ($result.stderr)
    directive: ($stdout_lines | last)
  })

  $result.completions
}
```

The process would be:
1. Run `<cmd> completion nushell` to generate the above block
2. Add the above block to the $nu.config-path (location of the nushell config file)
3. Configure the entry external_configurator under config to be $cobra_configurator

There will be no need to change the configuration for any subsequent cobra based app. The above configuration will work with any cobra based app without any further change.

marckhouzam:
Wow!  After sourcing the script in the latest comment and then doing 
```
let-env config = {external_completer: $cobra_configurator}
```
I was able to do completion for helm and kubectl!

One question with this approach: are we going to break completion for any non-cobra program that uses the external_completer? 

@ayax79 Could you update the PR with the new script and then I can start testing more thoroughly?
Also, there are the shell completion directives we'll have to try to support.

One thing to investigate is that flag value completion does not seem to work:
```
$ kubectl --namespace [tab]
NO RECORDS FOUND
```
We should be getting a list of namespaces.

But this is a great start!

ayax79:
The external completer is a fallback. Completions using [externs](https://www.nushell.sh/book/externs.html) will load first. If someone wants to use more than one external completer they could write a completer that wraps two:

```
let chaining_completer = {|spans|
  let completions = (do $cobra_completer $spans)
  let completions = if ($completions | is-empty) {
    (do $other_completer $spans)
  } else {
    $completions
  }
  $completions
}
```

I'll poke around with the flag completions. It should work the same as anything else.

ayax79:
There was a bug in the line parsing causing it to drop entries that weren't exactly {value}\t{description}. Anything without a description was getting dropped. I changed to parse via regex. The only thing that might get dropped with this one is values (commands/flags/options) that have particularly exotic characters, I am matching [\w\-\.:\+]. This is the corrected version:

```
# An external configurator that works with any cobra based
# command line application (e.g. kubectl, minikube)
let cobra_configurator = {|spans| 
 
  let cmd = $spans.0

  # skip the first entry in the span (the command) and join the rest of the span to create __complete args
  let cmd_args = ($spans | skip 1 | str join ' ') 

  # If the last span entry was empty add "" to the end of the command args
  let cmd_args = if ($spans | last | str trim | is-empty) {
    $'($cmd_args) ""'
  } else {
    $cmd_args
  }

  # The full command to be executed
  let full_cmd = $'($cmd) __complete ($cmd_args)'

  # Since nushell doesn't have anything like eval, execute in a subshell
  let result = (do -i { nu -c $"'($full_cmd)'" } | complete)

  # Create a record with all completion related info. 
  # directive and directive_str are for posterity
  let stdout_lines = ($result.stdout | lines)

  let $completions = ($stdout_lines | drop | parse -r '([\w\-\.:\+]*)\t?(.*)' | rename value description)

  let result = ({
    completions: $completions
    directive_str: ($result.stderr)
    directive: ($stdout_lines | last)
  })

  $result.completions
}
```

I will have an updated pull request shortly.

nushell_completions.go
```diff
@@ -0,0 +1,74 @@
...
+
+package cobra
+
+import (
+	"bytes"
+	"io"
+	"os"
+)
+
+func (c *Command) GenNushellCompletion(w io.Writer) error {
```
marckhouzam, line 23: I suggest this function accept a bool parameter to turn off descriptions.  We do that for the other shells.
So, if the program call this function with `false` (GenNushellCompletion(w, false)`) we should generate the same script but instead of using `__complete` we should use `__completeNoDesc`.   This will tell cobra not to include any descriptions with the completions.

nushell_completions.go
```diff
@@ -0,0 +1,74 @@
...
+	buf := new(bytes.Buffer)
+	WriteStringAndCheck(buf, `
+# An external configurator that works with any cobra based
+# command line application (e.g. kubectl, minikube)
+let cobra_configurator = {|spans|
```
marckhouzam, line 28: Let's name this `cobra_completer` as you have put in the documentation

nushell_completions.go
```diff
@@ -0,0 +1,74 @@
...
+
+  # If the last span entry was empty add "" to the end of the command args
+  let cmd_args = if ($spans | last | str trim | is-empty) {
+    $'($cmd_args) ""'
+  } else {
+    $cmd_args
+  }
+
+  # The full command to be executed
+  let full_cmd = $'($cmd) __complete ($cmd_args)'
```
marckhouzam, line 43: I don't know if nushell could support ActiveHelp (see https://github.com/spf13/cobra/blob/main/active_help.md) but unless we do, we should turn it off by setting the environment variable `<PROGRAM>_ACTIVE_HELP=0`.  For example: https://github.com/spf13/cobra/blob/main/fish_completions.go#L55-L56 where the `%[9]s` inserts the name of the variable as specified by the 9th argument to the function

marckhouzam (changes requested):
This is looking great.  Here are some things I noticed:
1. Could you add `nushell` to the default `completion` command? For example: https://github.com/spf13/cobra/blob/main/completions.go#L782-L785
2. The `=` form of flags does not quite work.  When doing `kubectl -n=<tab>`, the `-n=` part is removed
3. Probably linked to the previous point, doing `kubectl --namespace=<tab>` then selecting a completion choice, and then pressing backpace actually crashes the shell
4. When a completion is accepted a space should be added after it.  For example `kubectl comple<tab>` should become `kubectl completion ` (with a space after `completion`) ( except if the `ShellCompDirectiveNoSpace` is specified)
5. Would we be able to support the ShellComp directives?  Or maybe a subset of them? https://github.com/spf13/cobra/blob/7bb14400030441629c9c78b2317dfe9f2254e022/completions.go#L54-L78
6. File completion does not seem to work: if there are no completions choices, the shell should suggests file names (except if the `ShellCompDirectiveNoFileComp` is specified)

shell_completions.md
```diff
@@ -68,9 +69,25 @@ PowerShell:
...
+
+  # 4. Change the config block's external_completer line to be 
+  external_completer: $cobra_completer
+
+  # 5. You will need to start a new shell for this setup to take effect.
+
 `,cmd.Root().Name()),
 	DisableFlagsInUseLine: true,
-	ValidArgs:             []string{"bash", "zsh", "fish", "powershell"},
+	ValidArgs:             []string{"bash", "zsh", "fish", "powershell", "nushell"},
```
marckhouzam, line 90: Could you also add nushell to the example `Use:` line above?

nushell_completions.go
ayax79: Thanks for all the feedback, I work through as quickly as I can.

On the directives and adding a space, I'll dig around the nushell source and see if there is anything there. External completers are a pretty new feature, and I _think_ it really only allows you to return a list of completions of a list of strings for values (names) or a list of values and descriptions. I am not sure there is any way to effect behavior yet. I'll look around to be sure.

On descriptions, the description on/off flag will impact all cobra commands. I could make it easy to adjust after the fact. Thoughts?

nushell_completions.go
ayax79: Oops, good catch!

nushell_completions.go
marckhouzam: > On descriptions, the description on/off flag will impact all cobra commands. I could make it easy to adjust after the fact. Thoughts?

Good point.
Let's leave this point for later. Your solution may evolve in a way that helps that point. We'll see.

nushell_completions.go
ayax79: Active help isn't supported. Adding the env variable to the latest.

nushell_completions.go
ayax79: The file completion was breaking do to the completer returning an empty array instead of null. I was able to implement ShellCompDirectiveNoFileComp using this.

I wasn't able to implement ShellCompDirectiveFilterFileExt or ShellCompDirectiveFilterDirs. There isn't a way to handle this in nushell via external completers.

My fix for -n= is a little bit hacky. I returned the flag as part of the completion to keep it from replacing the flag. It looks a little odd but it worked:

Screen Shot 2022-11-27 at 11 31 14 AM

ayax79:
I just found a nasty bug.

When running things vim, it locks up because vim interprets ``` vim __complete``` as a valid command and never returns.

ayax79:
I think I am going to have to put a whitelist in for commands to execute the completer for:

```
# A list of cobra that completion will be attempted for.
# Add new apps to this list to enable completion for them.
let cobra_apps = ["kubectl", "minikube"]

# An external configurator that works with any cobra based
# command line application (e.g. kubectl, minikube)
let cobra_completer = {|spans| 
  let cmd = $spans.0

  if not ($cobra_apps | where $it == $cmd | is-empty) {
    let ShellCompDirectiveError = 1
    let ShellCompDirectiveNoSpace = 2
    let ShellCompDirectiveNoFileComp = 4
    let ShellCompDirectiveFilterFileExt = 8
    let ShellCompDirectiveFilterDirs = 16
   
    let last_span = ($spans | last | str trim)

    # skip the first entry in the span (the command) and join the rest of the span to create __complete args
    let cmd_args = ($spans | skip 1 | str join ' ') 

    # If the last span entry was empty add "" to the end of the command args
    let cmd_args = if ($last_span | is-empty) {
      $'($cmd_args) ""'
    } else {
      $cmd_args
    }

    # The full command to be executed with active help disable (Nushell does not support active help)
    let full_cmd = $'($cmd)_ACTIVE_HELP=0 ($cmd) __complete ($cmd_args)'

    # Since nushell doesn't have anything like eval, execute in a subshell
    let result = (do -i { nu -c $"'($full_cmd)'" } | complete)

    # Create a record with all completion related info. 
    # directive and directive_str are for posterity
    let stdout_lines = ($result.stdout | lines)
    let directive = ($stdout_lines | last | str trim | str replace ":" "" | into int)
    let completions = ($stdout_lines | drop | parse -r '([\w\-\.:\+\=]*)\t?(.*)' | rename value description)

    # Add space at the end of each completion
    let completions = if $directive != $ShellCompDirectiveNoSpace {
      ($completions | each {|it| {value: $"($it.value) ", description: $it.description}})
    } else {
      $completions
    }

    if $last_span =~ '=$' {
      # return flag as part of the completion so that it doesn't get replaced
      $completions | each {|it| $"($last_span)($it.value)" }
    } else if $directive == $ShellCompDirectiveNoFileComp {
      # Allow empty results as this will stop file completion
      $completions
    } else if ($completions | is-empty)  or  $directive == $ShellCompDirectiveError {
      # Not returning null causes file completions to break
      # Return null if there are no completions or ShellCompDirectiveError 
      null
    } else {
      $completions
    }
  } else {
    null
  }
}
```

I don't think it will be too onerous. The workflow will be:

**If you haven't setup cobra_completer**
1. edit config.nu
2. add generated contents
3. set external_completer = cobra_completer
4. start new shell

**If you have setup cobra_completer**
1. Update the cobra_apps array to contain your app, for example:
```
let cobra_apps = ["kubectl", "minikube", "helm"]
```
2. restart shell

nushell_completions.go
marckhouzam: > My fix for -n= is a little bit hacky. I returned the flag as part of the completion to keep it from replacing the flag. It looks a little odd but it worked

This is fine. We do exactly the same for the other shells 👍

marckhouzam:
> I think I am going to have to put a whitelist in for commands to execute the completer 

For other shells we don't use a "global" completer but instead have to register each program individually.  So if kubectl generates the completion script, it also register `kubectl` (on only `kubectl`) to use the functions in that script.

If nushell completions don't work like that then your allow-list idea sounds like the right approach.  I wonder if we could automate adding a program name to the list in the script itself?

marckhouzam:
@rsteube if you have any guidance for us don't hesitate to chime in and correct us.

ayax79:
We could certainly automate adding a line like:
`let cobra_apps = ($cobra_apps | append "newapp")`
to below the initial cobra_apps definition in the config.nu

marckhouzam (changes requested):
Things are moving very nicely @ayax79.  You probably didn't know what you were getting into when you first posted the PR 😄 but with a little more effort, we should get this in.

1. I now get a space after a single completion and the NoFileComp directive is respected.  Nice work!

2. Is it normal the completion does not work if I use a path along with the program?
For example: `kubectl [tab]` works but not `~/bin/kubectl [tab]`

3. Aliases work, nice! But...
```
alias k = kubectl
k [tab]
```
works, but most probably related to the previous point
```
alias k = ~/bin/kubectl
k [tab]
```
doesn't work

4. The = form of flags is starting to work but not completely.
For instance `kubectl -n=[tab]` does work, but not `kubectl -n=de[tab]`

5. I'm getting a weird behaviour where no records are found when I think they should:
```
$ kubectl get ns
NAME              STATUS   AGE
default           Active   23d
kube-system       Active   23d
kube-public       Active   23d
kube-node-lease   Active   23d
$ kubectl get namespaces k[tab]
NO RECORDS FOUND
$ kubectl get namespaces ku[tab]
kube-system
kube-public
kube-node-lease
```

6. New (advanced) requirement that I forgot about: the __complete command may
return completions that don't match the last argument the user typed.
For example:
```
$ helm __complete repo remove hello
test1       https://charts.bitnami.com/bitnami
test2   https://charts.bitnami.com/bitnami
test3   https://charts.bitnami.com/bitnami
bitnami https://charts.bitnami.com/bitnami
:4
```
Notice that although the last argument is "hello", the list of completions
does not filter on "hello".  This case should be supported by Cobra because
it allows for shells that support it to do fuzzy matching.  For example
with `zsh` and `fish` if I do `helm repo remove nami[tab]` the shell will
match `nami` with `bitnami`.
This however is causing problems with the current nushell script. So:
```
$ helm repo remove bitn[tab]
test1
test2
test3
bitnami
```
Notice that nushell completion still suggests all completions and does not
complete the `bitn` argument to `bitnami`.
Therefore it seems the nushell completion script will need to do the filtering by
itself to get rid of all completions that don't match what the user has typed.
The cool thing is that I tested and it seems nushell will support fuzzy matching,
which will be richer than just matching on prefix.  For example, nushell will allow:
```
$ helm repo remove nami[tab]
$ helm repo remove bitnami
```
The filtering rules in the script should be as follows:
1- try to match on prefix.  If any completions start with what the user has typed,
   then only those completions should be kept.
2- if no completion choices match on prefix, then we should look for any completions
   that contain what the user typed
For example if the list of completions obtained is:
```
test
mytest
bitnami
origami
```
if the last argument typed is `te[tab]` then we should only keep the completion
`test` because it starts with `te` (and remove the other 3 completions).
but if the user typed `ami[tab]`, since nothing starts with `ami` we match both
`bitnami` and `origami` which both contain `ami`.

completions.go
```diff
@@ -838,7 +881,7 @@ func CompDebug(msg string, printToStdErr bool) {
 	// variable BASH_COMP_DEBUG_FILE to the path of some file to be used.
 	if path := os.Getenv("BASH_COMP_DEBUG_FILE"); path != "" {
 		f, err := os.OpenFile(path,
-			os.O_APPEND|os.O_CREATE|os.O_WRONLY, 0644)
+			os.O_APPEND|os.O_CREATE|os.O_WRONLY, 0o644)
```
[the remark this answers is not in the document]
ayax79: odd.. going to blame that on neovim ;)

shell_completions.md
```diff
@@ -82,6 +111,8 @@ PowerShell:
 			cmd.Root().GenFishCompletion(os.Stdout, true)
 		case "powershell":
 			cmd.Root().GenPowerShellCompletionWithDesc(os.Stdout)
+		case "nushell":
+			cmd.Root().GenNushellCompletion(os.Stdout, true)
```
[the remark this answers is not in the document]
ayax79: Since we have the cobra_apps array now, there is one way could add support on a per app basis by doing something like this:

```
let cobra_apps = [
  {app: 'kubectl', include_help: true},
  {app: 'helm', include_help: false},
]
```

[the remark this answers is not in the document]
ayax79: @marckhouzam, thoughts?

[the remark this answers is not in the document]
marckhouzam: Will need to give it some thought. Will get back to the review over the weekend.  Thanks @ayax79

marckhouzam:
@ayax79 Sorry for the delay.  I'm having trouble finding enough time to do a thorough review, so let's start with the last remaining points I noticed.

1- First, things are looking really great!  I'm really impressed.  Excellent job!

2- I'm still getting the below weird behaviour.  I haven't had time to look into it myself.  Notice the "NO RECORDS FOUND` when I just type `k` as a prefix for the namespaces.

>I'm getting a weird behaviour where no records are found when I think they should:
```
$ kubectl get ns
NAME              STATUS   AGE
default           Active   23d
kube-system       Active   23d
kube-public       Active   23d
kube-node-lease   Active   23d
$ kubectl get namespaces k[tab]
NO RECORDS FOUND
$ kubectl get namespaces ku[tab]
kube-system
kube-public
kube-node-lease
```
3- The fuzzy matching works well!  Super nice!  However, it is a little too fuzzy 😄 .  See below:

> 6. New (advanced) requirement that I forgot about: the __complete command may
>    return completions that don't match the last argument the user typed.
> [...]
> The filtering rules in the script should be as follows: 
> 1- try to match on prefix. If any completions start with what the user has typed, then only those completions should be kept. 
> 2- if no completion choices match on prefix, then we should look for any completions that contain what the user typed

When I test, it seems the completions  don't first match on prefix, but immediately match on "contains".  Is it possible to first match on prefix and only if there are no matches to use "contains"?  It would make the following case better.

Say I have two namespaces "test" and "mytest".  If the user does `helm -n te<tab>` the completion should complete immediately to `test` since it is the only one left after matching on the prefix `te`.  Currently, both completions continue to match, so the user much select the `test` menu or type out `test` explicitly. 

4- I was testing the `ShellCompDirectiveNoSpace` and at least when the completion ends with a `/` things don't behave properly.  For example, the completion experience is bad when I have this output:
```
helm __complete install test ''                                                        12/11/2022 09:29:37 PM
bitnami/	https://charts.bitnami.com/bitnami
origami/	https://charts.bitnami.com/bitnami
artifact-hub/	https://artifacthub.github.io/helm-charts/
mytest/	https://charts.bitnami.com/bitnami
test/	https://charts.bitnami.com/bitnami
oci://	Chart OCI prefix
https://	Chart URL prefix
http://	Chart URL prefix
file://	Chart local URL prefix
./	Relative path prefix to local chart
/	Absolute path prefix to local chart
:2
Completion ended with directive: ShellCompDirectiveNoSpace
```
I can provide more info about this when I have a bit more time.

I'll stop here but will try to find time soon.

We're almost there!

nushell_completions.go
```diff
@@ -0,0 +1,115 @@
...
+
+    # If the last span entry was empty add "" to the end of the command args
+    let cmd_args = if ($last_span | is-empty) {
+      $'($cmd_args) ""'
+    } else {
+      $cmd_args
+    }
+
+    # The full command to be executed with active help disable (Nushell does not support active help)
+    let full_cmd = $'%[7]s=0 ($cmd) __complete ($cmd_args)'
```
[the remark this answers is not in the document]
ayax79: Added fixes for the fuzzy searching and active help env variable.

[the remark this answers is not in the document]
ayax79: Fixed the issue with slashes!

completions.go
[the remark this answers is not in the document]
marckhouzam: So this change is back.  But after looking into it, I suggest we keep it.  Both notation mean an octal number, but the new `0o` notation seems to be adding clarity.  It was introduced in Go 1.13, so we are good to include it in Cobra.

This issue description was a short read explaining things: https://github.com/golang/go/issues/33363#issue-474439626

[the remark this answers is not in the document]
ayax79: Interesting. I wonder if it is being added by the code formatter.

marckhouzam (commented):
Super nice!  Everything seems fixed 🎉 
I'm currently playing around with setting up the completions, trying to see if we could use `source`.

There are some cosmetics comments in-line.
And one more feature I forgot about: the other shells support doing completions from the middle of the line.  For example:
```
kubectl --namespace <TAB> get pod
```
So basically, I typed something on the command-line and then move my cursor back and then press tab.  I wonder if the nushell script could handle that?  What the other shells do is ignore whatever is after the cursor, and perform completion as normal.

ayax79:
Thanks for the updates!

> Super nice! Everything seems fixed 🎉 I'm currently playing around with setting up the completions, trying to see if we could use `source`.

It is possible to source the entire config.nu in instead of reloading your shell. The reason I didn't document it that way is that you have to pass in a static string into source. The location of this string will vary based on operating system, etc. See https://www.nushell.sh/book/thinking_in_nu.html#think-of-nushell-as-a-compiled-language

The instructions would be:
1. Run ```$nu.config-path```
2. Copy the outputted value manually
3. Run ```source <pasted value>```

Super annoying, but this how I reload my config. 


> ```
> kubectl --namespace <TAB> get pod
> ```
> 
> So basically, I typed something on the command-line and then move my cursor back and then press tab. I wonder if the nushell script could handle that? What the other shells do is ignore whatever is after the cursor, and perform completion as normal.

I'll take a look at the last issue with tabs when I have a few minutes. I'll have see what spans are based in at that point.

ayax79:
Completely off topic... I wrote nushell plugin for loading mac plists, if you are curious how nushell plugins work: https://github.com/ayax79/nu_plugin_plist

completions_test.go
```diff
@@ -2466,6 +2466,7 @@ func TestCompleteCompletion(t *testing.T) {
 	expected := strings.Join([]string{
 		"bash",
 		"fish",
+    "nushell",
```
marckhouzam, line 2469: The indentation is wrong here, which makes the linter fail

marckhouzam:
@ayax79 I've been thinking about how users will setup nushell completions.  The shared script between programs is an efficient approach and I'll be honest that I've been tempted to do the same for the other shells.  The reason I haven't is because it may cause problems in the future.

If we choose to evolve the script later on, how will we handle programs that use the new version of the script and programs that have not moved to it?  For example, say that in Cobra 1.8 we decide to add a new `__completeNewFeature` command to cobra programs to support some new feature, and that each shell script is modified to use it.  Then a program that uses cobra 1.8 will need the new script, but a program that is still on older cobra will need the old script.  With a common script in nushell, this will cause problems.

I was thinking we could go back to using something that you started with (sorry about the wrong syntax):
```
let external_completer = {|spans|
  {
    $spans.0: { }
    kubectl: { $kubectl_completer }
    helm: { $helm_completer }
    minikube: { $minikube_completer }
  } | get $spans.0 | each {|it| do $it}
}
```
and have each program generate their own completion variable (instead of `$cobra_completer` we'd have `$helm_completer`, `$kubectl_completer`, etc).

We then would also no longer need an allow-list of cobra-programs.

And instead of inserting the entire many script into `config nu`, we would tell the user to do
```
helm completion nushell | save ~/.completions/helm
```
and then in `config nu` the user would add:
```
source ~/.completions/helm
```
The user would do this for each program.

Does that sound reasonable?

ayax79:
Sure, sounds good. I'll get the changes in when I have some time.

completions_test.go
ayax79: Ugh... neovim shows the spacing correct, but yes different whitespace. fixed.

ayax79:
> There are some cosmetics comments in-line. And one more feature I forgot about: the other shells support doing completions from the middle of the line. For example:
> 
> ```
> kubectl --namespace <TAB> get pod
> ```

I think there must be a bug in nushell itself when trying to handle completions in the middle of the command. The completer is getting some really crazy span information:

When I hit <TAB> right in front of pods, I get span of apod.
Screenshot 2022-12-27 at 12 28 35 PM


When I type in ```--name``` and hit <TAB>, I receive a span of --namea
Screenshot 2022-12-27 at 12 27 58 PM

When I hit <TAB> in whitespace between kubectl and pods I get a span of "a" and it tries to complete it.
Screenshot 2022-12-27 at 12 29 02 PM

@rsteube, have you seen anything like this before?

marckhouzam:
> I think there must be a bug in nushell itself when trying to handle completions in the middle of the command

Ok. Until that is fixed it just won't work for Cobra. It's not a blocker to get this merged.

nushell_completions.go
```diff
@@ -0,0 +1,175 @@
...
+    # There is no way to currently support this in a nushell external completer
+    let completions = if $directive == $ShellCompDirectiveFilterFileExt {
+      []
+    } else {
+      $completions
+    }
+
+    let return_val = if $last_span =~ '=' {
+      # if the completion is of the form -n= return flag as part of the completion so that it doesn't get replaced
+      $completions | each {|it| $"($last_span | split row '=' | first)=($it.value)" }
```
marckhouzam, line 143: It is valid for a program to request file completion for a flag value.  For example:
```
kubectl apply -f <tab>
```
This currently works.
However, if we use the `=` for it does not:
```
kubectl apply -f=<tab>
```
I'm not sure if you will be to handle that case or not.  If not, we'll just have to live with it.  It's not critical.

nushell_completions.go
```diff
@@ -0,0 +1,175 @@
...
+    # Add space at the end of each completion
+    let completions = if $directive != $ShellCompDirectiveNoSpace {
+      $completions | each {|it| {value: $"($it.value) ", description: $it.description}}
+    } else {
+      $completions
+    }
+
+    # Cobra returns a list of completions that are supported with this directive
+    # There is no way to currently support this in a nushell external completer
+    let completions = if $directive == $ShellCompDirectiveFilterFileExt {
```
marckhouzam, line 135: Might as well check for `ShellCompDirectiveFilterDirs` here also, since it is also not supported.

rsteube:
> When I hit in whitespace between kubectl and pods I get a span of "a" and it tries to complete it. Screenshot 2022-12-27 at 12 29 02 PM
> 
> @rsteube, have you seen anything like this before?

Yeah, got the same behaviour. Seems to be caused by nushell. So best to open an issue there.

```sh
gh <TAB> issue
# 2022/12/30 10:45:45 nushell ["carapace","_carapace","nushell","gh","a"]
```

If i am not mistaken this is related to space suffix determination. I think it was nushell that was adding a character (`a`) suffix before doing word splitting to avoid the empty word to be skipped. It is removed later but this seems to not work when invoked between words (thus the single `a` parameter). Then there's even a subsequent `aa` for me due to it's auto-invocation of completion.

marckhouzam:
@ayax79 Just checking about this PR.  I believe the ball was in your court, or am I mistaken?

ayax79:
Hi @marckhouzam,

I know.. I have been pretty slammed with starting a new job. I am hoping to get back to it soon.

marckhouzam:
> Hi @marckhouzam,
> 
> I know.. I have been pretty slammed with starting a new job. I am hoping to get back to it soon.

No worries, I am in the same situation.  I was just asking to be sure.

Bios-Marcel:
Hey, may I ask for a summary of what is actually still to do here? Is it just what is mentioned here? https://github.com/spf13/cobra/pull/1857#issuecomment-1364744495

Maybe I could chime in and help. It is sad to see this sitting around here for so long :o

marckhouzam:
> Hey, may I ask for a summary of what is actually still to do here? Is it just what is mentioned here? [#1857 (comment)](https://github.com/spf13/cobra/pull/1857#issuecomment-1364744495)
> 
> Maybe I could chime in and help. It is sad to see this sitting around here for so long :o

Yes I believe that’s the main part that remains.  
I was even tempted to do it on my own, but I am too uncomfortable with nushell syntax. Anyone that wants to pick this up is more than welcomed

Bios-Marcel:
Well, I don't know nushell very well either, but it is a nice cross platform shell and I see more and more people using it. So yeah, I'd have to read the docs first as well :D

Thanks for the quick reply

ayax79:
I can try to pick it up again. It's been on the back of my mind lately.

ayax79:
I'm also on the nushell core team now.

ayax79:
@marckhouzam, I have been thinking more about the direction this PR has currently taken of generating a script. A lot has matured around external completers and nushell since I first began this. Other external completers such as nushell, fish, etc handle most of the logic within the command themselves (from [external completers nushell cookbook chapter](https://www.nushell.sh/cookbook/external_completers.html)):

```nushell
let carapace_completer = {|spans|
    carapace $spans.0 nushell ...$spans | from json
}

let fish_completer = {|spans|
    fish --command $'complete "--do-complete=($spans | str join " ")"'
    | $"value(char tab)description(char newline)" + $in
    | from tsv --flexible --no-infer
}
```
What if instead of generating a rather long nushell script, if `<cobra command> completions nushell` actually takes in nushell span arguments, handle most of the bulk of the logic internally, and returns something easily parseable by a much smaller completer hook?

```nushell
let minikube_completer = {|spans| 
   minikube completions nushell $spans | from json
}
```

thoughts?

marckhouzam:
@ayax79 

> What if instead of generating a rather long nushell script, if <cobra command> completions nushell actually takes in nushell span arguments, handle most of the bulk of the logic internally, and returns something easily parseable by a much smaller completer hook?

This is actually what cobra does right now.
Every cobra program has a `__complete` command which provides all the completions.
All the script needs to do is call that command.
I'll refer you to this [comment](https://github.com/spf13/cobra/pull/1857#issuecomment-1316267603).  I believe nushell may have evolved enough to handle this exactly like the other shells do.

If you look at the generated `bash`, `zsh`, `fish` or `powershell` scripts, they all call the `__complete` command.
The scripts may look bigger than you would expect (200 to 300 lines) but it is because they also interpret the ShellCompletionDirective that the `__complete` command specifies.  This allows the script to tell the shell to:
- not perform file completion
- not add a space
- keep the order of the completions
- etc

If memory serves, we had moved towards such a solution in this PR and were very close to having everything working.

Le me know if you need clarifications.

ayax79:
@marckhouzam, I finally got back around to this. I did a complete rewrote the nushell side. I think I have addressed all of the outstanding issues that can be addressed. 

This version can even be run as the default external completer and will noop for non-cobra commands.

ayax79:
Completions from aliases still aren't working. It has to be an underlying nushell issue. I'll investigate further when I have the chance.

marckhouzam:
@ayax79 I'm not sure what git/Github is doing but I'm seeing changes in this PR that belong to the main branch.
Could you rebase on top of the most recent `main` and force push?

nushell_completions.go
```diff
@@ -0,0 +1,128 @@
...
+import (
+	"bytes"
+	"fmt"
+	"io"
+	"os"
+)
+
+func (c *Command) GenNushellCompletion(w io.Writer, includeDesc bool) error {
+	buf := new(bytes.Buffer)
+	WriteStringAndCheck(buf, fmt.Sprintf(`
```
marckhouzam, line 26: Let's start the file with an explanatory comment similar to the other shells, so:
```
WriteStringAndCheck(buf, fmt.Sprintf(`# nushell completion -*- shell-script -*-
```

nushell_completions.go
```diff
@@ -0,0 +1,128 @@
...
+# An external completer that works with any cobra based
```
marckhouzam, line 27: Please replace all tabs with four spaces in the shell script.

nushell_completions.go
```diff
@@ -0,0 +1,128 @@
...
+# command line application (e.g. kubectl, minikube)
```
marckhouzam, line 28: Left over comment?

nushell_completions.go
```diff
@@ -0,0 +1,128 @@
...
+let cobra_completer = {|spans|
+	let ShellCompDirectiveError = %[1]d
+	let ShellCompDirectiveNoSpace = %[2]d
+	let ShellCompDirectiveNoFileComp = %[3]d
+	let ShellCompDirectiveFilterFileExt = %[4]d
+	let ShellCompDirectiveFilterDirs = %[5]d
```
marckhouzam, line 34: There is now a new directive: `ShellCompDirectiveKeepOrder`

nushell_completions_test.go
```diff
@@ -0,0 +1,97 @@
...
+	if os.Getenv("MSYSTEM") == "MINGW64" {
+		if got.Error() != "open ./tmp/test: Access is denied." {
+			t.Errorf("got: %s, want: %s", got.Error(), "open ./tmp/test: Access is denied.")
+		}
+	} else {
+		if got.Error() != "open ./tmp/test: permission denied" {
+			t.Errorf("got: %s, want: %s", got.Error(), "open ./tmp/test: permission denied")
+		}
+	}
+}
```
marckhouzam, line 97: Can you add the following test also:
```
func TestNushellCompletionNoActiveHelp(t *testing.T) {
	c := &Command{Use: "c", Run: emptyRun}

	buf := new(bytes.Buffer)
	assertNoErr(t, c.GenNushellCompletion(buf, true))
	output := buf.String()

	// check that active help is being disabled
	activeHelpVar := activeHelpGlobalEnvVar
	check(t, output, fmt.Sprintf("%s=0", activeHelpVar))
}
```

nushell_completions.go
```diff
@@ -0,0 +1,128 @@
...
+
+	let cmd = $spans | first 
+	let rest = $spans | skip
+
+	def exec_complete [
+		spans: list<string>
+	] {
+		# This will catch the stderr message related to the directive and any other errors,
+		# such as the command not being a cobra based command
+		let result = do --ignore-errors { cobra_active_help=0 run-external $cmd "__complete" ...$spans | complete }
```
marckhouzam, line 44: `cobra_active_help` -> `COBRA_ACTIVE_HELP`.
It doesn't work if it is not in uppercase.

marckhouzam (commented):
I've started testing and things are looking good!
I'm going to move on to more detailed testing but until then, here are some minor comments that need your attention @ajax79

nushell_completions.go
```diff
@@ -0,0 +1,175 @@
...
+      } else {
+        $params
+      }
+
+      let last_span = $params.last_span
+      let spans = $params.spans
+
+      # Drop the last param so we can fuzzy search on it
+      let spans = if $fuzzy {
+        $spans | drop
```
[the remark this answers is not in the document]
marckhouzam: this is a really old comment that was left over as pending. Please ignore

nushell_completions.go
ayax79: Will do.. I had them as tabs before as that was the LSP default and the go style guide says tabs over spaces.. I like spaces better myself.

nushell_completions.go
ayax79: removed

nushell_completions.go
ayax79: done.

nushell_completions.go
ayax79: Added to the list for posterity. I do think there is anything else needed for the implementation as I already keep the order and there is no way to report it back up to nushell.

nushell_completions.go
ayax79: hmm.. I am not sure how that go lowercases 🤷🏻

site/content/completions/_index.md
```diff
@@ -6,6 +6,7 @@ The currently supported shells are:
 - Zsh
 - fish
 - PowerShell
+- Nushell
```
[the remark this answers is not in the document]
ayax79: Is it alright if I also link to the nushell documentation?

nushell_completions_test.go
ayax79: done

site/content/completions/_index.md
```diff
@@ -68,9 +69,29 @@ PowerShell:
...
+Nushell:
+  
+  # To configure the Nushell cobra external completer for the first time:
+  # 1. Copy the output of the command below:
+  > %[1]s completion nushell 
+  # 2. Edit the nushell config file:
+  > config nu
+  # 3. Paste above the "let-env config" line.
+  # 4. Change the config block's external_completer line to be 
+  external_completer: $cobra_completer
```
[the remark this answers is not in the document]
ayax79: rewrote instructions

site/content/completions/_index.md
```diff
@@ -68,9 +69,29 @@ PowerShell:
...
+  # 5. You will need to start a new shell or for this setup to take effect.
+
+  # If you have already setup the cobra external completer:
+  # 1. Edit the nushell config file:
+  > config nu
+  # 2. Modify the cobra_apps varible to contain this application:
+  > let cobra_apps = [ "othercobraapp", "%[1]s" ]
```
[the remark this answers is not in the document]
ayax79: removed

site/content/completions/_index.md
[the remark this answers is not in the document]
ayax79: The instructions are much easier now and I have added links to relevant spots in the nushell documentation.

nushell_completions_test.go
```diff
@@ -0,0 +1,97 @@
...
+
+	buf := new(bytes.Buffer)
+	assertNoErr(t, rootCmd.GenNushellCompletion(buf, true))
+	output := buf.String()
+
+	check(t, output, fmt.Sprintf("let cobra_apps = [\"%[1]s\"]", rootCmd.Name()))
+}
+
+func TestGenNushellCompletionFile(t *testing.T) {
+	err := os.Mkdir("./tmp", 0o755)
```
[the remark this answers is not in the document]
ayax79: fixed!

nushell_completions_test.go
```diff
@@ -0,0 +1,97 @@
...
+		ValidArgsFunction: validArgsFunc,
+		Run:               emptyRun,
+	}
+	rootCmd.AddCommand(child)
+
+	assertNoErr(t, rootCmd.GenNushellCompletionFile("./tmp/test", true))
+}
+
+func TestFailGenNushellCompletionFile(t *testing.T) {
+	err := os.Mkdir("./tmp", 0o755)
```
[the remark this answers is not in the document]
ayax79: fixed!

nushell_completions.go
ayax79: fixed!

ayax79:
@marckhouzam I believe I have addressed all the feedback. Can you give this another look?

marckhouzam:
> I just found a nasty bug.
> 
> When running things vim, it locks up because vim interprets ` vim __complete` as a valid command and never returns.

@ayax79 I think this bug is still present.  Maybe we have to tell the user to use `multiple_completers`?

marckhouzam:
If we have to use `multiple_completers` can we try again to do it like suggested in this comment: https://github.com/spf13/cobra/pull/1857#issuecomment-1364744495

nushell_completions.go
```diff
@@ -0,0 +1,128 @@
...
+package cobra
+
+import (
+	"bytes"
+	"fmt"
+	"io"
+	"os"
+)
+
+func (c *Command) GenNushellCompletion(w io.Writer, includeDesc bool) error {
```
marckhouzam, line 24: The `includeDesc` variable is not used.

What you need to do is simple, if `includeDesc == false` you want to use `__completeNoDesc` in the script, instead of `__complete`.
If we have a single `cobra_completer` this won't really work as it will affect every program.
Let's see where we end up with having different completers and revisit at that time.

nushell_completions.go
```diff
@@ -0,0 +1,128 @@
...
+        # Cobra returns a list of completions that are supported with this directive
+        # There is no way to currently support this in a nushell external completer
+        let completions = if $directive == $ShellCompDirectiveFilterFileExt {
+            []
+        } else {
+            $completions
+        }
+
+        if $directive == $ShellCompDirectiveNoFileComp {
+            # Allow empty results as this will stop file completion
```
marckhouzam, line 94: This does not work for me.  
I'm running nushell 0.100.0

You can try it:

```
# Notice the directive returned is ShellCompDirectiveNoFileComp
> kubectl __complete help annotate ''
:4

# But the following does shell completion
> kubectl help annotate <tab>
```

If we can have a `debug` function to print to a file like my above comments, then I can try to debug such things.

marckhouzam (commented):
We're getting so close I can taste it!

1- Is there a way to add debug statements to the nushell completion `cobra_completer`.  I want to make sure I can debug this in case there is a bug reported.

I was thinking we could add a function like this, maybe?  But I'm having trouble calling it in the completer.
```
    def debug [message] {
        let file = $env.BASH_COMP_DEBUG_FILE?
        if (not ($file == "")) {
            echo $"($message)\n" | save $file --append
        }
    }
```

2- I'm hitting different bugs but as long as they are minor, I was thinking we can commit this in "beta level" and keep working on fixing it.  However we need to fix the `vim <tab>` hanging.  In the mean time, I'll post a small program and give examples of the small issues I'm noticing.

3- Is there no way in nushell to interrupt the shell completion?  I was testing with `tanzu` and in some cases the `tanzu __complete` command will trigger an interactive login and will wait there until I login.  If I don't login, it will hand there forever.  In `zsh` I can press ^C to interrupt the ongoing completion.

fdncred:
I'm not sure. I haven't worked with the completions much. I think the external completer is built in nushell but uses some reedline primitives.

ayax79:
@marckhouzam, I added a logging function.

marckhouzam:
>> 3- Is there no way in nushell to interrupt the shell completion? I was testing with tanzu and in some cases the tanzu __complete command will trigger an interactive login and will wait there until I login. If I don't login, it will hand there forever. In zsh I can press ^C to interrupt the ongoing completion.

> I don't think so currently. I can investigate if it is possible to handle this in nushell core. The currently external completer stuff is pretty limited. You can define a block that is handled a list of spans and you can either return a list of records in the format of {value: "foo", description: "description of foo"} or null to no op.

Should we put a timeout when running the `__complete` command in the `cobra_completer`?
Maybe a timeout of 10 seconds?

nushell_completions.go
ayax79: fixed

nushell_completions.go
ayax79: We could allow this to be set via an environment variable? Maybe something like this:

```nushell
$env.COBRA_NO_DESC_COMMANDS = [kubectl minikube]
```

ayax79: @marckhouzam, thoughts?

marckhouzam: Using an environnement variable means it is the end user that will decide if there will be completion descriptions or not. This is not a bad thing, but it doesn’t address the fact that the current proposed API (the function above) does not respect its `includeDesc` argument. 

If we have a single completer for all cobra programs instead of one per program, I don’t think we can do any better than that. But in that case, maybe we should remove the extra useless argument from this function?

marckhouzam:
CI is failing because it was broken on main. If you just rebase the PR to the latest main it will pass again.

nushell_completions.go
ayax79: Removed!

marckhouzam:
@ayax79 I want to get this in for the next release but finding time is a challenge.
One thing that I keep thinking about is the user experience to set up completion for multiple different tools.  Many many projects use Cobra, and it will be highly likely that a user will want to setup multiple cobra programs to perform completion.

Do you think what I proposed in this comment would be possible: https://github.com/spf13/cobra/pull/1857#issuecomment-1364744495

ayax79:
> @ayax79 I want to get this in for the next release but finding time is a challenge. One thing that I keep thinking about is the user experience to set up completion for multiple different tools. Many many projects use Cobra, and it will be highly likely that a user will want to setup multiple cobra programs to perform completion.
> 
> Do you think what I proposed in this comment would be possible: [#1857 (comment)](https://github.com/spf13/cobra/pull/1857#issuecomment-1364744495)

That is very outdated. The completer will work with all commands that support cobra, and you only need to add it once. 

This is my current config (borrowed from the nushell cookbook):
```nushell
  let external_completer = {|spans|
      let expanded_alias = scope aliases
      | where name == $spans.0
      | get -i 0.expansion

      let spans = if $expanded_alias != null {
          $spans
          | skip 1
          | prepend ($expanded_alias | split row ' ' | take 1)
      } else {
          $spans
      }

      match $spans.0 {
          databricks => $cobra_completer
          npectl => $cobra_completer
          # carapace completions are incorrect for nu
          nu => $fish_completer
          # fish completes commits and branch names in a nicer way
          # git => $fish_completer
          # carapace doesn't have completions for asdf
          asdf => $fish_completer
          # use zoxide completions for zoxide commands
          __zoxide_z | __zoxide_zi => $zoxide_completer
          _ => $carapace_completer
      } | do $in $spans
  }
```

It would also be completely acceptable to only use the cobra completer, it would then check to see if the command supports __complete.

External completer usage is in the documentation:
https://www.nushell.sh/book/custom_completions.html#external-completions

I will also make sure that there is a section dedicated to cobra in the cookbook once this is merged:
https://www.nushell.sh/cookbook/external_completers.html#zoxide-completer

marckhouzam:
> It would also be completely acceptable to only use the cobra completer, it would then check to see if the command supports __complete.

Aha I hadn’t seen that update yet. That sounds ok. It won’t be possible to configure the completer depending on the program; for example, turning off description only for one cobra program and not another. But I think that’s OK.

I’ll review the latest version that you posted. Thanks for the answers.

ayax79:
@marckhouzam, just checking in. Have you had a chance to look at this?

marckhouzam:
It's my next one.  I'll try in the next couple of days in the hopes of having it in the next release.

marckhouzam:
@ayax79 I really want to merge this. Although I’m still seeing some things to fix this new solution doesn’t have to be perfect to get something in the hands of users.

However, I’m still seeing that file completion breaks for all other commands, which I don’t think is acceptable. 

I’m hoping I’m doing something wrong. @ayax79 can you try setting up the cobra completer and then trying `cat ~/<TAB>`. Does it work for you?

AtifChy:
It would be nice to have this get merged.

ofek:
Hello everyone! I was [directed here](https://github.com/cue-lang/cue/issues/4265#issuecomment-3909806533) after submitting the same feature request to CUE, a downstream project.

It's worth noting that Nushell released a redesigned completion mechanism for external commands in version 0.108.0 which was released in October. Although it's not yet documented due to an [oversight](https://github.com/nushell/nushell/pull/16765#issuecomment-3909640883), the approach is prominently shown in the [release note](https://www.nushell.sh/blog/2025-10-15-nushell_v0_108_0.html#command-wide-completion-handler-16765-toc). Clap has an [example](https://github.com/clap-rs/clap/issues/5840#issuecomment-3423864439) in their ongoing discussion from its corresponding [implementation](https://github.com/clap-rs/clap/pull/5841).

This should significantly trim down what is required for completion support 🙂

rsteube:
@ofek you can bridge these either directly or via fish:

http://www.nushell.sh/cookbook/external_completers.html
https://github.com/carapace-sh/carapace-bridge

ayax79:
There have been a lot of changes to the completion system recently. I am going to try reimplementing with this https://www.nushell.sh/blog/2025-10-15-nushell_v0_108_0.html#command-wide-completion-handler-16765-toc when I have some time.

PR #1863, opened 2022-11-21
Update stale.yml

enrichman:
Fix typo

### DIFF
--- a/.github/workflows/stale.yml
+++ b/.github/workflows/stale.yml
@@ -28,7 +28,7 @@ jobs:
           You can:
 
           - Make a comment to remove the stale label and show your support. The 60 days reset.
-          - If an issue has lifecycle/rotten and is closed, comment and ask maintainers if they'd be interseted in reopening"
+          - If an issue has lifecycle/rotten and is closed, comment and ask maintainers if they'd be interested in reopening"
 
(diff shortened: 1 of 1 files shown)


marckhouzam (approved):
Thanks @enrichman !

marckhouzam:
The go 1.15 test is failing with 
```
../../../go/pkg/mod/github.com/kyoh86/richgo@v0.3.11/config/load.go:64:9: undefined: os.ReadFile
```
Looks like `richgo` 0.3.11, which was released 6 days ago, requires go 1.16.
We probably need to fix the version to 0.3.10 to continue working with go 1.15

@umarcor what do you think?

Since it is unrelated to the PR I will merge anyway.

enrichman:
@marckhouzam you're welcome, thanks for the quick review!

umarcor:
@marckhouzam go 1.15 is EOL. I believe we can remove it from the CI. However, this repo has a very unclear criteria to deprecate golang versions. In theory, https://github.com/spf13/cobra/blob/main/CONDUCT.md#deprecation. In practice, it's arbitrary upon how do maintainers (or employers) feel.

marckhouzam:
Let's move the discussion to #1864

PR #1865, opened 2022-11-23
Makefile: add target richtest

umarcor:
Reduce the barrier to entry by making makefile target `test` use `go test` and adding target `richtest` to be used in CI.


marckhouzam (approved):
Nice idea @umarcor !
LGTM

PR #1866, opened 2022-11-23
ci: deprecate go 1.15

umarcor:


marckhouzam:
So this change allows Cobra to continue working with Go 1.15 (since we don't update the `go.mod` file), but removes the testing for it.

I think this is ok and we need to get CI working again.  So I will merge.

marckhouzam (approved):

andreaangiolillo:
Hi @marckhouzam,
Could you explain the reasons behind not updating the `go.mod` file 🤔? 
Thanks!

marckhouzam:
> Hi @marckhouzam,
> Could you explain the reasons behind not updating the `go.mod` file 🤔?
> Thanks!

Sure. By not updating the go.mod file, if a program using Cobra is still using Go 1.15, it can still upgrade to the latest Cobra version.  It's not a good idea to be on 1.15 but with hundreds of projects using Cobra, it may exist.

The day Cobra doesn't compile with 1.15, then we might as well update the go mod file.

PR #1867, opened 2022-11-24
[RFC] Redirect bash completion v1 to v2 when possible

marckhouzam:
We are no longer actively maintaining bash completion v1 in favor of its more rich v2 version.  Previously, using bash completion v2 required projects to be aware of its existence and to explicitly call `GenBashCompletionV2()`.

With this commit, any projects calling `GenBashCompletion()` will automatically be redirected to using the v2 version.
One exception is if the project uses the legacy custom completion logic which is not supported in v2.  We can detect that by looking for the use of the field `BashCompletionFunction` on the root command.

If we ignore the tests and doc, this is a 2 line change in `bash_completions.go`

Note that descriptions are kept off when calling `GenBashCompletion()`. This means that to enable completion descriptions for bash, a project must still explicitly call `GenBashCompletionV2()`.  I chose this approach because I was hesitant to enable descriptions without getting projects to make the decision themselves. 

cc @jpmcb @scop @Luap99

### DIFF
--- a/bash_completions.go
+++ b/bash_completions.go
@@ -684,11 +684,19 @@ func gen(buf io.StringWriter, cmd *Command) {
 
 // GenBashCompletion generates bash completion file and writes to the passed writer.
 func (c *Command) GenBashCompletion(w io.Writer) error {
+	if len(c.BashCompletionFunction) == 0 {
+		// If the program does not define any legacy custom completion (which is not
+		// supported by bash completion V2), use bash completion V2 which is the version
+		// that is maintained.  However, we keep descriptions off to behave as much as v1
+		// as possible to avoid changing things unexpectedly for projects
+		return c.GenBashCompletionV2(w, false)
+	}
+
 	buf := new(bytes.Buffer)
 	writePreamble(buf, c.Name())
-	if len(c.BashCompletionFunction) > 0 {
-		buf.WriteString(c.BashCompletionFunction + "\n")
-	}
+
+	buf.WriteString(c.BashCompletionFunction + "\n")
+
 	gen(buf, c)
 	writePostscript(buf, c.Name())
 
--- a/bash_completions_test.go
+++ b/bash_completions_test.go
@@ -227,39 +227,72 @@ func TestBashCompletions(t *testing.T) {
 }
 
 func TestBashCompletionHiddenFlag(t *testing.T) {
-	c := &Command{Use: "c", Run: emptyRun}
+	c := &Command{
+		Use: "c",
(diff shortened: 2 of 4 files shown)


bash_completions.go
```diff
@@ -684,6 +684,14 @@ func gen(buf io.StringWriter, cmd *Command) {
...
 // GenBashCompletion generates bash completion file and writes to the passed writer.
 func (c *Command) GenBashCompletion(w io.Writer) error {
+	if len(c.BashCompletionFunction) == 0 {
+		// If the program does not define any legacy custom completion (which is not
+		// supported by bash completion V2), use bash completion V2 which is the version
+		// that is maintained.  However, we keep descriptions off to behave as much as v1
+		// as possible to avoid changing things unexpectedly for projects
+		return c.GenBashCompletionV2(w, false)
+	}
+
```
Luap99, line 694: with this change you can remove the `len(c.BashCompletionFunction) > 0` check on line 697

Luap99 (commented):
I am ok with the change but I wonder what the motivation is? Assuming that you will not remove the old script until a major version bump I don't see the benefit.

marckhouzam:
> I am ok with the change but I wonder what the motivation is? Assuming that you will not remove the old script until a major version bump I don't see the benefit.

I should have explained in the PR description, sorry.
Let me start by saying that I agree this is not a clearcut decision, and I welcome any input for or against it.

Basically I'm trying to improve the quality of bash completion for projects using Cobra.  I don't think it is easy for projects to know that there is such a thing as a V2 version and therefore I feel projects are missing out on bug fixes and features.
I opted to propose this change after opening this bug https://github.com/vmware-tanzu/tanzu-framework/issues/3963; the bug made me realize that bash completion can be affected by subtle bugs that may go unreported/unfixed although they have already been addressed by Cobra's bash completion V2 solution.

But maybe this is overreaching.  I'd like people's honest opinion on this and if there is too much hesitation, we'll just close it.

Luap99:
I think it is a good idea overall. IMO there shouldn't be a reason to use the V1 version but then again I don't really know how and where is is used. One thing that we can do without changing behaviour is to add a `// Deprecated:` comment which will be detected by some linters, i.e. https://staticcheck.io/docs/checks/#SA1019, and make authors aware that they should switch.

scop:
Seems fine to me too. `Deprecated:` on the original functions seconded, more info at https://github.com/golang/go/wiki/Deprecated

marckhouzam:
Let’s give up on this.  I might add a deprecation on the completion v1 instead

Issue #1864, opened 2022-11-22
CI fails for Go 1.15


marckhouzam:
As was seen in #1863, the go 1.15 CI is failing with 
```
../../../go/pkg/mod/github.com/kyoh86/richgo@v0.3.11/config/load.go:64:9: undefined: os.ReadFile
```
Looks like `richgo` 0.3.11, which was recently released (Nov 15, 2022), requires go 1.16.

marckhouzam:
I see two options here:
1. Fix richgo to v0.3.10 until we drop support for Go 1.15
2. Drop support for Go 1.15 now (as suggested by @umarcor in https://github.com/spf13/cobra/pull/1863#issuecomment-1323567791)

I'm ok with either option.

We also need to update our deprecation document as we have agreed we will not do a major Cobra release when removing support for Go versions.

I suggest our support policy for Go should be something like: "We guarantee to support the latest 2 versions of Go, just like the Go project does.  To give more flexibility to Cobra consumers, we also won't remove support for older Go versions until a valid reason arises; valid reasons are at the discretion of the maintainers and include but are not limited to, hindering the evolution of the project, limiting the use of CI tools, causing a noticeable added burden of maintenance."

What do you think @jpmcb @umarcor ?

jpmcb:
Thanks for opening this discussion Marc - I have a few thoughts:

I think we should remove usage of `richgo` entirely since it only provides us with colored / paged / enriched test output which is non-critical to the cobra library. It also creates a weird barrier to entry to contribute to cobra which I would _personally_ like removed.

If that is unacceptable, I think we should visit deprecating Go 1.15 from cobra. It's been EOL for awhile now so I'm not concerned with breaking consumers of cobra but it does deserve some deep consideration since cobra's posture towards old versions of Go has been "it should just work".

marckhouzam:
> I think we should remove usage of `richgo` entirely since it only provides us with colored / paged / enriched test output which is non-critical to the cobra library. It also creates a weird barrier to entry to contribute to cobra which I would _personally_ like removed.

You had me at “barrier to entry “ 😜

Also, I personally prefer keeping go versions as long as possible, even if they are EOL.  We don't know the reality of the many projects using Cobra so let's help as much as we can.

So +1 to remove our use of richgo.  Thanks for thinking outside the box @jpmcb

marckhouzam:
[...]

PR #1870, opened 2022-12-05
go.mod: github.com/inconshreveable/mousetrap v1.1.0

thaJeztah:
- Brings in the changes from https://github.com/inconshreveable/mousetrap/pull/7

- Removes compatibility code for go1.3
- Updates build-tag format for current go versions

full diff: https://github.com/inconshreveable/mousetrap/compare/v1.0.1...v1.1.0

PR #1871, opened 2022-12-05
update minimum go version to go1.17

thaJeztah:
- relates to https://github.com/spf13/cobra/pull/1870#issuecomment-1337033791
- supersedes / closes https://github.com/spf13/cobra/pull/1866
- closes https://github.com/spf13/cobra/issues/1864

Versions of go below go1.18 have reached EOL, but some projects may be slightly behind and still on go1.17.

This patch updates the minimum Go version to go1.17, as older versions are broken in CI:

    Error: ../../../go/pkg/mod/github.com/kyoh86/richgo@v0.3.11/config/load.go:64:9: undefined: os.ReadFile
    note: module requires Go 1.17
    Error: Process completed with exit code 2.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -63,8 +63,6 @@ jobs:
         - ubuntu
         - macOS
         go:
-        - 15
-        - 16
         - 17
         - 18
         - 19
@@ -81,12 +79,8 @@ jobs:
 
     - run: |
         export GOBIN=$HOME/go/bin
-        case "${{ matrix.go }}" in
-          14|15) _version='';;
-          *) _version='@latest';;
-        esac
-        go install github.com/kyoh86/richgo"${_version}"
-        go install github.com/mitchellh/gox"${_version}"
+        go install github.com/kyoh86/richgo@latest
+        go install github.com/mitchellh/gox@latest
 
     - run: RICHGO_FORCE_COLOR=1 PATH=$HOME/go/bin/:$PATH make richtest
 
--- a/go.mod
+++ b/go.mod
@@ -1,6 +1,6 @@
 module github.com/spf13/cobra
 
-go 1.15
+go 1.17
 
 require (
 	github.com/cpuguy83/go-md2man/v2 v2.0.2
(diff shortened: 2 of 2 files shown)


marckhouzam:
Thanks @thaJeztah. We're in discussion right now about dropping support for go 1.15 or not.  CI is failing because of a tool we use for CI only so technically Cobra can continue to support go 1.15.

Although these versions of go are EOL, it may not be cobra's place to force projects to upgrade unless we have a legitimate need to stop using these old versions.

We'll get this sorted out soon.

Sorry about the hassle.

umarcor:
@marckhouzam this seems to be motivated by a direct dependency of cobra (github.com/inconshreveable/mousetrap), unrelated to richgo. So this is not about deprecating 1.15 but <1.17.

marckhouzam:
I prefer not to force projects to upgrade their go version unless Cobra really needs it.  Even if go 1.15 is EOL for a while, Cobra still works with it, the CI failure was due to a CI dependency not a Cobra dependency.

I'll close this, but if you feel it still should be considered please reopen.

umarcor:
@marckhouzam as far as I understand, since #1872 was merged, cobra does require go <1.17.

However, on Windows we do test a single version, which is currenly 1.19 (see https://packages.msys2.org/base/mingw-w64-go). And the code on "others" is almost none: https://github.com/inconshreveable/mousetrap/blob/master/trap_others.go.

Does golang ignore the go version specified in the dependencies (https://github.com/inconshreveable/mousetrap/blob/master/go.mod#L3)? Otherwise, building the current cobra main using go 1.16 should be failing.

On the other hand, maybe mousetrap does not use any feature that forces to use go <1.16 and it is specified `go 1.18` just because the file was added in august: https://github.com/inconshreveable/mousetrap/commits/master/go.mod. If that's the case, cobra might keep ignoring it.

thaJeztah:
Go currently uses the version specified in the `go.mod` as a "recommended minimum version" (it doesn't block other versions of go to be used, but considers the version to be the "known to work with, and older versions may not work".

That said, Go (and other tools) may look at the go version in `go.mod`. For example, when using `go mod tidy` or `go mod vendor`, it may use that version to determine whether to use "module pruning" in the resulting `go.mod`.

PR #1875, opened 2022-12-09
Always propagate root context to child command

pietern:
The context passed to the root command should propagate to its children not only on the first execution but also subsequent calls. Calling the same command multiple times is common when testing cobra applications.


marckhouzam:
Thanks @pietern.
Can you first sign the CLA: https://github.com/spf13/cobra/pull/1875#issuecomment-1344437980

command_test.go
```diff
@@ -232,6 +232,44 @@ func TestExecuteContextC(t *testing.T) {
 	}
 }
 
+// This tests that the context passed to the root command propagates to its children
+// not only on the first execution but also subsequent calls.
+// Calling the same command multiple times is common when testing cobra applications.
+func TestExecuteContextMultipleTimes(t *testing.T) {
+	ctxRun := func(cmd *Command, args []string) {
+		err := cmd.Context().Err()
```
marckhouzam, line 240: Cool trick to cancel the context for this test.

However, I think we should still test that the we are really receiving the correct context, not just that it hasn't been cancelled.  I think it would cover more cases.

Could you check we have the right context, similarly to the test right above this one?

command.go
```diff
@@ -1059,11 +1059,8 @@ func (c *Command) ExecuteC() (cmd *Command, err error) {
 		cmd.commandCalledAs.name = cmd.Name()
 	}
 
-	// We have to pass global context to children command
-	// if context is present on the parent command.
-	if cmd.ctx == nil {
-		cmd.ctx = c.ctx
-	}
+	// Pass context of root command to child command.
+	cmd.ctx = c.ctx
```
marckhouzam, line 1063: I'm worried think this could be a breaking change...
If a program calls `ExecuteContext` on another command than that command would get a different context on the root if we always pass it down.

I'll be honest that I don't know if this is a valid scenario.  But we have to be very careful since Cobra is used by so many programs.  I'll have to test to see if there is a potential problem.

marckhouzam (changes requested):

command_test.go
pietern: Thanks for the review. I updated it to test for equality.

command.go
pietern: I understand what you mean. In fact without other changes executing a subcommand with a context would not have effect and it would use the root's context. The line of code that makes that happen turned out to not be covered by tests: https://github.com/spf13/cobra/blob/923592041e4ae38e2ae5c3cc66c3eb1e0c27261f/command.go#L1012

I added a test to cover this scenario and changed the line such that a command's context is also passed down to the root if execute was called on a subcommand.

pietern:
Not stale,

command.go
liboz1: I had a similar pull request: https://github.com/spf13/cobra/issues/2077
We can combine these two to one I think. Thanks.

libozh:
I had a similar request: https://github.com/spf13/cobra/issues/2077 . I think they are trying to cover similar use cases.

PR #1879, opened 2022-12-15
fix: conflict import name with variable

yashLadha:
`template` is an import in `cobra.go` file and also used as a variable name, which masks the library in the scope of that function.


marckhouzam (approved):
Thanks @yashLadha

Issue #1880, opened 2022-12-16
Front page of the GitHub project: image less readable in dark theme


Deleplace:
GitHub has provided dark theme for users like me who prefer dark, which is great and sometimes comes with challenges for existing images.

Here is what the GitHub project looks like in my browser:

<img width="1043" alt="image" src="https://user-images.githubusercontent.com/13508141/208109121-26009319-15d1-4e1b-9cf0-1ba6c77818a1.png">

There a several ways to fix this, I think the most straightforward would be a white shadow on the "cobra" text. I will make a PR for this.

Deleplace:
The png is actually not inside the repo, so instead of a PR, I'm attaching my suggested modified png to this issue.

![cobra_shadow](https://user-images.githubusercontent.com/13508141/208113069-6bdccb85-1797-4946-9fff-c35443fbb29e.png)

marckhouzam:
Very nice @Deleplace, thank you!

I think this should be approved by @spf13 himself.

spf13:
Great idea @Deleplace . Approved!!

PR #1883, opened 2022-12-22
Update main image to better handle dark background

marckhouzam:
This commit uses the images provided by @Deleplace in #1880 

~I'm not too sure if it is safe to point to that image URL directly as provided in the issue?  Seems to work...~
The image is part of a new `assets` directory

In the commit message I used what I found as name and email, please confirm this is ok with you @Deleplace.

Edit: email address removed.  Sorry for having posted it.


marckhouzam:
You can see the resulting page in my repo: https://github.com/marckhouzam/cobra/tree/fix/image

stmcginnis:
I think it would be better to copy the image over and check it in to the repo. Especially with using someone else's image, there would be no control over whether they decided to delete the source image - either accidentally or on purpose.

Image looks great though! Definitely easier to read when using dark mode.

marckhouzam:
You're right @stmcginnis . I don't know why I didn't think of it. I was following the current pattern.  But it's not a good idea.

I'll update the PR.

marckhouzam:
I created a new `assets` directory and put the image in there.

stmcginnis (approved):
Nice!

marckhouzam:
@jpmcb I need your approval to merge this 🙂

umarcor (approved):

PR #1884, opened 2022-12-23
Update badge route

sh-cho:
https://github.com/badges/shields/issues/8671

and fixed typo (tru -> true)

### DIFF
--- a/README.md
+++ b/README.md
@@ -6,7 +6,7 @@ Cobra is used in many Go projects such as [Kubernetes](https://kubernetes.io/),
 [Hugo](https://gohugo.io), and [GitHub CLI](https://github.com/cli/cli) to
 name a few. [This list](./projects_using_cobra.md) contains a more extensive list of projects using Cobra.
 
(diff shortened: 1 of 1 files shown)


marckhouzam:
Thanks @sh-cho .
I saw this in the bug report you refer to: "If your workflow is in a subdirectory under .github, include the subdir in the route.".

That applies to Cobra, no?

sh-cho:
@marckhouzam 

I think it is the case where the workflow file is in a folder other than `.github/workflows` folder. (e.g. `.github/some_dir/test2.yml`)

marckhouzam:
> I think it is the case where the workflow file is in a folder other than `.github/workflows` folder. (e.g. `.github/some_dir/test2.yml`)

Their message is confusing if that is the case since they say "If your workflow is in a subdirectory under .github...", but looking at your repo with this change, it works, so you must be right.

marckhouzam (approved):
Thanks @sh-cho

PR #1885, opened 2022-12-24
fix: func name in doc strings

yashLadha:
Corrected the function name at the start of doc strings, as per the convention
outlined in official go documentation: https://go.dev/blog/godoc

Though a minor change, but found this while browsing through the code.

### DIFF
--- a/command.go
+++ b/command.go
@@ -35,7 +35,7 @@ const FlagSetByCobraAnnotation = "cobra_annotation_flag_set_by_cobra"
 // FParseErrWhitelist configures Flag parse errors to be ignored
 type FParseErrWhitelist flag.ParseErrorsWhitelist
 
-// Structure to manage groups for commands
+// Group Structure to manage groups for commands
 type Group struct {
 	ID    string
 	Title string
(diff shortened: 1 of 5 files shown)


command.go
```diff
@@ -321,7 +321,7 @@ func (c *Command) SetHelpCommand(cmd *Command) {
 	c.helpCommand = cmd
 }
 
-// SetHelpCommandGroup sets the group id of the help command.
+// SetHelpCommandGroupID SetHelpCommandGroup sets the group id of the help command.
```
marckhouzam, line 324:
```suggestion
// SetHelpCommandGroupID sets the group id of the help command.
```

command.go
```diff
@@ -330,7 +330,7 @@ func (c *Command) SetHelpCommandGroupID(groupID string) {
 	c.helpCommandGroupID = groupID
 }
 
-// SetCompletionCommandGroup sets the group id of the completion command.
+// SetCompletionCommandGroupID SetCompletionCommandGroup sets the group id of the completion command.
```
marckhouzam, line 333:
```suggestion
// SetCompletionCommandGroupID sets the group id of the completion command.
```

marckhouzam (commented):
Thanks @yashLadha .
Couple of suggestions inline

command.go
yashLadha: Updated

command.go
yashLadha: Updated

marckhouzam (approved):

PR #1903, opened 2023-01-30
added keep order to bash, zsh, pwsh & fish

### LINKED ISSUE #1900: preserve ordering of completions passed to zsh autocomplete
https://github.com/spf13/cobra/blob/6d978a911e7fff69b1ca2a873dd91d78ebca44cf/zsh_completions.go#L265

The way describes works with the following input - `apple cat bat`, it would sort it for you, and suggest `apple bat cat`. If we pass the `-V` flag we end up preserving the ordering of the passed completion,

a real world example,

```
CURRENT: 4, words[*]: your favorite command 
Truncated words[*]: your favorite command ,
lastParam: , lastChar: 
Adding extra empty parameter
About to call: eval your __complete favorite command  ""
completion output: dry-thunder
patient-field
twilight-shadow
edf1abc0bfc445b38d46a82585a8bfd0
edf78299bbbd453f9552a3b730d68834
edffa85eb3b14e22951f47e252c5fbbf
:4
last line: :4
directive: 4
completions: dry-thunder
patient-field
twilight-shadow
edf1abc0bfc445b38d46a82585a8bfd0
edf78299bbbd453f9552a3b730d68834
edffa85eb3b14e22951f47e252c5fbbf
flagPrefix: 
Adding completion: dry-thunder
Adding completion: patient-field
Adding completion: twilight-shadow
Adding completion: edf1abc0bfc445b38d46a82585a8bfd0
Adding completion: edf78299bbbd453f9552a3b730d68834
Adding completion: edffa85eb3b14e22951f47e252c5fbbf
Calling _describe
_describe found some completions
```

on pressing tab you get, dry-thunder edf1abc0bfc445b38d46a82585a8bfd0 edf78299bbbd453f9552a3b730d68834 edffa85eb3b14e22951f47e252c5fbbf patient-field twilight-shadow

if we change the above line to 

`if eval _describe -V "completions" completions $flagPrefix $noSpace; then`

we maintain the passed ordering. Would you be open to a contribution that gives users an option to preserve ordering?

h4ck3rk3y:
Closes #1900

### DIFF
--- a/bash_completions.go
+++ b/bash_completions.go
@@ -81,6 +81,7 @@ __%[1]s_handle_go_custom_completion()
     local shellCompDirectiveNoFileComp=%[5]d
     local shellCompDirectiveFilterFileExt=%[6]d
     local shellCompDirectiveFilterDirs=%[7]d
+    local shellCompDirectiveKeepOrder=%[8]d
 
     local out requestComp lastParam lastChar comp directive args
 
@@ -88,7 +89,7 @@ __%[1]s_handle_go_custom_completion()
     # Calling ${words[0]} instead of directly %[1]s allows to handle aliases
     args=("${words[@]:1}")
     # Disable ActiveHelp which is not supported for bash completion v1
-    requestComp="%[8]s=0 ${words[0]} %[2]s ${args[*]}"
+    requestComp="%[9]s=0 ${words[0]} %[2]s ${args[*]}"
 
     lastParam=${words[$((${#words[@]}-1))]}
     lastChar=${lastParam:$((${#lastParam}-1)):1}
@@ -125,6 +126,12 @@ __%[1]s_handle_go_custom_completion()
             if [[ $(type -t compopt) = "builtin" ]]; then
                 __%[1]s_debug "${FUNCNAME[0]}: activating no space"
                 compopt -o nospace
+            fi
+        fi
+        if [ $((directive & shellCompDirectiveKeepOrder)) -ne 0 ]; then
+            if [[ $(type -t compopt) = "builtin" ]]; then
+                __%[1]s_debug "${FUNCNAME[0]}: activating keep order"
+                compopt -o nosort
             fi
         fi
         if [ $((directive & shellCompDirectiveNoFileComp)) -ne 0 ]; then
@@ -398,7 +405,7 @@ __%[1]s_handle_word()
 
 `, name, ShellCompNoDescRequestCmd,
 		ShellCompDirectiveError, ShellCompDirectiveNoSpace, ShellCompDirectiveNoFileComp,
-		ShellCompDirectiveFilterFileExt, ShellCompDirectiveFilterDirs, activeHelpEnvVar(name)))
+		ShellCompDirectiveFilterFileExt, ShellCompDirectiveFilterDirs, ShellCompDirectiveKeepOrder, activeHelpEnvVar(name)))
 }
 
 func writePostscript(buf io.StringWriter, name string) {
@@ -437,7 +444,7 @@ func writePostscript(buf io.StringWriter, name string) {
 	WriteStringAndCheck(buf, fmt.Sprintf(`if [[ $(type -t compopt) = "builtin" ]]; then
     complete -o default -F __start_%s %s
 else
-    complete -o default -o nospace -F __start_%s %s
+    complete -F __start_%s %s
 fi
 
 `, name, name, name, name))
--- a/bash_completions.go
+++ b/bash_completions.go
@@ -81,7 +81,6 @@ __%[1]s_handle_go_custom_completion()
     local shellCompDirectiveNoFileComp=%[5]d
     local shellCompDirectiveFilterFileExt=%[6]d
     local shellCompDirectiveFilterDirs=%[7]d
-    local shellCompDirectiveKeepOrder=%[8]d
 
     local out requestComp lastParam lastChar comp directive args
 
@@ -89,7 +88,7 @@ __%[1]s_handle_go_custom_completion()
     # Calling ${words[0]} instead of directly %[1]s allows to handle aliases
     args=("${words[@]:1}")
     # Disable ActiveHelp which is not supported for bash completion v1
-    requestComp="%[9]s=0 ${words[0]} %[2]s ${args[*]}"
+    requestComp="%[8]s=0 ${words[0]} %[2]s ${args[*]}"
 
     lastParam=${words[$((${#words[@]}-1))]}
     lastChar=${lastParam:$((${#lastParam}-1)):1}
@@ -126,12 +125,6 @@ __%[1]s_handle_go_custom_completion()
             if [[ $(type -t compopt) = "builtin" ]]; then
                 __%[1]s_debug "${FUNCNAME[0]}: activating no space"
                 compopt -o nospace
-            fi
-        fi
-        if [ $((directive & shellCompDirectiveKeepOrder)) -ne 0 ]; then
-            if [[ $(type -t compopt) = "builtin" ]]; then
-                __%[1]s_debug "${FUNCNAME[0]}: activating keep order"
-                compopt -o nosort
             fi
         fi
         if [ $((directive & shellCompDirectiveNoFileComp)) -ne 0 ]; then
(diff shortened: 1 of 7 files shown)


bash_completions.go
```diff
@@ -437,7 +444,7 @@ func writePostscript(buf io.StringWriter, name string) {
 	WriteStringAndCheck(buf, fmt.Sprintf(`if [[ $(type -t compopt) = "builtin" ]]; then
     complete -o default -F __start_%s %s
 else
-    complete -o default -o nospace -F __start_%s %s
+    complete -F __start_%s %s
```
h4ck3rk3y, line 447: `compopts` should have set (-o nospace, +o default, -nosort) for us at

marckhouzam: This entry is for bash 3. The `nospace` is annoying but I believe it's for the `--flag=` completions. Is rather not change this old behavior.

Any reason you are changing it?

h4ck3rk3y: I was of the opinion that if the `-o nospace` is required then https://github.com/spf13/cobra/blob/87cd9437e12b95810770f02ec1e55bcd774c2256/bash_completions.go#L125-L130 would have set it for any completion downstream

h4ck3rk3y: I am happy reverting this change if we want to keep the legacy bash completions as is.

marckhouzam: So the problem with bash 3 is that `compopt` does not exist.  So we can't control those options.  That is why `nospace` was on by default.

So yeah, please revert 😄

h4ck3rk3y: Reverted! Thank you :-)

marckhouzam (changes requested):
This is really great!
I haven't gone through the code in detail yet, but the overall change looks great.

Two things I noticed while testing:
1. `zsh` seems to order all completions with descriptions first, then the ones without them, even with the -V flag.  I don't think it is a big deal, but it does mean it does not totally respect the "keep order" directive.  The other shells don't behave like this.  Is this ok with your usecase?
1. When trying `fish` it looks like the completions are never sorted anymore, even if the new directive is not used.  I haven't looked at the code yet to see where things broke.

h4ck3rk3y:
@marckhouzam `fish` was indeed buggy, my bad. I have fixed it and tried it with the directive set & unset, have another run with it :-) 

The `zsh` behavior works for my use case! We set the `-V` flag for non description dynamic completions.

fish_completions.go
```diff
@@ -207,11 +234,13 @@ complete -c %[2]s -e
 
 # The call to __%[1]s_prepare_completions will setup __%[1]s_comp_results
 # which provides the program's completion choices.
-complete -c %[2]s -n '__%[1]s_prepare_completions' -f -a '$__%[1]s_comp_results'
-
+# if this doesn't require order preservation, we dont use the -k flag
+complete -c %[2]s -n '__%[1]s_doesnt_requires_order_preservation' -n '__%[1]s_prepare_completions' -f -a '$__%[1]s_comp_results'
```
marckhouzam, line 238: This wasn't working for me.  After troubleshooting I found out that my version of fish didn't support using multiple conditions (`-n` flag).  Apparently this was added/fixed in fish 3.5: https://github.com/fish-shell/fish-shell/pull/8967

So, this change wouldn't work for older fish version.  But good news, we can replace the multiple `-n` logic with a single `-n` using a logical `&&`.  So we can instead do:
```
complete -c %[2]s -n '__%[1]s_doesnt_requires_order_preservation && __%[1]s_prepare_completions' -f -a '$__%[1]s_comp_results'
```

fish_completions.go
```diff
@@ -89,6 +89,33 @@ function __%[1]s_perform_completion
     printf "%%s\n" "$directiveLine"
 end
 
+function __%[1]s_doesnt_requires_order_preservation
+    __%[1]s_debug ""
+    __%[1]s_debug "========= checking if order preservation is required =========="
+
+    set -l results (__%[1]s_perform_completion)
```
marckhouzam, line 96: With the new keeporder logic `__%[1]s_perform_completion` ends up being called 3 times instead of one.  That function calls the program to get the completions, so there is some expense to that call.

I haven't thought it through yet, but it would be nice if we could achieve all the logic by using only a single call to the program.

bash_completionsV2.go
```diff
@@ -115,6 +116,14 @@ __%[1]s_process_completion_results() {
                 __%[1]s_debug "No space directive not supported in this version of bash"
             fi
         fi
+        if (((directive & shellCompDirectiveKeepOrder) != 0)); then
+            if [[ $(type -t compopt) == builtin ]]; then
+                __%[1]s_debug "Activating keep order"
+                compopt -o nosort
```
marckhouzam, line 122: the `nosort` was added with bash 4.4 according to the answer to https://unix.stackexchange.com/questions/215937/disable-sorting-of-compreply-in-bash-complete-function

When I use bash 4.2 the script throws an error:
```
bash4-4.2$ ./prog orderfromshell [tab][tab]
bash4-4.2$ ./prog orderfromprogram bash4: compopt: nosort: invalid option name
```
Can we protect against that?

marckhouzam (changes requested):
Almost ready.  thanks for the efforts @h4ck3rk3y !
Now we get into some of the annoying "supporting older shell" details.

powershell_completions.go
```diff
@@ -182,6 +183,11 @@ filter __%[1]s_escapeStringWithSpecialChars {
         }
     }
 
+    # we sort the values in ascending order by name if keep order isn't passed
+    if (($Directive -band $ShellCompDirectiveKeepOrder) -eq 0 ) {
+        $Values = $Values | Sort-Object -Property Name
+    }
```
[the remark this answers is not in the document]
Luap99: sorting alphabetically by default sounds good to me

bash_completionsV2.go
```diff
@@ -354,13 +363,13 @@ __start_%[1]s()
 if [[ $(type -t compopt) = "builtin" ]]; then
     complete -o default -F __start_%[1]s %[1]s
 else
-    complete -o default -o nospace -F __start_%[1]s %[1]s
+    complete -F __start_%[1]s %[1]s
```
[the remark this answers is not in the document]
h4ck3rk3y: That makes sense, I have reverted this line 👍

fish_completions.go
h4ck3rk3y: I am not super familiar with fish, a work around I could think off was,

```
function __cli_perform_completion_once
    if test -n "$results"
        __cli_debug "seems like its already set skipping"
        return 0
    end

    set --global results (__cli_perform_completion)
    __cli_debug "set global results"
    return 0
end
```

and then unsetting it with `__cli_clear_set_globals`, seems like in fish, the order in which completions execute is a stack, so clear set globals will get called after the useful completions at the end

```
complete -c  && __cli_clear_set_globals'
# If this doesn't require order preservation, we don't use the -k flag
complete -c cli -n 'not __cli_requires_order_preservation && __cli_prepare_completions' -f -a '$__cli_comp_results'
# otherwise we use the -k flag
complete -k -c cli -n '__cli_requires_order_preservation && __cli_prepare_completions -f -a '$__cli_comp_results'
```

any thoughts on this?

bash_completionsV2.go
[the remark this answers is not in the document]
marckhouzam: It is up to you.  Many project still use the legacy bash completion, but we don't officially support it for new features.  You decide what you prefer.

fish_completions.go
marckhouzam: I like that and I don't have any other ideas.
Checking the doc here https://fishshell.com/docs/current/cmds/complete.html
we can read: "Invoking complete multiple times for the same command adds the new definitions on top of any existing completions defined for the command."

So it implies the `complete` commands we register are called in a specific order, which is what we need.

You can go ahead and implement it as you suggest 👍

h4ck3rk3y: I've implemented this 👍 !

bash_completionsV2.go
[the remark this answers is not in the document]
h4ck3rk3y: Our tool doesn't use the legacy completion! I am happy skipping this.

h4ck3rk3y:
@marckhouzam let me know if you get a chance to look at this

marckhouzam (approved):
Looks great 🎉 
Thanks @h4ck3rk3y

PR #1904, opened 2023-02-06
Improve MarkFlagsMutuallyExclusive example in User Guide

janhn:


marckhouzam (approved):
Thanks for the improvement @janhn !

Issue #1911, opened 2023-02-15
no support for removing group


valllabh:
https://github.com/spf13/cobra/blob/b4f979ae352828a0153281b590771ce9588d67f8/command.go#L1309

We can see there is a support to `AddGroup` but there is no support for `RemoveGroup`
Where as we also see support for `AddCommand` and `RemoveCommand`

marckhouzam:
This seems to make sense, but due completeness can you give an example of where this would be needed?

valllabh:
Sure @marckhouzam

Usecase:
Consider there is a Cobra based tool that is already there written and maintained by the community and developer needs to use it as library to add/remove more functionalities. In such cases developer should be able to modify (add, remove) `Groups`  just like commands.

Example:
I am facing such issue right now, where I am wrapping `restish` with some more functionalities.
Forking `restish` will lead to unnecessary maintenance.
I need to add extra functionality and make the tool more simple for the my use case .
It will also not make sense to move that code into `restish`.

`restish` is built using `cobra` and it has exposed root command for such use cases.
Where developer can add/remove more commands and groups. But `cobra` do not support removal of groups.

JunNishimura:
@marckhouzam @valllabh 

Sorry, I missed that a PR(https://github.com/spf13/cobra/pull/1912) has already been submitted regarding this Issue.
However, it appears that a PR has not been updated, so I am submitting new PR(#1997) regarding this issue just in case.

I would appreciate it if you check the PR.

PR #1912, opened 2023-02-15
Missing support for removing group #1911

valllabh:
Fixed Missing support for removing group #1911

### DIFF
--- a/command.go
+++ b/command.go
@@ -1312,8 +1312,8 @@ func (c *Command) AddGroup(groups ...*Group) {
 }
 
 // RemoveGroup removes command group from parent command.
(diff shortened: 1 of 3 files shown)


marckhouzam:
Thanks @valllabh.
When you remove a group I assume you must also remove all subcommands from that group?

So without your PR do I understand correctly that you are able to remove all your sub commands from a group but that then the group appears empty in the help text. Is that right?

If that is the case, do we really need to allow removing groups? Maybe we should simply not display empty groups in the help?

valllabh:
Closing. Duplicated to #1997

PR #1914, opened 2023-02-16
OnKillRun feature

decanus:
A helper to run a function if a command is killed.

### DIFF
--- a/command.go
+++ b/command.go
@@ -23,9 +23,11 @@ import (
 	"fmt"
 	"io"
 	"os"
+	"os/signal"
 	"path/filepath"
 	"sort"
 	"strings"
+	"syscall"
 
 	flag "github.com/spf13/pflag"
 )
@@ -136,6 +138,8 @@ type Command struct {
 	PersistentPostRun func(cmd *Command, args []string)
 	// PersistentPostRunE: PersistentPostRun but returns an error.
 	PersistentPostRunE func(cmd *Command, args []string) error
+	// OnKillRun: run if a commands execution is exited
+	OnKillRun func(cmd *Command, args []string)
 
 	// groups for subcommands
 	commandgroups []*Group
@@ -904,6 +908,23 @@ func (c *Command) execute(a []string) (err error) {
 	argWoFlags := c.Flags().Args()
 	if c.DisableFlagParsing {
 		argWoFlags = a
+	}
+
+	if c.OnKillRun != nil {
+		sigchan := make(chan os.Signal)
+		signal.Notify(
+			sigchan,
+			syscall.SIGINT,
+			syscall.SIGKILL,
+			syscall.SIGTERM,
+			syscall.SIGQUIT,
+		)
+
+		go func() {
+			_ = <-sigchan
+
+			c.OnKillRun(c, argWoFlags)
+		}()
 	}
 
 	if err := c.ValidateArgs(argWoFlags); err != nil {


marckhouzam:
Cool idea @decanus !
I haven't tried it out yet but it looks useful.
This will need some tests and documentation.
cc @jpmcb

talbright:
This is a good idea...we had to implement this outside of cobra (which is fine.) But I think this establishes a good practice that all CLI's should have OOB.

marckhouzam:
@decanus Sorry for the delay.  Can you rebase this to make CI pass?

decanus:
@marckhouzam done!

command.go
```diff
@@ -930,6 +934,23 @@ func (c *Command) execute(a []string) (err error) {
 		argWoFlags = a
 	}
 
+	if c.OnKillRun != nil {
+		sigchan := make(chan os.Signal)
+		signal.Notify(
+			sigchan,
+			syscall.SIGINT,
+			syscall.SIGKILL,
```
marckhouzam, line 942: By definition, at least on POSIX, SIGKILL cannot be caught.  I've tested it on my Mac and I can confirm an `OnKillRun` function does not get called.  So I suggest removing this one.

In fact, the linter reports it should be removed 😄

command.go
```diff
@@ -930,6 +934,23 @@ func (c *Command) execute(a []string) (err error) {
...
+			syscall.SIGTERM,
+			syscall.SIGQUIT,
+		)
+
+		go func() {
+			_ = <-sigchan
+
+			c.OnKillRun(c, argWoFlags)
```
marckhouzam, line 950: I wonder if the function would be interested in knowing the signal that was received?
For example, `SIGQUIT` is a request for a core dump; the `OnKillRun` function may want to honour that.
What do you think of passing the received signal to the `OnKillRun` function?

marckhouzam (changes requested):
This is really great @decanus !  I love it!
I would like to include it in the next release of Cobra.

Would you be able to add a small paragraph of documentation after this section https://github.com/spf13/cobra/blob/main/site/content/user_guide.md#prerun-and-postrun-hooks

And a unit test would be important to make sure this does not break in the future.

Finally, there are a couple of linter errors to fix.

Thanks a lot!

marckhouzam:
I was thinking about how this new feature would be used. I like the fact that the `OnKillRun` can be defined for a specific command. But what if I wanted to create a global `OnKillRun` for all the commands?  For example to always intercept `^C`?

Would that be a likely scenario?

marckhouzam:
@decanus Are you still interested in getting this merged?

command.go
```diff
@@ -930,6 +934,23 @@ func (c *Command) execute(a []string) (err error) {
...
 	}
 
+	if c.OnKillRun != nil {
+		sigchan := make(chan os.Signal)
+		signal.Notify(
+			sigchan,
+			syscall.SIGINT,
+			syscall.SIGKILL,
+			syscall.SIGTERM,
+			syscall.SIGQUIT,
```
marckhouzam, line 944: I think we also need `os.Interrupt` for things to work on Windows

rojer:
found this through google - would be nice to get this finished and merged, but i guess for now i'll have to roll my own signal handler

marckhouzam:
I'm still interested in merging this @decanus but CI is failing.
Also, a very small doc and test would be important.

command.go
prochac: They seem to be supported for Windows, yet use of `syscall` package can be tricky for cross-platform Go.
/usr/lib/go/src/syscall/types_windows.go

PR #1916, opened 2023-02-24
Add support for PowerShell 7.2+

### LINKED ISSUE #1849: Completer scripts broken in PowerShell 7.3+
PowerShell 7.3 has corrected a long-standing bug with quote escaping during native argument passing. This causes existing quoting work-arounds to break. In it's current form, the completer scripts no longer work in PowerShell 7.3+ when using the new default settings and attempting to complete an empty command (ie. `command [tab]` vs `command s[tab]`). 

The offending lines:
https://github.com/spf13/cobra/blob/6b0bd3076cfafd1c108264ed1e4aa0c0fe3f8537/powershell_completions.go#L109-L110

It's sort of hard to see, but the single back ticks escaping the quotes near the end of line 110 need to be removed. However it's dependent on _both_ the version of PowerShell _and_ the new preference variable for native command arguments, `$PSNativeCommandArgumentPassing`, so a better fix is likely the following (note this is in PowerShell, not as an embedded go string) :

```powershell
# We need to use `"`" to pass an empty argument a "" or '' does not work!!!
# PowerShell 7.3+ need to use "" if Argument Passing is not Legacy	
if ($PSVersionTable.PsVersion -lt [version]'7.3.0' -or $PSNativeCommandArgumentPassing -eq 'Legacy') {
     $RequestComp="$RequestComp" + ' `"`"'
} else {
     $RequestComp="$RequestComp" + ' ""'
}
```
I'm not sure if that's the most elegant fix nor if it is 100% correct given that there's multiple modes (`Legacy`, `Standard` and `Windows`). I'm also not sure if the script needs any other modifications, though in my limited testing it is has fixed all of the completer scripts I use that were generated via this library (ie. flux/helm/kustomize)

See GitHub issues [here](https://github.com/PowerShell/PowerShell/issues/18540) and [here](https://github.com/PowerShell/PowerShell/issues/13428#issuecomment-1297723605)  as well as [documentation for PSNativeCommandArgumentPassing](https://learn.microsoft.com/en-us/powershell/scripting/learn/experimental-features?view=powershell-7.3#psnativecommandargumentpassing) (which despite being on the experimental features page is actually mainstream/GA in 7.3)

oldium:
PowerShell 7.2 has changed the way arguments are passed to executables. This was originally an experimental feature in 7.2, but as of 7.3 it is built-in. A simple "" is now sufficient for passing empty arguments, no back-tick escaping is required.

Fixes #1849

### DIFF
--- a/powershell_completions.go
+++ b/powershell_completions.go
@@ -106,8 +106,17 @@ filter __%[1]s_escapeStringWithSpecialChars {
         # If the last parameter is complete (there is a space following it)
         # We add an extra empty parameter so we can indicate this to the go method.
         __%[1]s_debug "Adding extra empty parameter"
-`+"        # We need to use `\"`\" to pass an empty argument a \"\" or '' does not work!!!"+`
-`+"        $RequestComp=\"$RequestComp\" + ' `\"`\"'"+`
+        # PowerShell 7.2+ changed the way how the arguments are passed to executables,
+        # so for pre-7.2 or when Legacy argument passing is enabled we need to use
+`+"        # `\"`\" to pass an empty argument, a \"\" or '' does not work!!!"+`
+        if ($PSVersionTable.PsVersion -lt [version]'7.2.0' -or
+            ($PSVersionTable.PsVersion -lt [version]'7.3.0' -and -not [ExperimentalFeature]::IsEnabled("PSNativeCommandArgumentPassing")) -or
+            (($PSVersionTable.PsVersion -ge [version]'7.3.0' -or [ExperimentalFeature]::IsEnabled("PSNativeCommandArgumentPassing")) -and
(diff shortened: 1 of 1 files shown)


oldium:
Tested separately the `if` block on PowerShell 5.1, 7.1, 7.2 (with experimental feature disabled and enabled) and 7.3.

marckhouzam (approved):
Thanks @oldium !
I have tested with pwsh 7.3.2 before this PR and I can see the problem for the very first set of completions `./prog <TAB>`, which don't appear.  I have then applied your PR and it does fix the problem.

> Tested separately the `if` block on PowerShell 5.1, 7.1, 7.2 (with experimental feature disabled and enabled) and 7.3.

Thanks for testing those cases, I am not able to do so myself, so I'll rely on your testing.

PR #1922, opened 2023-03-03
provide a message in case of an error

gusega:


marckhouzam:
Thanks @gusega.  Could you give an example of how you imagine this being used?
It is also a bit tricky because we would start showing a message based on the completions being returned, but some projects may be blindly returning something but expecting it to be ignored when the ShellCompDirectiveError is used.

An alternative could be to use ActiveHelp in this case, which is already supported for both zsh and bash.
See this doc for ActiveHelp details: https://github.com/spf13/cobra/blob/main/active_help.md

gusega:
Hi @marckhouzam 

So, in your program you'd return some error description as a single value of the completion slice, like this: `[]string{"error description"}, cobra.ShellCompDirectiveNoFileComp`. 

I will look into active help.

Thanks.

PR #1926, opened 2023-03-05
ci: deprecate go 1.16

umarcor:
As #1866, this PR removes go 1.16 from CI (since it's failing due to a test dependency) but keeps the `go.mod` unmodified.


marckhouzam:
I was hoping you would open something to fix CI. Thanks!

marckhouzam (approved):

PR #1930, opened 2023-03-10
Document suggested layout for subcommands

lcarva:
Document the approach suggested in https://github.com/spf13/cobra/issues/266#issuecomment-817295792.

### DIFF
--- a/user_guide.md
+++ b/user_guide.md
@@ -188,6 +188,37 @@ var versionCmd = &cobra.Command{
 }
 ```
 
+### Organizing subcommands
+
+A command may have subcommands which in turn may have other subcommands. This is achieved by using
+`AddCommand`. In some cases, especially in larger applications, each subcommand may be defined in
+its own go package.
+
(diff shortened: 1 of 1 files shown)


user_guide.md
```diff
@@ -188,6 +188,37 @@ var versionCmd = &cobra.Command{
...
 ```
 
+### Organizing subcommands
+
+A command may have subcommands which in turn may have other subcommands. This is achieved by using
+`AddCommand`. In some cases, especially in larger applications, each subcommand may be defined in
+its own go package.
+
+The suggested approach is for the parent command to use `AddCommand` to add its most immediate
+subcommands. For example, consider the following directory structured:
```
marckhouzam, line 198:
```suggestion
subcommands. For example, consider the following directory structure:
```

marckhouzam (changes requested):
This looks useful. Thanks.
One typo before we can merge.

marckhouzam (approved):
Thanks

PR #1932, opened 2023-03-14
Update `projects_using_cobra.md` add `Vitess` and `Arewefastyet`

frouioui:
This Pull Request updates the `projects_using_cobra.md` file to add Vitess and Arewefastyet as projects that are using Cobra.


marckhouzam (approved):
Thanks!

PR #1935, opened 2023-03-23
fix(bash,bash-v2): avoid process substitutions for the POSIX mode

akinomyoga:
Bash 5.0 and lower do not support process substitutions `<(...)` when the POSIX mode is turned on. On the other hand, we can always use here strings and command substitutions. In this PR, we rewrite process substitutions for Bash (`bash_completions{,V2}.go`) to use here strings and command substitutions instead.


marckhouzam:
cc @scop 
This is beyond my comfort zone 😅

scop:
TBH, can't say it would be within mine either :) But I've grown to trust things submitted by @akinomyoga as if I had written them. Often more so.

For this particular case, I don't see any functional problems offhand. But one generic drawback of here-strings is that their bash implementation uses temporary files. So in disk full scenarios the process substitution approach might succeed and the here-string one fail. (I'm not sure if the process substitution one can end up using temporary files too, but at least it doesn't for me given relatively small inputs.)

So this is kind of a judgement call. The disk full corner case might not be a biggie. But then again, I'm not sure how relevant it is to support interactive shells and completions in POSIX mode bash. Nice to have surely, but :shrug:. (We have that as a goal in bash-completion, but as far as I'm concerned it's mostly just cargo culted goal these days.)

scop:
On the other hand, @akinomyoga's suggestion of using `eval` and command substitution instead of process substitution or here-strings elsewhere should not have either the POSIX mode or the temp file drawbacks. In the particular case of this PR, it would result in a somewhat larger diff and perhaps not as clean code, but I guess that's what I'd vouch for.

akinomyoga:
I think It is fair to provide more context and other possibilities. In this PR, I tried to minimize the changes, but there are also other alternative options. I can adjust the PR if necessary.

**Option: use word splitting**--The current uses of the process substitutions in cobra have the following forms:

```bash
while IFS='' read -r comp; do
    COMPREPLY+=("$comp")
done < <(compgen -W "${out}" -- "$cur")
```

This can be rewritten by the word splitting:

```bash
local IFS=$'\n' reset=$(shopt -po noglob)
set -noglob
COMPREPLY+=($(compgen -W "${out}" -- "$cur"))
IFS=$' \t\n'
$reset
```

This doesn't require temporary files. On the other hand, this adds a fork cost for `$(shopt -po noglob)` which is usually much slower than the creation of temporary files. The fork can be actually avoided by instead recording `$-` and later checking the saved value to determine whether we should run `set +noglob`. If we ignore the fork cost, this approach should be faster than the `while read` loop approach.

**Option: not support the POSIX mode**--It is also an option not to support the POSIX mode, in which case we should check the POSIX mode at the beginning of the produced script and return quietly. (Any error messages shouldn't be usually outputted from bashrc unless it is a fatal error for the Bash startup. Any outputs to `stdout` would make `scp` or other operations using the output of `ssh` commands. Any outputs to `stderr` would cause startup errors in some of the window managers which load bashrc through `.bash_profile` or `.profile`.)

----

Also, there are some background information.

**A case where the POSIX mode is relevant**--I'm not sure if we should care about such a case, but even if the user doesn't explicitly turn on the POSIX mode, the POSIX mode can be turned on when the user starts Bash through `/bin/sh` (which would be typically a hard or symbolic link to Bash or Dash). By default, `.bashrc` will not be loaded if the bash instance is started from `/bin/sh`, but some users source `.bashrc` from `.profile` when `BASH_VERSION` is defined. With such a setup, the completion settings in `.bashrc` would be active even in Bash started from `sh`. I once received a report in my project caused by this situation.

**Here strings and temporary files in Bash 5.1+**--Bash had been always using temporary files internally for here documents and here strings but, from Bash 5.1, started to use a pipe when the content size is smaller than the system's default pipe size (e.g. 64 KiB in Linux). This means that Bash >= 5.1 does not use temporary files internally for here strings as far as we do not generate a very large number of completion candidates.

**Support for process substitutions**--Although I don't know the actual systems without the support for the process substitutions, but the process substitutions are only supported in systems where named pipes or `/dev/fd` is supported according to Bash Reference Manual:

> *From [3.5.6 Process Substitution](https://www.gnu.org/software/bash/manual/bash.html#Process-Substitution)*
>
> Process substitution is supported on systems that support named pipes (FIFOs) or the /dev/fd method of naming open files.

Issue #1937, opened 2023-03-26
RFE: env var for disabling descriptions


scop:
In https://github.com/scop/bash-completion/pull/905, we've added fallback 3rd party completion loaders for various known commands, including some cobra powered ones, and more are expected to be added later. However, as it stands, this mechanism does not have any way for users to specify `--no-descriptions` if that's their preference.

We could create a bash-completion specific way to accomplish this, but in my opinion this would be better placed in cobra proper. An environment variable would serve this nicely I think: something like `COBRA_COMPLETION_NO_DESCRIPTIONS(=non-empty-value)` could be equivalent as if `--no-descriptions` was passed in the generator command line.

marckhouzam:
Agreed. I've been thinking about this for a long time but never actually pulled the trigger.  Thanks for bringing it up.

scop:
Quick take on it in https://github.com/spf13/cobra/pull/1938

PR #1938, opened 2023-03-27
Add env variable to suppress completion descriptions on create

scop:
Closes https://github.com/spf13/cobra/issues/1937

### DIFF
--- a/completions.go
+++ b/completions.go
@@ -212,7 +212,7 @@ func (c *Command) initCompleteCmd(args []string) {
 				// 2- Even without completions, we need to print the directive
 			}
 
-			noDescriptions := cmd.CalledAs() == ShellCompNoDescRequestCmd || GetEnvConfig(cmd, configEnvVarSuffixDescriptions) == configEnvVarDescriptionsOff
+			noDescriptions := cmd.CalledAs() == ShellCompNoDescRequestCmd || getEnvConfig(cmd, configEnvVarSuffixDescriptions) == configEnvVarDescriptionsOff
 			noActiveHelp := GetActiveHelpConfig(finalCmd) == activeHelpGlobalDisable
 			out := finalCmd.OutOrStdout()
 			for _, comp := range completions {
@@ -920,12 +920,12 @@ func configEnvVar(name, suffix string) string {
 	return v
 }
 
-// GetEnvConfig returns the value of the configuration environment variable
+// getEnvConfig returns the value of the configuration environment variable
 // <PROGRAM>_<SUFFIX> where <PROGRAM> is the name of the root command in upper
 // case, with all non-ASCII-alphanumeric characters replaced by `_`.
 // If the value is empty or not set, the value of the environment variable
 // COBRA_<SUFFIX> is returned instead.
-func GetEnvConfig(cmd *Command, suffix string) string {
+func getEnvConfig(cmd *Command, suffix string) string {
 	v := os.Getenv(configEnvVar(cmd.Root().Name(), suffix))
 	if v == "" {
 		v = os.Getenv(configEnvVar(configEnvVarGlobalPrefix, suffix))
--- a/completions.go
+++ b/completions.go
@@ -212,7 +212,7 @@ func (c *Command) initCompleteCmd(args []string) {
 				// 2- Even without completions, we need to print the directive
 			}
 
-			noDescriptions := (cmd.CalledAs() == ShellCompNoDescRequestCmd)
+			noDescriptions := cmd.CalledAs() == ShellCompNoDescRequestCmd || GetEnvConfig(cmd, configEnvVarSuffixDescriptions) == configEnvVarDescriptionsOff
 			noActiveHelp := GetActiveHelpConfig(finalCmd) == activeHelpGlobalDisable
 			out := finalCmd.OutOrStdout()
 			for _, comp := range completions {
@@ -901,8 +901,12 @@ func CompErrorln(msg string) {
 	CompError(fmt.Sprintf("%s\n", msg))
 }
 
-// configEnvVarGlobalPrefix should not be changed: users will be using it explicitly.
-const configEnvVarGlobalPrefix = "COBRA"
+// These values should not be changed: users will be using them explicitly.
+const (
+	configEnvVarGlobalPrefix       = "COBRA"
+	configEnvVarSuffixDescriptions = "COMPLETION_DESCRIPTIONS"
+	configEnvVarDescriptionsOff    = "off"
+)
 
 var configEnvVarPrefixSubstRegexp = regexp.MustCompile(`[^A-Z0-9_]`)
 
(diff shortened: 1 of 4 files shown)


marckhouzam:
Thanks @scop !
This approach will only work for programs that use the default `completion` command.  Also it would require regenerating the completion script to toggle the descriptions.

Instead we could have the go code itself check the new variable and not return descriptions with the list of completions.

This is what we do for activehelp and probably what I should have done for descriptions from the start.

This go logic would be here: https://github.com/spf13/cobra/blob/4dd4b25de38418174a6e859e8a32eaccca32dccc/completions.go#L200

Luap99:
I agree we should follow what active helps does , this includes support for two env vars, i.e. `COBRA_COMPLETION_NO_DESCRIPTIONS` and `<PROGRAM>_COMPLETION_NO_DESCRIPTIONS`, this allows users to still configure it on a per application basis without having to regenerate the completion script.

marckhouzam:
Thanks for the input @Luap99 👍

> COBRA_COMPLETION_NO_DESCRIPTIONS and <PROGRAM>_COMPLETION_NO_DESCRIPTIONS

Let's "go" 😆 with this flexible and"standard" approach if @scop  is willing.

scop:
Makes sense. I suggest we take a look at #1940 before reworking this though -- if/after that's in, the env var prefix handling could be generalized and shared between active help and this feature.

scop:
I suppose this is starting to take shape, however it requires #1940, so keeping as a draft until that's in.

This will also cause conflicts with #1957, so either this or that will eventually need rebasing. This one's still blocked and #1957 is not, so perhaps that one could be applied first? Then again, I would not want to delay this one any more.

scop:
Resolved conflicts, rebased.

completions.go
```diff
@@ -211,7 +212,7 @@ func (c *Command) initCompleteCmd(args []string) {
 				// 2- Even without completions, we need to print the directive
 			}
 
-			noDescriptions := (cmd.CalledAs() == ShellCompNoDescRequestCmd)
+			noDescriptions := cmd.CalledAs() == ShellCompNoDescRequestCmd || GetEnvConfig(cmd, configEnvVarSuffixDescriptions) == configEnvVarDescriptionsOff
```
marckhouzam, line 215: Could you add a test to make sure descriptions are deactivated when either the program variable or the global variable is properly set?

marckhouzam (commented):
Thanks for your patience @scop

completions_test.go
```diff
@@ -3517,3 +3518,78 @@ func TestGetFlagCompletion(t *testing.T) {
...
+			globalVar: "COBRA_TEST",
+			cmdVal:    "cmd",
+			globalVal: "global",
+			expected:  "cmd",
+		},
+	}
+
+	for _, tc := range testCases {
+		// Could make env handling cleaner with t.Setenv with Go >= 1.17
+		func() {
```
marckhouzam, line 3581: To be able to differentiate between the different tests upon a failure, you can replace `func()` with `t.Run(tc.name, func(t *testing.T)` and add a field `name` to your test struct.
Then you get (because I used `t.use` as the test case name because I was lazy):
```
=== RUN   TestGetEnvConfig/root
    completions_test.go:3591: expected: "cmd", got: ""
=== RUN   TestGetEnvConfig/root#01
    completions_test.go:3591: expected: "global", got: ""
=== RUN   TestGetEnvConfig/root#02
=== RUN   TestGetEnvConfig/foo.bar
    completions_test.go:3591: expected: "cmd", got: ""
=== RUN   TestGetEnvConfig/quux-BAZ
    completions_test.go:3591: expected: "cmd", got: ""
```

completions_test.go
```diff
@@ -3517,3 +3518,78 @@ func TestGetFlagCompletion(t *testing.T) {
...
+			err := os.Setenv(tc.cmdVar, tc.cmdVal)
+			defer assertNoErr(t, os.Unsetenv(tc.cmdVar))
```
marckhouzam, line 3583: I believe the tests fail because the call to ` os.Unsetenv()` is done right away, instead of in the `defer`.  Only `assertNoErr` is deferred.  If you want to achieve your intent you can do
```
defer func() {
   assertNoErr(t, os.Unsetenv(tc.cmdVar))
}()
```

completions.go
```diff
@@ -899,3 +900,35 @@ func CompError(msg string) {
...
+	v = configEnvVarPrefixSubstRegexp.ReplaceAllString(v, "_")
+	return v
+}
+
+// GetEnvConfig returns the value of the configuration environment variable
+// <PROGRAM>_<SUFFIX> where <PROGRAM> is the name of the root command in upper
+// case, with all non-ASCII-alphanumeric characters replaced by `_`.
+// If the value is empty or not set, the value of the environment variable
+// COBRA_<SUFFIX> is returned instead.
+func GetEnvConfig(cmd *Command, suffix string) string {
```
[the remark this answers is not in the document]
scop: 7aa559d693ce60d79c83942b31e1d0f76c7530bd

completions.go
```diff
@@ -899,3 +900,35 @@ func CompError(msg string) {
 func CompErrorln(msg string) {
 	CompError(fmt.Sprintf("%s\n", msg))
 }
+
+// These values should not be changed: users will be using them explicitly.
+const (
+	configEnvVarGlobalPrefix       = "COBRA"
+	configEnvVarSuffixDescriptions = "COMPLETION_DESCRIPTIONS"
+	configEnvVarDescriptionsOff    = "off"
```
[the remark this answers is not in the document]
scop: Those suggestions make sense to me, I'm not that hung up on `on/off`.

One issue of starting to accept true values for this is that setting a true value won't make e.g. descriptions magically appear if they haven't been written, whereas a false value will definitely turn then off no matter if they're implemented or not. But IMO that's a tiny one that can be addressed by wording related docs appropriately.

I'd keep the full `COMPLETION_DESCRIPTIONS` for clarity and not abbreviate it. Although `_ON` would be good readabilitywise, it would be inconsistent with `ACTIVE_HELP`.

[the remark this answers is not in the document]
scop: Also, `strconv.ParseBool` returns an error if the value is not one of the acceptable ones. We could handle it for descriptions, but if we want to make `ACTIVE_HELP` behave the same, there is a backwards incompatibility chance.

completions_test.go
scop: c96f1a622a4db0f6e407762be6852ea44247866d

completions.go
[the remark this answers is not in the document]
scop: 138b0ab96495545094a0576d989241872687f886

completions_test.go
scop: 977823cf93624546e86250c048a94c3f1a4a1f41

completions.go
scop: 1107319c750f917bc3bf1c74a7705fe6e93123be

marckhouzam (approved):
Very nice!
Thanks @scop.

Issue #1895, opened 2023-01-13
Enhance arg parsing for commands which expect arbitrary input after the '--' dash dash terminator


nq-2:
I'm working on an exec command which wants to take in a few well defined and validated args/flags, followed by arbitrary args/flags after the '--' terminator which will be passed on to another binary:
`rootCmd exec arg1 -flag1 -- arg2 arg3`.

The following functionality would have saved me a lot of boiler plate:
1.  Providing something like `func (*Command) SplitArgsAtDash ([]string, []string)` that would make it easier for commands to:
    * apply custom validation to only the pre-terminator args
    * pass the post-terminator args/flags to the child binary

I've copy pasted this tedious code into multiple cobra apps in separate repos, and each time I second guess myself because of how `pflags ArgsLenAtDash()` returns (#166) 

2. It would be nice if users could specify that positional arg validators like `NoArgs` or `ExactArgs(1)` should ignore all arguments after the `--` terminator, similar to #17
```go
var rootCmd = &cobra.Command{
	Use:   "rootCmd",
	Args: cobra.ExactArgs(1),
	Run: func(cmd *cobra.Command, args []string) {
		fmt.Println(args)
	},
}
```
Currently calling `rootCmd arg1 -- arg2`  results in:
> Error: accepts 1 arg(s), received 2

Not sure if 2. makes sense given how `Command.Args`/ custom validators work, but it would have saved me a lot of time.

malko:
To pass the post-terminator args flags to the child binary and remove them from args here's how I do it right now:

```go
func splitAdditionalArgs(args *[]string) []string {
	additionalArgs := getAdditionalArgs()
	length := len(additionalArgs)
	if length > 0 {
		tmp := *args
		tmp = tmp[0 : len(*args)-length]
		*args = tmp
	}
	return additionalArgs
}
```

And when declaring run I use it like this: 
```go
Run: func(cmd *cobra.Command, args []string) {
		// remove additional args from the command and populate additional args in child command
		additionalArgs := splitAdditionalArgs(&args)
		// from here args doesn't contains additional args anymore
		child := exec.Command("mybin", additionalArgs...)
		// ...
}
```

This let the problem that we can't use builtin validators. 

Regarding point 2 I wish to have this feature too :)

I think the best options would be to have a cobra.Cmd field like IgnoreAfterDDashArgs and then to have either a cmd.DDashArgs property set or a cmd.GetDDashArgs method to retrieve them easily.

PR #1941, opened 2023-03-30
Add Post terminator args support

malko:
This PR addresses #1895, adding an easy way to ignore args after the arg terminator --.
It also add easy retrieving of such arguments to easily manipulate them.
It is an optin feature, without setting IgnorePostTerminatorArgs to true it should not change anything, so no breaking change.

this PR includes:
- Adding a field IgnorePostTerminatorArgs to the Command struct
- Adding a method PostTerminatorArgs to the Command struct 
- test we don't break normal behaviour such as args validation
- documentation for the added feature

Hope this can be of any help to others.

### DIFF
--- a/args_test.go
+++ b/args_test.go
@@ -89,6 +89,22 @@ func TestNoArgs(t *testing.T) {
 	output, err := executeCommand(c)
 	expectSuccess(output, err, t)
 }
+func TestNoArgs_WithPostTerminatorArgs(t *testing.T) {
+	c := getCommand(NoArgs, false)
+	_, err := executeCommand(c, "--", "post", "args")
+	noArgsWithArgs(err, t, "post")
+	// got := c.PostTerminatorArgs()
+	// expected := []string{"post", "args"}
+	// if strings.Join(got, ",") != strings.Join(expected, ",") {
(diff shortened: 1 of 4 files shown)


args_test.go
```diff
@@ -33,78 +33,53 @@ func getCommand(args PositionalArgs, withValid bool) *Command {
 }
 
 func expectSuccess(output string, err error, t *testing.T) {
+	t.Helper()
```
malko, line 36: Adding t.Helper() avoid IDE to jump to that function when failing

malko:
Is there any chance for that to be part of next release ?

adrian-gierakowski:
I would really like to use this. Any chance for getting it accepted and released?

malko:
@adrian-gierakowski waiting for this since march 2023, the PR is now conflicting :cry: . 

I'm ok to make the extra work to get rid of conflict if this is not more that the user_guide.md to change, but the more maintainers wait to review this the more work risks to arise. 
What i'm trying to say, is that I won't make any more work on this until an official maintainer ask for it. This PR took me a lot of time, and for now even if I understand that maintainers have a lot to review and do to maintain the project, it looks to me like I've wasted my times trying to propose a PR.

Hope future will prove me wrong.

Issue #1942, opened 2023-03-30
Add getters for flagCompletionFunctions map to support interactive flag completion suggestions in cobra-prompt


avirtopeanu-ionos:
The flagCompletionFunctions map (https://github.com/spf13/cobra/blame/main/completions.go#L36) is used to store flag completion functions for a given command

I am working on extending the cobra-prompt library (https://github.com/stromland/cobra-prompt/) to add support for interactive flag completion suggestions. To achieve this, I need to access the flagCompletionFunctions map and retrieve the flag completion functions for a given command.

avirtopeanu-ionos:
> I believe you should already be able to add new flag completions to another program. And I guess if you want to know if your flag completion already exists, you could try to register a new one and check for an error being returned?

I may be misunderstanding something here. I don't want to add suggestions to the cobra command itself, I want to add suggestions to the embedded `go-prompt` Document object, which `cobra-prompt` library is building here https://github.com/stromland/cobra-prompt/blob/main/cobra-prompt.go#L131

I would do that something like this 
```
flagName := d.GetWordBeforeCursorWithSpace()
if compFunc, exists := command.GetFlagCompletionByName(flagName); exists {
	completions, _ := compFunc(command, args, flagName)

	for _, completion := range completions {
		suggestions = append(suggestions, prompt.Suggest{Text: completion})
	}
}
```

however I haven't had the chance of testing it this much. there's a prototype here https://github.com/avirtopeanu-ionos/cobra-prompt/tree/feat/add-flag-completions where readers can play with this if they want to, it uses my fork of cobra.

avirtopeanu-ionos:
Any update on this? We were considering making a fork of cobra if this PR #1943 shouldn't be merged

avirtopeanu-ionos:
#1943 was merged. Thank you!

PR #1943, opened 2023-03-30
feat: add getters for flag completions

avirtopeanu-ionos:
Would provide a way for reading flagCompletionFunctions map

 https://github.com/spf13/cobra/issues/1942


marckhouzam (approved):
Sorry for the delay.
LGTM

marckhouzam:
I believe the use of `flag := c.Flags().Lookup(flagName)` does not work for persistent flags.
I will post a fix.

PR #1945, opened 2023-04-05
Fix typo in fish completions

twpayne:
This is a trivial fix. The typo broke CI for projects that run a spellchecker (e.g. https://github.com/reviewdog/action-misspell) on their committed code ([like this one](https://github.com/twpayne/chezmoi/actions/runs/4617766534/jobs/8164413060)).

### DIFF
--- a/fish_completions.go
+++ b/fish_completions.go
@@ -113,7 +113,7 @@ function __%[1]s_clear_perform_completion_once_result
     __%[1]s_debug ""
     __%[1]s_debug "========= clearing previously set __%[1]s_perform_completion_once_result variable =========="
     set --erase __%[1]s_perform_completion_once_result
-    __%[1]s_debug "Succesfully erased the variable __%[1]s_perform_completion_once_result"
+    __%[1]s_debug "Successfully erased the variable __%[1]s_perform_completion_once_result"
 end
 
 function __%[1]s_requires_order_preservation


marckhouzam (approved):
Thanks @twpayne .

Although trivial, from your description it could be impactful.

Can you clarify:
- why it's affecting your ci? Is it because you are vendoring Cobra?
- Can you work around it?

Also, off you can find time, I suggest opening an issue asking Cobra to use a spell check linter in its own CI.

twpayne:
> * why it's affecting your ci? Is it because you are vendoring Cobra?

The reasons are:
1. I use cobra to generate completions for multiple shells.
2. The generated completions for fish include this typo.
3. I commit the generated completion files to my GitHub repo.
4. I run a spell checker as part of my CI.
5. The spell checker complains about the spelling error in the generated fish completion file.

I'm not vendoring Cobra.

> * Can you work around it?

Yes, I can simply ignore the spelling error and merge anyway (which I've done). The spelling checker in CI only checks new code, so this is a satisfactory work-around for me.



> Also, if you can find time, I suggest opening an issue asking Cobra to use a spell check linter in its own CI.

Yes, I would like to do this, but, I'm not sure when I will have time to do this. From a quick investigation, it looks like Cobra's golangci-lint configuration is quite out of date, so it's not quite as simple as enabling the `misspell` linter and fixing the problems it finds.

Issue #1891, opened 2023-01-09
Flags mutually exclusive but one of them required


krystian-panek-wttech:
hey, thanks for cool library, I have only a one question:

```go
func pkgDefineFileAndUrlFlags(cmd *cobra.Command) {
	cmd.Flags().String("file", "", "Local ZIP path")
	cmd.Flags().String("url", "", "URL to ZIP")
	cmd.MarkFlagsMutuallyExclusive("file", "url")
}
```

how to make only one of flags defined required, I mean `--file` or `--url` ? but not at the same time?

I see that there is `MarkFlagsRequiredTogether` but I more want to have sth like `MarkFlagsRequiredSome("file", "url")` :)

Any ideas have to handle such case?
Greetings, Krystian

marckhouzam:
Sadly this is not supported. See #1761 .
You would have to implement that logic yourself in your program once the flags are parsed.

krystian-panek-wttech:
ok, thanks for confirmation; I hope I will find some time to improve it in Cobra someday :)

marckhouzam:
> ok, thanks for confirmation; I hope I will find some time to improve it in Cobra someday :)

That would be great!

marckhouzam:
Fixed by @marevers in #1952

Issue #1761, opened 2022-07-22
Mark flag mutually exclusive and required


shanduur-auto:
Hi!

I want to ask, if there is an option to mark one of `n` of the mutually exclusive flags required? This means there must appear exactly one of the `n` flags, otherwise an error should be printed.

For example:
```sh
# this should print contents from file:
$ echo -f ./file.txt
hello world

# this should act as a regular echo:
$ echo -i "hello world"
hello world

# this should print help and info about missing required flags:
$ echo
Error: one of mutually exclusive required flag(s) not set: "f", "i"
Usage:
  echo [flags]
```

marckhouzam:
Hi @shanduur-auto.  Currently there is no special support from Cobra for what you are asking.  You would have to do the verification yourself in your program.

marckhouzam:
Fixed by @marevers in #1952

PR #1952, opened 2023-04-25
Add 'one required flag' group

marevers:
This PR addresses #1761 and #1891 - it adds a new flag group which can be used to group a set of flags of which at least one is required. If the related command is invoked and none of the flags in the group are set, the user is informed of the flag group (e.g. with two flags called 'path' and 'id'):

```
Error: at least one of the flags in the group [path id] is required
```

It is used like the other flag groups (MarkFlagsMutuallyExclusive / MarkFlagsRequiredTogether):

```
getCmd.AddCommand(getEntryCmd)
getEntryCmd.MarkFlagsOneRequired("path", "id")
```

It can also be combined with MarkFlagsMutuallyExclusive with the same set of flags to enforce only one flag (and not more than one).


marckhouzam (commented):
Thanks @marevers, this looks great.
I've added this to the planned features for the next release.

1. Add you mentioned, could you add some tests please?
2. Could you also update `enforceFlagGroupsForCompletion()`. I think what we need there is that: if none of the flags in the new required group are present, all of them should be marked as required so the completion suggests all of them, but as soon as one of them is present, we do nothing more.

marckhouzam:
Nicely done @marevers !  Thanks.
I want to run a bit more tests, but this should be ready to merge soon.

marckhouzam (commented):
Everything is fine.
One last thing to ensure quality, can you also add a test like `TestCompletionForGroupedFlags` and `TestCompletionForMutuallyExclusiveFlags`
Then we can merge.

Thanks

marckhouzam:
@marevers Sorry for the confusion.  The `TestCompletionForGroupedFlags` and `TestCompletionForMutuallyExclusiveFlags` already exist in the `completions_test.go` file.  You can do something very similar for a new `TestCompletionForOneRequiredGroupFlags`

marevers:
@marckhouzam Thanks for the clarification. I added a new test function `TestCompletionForOneRequiredGroupFlags` based on `TestCompletionForGroupedFlags` with very similar subtests but also verifying that when any flag in the one-required group is present, other flags will not be suggested.

marckhouzam (approved):
LGTM
Thanks @marevers !

marckhouzam:
@marevers I was about to press merge, and then I remembered we need to update the doc.  It should be very quick for you.  Can you update this section please: https://github.com/spf13/cobra/blob/main/site/content/user_guide.md#flag-groups

marevers:
@marckhouzam Okay, that should do it. I had to resync my fork because `user_guide.md` had been moved since I made that fork.

site/content/user_guide.md
```diff
@@ -349,7 +349,16 @@ rootCmd.Flags().BoolVar(&ofYaml, "yaml", false, "Output in YAML")
 rootCmd.MarkFlagsMutuallyExclusive("json", "yaml")
 ```
 
-In both of these cases:
+If you want to require least one flag from a group to be present, you can use `MarkFlagsOneRequired`.
```
marckhouzam, line 352:
```suggestion
If you want to require at least one flag from a group to be present, you can use `MarkFlagsOneRequired`.
```

marckhouzam (changes requested):
One typo and we're good to go.

marevers:
> One typo and we're good to go.

Fixed!

marckhouzam (approved):
LGTM
Thanks @marevers !
There’s already scenarios where this new feature can be used in the CLI I work on. This will be great!

PR #1956, opened 2023-05-05
Allow linker to perform deadcode elimination for program using Cobra

### LINKED ISSUE #2015: The use of text/template disables dead code elimination in all users of cobra
Any use of `reflect.Value.MethodByName()` or `reflect.Type.MethodByName()` disables the DCE. The compiler assumes that these functions will look any method up, and cannot remove methods even if they are never called.

`text/template` is a very notable offender in this respect. It calls `MethodByName()` with random strings extracted from template texts.

`Command::UsageFunc()` and `Command::HelpFunc()` use `text/template`, hence they disable the DCE for all users of cobra. Since cobra is the go-to module to implement CLIs, this essentially disables DCE in all command line tools written in go.

A low-level library that is used by almost everyone should not pessimise the code generation by disabling the DCE. I'd avoid `text/template` altogether.

aarzilli:
Fixes #2015 

Cobra, in its default configuration, will execute a template to generate help, usage and version outputs. Text/template execution calls MethodByName and MethodByName disables dead code elimination in the Go linker, therefore all programs that make use of cobra will be linked with dead code elimination disabled, even if they end up replacing the default usage, help and version formatters with a custom function and no actual text/template evaluations are ever made at runtime.

Dead code elimination in the linker helps reduce disk space and memory utilization of programs. For example, for the simple example program used by TestDeadcodeElimination 40% of the final executable size is dead code. For a more realistic example, 12% of the size of Delve's executable is deadcode.

This PR changes Cobra so that, in its default configuration, it does not automatically inhibit deadcode elimination by:

1. changing Cobra's default behavior to emit output for usage and help using simple Go functions instead of template execution
2. quarantining all calls to template execution into SetUsageTemplate, SetHelpTemplate and SetVersionTemplate so that the linker can statically determine if they are reachable

### DIFF
--- a/cobra_test.go
+++ b/cobra_test.go
@@ -15,6 +15,10 @@
 package cobra
 
 import (
+	"os"
+	"os/exec"
+	"path/filepath"
+	"strings"
 	"testing"
 	"text/template"
 )
@@ -220,5 +224,59 @@ func TestRpad(t *testing.T) {
 				t.Errorf("Expected rpad: %v\nGot: %v", tt.expected, got)
 			}
 		})
+	}
+}
+
+func TestDeadcodeElimination(t *testing.T) {
+	// check that a simple program using cobra in its default configuration is
+	// linked with deadcode elimination enabled.
+	const (
+		dirname  = "test_deadcode"
+		progname = "test_deadcode_elimination"
+	)
+	os.Mkdir(dirname, 0770)
+	filename := filepath.Join(dirname, progname+".go")
+	err := os.WriteFile(filename, []byte(`package main
+
+import (
+	"fmt"
+	"os"
+
+	"github.com/spf13/cobra"
+)
+
+var rootCmd = &cobra.Command{
+	Version: "1.0",
+	Use:     "example_program",
+	Short:   "example_program - test fixture to check that deadcode elimination is allowed",
+	Run: func(cmd *cobra.Command, args []string) {
+		fmt.Println("hello world")
+	},
+	Aliases: []string{"alias1", "alias2"},
+	Example: "stringer --help",
+}
+
+func main() {
+	if err := rootCmd.Execute(); err != nil {
+		fmt.Fprintf(os.Stderr, "Whoops. There was an error while executing your CLI '%s'", err)
+		os.Exit(1)
+	}
+}
+`), 0660)
+	if err != nil {
+		t.Fatalf("could not write test program: %v", err)
+	}
+	defer os.RemoveAll(dirname)
+	buf, err := exec.Command("go", "build", filename).CombinedOutput()
+	if err != nil {
+		t.Fatalf("could not compile test program: %s", string(buf))
+	}
+	defer os.Remove(progname)
+	buf, err = exec.Command("go", "tool", "nm", progname).CombinedOutput()
+	if err != nil {
+		t.Fatalf("could not run go tool nm: %v", err)
+	}
+	if strings.Contains(string(buf), "MethodByName") {
+		t.Error("compiled programs contains MethodByName symbol")
 	}
 }
--- a/cobra_test.go
+++ b/cobra_test.go
@@ -227,6 +227,20 @@ func TestRpad(t *testing.T) {
 	}
 }
 
+// TestDeadcodeElimination checks that a simple program using cobra in its
+// default configuration is linked taking full advantage of the linker's
+// deadcode elimination step.
+//
+// If reflect.Value.MethodByName/reflect.Value.Method are reachable the
+// linker will not always be able to prove that exported methods are
+// unreachable, making deadcode elimination less effective. Using
+// text/template and html/template makes reflect.Value.MethodByName
+// reachable.
+// Since cobra can use text/template templates this test checks that in its
+// default configuration that code path can be proven to be unreachable by
+// the linker.
+//
+// See also: https://github.com/spf13/cobra/pull/1956
 func TestDeadcodeElimination(t *testing.T) {
 	// check that a simple program using cobra in its default configuration is
 	// linked with deadcode elimination enabled.
@@ -234,7 +248,8 @@ func TestDeadcodeElimination(t *testing.T) {
 		dirname  = "test_deadcode"
 		progname = "test_deadcode_elimination"
 	)
-	os.Mkdir(dirname, 0770)
+	_ = os.Mkdir(dirname, 0770)
+	defer os.RemoveAll(dirname)
 	filename := filepath.Join(dirname, progname+".go")
 	err := os.WriteFile(filename, []byte(`package main
 
@@ -262,11 +277,10 @@ func main() {
 		os.Exit(1)
 	}
 }
-`), 0660)
+`), 0600)
 	if err != nil {
 		t.Fatalf("could not write test program: %v", err)
 	}
-	defer os.RemoveAll(dirname)
 	buf, err := exec.Command("go", "build", filename).CombinedOutput()
 	if err != nil {
 		t.Fatalf("could not compile test program: %s", string(buf))
--- a/cobra_test.go
+++ b/cobra_test.go
@@ -18,6 +18,7 @@ import (
 	"os"
 	"os/exec"
 	"path/filepath"
+	"runtime"
 	"strings"
 	"testing"
 	"text/template"
--- a/cobra_test.go
+++ b/cobra_test.go
@@ -242,6 +242,10 @@ func TestRpad(t *testing.T) {
 //
 // See also: https://github.com/spf13/cobra/pull/1956
 func TestDeadcodeElimination(t *testing.T) {
+	if runtime.GOOS == "windows" {
+		t.Skip("go tool nm fails on windows")
+	}
+
 	// check that a simple program using cobra in its default configuration is
 	// linked with deadcode elimination enabled.
 	const (
--- a/command.go
+++ b/command.go
@@ -1978,7 +1978,7 @@ func defaultUsageFunc(w io.Writer, in interface{}) error {
 		fmt.Fprintf(w, "\n\nAdditional help topcis:")
 		for _, subcmd := range c.Commands() {
 			if subcmd.IsAdditionalHelpTopicCommand() {
-				fmt.Fprintf(w, "\n %s %s", rpad(subcmd.CommandPath(), subcmd.CommandPathPadding()), subcmd.Short)
+				fmt.Fprintf(w, "\n  %s %s", rpad(subcmd.CommandPath(), subcmd.CommandPathPadding()), subcmd.Short)
 			}
 		}
 	}
--- a/command.go
+++ b/command.go
@@ -32,6 +32,8 @@ import (
 const (
 	FlagSetByCobraAnnotation     = "cobra_annotation_flag_set_by_cobra"
 	CommandDisplayNameAnnotation = "cobra_annotation_command_display_name"
+
+	helpFlagName = "help"
 )
 
 // FParseErrWhitelist configures Flag parse errors to be ignored
@@ -1170,7 +1172,7 @@ func (c *Command) checkCommandGroups() {
 // If c already has help flag, it will do nothing.
 func (c *Command) InitDefaultHelpFlag() {
 	c.mergePersistentFlags()
-	if c.Flags().Lookup("help") == nil {
+	if c.Flags().Lookup(helpFlagName) == nil {
 		usage := "help for "
 		name := c.DisplayName()
 		if name == "" {
@@ -1178,8 +1180,8 @@ func (c *Command) InitDefaultHelpFlag() {
 		} else {
 			usage += name
 		}
-		c.Flags().BoolP("help", "h", false, usage)
-		_ = c.Flags().SetAnnotation("help", FlagSetByCobraAnnotation, []string{"true"})
+		c.Flags().BoolP(helpFlagName, "h", false, usage)
+		_ = c.Flags().SetAnnotation(helpFlagName, FlagSetByCobraAnnotation, []string{"true"})
 	}
 }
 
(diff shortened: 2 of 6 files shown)


aarzilli:
The CLA thing isn't working for some reason.

aarzilli:
Unless I made a mistake there shouldn't be any behavior changes (as in, observable from the outside). Happy to explain anything about this if needed.

aarzilli:
Ping?

aarzilli:
Ping?

marckhouzam:
@aarzilli I apologize for such a long delay, but I had misunderstood the value of this PR.  I now realize that it may help all programs using Cobra, so I'm very interested.

I'm trying to convince myself that that this change actually has an impact.
Would you be able to explain how I can see the difference in a program that has dead code, before and after this PR?

Here is what I tried.
- I used the program you have added to the tests in this PR
- I added an "Unused()" function that I don't call to that program.
- I ran `go build -o myprog .` (using cobra 1.8.1 *without* your PR
- I ran `go tool nm ./myprog | grep Unused`

I would have expected to see the `Unused` function but I don't.
Could you clarify what I should expect?

This is the program:
```
package main

import (
	"fmt"
	"os"

	"github.com/spf13/cobra"
)

func Unused() {
	fmt.Println("not called")
}

var rootCmd = &cobra.Command{
	Version: "1.0",
	Use:     "example_program",
	Short:   "example_program - test fixture to check that deadcode elimination is allowed",
		fmt.Println("hello world")
	},
	Aliases: []string{"alias1", "alias2"},
	Example: "stringer --help",
}

func main() {
		fmt.Fprintf(os.Stderr, "Whoops. There was an error while executing your CLI '%s'", err)
		os.Exit(1)
	}
}
```

aarzilli:
One easy way to see the impact of this change is to compile your example program with and without this PR:

```
$ ls -lh
total 8.4M
-rwxr-xr-x. 1 a a 5.2M Oct 14 14:45 example.withoutpr*
-rwxr-xr-x. 1 a a 3.2M Oct 14 14:45 example.withpr*
-rw-r--r--. 1 a a  596 Oct 14 14:44 main.go
```

This is an extreme example where the size of the executable is reduced by 38% but reductions of 10% are realistic.
I use go1.23 but you should get very similar results with go1.22.8.

As to your question, saying that deadcode elimination gets disabled is incorrect: in reality it continues to function but in a reduced capacity. Specifically if `reflect.Value.MethodByName` or `reflect.Value.Method` are reachable it can not always prove that exported methods are unreachable.

Your Unused function is not an exported method so, unless it is called by an exported method, it can always be deadcode eliminated.

If you want to see the Unused function make it all the way to the executable you have to do something like this:

```
package main

import (
	"fmt"
	"os"

	"github.com/spf13/cobra"
)

type Astruct struct {
	n int
}

//go:noinline
func Unused() {
	fmt.Println("not called")
}

//go:noinline
func (a *Astruct) Unused() {
	fmt.Println("not called", a.n)
	Unused()
}

//go:noinline
func (a *Astruct) Used() {
	fmt.Println("used", a.n)
}

var rootCmd = &cobra.Command{
	Version: "1.0",
	Use:     "example_program",
	Short:   "example_program - test fixture to check that deadcode elimination is allowed",
		fmt.Println("hello world")
	},
	Aliases: []string{"alias1", "alias2"},
	Example: "stringer --help",
}

func main() {
		fmt.Fprintf(os.Stderr, "Whoops. There was an error while executing your CLI '%s'", err)
		os.Exit(1)
	}
	var a Astruct
	fmt.Println("%s", a)
	a.Used()
}
```

All of the noinline directives are needed because small functions like this would be removed by the inliner and the fmt.Println call is needed to make the Astruct type itself reachable (if all calls to the methods of Astruct are static the linker can still prove that Unused is unreachable).

With these changes you get:

```
$ go build main2.go && go tool nm ./main2 | grep 'Unused'
  589c00 T main.(*Astruct).Unused
  589ba0 T main.Unused
  765570 D runtime.adviseUnused
  4179e0 T runtime.sysUnusedOS

```

without the PR and:

```
$ go build main2.go && go tool nm ./main2 | grep 'Unused'
  602564 D runtime.adviseUnused
  4144a0 T runtime.sysUnusedOS

```

with the PR.

marckhouzam:
Thanks for the explanation @aarzilli, I can now see the benefits.

So, with this PR, programs that don't set new templates (usage, help or version) will be able to do dead code elimination fully.

What about programs that do modify those templates?  If they want to get full usage of dead code elimination they should convert their template use to a go function like you have done, IIUC?   My next step for this review is to try to override the help/usage/version in that fashion and see if it works as expected.

I believe that using `SetHelpFunc()` and `SetUsageFunc()` will allow programs to do this, but I don't think we have a way to do this for the version template.  But it is ok, it is still better than not being able to do it at all.  Once this is merged, we can discuss adding a `SetVersionFunc()`.

marckhouzam:
What do you think about updating the documentation in the three sections starting here https://github.com/spf13/cobra/blob/main/site/content/user_guide.md#defining-your-own-help to explain the side-effect of overriding the template, and therefore that it is recommended to set a function instead?

aarzilli:
> Thanks for the explanation @aarzilli, I can now see the benefits.
> 
> So, with this PR, programs that don't set new templates (usage, help or version) will be able to do dead code elimination fully.
> 
> What about programs that do modify those templates? If they want to get full usage of dead code elimination they should convert their template use to a go function like you have done, IIUC?

Yes, this is correct. 

> I believe that using `SetHelpFunc()` and `SetUsageFunc()` will allow programs to do this, but I don't think we have a way to do this for the version template. But it is ok, it is still better than not being able to do it at all. Once this is merged, we can discuss adding a `SetVersionFunc()`.

It imagine should be trivial.

cobra_test.go
```diff
@@ -222,3 +226,57 @@ func TestRpad(t *testing.T) {
...
 }
+
+func TestDeadcodeElimination(t *testing.T) {
+	// check that a simple program using cobra in its default configuration is
+	// linked with deadcode elimination enabled.
+	const (
+		dirname  = "test_deadcode"
+		progname = "test_deadcode_elimination"
+	)
+	os.Mkdir(dirname, 0770)
```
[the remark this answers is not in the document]
aarzilli: Done

cobra_test.go
```diff
@@ -222,3 +226,57 @@ func TestRpad(t *testing.T) {
...
+	Example: "stringer --help",
+}
+
+func main() {
+	if err := rootCmd.Execute(); err != nil {
+		fmt.Fprintf(os.Stderr, "Whoops. There was an error while executing your CLI '%s'", err)
+		os.Exit(1)
+	}
+}
+`), 0660)
```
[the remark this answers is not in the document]
aarzilli: Done

cobra_test.go
```diff
@@ -222,3 +226,57 @@ func TestRpad(t *testing.T) {
...
+	if err != nil {
+		t.Fatalf("could not write test program: %v", err)
+	}
+	defer os.RemoveAll(dirname)
```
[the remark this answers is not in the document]
aarzilli: Done

cobra_test.go
```diff
@@ -222,3 +226,57 @@ func TestRpad(t *testing.T) {
 		})
 	}
 }
+
```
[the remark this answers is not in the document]
aarzilli: See if you like what I wrote.

marckhouzam:
The linter has failed

I looked in detail into this and it makes sense.
The problem is that in reality, so many projects use templates, that I wonder if any real program will benefit from this improvement?

For example, programs that use some k8s libraries will still end up using templates. I’ve tested this with both tanzu and helm and the binary size didn’t change. 

So I was just hesitant in accepting this because it complicates the Cobra code base and prevents any future use of templates. But it also feels wrong to force any program using cobra to have a dependency on templates. 

@aarzilli maybe you can convince me. Did you say Delve would be able to reduce its size with this change?

dominiquelefevre:
> The problem is that in reality, so many projects use templates, that I wonder if any real program will benefit from this improvement?

At my company, most of services see a reduction of their size by 40% to 45% when DCE is enabled.

> For example, programs that use some k8s libraries will still end up using templates

As for k8s, most of it depends on `text/template` only indirectly via `x/net/trace`. Its package init func registers some complex HTTP handlers that use templating. That causes other problems as well, and I think that fixing the init will fix the dependency for k8s, too. See https://github.com/golang/go/issues/69951.

I think that k8s should not influence the design of other libs. There is a huge go ecosystem outside k8s.

aarzilli:
> @aarzilli maybe you can convince me. Did you say Delve would be able to reduce its size with this change?

Delve uses templates (and reflection) in various places but cobra is the only place where it couldn't be easily circumvented.

I think you are right that any given large Go program is likely to end up having multiple different dependencies on reflection but cobra itself is not large, nor is it only recommended for large programs.

You say "prevents any future use of templates" but I don't think it's true: first if they aren't statically reachable they are allowed, secondly this decision could be reversed if a compelling use case for templates was found.

pgimalac:
To weight in, this change (and a few other similar ones in other dependencies) would reduce the size of various binaries I work on by 18 to 26% !

Despite using hundreds of dependencies, in particular from `k8s`, `docker`, `opentelemetry`, or `trivy`, I only had to change 6 (including `cobra`) to make it work, so it seems this is not that widespread.

Some known libraries provide explicit workarounds to avoid the import, eg. [google.golang.org/grpc](https://pkg.go.dev/google.golang.org/grpc) uses [golang.org/x/net/trace](https://pkg.go.dev/golang.org/x/net/trace), which disables the optimization, but they provide a build tag to remove the import (`grpcnotrace`).

command.go
```diff
@@ -1899,3 +1882,141 @@ func commandNameMatches(s string, t string) bool {
...
+	}
+	if c.HasAvailableInheritedFlags() {
+		fmt.Fprintf(w, "\n\nGlobal Flags:\n")
+		fmt.Fprint(w, trimRightSpace(c.InheritedFlags().FlagUsages()))
+	}
+	if c.HasHelpSubCommands() {
+		fmt.Fprintf(w, "\n\nAdditional help topcis:")
+		for _, subcmd := range c.Commands() {
+			if subcmd.IsAdditionalHelpTopicCommand() {
+				fmt.Fprintf(w, "\n %s %s", rpad(subcmd.CommandPath(), subcmd.CommandPathPadding()), subcmd.Short)
```
marckhouzam, line 1981: Missing one space here: `fmt.Fprintf(w, "\n %s %s"` -> `fmt.Fprintf(w, "\n  %s %s"`

marckhouzam: I have fixed it.

marckhouzam:
On MinGW were getting a CI error

START| DeadcodeElimination
     |     cobra_test.go:291: could not run go tool nm: exit status 1
FAIL | DeadcodeElimination (0.56s)

we could actually just turn that test off for windows.

command.go
```diff
@@ -434,7 +437,11 @@ func (c *Command) UsageFunc() (f func(*Command) error) {
 	}
 	return func(c *Command) error {
 		c.mergePersistentFlags()
-		err := tmpl(c.OutOrStderr(), c.UsageTemplate(), c)
+		fn := defaultUsageFunc
+		if c.usageTemplate != nil {
+			fn = c.usageTemplate.fn
+		}
```
marckhouzam, line 443: Unlike the previous code which called `UsageTemplate()` which would return the usage template of the first parent that has one, this new code does not check for parents.

I will push a couple of new tests to check this case.
I'll also try to fix it.

aarzilli: Doesn't the code above do it? This is the full func:

```
func (c *Command) UsageFunc() (f func(*Command) error) {
	if c.usageFunc != nil {
		return c.usageFunc
	}
	if c.HasParent() {
		return c.Parent().UsageFunc() // <-- this call?
	}
	return func(c *Command) error {
		c.mergePersistentFlags()
		fn := defaultUsageFunc
		if c.usageTemplate != nil {
			fn = c.usageTemplate.fn
		}
		err := fn(c.OutOrStderr(), c)
		if err != nil {
			c.PrintErrln(err)
		}
		return err
	}
}
```

marckhouzam: Hmm. I see your point. Let me dig a bit.

marckhouzam: So the fields `usageFunc` and `usageTemplate` are two different ways a program can override the usage text. If a program calls `SetUsageTemplate()` only, it will affect the call to `UsageFunc` but only through the code I’m pointing out. 

What had me confused, and possibly you as well, is that because of the line you point out, it gives the impression that the part calling `usageTemplate` can only work on the root command.  I had to debug to realize the the `c` command in the function is passed as an argument and is **not** the same as the `c` command in the code you point out.  So, in the function we start with a `c` being a sub-command, and we still need to go through the parents looking for a `usageTemplate`

marckhouzam: I will change the name of the parameter to the function, it is much too confusing.

marckhouzam: Turns out that it is nice to have this "confusing" shadowing.  It avoids the risk of using the other `c` command inside the function by mistake.  I'll leave it alone.

marckhouzam (approved):
This is what I feel is the correct final version for this PR. @aarzilli can you please look at the changes I pushed on top of yours and confirm you agree?

command.go
aarzilli: I agree.

aarzilli:
The changes look good to me.

marckhouzam:
> This is a very good improvement. @marckhouzam, could you please make a new release of cobra?

Nice to see immediate interest 😄 .
I'll work on a release this week.

aarzilli:
Thanks for getting this merged!

sylr:
@jpmcb Could we get a cobra release soon to be able to enjoy this change ?

PR #1957, opened 2023-05-09
Micro-optimizations

scop:
Noticed while looking into something else. No benchmarks/numbers done, but even if this wouldn't yield measurable improvements, the changes don't hurt readability IMO and actually in a sense they convey the intent better.

### DIFF
--- a/args.go
+++ b/args.go
@@ -54,7 +54,7 @@ func OnlyValidArgs(cmd *Command, args []string) error {
 		// A description is following a tab character.
 		var validArgs []string
 		for _, v := range cmd.ValidArgs {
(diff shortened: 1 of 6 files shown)


marckhouzam:
Thanks @scop !
I haven't had as much time as I would have like for Cobra recently but things should improve in a couple of weeks. I'll give this some attention then.

scop:
Rebased and added some more.

marckhouzam (approved):
LGTM
But I’m not near a computer right now so I’ll wait until I can test it to merge. 

@jpmcb  if you’re comfortable merging, go right ahead

PR #1960, opened 2023-05-16
powershell: escape variable with curly brackets

### LINKED ISSUE #1853: autocompletion not working on windows bash & powershell
its possible to remove the .exe extension to use binaries in windows git bash by specifying shopt -s completion_strip_exe
but bash still calls it with the exe extension. this leads to the issue that the completions are only registerd for the `binary.exe` and not also for just `binary`

e.g.: https://github.com/containers/podman/issues/16499

it registers only podman.exe: complete -o default -F __start_podman.exe podman.exe
it should also inlclude without .exe extension: complete -o default -F __start_podman.exe podman.exe podman.

Luap99:
This fixes an issue with program names that include a dot, in our case `podman.exe`. This was caused by the change in commit 6ba7ebbc.

Fixes #1853

### DIFF
--- a/powershell_completions.go
+++ b/powershell_completions.go
@@ -47,7 +47,7 @@ filter __%[1]s_escapeStringWithSpecialChars {
 `+"    $_ -replace '\\s|#|@|\\$|;|,|''|\\{|\\}|\\(|\\)|\"|`|\\||<|>|&','`$&'"+`
 }
 
-[scriptblock]$__%[2]sCompleterBlock = {
+[scriptblock]${__%[2]sCompleterBlock} = {
     param(
             $WordToComplete,
             $CommandAst,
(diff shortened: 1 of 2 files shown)


marckhouzam:
Thanks @Luap99 .
I've added this to our next milestone and will try to find some time soon.

marckhouzam:
@Luap99 I'm trying to reproduce the problem but I am unable to.   Could you provide instructions?

Luap99:
Looking at this more closely I think you need to have the dot in the root command name, i.e.
```
var rootCmd = &cobra.Command{
	Use: "testprog.exe",
...
```
Of course this looks weird and why would we do that but in podman we actually just pass `filepath.Base(os.Args[0])` to `Use:`. I guess we can fix that specific case in podman but the real problem is that the logic does not account for a dot in the command name.
And now properly testing having a dot in your command name seems to break bash and fish as well. AFAICT only zsh works with it. I think having a dot in command name is something we need to support, also for that matter I think most other special chars cause similar problems.

Luap99:
> Thanks for the info @Luap99. This seems slightly similar to #1940.

Ah yes that sounds reasonable. I will check if this PR is enough or if other cases are missing.

marckhouzam (approved):
LGTM

PR #1972, opened 2023-06-06
feat: add MarkFlagsMutuallyExclusiveAndRequired

### LINKED ISSUE #1216: Any best practice to imitate the "either-or" flags?
### Background
Suppose you want to make a simple app that lets you migrate from database to database, and I want to allow the end user to **supply either _a DSN string_ or _fill in the connection structure_**, both represented as flags. If you know Java, this is comparable to Spring DB config. 

Now, I have problems trying to imitate such relation: it's an either-or relation, i.e. logically mutually exclusive. However, whether the (colloquial) either-or is XOR or IMPLY is still debatable, so I want to discuss with others if they had any more clues here.

### Current workarounds
1. Use priority, so we basically allow both DSN and struct to be filled in, and weight in which one to override which during tiebreaking. Usually DSN takes the precedence.
2. Error at runtime if both are supplied. This is the intuitive solution that most people including me can figure out instantly. This is also the least effort cobra needs to do. You can even panic if you will because this is an undefined, exceptional case.
3. Error at cobra init time. This can be tricky because you have to make up your own generic error message and figure out the condition.

### Ideal solution
Implement a MarkMutualExclusionForFlagGroups method to let it correctly represent the relation: **only one flag group among the "mutual exclusion set" can be activated**. 

Before that, we also need to implement FlagGroups -- **flags that are chained to be required**. It should set error if more than one groups are fulfilled. This is the exact solution to this problem but it is a very heavy engineering, i.e. no one will do this.

baruchiro:
Fixes #1216

I chose the most naive way.

Just like the flow of `mutuallyExclusive`, I cloned it to another flow for `mutuallyExclusiveAndRequired`, with only different condition, to return an error also if the `set` slice is empty.

Example:

```shell
❯ go run .
Error: exactly one of the flags in the group [op1 op2] must be set; none were set
Usage:
  cobra-temp [flags]

Flags:
  -h, --help         help for cobra-temp
      --op1 string   Option1 mutually exclusive and required with Option2
      --op2 string   Option2 mutually exclusive and required with Option1

exit status 1
❯ go run . --op1 foo
❯ go run . --op2 foo
❯ go run . --op1 foo --op2 foo
Error: exactly one of the flags in the group [op1 op2] must be set; [op1 op2] were all set
Usage:
  cobra-temp [flags]

Flags:
  -h, --help         help for cobra-temp
      --op1 string   Option1 mutually exclusive and required with Option2
      --op2 string   Option2 mutually exclusive and required with Option1

exit status 1
```

### DIFF
--- a/flag_groups.go
+++ b/flag_groups.go
@@ -23,8 +23,9 @@ import (
 )
 
 const (
-	requiredAsGroup   = "cobra_annotation_required_if_others_set"
-	mutuallyExclusive = "cobra_annotation_mutually_exclusive"
+	requiredAsGroup              = "cobra_annotation_required_if_others_set"
+	mutuallyExclusive            = "cobra_annotation_mutually_exclusive"
+	mutuallyExclusiveAndRequired = "cobra_annotation_mutually_exclusive_and_required"
 )
 
 // MarkFlagsRequiredTogether marks the given flags with annotations so that Cobra errors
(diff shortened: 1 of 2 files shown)


marckhouzam:
Thanks @baruchiro !
The issue this PR fixes has a long discussion.
Could you add in the description of this PR an example of how this new group can be used?

baruchiro:
@marckhouzam Description updated

marckhouzam:
> @marckhouzam Description updated

Thanks. Now that understand better what this does I think it overlaps with #1952 which allows to require one flag out of a group and then use the existing API `MarkFlagsMutuallyExclusive` to force mutual exclusiveness.

What do you think?

baruchiro:
I think #1952 is better, thanks!

PR #1980, opened 2023-06-16
test: make fish_completions_test more robust

branchvincent:
Use temporary files instead of assuming the current directory is writable. Also, if creating a temporary file still returns an error, prevent the test from failing silently by replacing `log.Fatal` with `t.Fatal`.

### DIFF
--- a/fish_completions_test.go
+++ b/fish_completions_test.go
@@ -16,9 +16,10 @@ package cobra
 
 import (
 	"bytes"
+	"errors"
 	"fmt"
-	"log"
 	"os"
+	"path/filepath"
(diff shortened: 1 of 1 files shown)


marckhouzam:
Thanks @branchvincent ! This seems better.
You need to first sign the CLA. Please see this comment https://github.com/spf13/cobra/pull/1980#issuecomment-1594080528

marckhouzam (approved):
LGTM pending signing of the CLA

branchvincent:
done @marckhouzam, thanks for the quick review and all your work on this project!

PR #1428, opened 2021-06-29
Move documentation sources to site/content

umarcor:
In preparation for #1245, this PR moves documentation sources (markdown files) into `site/content`:

- `projects_using_cobra.md` is moved to `site/content/`.
- All `*_completions.md` are moved to `site/content/completions/*.md`.
- All `doc/*_docs.md` are moved to `site/content/docgen/*.md`.
- `doc/README.md` is moved to `site/content/CLI.md`.

References are updated accordingly.

Issue #1979, opened 2023-06-14
Broken link


guettli:
On https://cobra.dev/

The link to "[Here](https://github.com/spf13/cobra/blob/master/cobra/README.md) you can find more information about it." is broken.

404: file not found.

umarcor:
See #1674.

tlarnold10:
@guettli @umarcor this appears to still be an issue... the issue you referenced has a merged PR, was that supposed to fix this issue?

Issue #1959, opened 2023-05-12
Broken links in cobra.dev site


Parthiba-Hazra:
The cobra-generator link is not working, seems older version -- https://cobra.dev/#using-the-cobra-generator
also same case with -- https://cobra.dev/#generating-documentation-for-your-command

umarcor:
See #1674.

PR #1984, opened 2023-06-20
site: add gohugo configuration and CI

umarcor:
Following #1428, this PR is a subset of #1245.

This PR adds the minimal configuration and a CI workflow to build a documentation site using hugo and push it to branch `gh-pages` (to be hosted at USERNAME.github.io/cobra).


umarcor:
@marckhouzam can we please have this milestoned and merged?

marckhouzam:
@umarcor when we merge this what should we see?  Still there be a new gh site for users?  Is there already one on your account?

umarcor:
@marckhouzam the CI workflow will:

- Build an HTML site using Hugo.
- Upload it as an artifact (zipfile).
- Push it to branch `gh-pages` (always overwriting the content of that branch but showing the commit that it corresponds to; see https://github.com/umarcor/cobra/blob/umarcor/pages/.github/workflows/site.yml#L41 and https://github.com/umarcor/cobra/blob/umarcor/pages/.github/workflows/site.yml#L47).

GitHub Pages allows serving the content of branch `gh-pages` at `USERNAME.github.io/REPONAME`. Actually, it can be served from any branch, either from the root of the branch or from subdir `/docs`.
However, the site might not be served by default. When the default token is used to push to `gh-pages`, it does not trigger the activation. A user needs to enable it through https://github.com/USERNAME/cobra/settings/pages, or push a commit to `gh-pages` with user credentials (just once; it does not matter if all future commits are created by the token, neither that it overwrites the commit of the user).

I suggest you push this branch to your fork, let CI run, and then check https://github.com/marckhouzam/cobra/settings/pages.

When GitHub Pages is enabled in a repo, it will show in the sidebar, in section 'Environtments'. See my fork: https://github.com/umarcor/cobra. Also, in section 'About', the website can be set to the Pages (without manually typing it).
Nonetheless, both of those references can be hidden (by unchecking the 'Environments' section and leaving the URL empty).
That is, we can merge this PR in order to have CI for the docs and start working on improving them, but we do not need to advertise it yet, so that users still find the "official" cobra.dev.

GitHub Pages does also support setting a custom domain. Therefore, eventually we could make spf13.github.io/cobra be an alias of cobra.dev (or the other way round). I.e. users navigating to cobra.dev will see the content of the gh-pages of this repo.
Alternatively, we can make the CI of this repo push to the `master` branch of spf13/cobra.dev. However, that would require a token with write permissions in that repo, while doing it from the gh-pages branch of this repo allows using the default token in CI and not dealing with PATs. Overall, when the gh-pages of this repo is good enough to be served at cobra.dev, I believe we should archive spf13/cobra.dev.

Yet, I don't think this PR alone is enough to replace cobra.dev. This is just to have CI and start treating documentation-as-code (e.g. checking for broken links/refs). I believe we should then discuss what theme we want, and reorganise the (sub)sections to better use the sidebar(s).

The main reason to push this PR forward is the amount of issues and PRs related to cobra.dev being broken for several years already:

- https://github.com/spf13/cobra.dev/issues
- https://github.com/spf13/cobra.dev/pulls
- #1996
- #1979
- #1959
- #1951
- #1674

We could close all those issues/PRs with "The automatically generated and up-to-date content can be found in spf13.github.io/cobra. There is work-in-progress to make cobra.dev serve that content".

umarcor:
*ping* @marckhouzam

site/go.mod
```diff
@@ -0,0 +1,10 @@
+module github.com/bep/docuapi
```
marckhouzam, line 1: Is this correct?

Why do we need a sub module? And if we do, isn’t there a `go.sum` file?

marckhouzam (commented):
I’m finally going to give this PR some attention. I’ve just started but I’ll continue in the next little while

sorry for all the delay @umarcor

.github/workflows/site.yml
```diff
@@ -0,0 +1,47 @@
...
+jobs:
+  doc:
+    runs-on: ubuntu-latest
+    steps:
+
+    - uses: actions/checkout@v3
+
+    - name: Get Hugo and Theme
+      run: |
+        curl -fsSL https://github.com/gohugoio/hugo/releases/download/v0.114.0/hugo_extended_0.114.0_Linux-64bit.tar.gz | sudo tar xzf - -C /usr/local/bin hugo
```
marckhouzam, line 16: How about using a github action?
This seems to be the most popular: https://github.com/marketplace/actions/hugo-setup
This example should suit our needs I think: https://github.com/peaceiris/actions-hugo?tab=readme-ov-file#%EF%B8%8F-workflow-for-autoprefixer-and-postcss-cli

.github/workflows/site.yml
```diff
@@ -0,0 +1,47 @@
...
+      with:
+        name: site
+        path: site/public
+
+    - name: Publish site to gh-pages
+      if: github.event_name != 'pull_request'
+      run: |
+        cd site/public/
+        touch .nojekyll
+        git init
```
marckhouzam, line 41: I wanted to discuss this approach with you, but I'll wait to see what you think about the GH Action, which may change this approach.

marckhouzam (commented):
Ok, I'm on board with this.
Sorry @umarcor this took sooo long: I hadn't used Hugo before so I kept pushing this off as I needed to ramp up.

It has been so long that some parts of the PR would benefit with modernizing.  I've put comments in those locations.

Some of my comments may be off as I don't have much experience with Hugo, but hopefully we can work out the details together.

Issue #1918, opened 2023-02-28
"help" subcommand does not work when MarkPersistentRequiredFlag is used.


makiuchi-d:
This issue seems to be similar to #1093.

The auto-implemented subcommands (help and completion) don't work when `MarkPersistentRequiredFlag()` called on root command.
Of course it works if the required flag is specified.
IMHO, these auto-implemented subcommands should be able to work without flags defined by user.
Additionaly, these subcommends should not ignore auto-implement  flags like `-h, --help` (this differs from #1093).

```go
package main

import (
	"github.com/spf13/cobra"
)

var rootCmd = &cobra.Command{
	Use: "mycmd",
}

var subCmd = &cobra.Command{
	Use: "subcmd",
	Run: func(cmd *cobra.Command, args []string) {},
}

func main() {
	rootCmd.PersistentFlags().Bool("required", true, "required")
	rootCmd.MarkPersistentFlagRequired("required")
	rootCmd.AddCommand(subCmd)
	rootCmd.Execute()
}
```

```console
$ go run . help
Error: required flag(s) "required" not set
Usage:
  mycmd help [command] [flags]

Flags:
  -h, --help   help for help

Global Flags:
      --required   required (default true)
```

marckhouzam:
Thanks @makiuchi-d. You're point sounds reasonable to me.  I would also consider this a bug.

payxn:
Hello,

Do you have a workaround for this ?

makiuchi-d:
No, I always specify the required flags. (although I often use aliases or shell scripts)

JunNishimura:
I made PR for this Issue!

@marckhouzam 
It would be helpful if you could check my [PR](https://github.com/spf13/cobra/pull/1992)!

PR #1992, opened 2023-07-09
Disable mark persistent required flag for help and completion command

JunNishimura:
# Overview
This PR addresses #1918.

The auto-implemented subcommands (help and completion) did't work when MarkPersistentRequiredFlag() called on root command. So I set DisableFlagParsing true for HelpCommand and CompletionCommand (bash, zsh, fish, powershell) to avoid required flags validation.

I also added unit test to see if the code works as I expected.

# Related Issue
- https://github.com/spf13/cobra/issues/1918

### DIFF
--- a/command.go
+++ b/command.go
@@ -1190,6 +1190,7 @@ func (c *Command) InitDefaultHelpCmd() {
 			Short: "Help about any command",
 			Long: `Help provides help for any command in the application.
 Simply type ` + c.Name() + ` help [path to command] for full details.`,
+			DisableFlagParsing: true,
 			ValidArgsFunction: func(c *Command, args []string, toComplete string) ([]string, ShellCompDirective) {
 				var completions []string
 				cmd, _, e := c.Root().Find(args)
--- a/command_test.go
+++ b/command_test.go
@@ -885,6 +885,18 @@ func TestHelpCommandExecuted(t *testing.T) {
 	checkStringContains(t, output, rootCmd.Long)
 }
 
+func TestHelpCommandExecutedWithPersistentRequiredFlags(t *testing.T) {
+	rootCmd := &Command{Use: "root", Run: emptyRun}
+	rootCmd.PersistentFlags().Bool("foo", false, "")
+	childCmd := &Command{Use: "child", Run: emptyRun}
+	rootCmd.AddCommand(childCmd)
+	assertNoErr(t, rootCmd.MarkPersistentFlagRequired("foo"))
+
+	if _, err := executeCommand(rootCmd, "help"); err != nil {
+		t.Errorf("unexpected error: %v", err)
+	}
+}
+
 func TestHelpCommandExecutedOnChild(t *testing.T) {
 	rootCmd := &Command{Use: "root", Run: emptyRun}
 	childCmd := &Command{Use: "child", Long: "Long description", Run: emptyRun}
--- a/completions.go
+++ b/completions.go
@@ -713,6 +713,7 @@ You will need to start a new shell for this setup to take effect.
 `, c.Root().Name()),
 		Args:                  NoArgs,
 		DisableFlagsInUseLine: true,
+		DisableFlagParsing:    true,
 		ValidArgsFunction:     NoFileCompletions,
 		RunE: func(cmd *Command, args []string) error {
 			return cmd.Root().GenBashCompletionV2(out, !noDesc)
@@ -748,8 +749,9 @@ To load completions for every new session, execute once:
 
 You will need to start a new shell for this setup to take effect.
 `, c.Root().Name()),
-		Args:              NoArgs,
-		ValidArgsFunction: NoFileCompletions,
+		Args:               NoArgs,
+		DisableFlagParsing: true,
+		ValidArgsFunction:  NoFileCompletions,
 		RunE: func(cmd *Command, args []string) error {
 			if noDesc {
 				return cmd.Root().GenZshCompletionNoDesc(out)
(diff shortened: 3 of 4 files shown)


marckhouzam:
Thanks @JunNishimura.

However this change causes a couple of regressions (that are apparently not tested).
1. `./prog completion bash --no-descriptions` no longer works, and the same for the other three shells
2. `./prog completion bash -h` no longer works, and the same for the other three shells
3. `./prog help -h` no longer works (it used to show a help for the help itself)

I don't think we can disable flag parsing on those commands.  We have to find another approach.

You may be able to add a `PersistentPreRun` function to those two commands which would disable the annotation that marks a flag as required.  Something like:
```
			PersistentPreRun: func(cmd *Command, args []string) {
				cmd.Flags().VisitAll(func(pflag *flag.Flag) {
					requiredAnnotation, found := pflag.Annotations[BashCompOneRequiredFlag]
					if !found {
						return
					}
					if requiredAnnotation[0] == "true" {
						pflag.Annotations[BashCompOneRequiredFlag] = []string{"false"}
					}
				})
			},
```

marckhouzam (changes requested):
See above comment

JunNishimura:
Thanks, @marckhouzam 

Your point was right.
I didn't notice that my previous code caused some regressions.

And, I appreciate the alternative suggestions.
I implemented your suggestions in the code.

I would appreciate it if you check the commits again.

marckhouzam:
@JunNishimura This looks pretty good.
There is a subtle backwards-compatibility impact that we should probably protect against.
If you look at where the `PersistentPreRun` functions are called you will notice that if a command has a `PersistentPreRun`, the `PersistentPreRun` of its parent is not called: 
https://github.com/spf13/cobra/blob/dcb405a9399ed307eaa0e50f5cb1c290c8979dd4/command.go#L913-L923)

So here is the scenario:
1. a CLI tool using Cobra has defined a `PersistentPreRun` function on its _root_ command which does something that should be done for every command (e.g., collect some statistics about which command is being called)
2. this PR adds a `PersistentPreRun` on the sub-commands `completion` and `help`

Because of the new `PersistentPreRun` on `completion` and `help`, the `root.PersistentPreRun` will no longer be called for those two sub-commands.  This is a change in behaviour and could affect projects.

I suggest that at the end of the two new `PersistentPreRun` functions, we actually call `root.PersistentPreRun` to get the original behaviour back.

Does that make sense?

JunNishimura:
Thanks, @marckhouzam 

I understand the situation. 
I'm so glad to know the inner algorithm of cobra thanks to your points.

I implemented to call root's PersistentPreRun in sub-command's (help and completion) PersistentPreRun.
I also added unit tests to the above implementation just in case.

I would appreciate it if you check the commits again.

marckhouzam:
@JunNishimura The backwards-compatibility fix is a little more complicated.  For example the `root.PersistentPreRun` could be nil, so we cannot blindly call it.  Also, the root command may not use `PersistentPreRun` but instead `PersistentPreRunE` (notice the final `E` on the second one).  So we should mimic what cobra does already to make this work properly.  I'm thinking of something like this:
```
  if cmd.Root().PersistentPreRunE != nil {
      return cmd.Root().PersistentPreRunE(cmd, args)
  } else if cmd.Root().PersistentPreRun != nil {
      cmd.Root().PersistentPreRun(cmd, args)
  }
  return nil
```
To be able to propagate any error returned by `cmd.Root().PersistentPreRunE` you will probably have to use `PersistentPreRunE` for the `help` and `completion` commands.

Let me know if you need any clarifications, and thanks again for your efforts!

JunNishimura:
Thanks, @marckhouzam 

Your point is right.
My implementation did not take into account the possibility that root.PersistentPreRun is nil.

I modified to use PersitentPreRunE to be able to propagate any error returned by `root.PersistentPreRunE` instaed of PersistentPreRun. 
I also added a check to see if the function is not nil.

Sorry for repeating myself, but I'd appreciate it if you'd check commits again.

marckhouzam (approved):
This LGTM.  Thanks @JunNishimura !
I will mark it as approved but I'll wait a little while to allow @jpmcb to have a look if he wants.

marckhouzam:
@JunNishimura The linter is failing 😢 .  It is just that this PR introduce a third string `"true"` which the linter wants us to define a constant for.  Could you please do that and use it in the three instance of `"true"`

Thanks

JunNishimura:
Thanks, @marckhouzam 

I define a constant for "true" named `trueString`.
The name `trueString` might be too direct, so if you have a better name, please let me know and I will change it.

marckhouzam:
Thanks for the quick turnaround with the linter @JunNishimura .
This looks good now. 

As this change had backwards-compatibility implications
> I'll wait a little while to allow @jpmcb to have a look if he wants.

jpmcb:
Planning to take a look this weekend. Thanks @marckhouzam for the heads up!

JunNishimura:
@jpmcb 
Hi! How about this PR?

If you haven't seen it because you are busy or for some reasons, that's perfectly fine, as I do not want to rush you. I would appreciate it if you could take a look when you have time.

I reminded you because I thought it was possible that you simply forgot.

marckhouzam:
This needs a rebase

saltbo:
Any update？

JunNishimura:
@marckhouzam @jpmcb 
Sorry for the late reply. I resolved the conflict.

I would appreciate it if you check this PR again!!

PR #1997, opened 2023-07-16
Add remove group method for command

JunNishimura:
## Overview
This PR is about #1911.
Sorry, I missed that a PR(#1912) has already been submitted regarding this Issue.
However, it appears that a PR has not been updated, so I am submitting this just in case.

I added `RemoveGroup` method for command. RemoveGroup method removes groups from commandgroups and at the same time initializes the command's groupID.

## Related Issue
- https://github.com/spf13/cobra/issues/1911

### DIFF
--- a/command.go
+++ b/command.go
@@ -1311,6 +1311,30 @@ func (c *Command) AddGroup(groups ...*Group) {
 	c.commandgroups = append(c.commandgroups, groups...)
 }
 
+// RemoveGroup removes one or more command groups to this parent command.
+func (c *Command) RemoveGroup(groupIDs ...string) {
+	// remove groups from commandgroups
+	groups := []*Group{}
+main:
+	for _, group := range c.commandgroups {
+		for _, groupID := range groupIDs {
+			if group.ID == groupID {
+				continue main
+			}
+		}
+		groups = append(groups, group)
+	}
+	c.commandgroups = groups
+	// remove the groupID from the target commands
+	for _, command := range c.commands {
+		for _, groupID := range groupIDs {
+			if command.GroupID == groupID {
+				command.GroupID = ""
+			}
+		}
+	}
+}
+
 // RemoveCommand removes one or more commands from a parent command.
 func (c *Command) RemoveCommand(cmds ...*Command) {
 	commands := []*Command{}
--- a/command_test.go
+++ b/command_test.go
@@ -1862,6 +1862,61 @@ func TestAddGroup(t *testing.T) {
 	checkStringContains(t, output, "\nTest group\n  cmd")
 }
 
+func TestRemoveSingleGroup(t *testing.T) {
+	var rootCmd = &Command{Use: "root", Short: "test", Run: emptyRun}
+
+	rootCmd.AddGroup(
+		&Group{ID: "group", Title: "Test group"},
+		&Group{ID: "help", Title: "help"},
+		&Group{ID: "comp", Title: "comp"},
+	)
+
+	rootCmd.AddCommand(&Command{Use: "sub", GroupID: "group", Run: emptyRun})
+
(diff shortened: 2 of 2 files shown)


valllabh (approved):

command_test.go
```diff
@@ -1862,6 +1862,61 @@ func TestAddGroup(t *testing.T) {
 	checkStringContains(t, output, "\nTest group\n  cmd")
 }
 
+func TestRemoveSingleGroup(t *testing.T) {
```
jpmcb, line 1865: Thanks for adding tests! Definitely makes it easier to validate!!

jpmcb (approved):
This makes sense to me and looks to be on par with how `RemoveCommand` works 👍🏼

JunNishimura:
@jpmcb 
Thanks for checking the PR!!

One question, do I need to ask one more person to review this PR to merge it into the main branch?

site/content/user_guide.md
```diff
@@ -525,7 +525,7 @@ around it. In fact, you can provide your own if you want.
 
 Cobra supports grouping of available commands in the help output.  To group commands, each group must be explicitly
 defined using `AddGroup()` on the parent command.  Then a subcommand can be added to a group using the `GroupID` element
-of that subcommand. The groups will appear in the help output in the same order as they are defined using different
+of that subcommand. `RemoveGroup()` can be used to remove groups and initialize the GroupID of commands belonging to the removed group. The groups will appear in the help output in the same order as they are defined using different
```
marckhouzam, line 528: initialize -> reset

command.go
```diff
@@ -1311,6 +1311,30 @@ func (c *Command) AddGroup(groups ...*Group) {
 	c.commandgroups = append(c.commandgroups, groups...)
 }
 
+// RemoveGroup removes one or more command groups to this parent command.
+func (c *Command) RemoveGroup(groupIDs ...string) {
```
marckhouzam, line 1315: i’m not sure, but I don’t think this will allow removing a group for the help or completion commands. Do we want to support removing groups for those two commands?

marckhouzam (commented):
A thought I had. If the user calls remove group with an invalid group ID, the current logic will simply ignore it. Is that what we want or do you prefer to help the developer notice their mistake?

command.go
jpmcb: Interesting use case. I guess in theory someone might want to remove those from a group that have been added. But they are also sort of the "default" commands you get. What do you think Marc?

JunNishimura: In my opinion, I would like cobra to give users as much freedom of development as possible. It is up to the developers to decide whether they want to keep the `help` and `completion` commands as existing groupies or remove them.

jpmcb: I agree and I'm 100% for freedom of use. It more just becomes a matter of not breaking long standing existing features. But this is new functionality anyways. So I'm good with it!

JunNishimura: I fixed as follows. If the group id passed as argument match, the group id of the help and completion command is reset in the RemoveGroup method.  I would appreciate it if you check the code again.

JunNishimura:
@marckhouzam @jpmcb 
> Is that what we want or do you prefer to help the developer notice their mistake?

Thanks for pointing that out.

In my opinion, I though it would be better to help the developer notice their mistake.
And I think using `panic` to let the developer know as a coding error that should be corrected is appropriate in this case.

What do you think?

JunNishimura:
@marckhouzam @jpmcb 
Sorry it took so long to reply.

> Is that what we want or do you prefer to help the developer notice their mistake?

I thought it would be better to output an error if the expected group does not disappear as a result of calling the `RemoveGroup` command with invalid IDs, since that is not the result expected by the developer. Therefore I fixed the `RemoveGroup` command to return an error.

What do you think.

PR #1999, opened 2023-07-18
[WIP] Cound localFlags correctly

JunNishimura:


JunNishimura:
@johnSchnake 
Due to the large number of revisions, there is no urgent need for a reply, but I would appreciate it if you check this PR.

PR #2001, opened 2023-07-20
all: refactor to use %q instead of "%s"

gocurr:
This PR simplifies code using %q rather than "%s".

Please take a look. Thanks.

### DIFF
--- a/doc/man_docs.go
+++ b/doc/man_docs.go
@@ -146,7 +146,7 @@ func manPreamble(buf io.StringWriter, header *GenManHeader, cmd *cobra.Command, 
 		description = cmd.Short
 	}
 
-	cobra.WriteStringAndCheck(buf, fmt.Sprintf(`%% "%s" "%s" "%s" "%s" "%s"
+	cobra.WriteStringAndCheck(buf, fmt.Sprintf(`%% %q %q %q %q %q
 # NAME
 `, header.Title, header.Section, header.date, header.Source, header.Manual))
 	cobra.WriteStringAndCheck(buf, fmt.Sprintf("%s \\- %s\n\n", dashedName, cmd.Short))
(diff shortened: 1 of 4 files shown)


marckhouzam (approved):
Looks safe.  Thanks!

marckhouzam (changes requested):
CI disagrees that it’s safe. I didn’t look at what the error was though.

gocurr:
Thanks. I think it's this code.
```go
fmt.Errorf(`required flag(s) "%s" not set`, strings.Join(missingFlagNames, `", "`))
```
It is now reverted. Thanks again.

gocurr:
Please take a look at this.
```go
package main

import (
	"fmt"
	"path"
	"path/filepath"
	"strings"
	"time"
)

func main() {
	const fmTemplate = `---
date: %s
title: %q
slug: %s
url: %s
---
`
	filePrepender := func(filename string) string {
		now := time.Now().Format(time.RFC3339)
		name := filepath.Base(filename)
		base := strings.TrimSuffix(name, path.Ext(name))
		url := "/commands/" + strings.ToLower(base) + "/"
		return fmt.Sprintf(fmTemplate, now, strings.Replace(base, "_", " ", -1), base, url)
	}
	fmt.Printf("[%v]\n", filePrepender(`"abc".txt`))
}
```

The output is
```text
[---
date: 2023-07-21T13:37:27+08:00
title: "\"abc\""
slug: "abc"
url: /commands/"abc"/
---
]
```

And the original output of "%s" is
```text
[---
date: 2023-07-21T13:38:14+08:00
title: ""abc""
slug: "abc"
url: /commands/"abc"/
---
]
```

I don't know which style is better. Maybe this PR is unnecessary and I'll close it.

marckhouzam:
> I don't know which style is better. Maybe this PR is unnecessary and I'll close it.

yeah, with the extra escaping being done I see that %q is not exactly “%s”.  Although it may be better in some cases let’s avoid changing anything that may affect existing projects. 

Thanks anyway.

Issue #1281, opened 2020-11-22
Begin a command error message with a custom string


ashpooljh:
It is a convention for CLI utilities in Unix-like environments to prefix the name of the utility to any error messages printed. For example, notice how the `find` command outputs the `find:` prefix:
```
$ find nemo
find: ‘nemo’: No such file or directory
```
This is useful when debugging shell scripts for determining exactly which of the commands used in the script has failed.

Cobra has a convenient facility for declaring commands which might fail and return an error: the `Command.RunE()` method and friends. However, upon returning an error from such a method, we find that it is [hard-coded](https://github.com/spf13/cobra/blob/08c51e585ca78e4b9408ae034d0cdacdd81aad41/command.go#L970) in `Command.ExecuteC()` to prefix the string `Error:` to any error message that results from this.

My proposal is to add a way to change that hard-coded value to a custom string. Maybe use `argv[0]` by default to follow the convention. Without this feature, a programmer who wishes to adhere to the convention I mentioned is forced to implement their own error output in `Run()` or `RunE()` and/or silence the one provided by Cobra.

marckhouzam:
Sounds like a good idea.  But you'll have to keep the existing behavior as the default to remain backwards-compatible.

PR #2023, opened 2023-08-31
Customizable error message prefix

5ouma:
You can change the prefix before the error message by `SetErrPrefix()`.
It's currently `Error:`, and now it's read from `errPrefix`.
It's just like the `versionTemplate` and `SetVersionTemplate()` relationship.

Resolve: #1281

### DIFF
--- a/command.go
+++ b/command.go
@@ -1070,7 +1070,7 @@ func (c *Command) ExecuteC() (cmd *Command, err error) {
 			c = cmd
 		}
 		if !c.SilenceErrors {
-			c.PrintErrln(c.ErrPrefix(), err.Error())
+			c.PrintErrln(cmd.ErrPrefix(), err.Error())
 			c.PrintErrf("Run '%v --help' for usage.\n", c.CommandPath())
 		}
 		return c, err
--- a/command.go
+++ b/command.go
@@ -603,6 +603,18 @@ func (c *Command) VersionTemplate() string {
 `
 }
 
+// ErrPrefix return error message prefix for the command
+func (c *Command) ErrPrefix() string {
+	if c.errPrefix != "" {
+		return c.errPrefix
+	}
+
+	if c.HasParent() {
+		return c.parent.ErrPrefix()
+	}
+	return "Error:"
+}
+
 func hasNoOptDefVal(name string, fs *flag.FlagSet) bool {
 	flag := fs.Lookup(name)
 	if flag == nil {
@@ -1058,7 +1070,7 @@ func (c *Command) ExecuteC() (cmd *Command, err error) {
 			c = cmd
 		}
 		if !c.SilenceErrors {
-			c.PrintErrln(c.errPrefix, err.Error())
+			c.PrintErrln(c.ErrPrefix(), err.Error())
 			c.PrintErrf("Run '%v --help' for usage.\n", c.CommandPath())
 		}
 		return c, err
@@ -1087,7 +1099,7 @@ func (c *Command) ExecuteC() (cmd *Command, err error) {
 		// If root command has SilenceErrors flagged,
 		// all subcommands should respect it
 		if !cmd.SilenceErrors && !c.SilenceErrors {
-			c.PrintErrln(c.errPrefix, err.Error())
+			c.PrintErrln(c.ErrPrefix(), err.Error())
 		}
 
 		// If root command has SilenceUsage flagged,
--- a/command.go
+++ b/command.go
@@ -1070,7 +1070,7 @@ func (c *Command) ExecuteC() (cmd *Command, err error) {
 			c = cmd
 		}
 		if !c.SilenceErrors {
-			c.PrintErrln(cmd.ErrPrefix(), err.Error())
+			c.PrintErrln(c.ErrPrefix(), err.Error())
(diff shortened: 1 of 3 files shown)


command.go
```diff
@@ -1050,7 +1058,7 @@ func (c *Command) ExecuteC() (cmd *Command, err error) {
 			c = cmd
 		}
 		if !c.SilenceErrors {
-			c.PrintErrln("Error:", err.Error())
+			c.PrintErrln(c.errPrefix, err.Error())
```
marckhouzam, line 1061: Won’t the new variable be empty by default?  We probably need a setter for it and make it default to “Error:”

marckhouzam (changes requested):
Thanks !
 This looks reasonable:
- can you add a getter (you can look at `VersionTemplate`)
- add some short doc
- add a test

5ouma:
May I ask you which function to run to get an error message with a prefix?
I wrote this test code, but it said just an error message without `errPrefix` on the test.
It said an error message with a prefix on CLI though.

```golang
func TestErrPrefix(t *testing.T) {
	rootCmd := &Command{
		Use: "root",
			return errors.New("error")
		},
	}
	rootCmd.SilenceUsage = true
	rootCmd.SetErrPrefix("customized error prefix:")

	checkStringContains(t, fmt.Sprint(rootCmd.Execute()), "customized error prefix: error")
}
```

```
customized error prefix: error
--- FAIL: TestErrPrefix (0.00s)
    command_test.go:77: Expected to contain: 
         customized error prefix: error
        Got:
         error
FAIL
exit status 1
FAIL    github.com/spf13/cobra  0.139s
```

<br />

When I run `executeCommand()` with an undefined flag, it says an error message with a prefix.

```golang
func TestErrPrefix(t *testing.T) {
	rootCmd := &Command{Use: "root", Run: emptyRun}
	rootCmd.SilenceUsage = true
	rootCmd.SetErrPrefix("customized error prefix:")

	output, _ := executeCommand(rootCmd, "--flag")

	checkStringContains(t, output, "customized error prefix: unknown flag: --flag")
}
```

command.go
```diff
@@ -1079,7 +1099,7 @@ func (c *Command) ExecuteC() (cmd *Command, err error) {
 		// If root command has SilenceErrors flagged,
 		// all subcommands should respect it
 		if !cmd.SilenceErrors && !c.SilenceErrors {
-			c.PrintErrln("Error:", err.Error())
+			c.PrintErrln(c.ErrPrefix(), err.Error())
```
marckhouzam, line 1102: I believe this should be `cmd.ErrPrefix()` instead.  The `c` command is always the root command here, so if we want to support setting a different prefix for different command, it is `cmd` that should be used here.

command_test.go
```diff
@@ -1099,6 +1099,18 @@ func TestShorthandVersionTemplate(t *testing.T) {
 	checkStringContains(t, output, "customized version: 1.0.0")
 }
 
+func TestErrPrefix(t *testing.T) {
```
marckhouzam, line 1102: Can you add two more tests:
1. only setting an error prefix on the root, but triggerring it by a sub command to see that it still gets printed
2. setting a different error prefix for a sub command and checking it gets printed when the sub command fails

marckhouzam (changes requested):
This looks pretty good @5ouma.
A couple of things to fix, but after that I think we'll be good.

command_test.go
5ouma: - When I generate an object with `&Command{}`, is `Run: emptyRun` needed?
- What names do you think are suitable for functions 1 and 2?

marckhouzam: > * When I generate an object with `&Command{}`, is `Run: emptyRun` needed?

you probably need it for the deepest child command but it’s ok to have it for every command

> * What names do you think are suitable for functions 1 and 2?

You mean for the sub command?
How about child1 and child2?

5ouma: Thank you!

I mean 2 test functions, `TestRootErrPrefixExecutedOnSubcommand()` and `TestRootAndSubErrPrefix()` temporarily named are.

marckhouzam: Oh that makes more sense😁
Those names are fine

command.go
```diff
@@ -1050,7 +1070,7 @@ func (c *Command) ExecuteC() (cmd *Command, err error) {
 			c = cmd
 		}
 		if !c.SilenceErrors {
-			c.PrintErrln("Error:", err.Error())
+			c.PrintErrln(cmd.ErrPrefix(), err.Error())
```
marckhouzam, line 1073: This line should have stayed with `c.ErrPrefix()`.  Notice that right above, `c = cmd` as long as `cmd != nil`.  So, in case `cmd == nil` we should use `c` here.

marckhouzam (changes requested):
One more comment and we're good

command.go
5ouma: Oops, thank you for describing!

marckhouzam (approved):
LGTM
Thanks @5ouma !

5ouma:
Thank you!!

supriyo-biswas:
Can we please cut over a release? I'd like to use this feature.

Issue #2029, opened 2023-09-20
GenMarkdownCustom adds extra newline when DisableAutoGenTag is true


dramich:
When running `GenMarkdownCustom` with `DisableAutoGenTag=true` an extra newline is added to the end of the file. This can be seen in the md file that is output and tools like pre-commit will also attempt to fix the issue

Expected: No extra newline is added when `DisableAutoGenTag=true`
Cobra version v1.7.0
New line https://github.com/spf13/cobra/blob/v1.7.0/doc/md_docs.go#L108C8

PR #2030, opened 2023-09-20
Move newline to inside the DisableAutoGenTag block

dramich:
Fixes https://github.com/spf13/cobra/issues/2029


nirs (approved):

jpmcb (commented):
This seems fine. Is this more stylistic than anything? I'm surprised there are tools that have an opinion on markdown formatting. Or is it more that some version control tools don't want extra newlines at the end of files?

I'd like to understand more on why this is needed.

doc/md_docs.go
```diff
@@ -105,9 +105,9 @@ func GenMarkdownCustom(cmd *cobra.Command, w io.Writer, linkHandler func(string)
 			link = strings.ReplaceAll(link, " ", "_")
 			buf.WriteString(fmt.Sprintf("* [%s](%s)\t - %s\n", cname, linkHandler(link), child.Short))
 		}
-		buf.WriteString("\n")
 	}
 	if !cmd.DisableAutoGenTag {
+		buf.WriteString("\n")
 		buf.WriteString("###### Auto generated by spf13/cobra on " + time.Now().Format("2-Jan-2006") + "\n")
```
jpmcb, line 111: If we fix not having newlines at the end of the file, we should also remove the newline from the autogenerated tag at the end of the file:

```suggestion
		buf.WriteString("###### Auto generated by spf13/cobra on " + time.Now().Format("2-Jan-2006"))
```

dramich: This fixes adding a second new line at the end of the file so this suggestion isn't needed

dramich:
There are tools that care about whitespace regardless of the file type, as stated in the original issue pre-commit has a `end-of-file-fixer` which looks for things like double returns. https://github.com/pre-commit/pre-commit-hooks?tab=readme-ov-file#end-of-file-fixer

The fix ensures the generated files only have a single return at the end of the file, it does not remove all of them.

If this is deemed not needed feel free to close it or let me know and I can.

nirs:
It would be nice to add a little test to keep the output correct after this change.

dramich:
Added a new assertion to the tests
Test fails before the change with:
```
go test github.com/spf13/cobra/doc
--- FAIL: TestGenMdNoTag (0.00s)
    cmd_test.go:110: Expected to not have suffix:
         "\n\n"
        Got: "## root\n\nRoot short description\n\n### Synopsis\n\nRoot long description\n\n```\nroot [flags]\n```\n\n### Options\n\n```\n  -h, --help              help for root\n  -r, --rootflag string    (default \"two\")\n  -t, --strtwo string     help message for parent flag strtwo (default \"two\")\n```\n\n### SEE ALSO\n\n* [root echo](root_echo.md)\t - Echo anything to the screen\n\n"
```
After fix the test passes. All other tests already pass as they don't have the double line break.

doc/md_docs_test.go
```diff
@@ -40,6 +42,7 @@ func TestGenMdDoc(t *testing.T) {
 	checkStringContains(t, output, echoSubCmd.Short)
 	checkStringOmits(t, output, deprecatedCmd.Short)
 	checkStringContains(t, output, "Options inherited from parent commands")
+	checkStringOmitsSuffix(t, output, doubleLineBreak)
```
nirs, line 45: This is nice, but don't we want the more general check, that the output always ends with a single newline?

PR #2037, opened 2023-10-04
fix documentation of Command.SetOut method

kanaksinghal:
This change fixes a misleading documentation issue for methods: `SetOut`, `SetErr`, `SetIn` & `SetOutput`

In `Command.SetOut`, the statement "If newOut is nil, os.Stdout is used" was misleading as `Command.Print` prints to os.Stderr when `outWriter = nil`

The defaults are based on the getter being used i.e. one of `OutOrStdout` or `OutOrStderr`. i.e. there is no default.

### DIFF
--- a/command.go
+++ b/command.go
@@ -273,7 +273,6 @@ func (c *Command) SetArgs(a []string) {
 }
 
 // SetOutput sets the destination for usage and error messages.
-// If output is nil, os.Stderr is used.
 // Deprecated: Use SetOut and/or SetErr instead
 func (c *Command) SetOutput(output io.Writer) {
 	c.outWriter = output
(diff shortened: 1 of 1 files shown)


marckhouzam:
I agree that there is an inconsistency here.  I also feel the use of `OutOrStderr` by `Print()` is a bug we are stuck with due to backwards-compatibility.

Now, if you don't call `SetOut()` or call it with `nil`, the existing printouts of cobra will default to `stdout`, so the comment feels right to me.  If you add printout using `Pint()` then you justifiably get confused, but there isn't much we can do about it.

So I don't think changing the comment will clarify things.  That is why the comment for `Print()` mention stderr.

kanaksinghal:
I see, thanks for clarifying 🙂 👍

Issue #2039, opened 2023-10-05
PersistentPreRun was only executed if the command has a Run function 


Techming:
I'm not sure if this is intentional but if the command does not have a `Run` function set up, the `PersistentPreRun` function will not be triggered on this command.

I had this issue where I don't need to `Run` anything for a bunch of my subcommands but I still want to be able to show `--version` for the subcommand via `PersistentPreRun`.

If this is intentional, can we update the user guide to reflect this? If not, I can try to fix this issue. Just spent hours debugging until realized might have something to do with Run.

## Steps to reproduce
* Setup the project with `go mod init example-project`
* Setup external modules by `go mod tidy`
* Get the cobra package
* Build and run

* Run `./executable --version` will print out 

```bash
PersistentPreRun in rootcmd runs
inside the flag
``` 

* Run `./executable child --version` will print out 

```bash
PersistentPreRun in rootcmd runs
inside the flag
``` 

* Run `./executable grandchild --version` will print out 

```bash
PersistentPreRun in rootcmd runs
inside the flag
``` 

However, if you comment out any of the Run function from any of the command, the output will be the content from --help

```bash
Hugo child is a very fast static site generator

Usage:

Flags:
  -h, --help   help for child

Global Flags:
  -v, --version   Show version

Additional help topics:
  hugo child grandchild Hugo grand child is a very fast static site generator
```

## Code to reproduce

```go
package main

import (
	"fmt"
	"os"

	"github.com/spf13/cobra"
)

var showVersion bool

var rootCmd = &cobra.Command{
	Use:   "hugo",
	Short: "Hugo is a very fast static site generator",
	Long: `A Fast and Flexible Static Site Generator built with
                love by spf13 and friends in Go.
                Complete documentation is available at https://gohugo.io/documentation/`,
	Run: func(cmd *cobra.Command, args []string) {
		// Do Stuff Here
	},
	PersistentPreRun: func(cmd *cobra.Command, args []string) {
		// Do Stuff Here
		fmt.Println("PersistentPreRun in rootcmd runs")
		if showVersion {
			fmt.Println("Inside the flag")
		}
	},
}

var childCmd = &cobra.Command{
	Use:   "child",
	Short: "Hugo child is a very fast static site generator",
	Run: func(cmd *cobra.Command, args []string) {
		// Do Stuff Here
	},
}

var grandChildCmd = &cobra.Command{
	Use:   "grandchild",
	Short: "Hugo grand child is a very fast static site generator",
[...]

PR #2041, opened 2023-10-07
Add notes to doc about the execution condition of *PreRun and *PostRun functions

Techming:
According to #2039, As per code, if the current command does not have a `Run` function defined, the `PersistentPreRun`, `PreRun`, `PostRun`, or `PersistentPostRun` functions will not be executed at the current command. It's true for `PersistentPreRun` and `PersistentPostRun` functions that child commands inherited from the parent. I added a note to the function comment and user guide to clarify this.


marckhouzam (approved):
Thanks @Techming !

PR #2044, opened 2023-10-12
Feature: allow running persistent run hooks of all parents

vkhoroz:
Currently, only one of the persistent pre-runs and post-runs is executed.
It is always the first one found in the parents chain, starting at this command.
Expected behavior is to execute all parents' persistent pre-runs and post-runs.

Dependent projects implemented various workarounds for this:
- manually building persistent hook chains (in every hook).
- applying some kind of monkey-patching on top of Cobra.

This change eliminates the necessity for such workarounds
by allowing to set a global variable EnableTraverseRunHooks.

Tickets:
- https://github.com/spf13/cobra/issues/216
- https://github.com/spf13/cobra/issues/252
- https://github.com/spf13/cobra/issues/1198

Signed-off-by: Volodymyr Khoroz <volodymyr.khoroz@foundries.io>

### DIFF
--- a/cobra.go
+++ b/cobra.go
@@ -43,9 +43,10 @@ var initializers []func()
 var finalizers []func()
 
 const (
-	defaultPrefixMatching  = false
-	defaultCommandSorting  = true
-	defaultCaseInsensitive = false
+	defaultPrefixMatching   = false
+	defaultCommandSorting   = true
+	defaultCaseInsensitive  = false
+	defaultTraverseRunHooks = false
 )
 
 // EnablePrefixMatching allows setting automatic prefix matching. Automatic prefix matching can be a dangerous thing
@@ -59,6 +60,10 @@ var EnableCommandSorting = defaultCommandSorting
 
 // EnableCaseInsensitive allows case-insensitive commands names. (case sensitive by default)
 var EnableCaseInsensitive = defaultCaseInsensitive
+
+// EnableTraverseRunHooks executes persistent pre-run and post-run hooks from all parents.
+// By default this is disabled, which means only the first run hook to be found is executed.
+var EnableTraverseRunHooks = defaultTraverseRunHooks
 
 // MousetrapHelpText enables an information splash screen on Windows
 // if the CLI is started from explorer.exe.
--- a/command.go
+++ b/command.go
@@ -934,15 +934,31 @@ func (c *Command) execute(a []string) (err error) {
 		return err
 	}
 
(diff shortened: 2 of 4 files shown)


vkhoroz:
**Note to mainteiners:**

There were quite a few attempts by the community to resolve this issue, as many see it as a bug.
However, I the mainteiners position to keep the implementation simple and stable also makes sense.

In my view, previous attempts to resolve this had various problems:
- introducing a breaking change which changes behavior.
- introducing the TraverseRunHooks at command level which is inconvenient to use.
- implementing this as a complex logic which changes a lot of code paths.

I tried to correct all of the above flaws introducing a change which is:
- Minimal by design, only modifying a couple of lines of code.
- Fully backward compatible, breaking no package users.
- Introduces the new functionality in a convenient way by globally swithing the package behavior for those who want it.
- Attempts to satisfy both sides in a way which does not behavior fracturing across separate commands.
- Allow future extensions on per-command level _if mainteiners decide later to do that_.

Please, accept the PR, as I believe the community would benefit from it.

marckhouzam:
Thanks @vkhoroz !  I feel this approach is reasonable.  Nicely done!
Let me play around with it.

cc @jpmcb This is something the community has been complaining about.

cobra.go
```diff
@@ -60,6 +61,10 @@ var EnableCommandSorting = defaultCommandSorting
 // EnableCaseInsensitive allows case-insensitive commands names. (case sensitive by default)
 var EnableCaseInsensitive = defaultCaseInsensitive
 
+// EnableTraverseRunHooks executes persistent pre-run and post-run hooks from all parents.
+// By default, only the first run hook to be found is executed.
```
marckhouzam, line 65: Can I suggest to change the second comment line to be:
// By default this is disabled, which means only the first run hook to be found is executed.

marckhouzam (approved):
This looks great!

Could you add a note in the documentation about this new option?
The section is here: https://github.com/spf13/cobra/blob/main/site/content/user_guide.md#prerun-and-postrun-hooks

cobra.go
vkhoroz: Yes, sure.

I changed this and added some documentation in the force-push.

marckhouzam:
The linter is failing because the version of golanci-lint 1.55.0 reports new error.
I've opened #2052 to fix this.
Since this is not a problem with this PR, I will attempt to merge anyway.

PR #2049, opened 2023-10-18
Powershell completion alias handling

bartoncasey:
Currently the Powershell completion generator registers the completion for the default command name. Any aliases of the command will not have completion enabled.  (e.g. `k -> kubectl`)

This change will include code to register the completer for all current aliases of the command in addition to the command itself.

### DIFF
--- a/powershell_completions.go
+++ b/powershell_completions.go
@@ -279,7 +279,12 @@ filter __%[1]s_escapeStringWithSpecialChars {
     }
 }
 
-Register-ArgumentCompleter -CommandName '%[1]s' -ScriptBlock ${__%[2]sCompleterBlock}
+# Register the completer for the command and for all aliases of the command
+'%[1]s', (Get-Alias -Definition '%[1]s' -ErrorAction Ignore).Name | ForEach-Object {
+    if ($_) {
+        Register-ArgumentCompleter -CommandName $_ -ScriptBlock ${__%[2]sCompleterBlock}
+    }
+}
 `, name, nameForVar, compCmd,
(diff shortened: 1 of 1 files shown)


marckhouzam:
Thanks @bartoncasey.
There is #1977 which is trying to do this.  I don't have much expertise with PowerShell, so maybe you can help figure out if #1977 solves things for you?

marckhouzam:
> Set-Alias -Name k -Value "C:\Program Files\Docker\Docker\resources\bin\kubectl.exe"

I've tested this PR and although it is working very nicely for normal aliases, I don't believe it works for the above form.

So now I'm going back and forth between this PR and #1977 hoping to find the sweet spot.

marckhouzam:
After spending quite a bit of time trying things out, I think this solution is the simplest and I believe can be made to cover every case.  I was thinking of these cases:
1. any alias pointing to kubectl or kubectl.exe
1. any alias pointing to /<path>/kubectl or /<path>/kubectl.exe

I think we can achieve this with a slight modification do have instead:
```
# Register the completer for the command and for all aliases of the command
'%[1]s', (Get-Alias -Definition %[1]s,*/%[1]s,%[1]s.exe,*/%[1]s.exe -ErrorAction Ignore).Name | ForEach-Object {
    if ($_) {
        Register-ArgumentCompleter -CommandName $_ -ScriptBlock ${__%[2]sCompleterBlock}
    }
}
```
This translates to something like
`Get-Alias -Definition kubectl,*/kubectl,kubectl.exe,*/kubectl.exe`
which covers the cases listed above.

@bartoncasey What do you think?

PR #2053, opened 2023-10-23
Add tests for flag completion registration

marckhouzam:
Different problems have been reported about flag completion registration. 
These two tests are the cases that were not being verified but had been mentioned as problematic.

The test `TestFlagCompletionConcurrentRegistration` reproduces the `"concurrent map writes"` errors if concurrency is not properly handed, as reported by #1320.
I have verified this test does fail by running it on code before the fix from #1438.

The test `TestFlagCompletionForPersistentFlagsCalledFromSubCmd` makes sure persistent flags get completed.
This is based on problems mentions here:
- https://github.com/spf13/cobra/pull/1438#issuecomment-872928669
- https://github.com/spf13/cobra/pull/2012#pullrequestreview-1627336420

I have verified this test does fail based on the interim commit that was reported to have this issue in this [comment](https://github.com/spf13/cobra/pull/1438#issuecomment-872928669).

### DIFF
--- a/completions_test.go
+++ b/completions_test.go
@@ -17,7 +17,9 @@ package cobra
 import (
 	"bytes"
 	"context"
+	"fmt"
 	"strings"
+	"sync"
 	"testing"
 )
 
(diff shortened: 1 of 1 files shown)


Luap99 (approved):
LGTM

jpmcb (approved):
Thanks for the well documented tests!! 👏🏼

PR #2061, opened 2023-10-27
Don't complete --help flag when flag parsing disabled

### LINKED ISSUE #2060: Completion should not complete the --help/-h and --version/-v flags if DisableFlagParsing==true
When a command sets `DisableFlagParsing = true` it requests the responsibility of doing all the flag parsing.
If there are flags for that command, Cobra may not even know about them.
I've seen this used to provide plugins to a Cobra-based CLI.  The plugins must handle their own flags as the original CLI knows nothing about such flags.

The problem is that with #1813 the shell completion logic completes the `--help/-h` and `--version/-v` for all commands, even those that do their own flag handling (i.e., `DisableFlagParsing == true`).  I've seen this cause the `--help/-h` be completed twice for plugins: one time from cobra and one time from the plugin.

I therefore feel that Cobra should not complete the `--help/-h` and `--version/-v` if `DisableFlagParsing==true`.

Somewhat related to #1328 but for completion.

marckhouzam:
Fixes #2060

When a command sets `DisableFlagParsing = true` it requests the responsibility of doing all the flag parsing. Therefore even the `--help/-f/--version/-v` flags should not be automatically completed by Cobra in such a case.

Without this change the `--help/-h/--version/-v` flags can end up being completed twice for plugins: one time from cobra and one time from the plugin (which has set `DisableFlagParsing = true`).

### DIFF
--- a/completions.go
+++ b/completions.go
@@ -302,9 +302,13 @@ func (c *Command) getCompletions(args []string) (*Command, []string, ShellCompDi
 
 	// These flags are normally added when `execute()` is called on `finalCmd`,
 	// however, when doing completion, we don't call `finalCmd.execute()`.
-	// Let's add the --help and --version flag ourselves.
-	finalCmd.InitDefaultHelpFlag()
-	finalCmd.InitDefaultVersionFlag()
+	// Let's add the --help and --version flag ourselves but only if the finalCmd
+	// has not disabled flag parsing; if flag parsing is disabled, it is up to the
+	// finalCmd itself to handle the completion of *all* flags.
+	if !finalCmd.DisableFlagParsing {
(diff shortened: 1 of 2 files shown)


marckhouzam:
@jpmcb This is something I noticed in the `tanzu` CLI and its use of plugins.  The same happens for `kubectl` and `helm` plugins.  

For example, notice the double `--help/-h` flag completions:

```
$ kubectl krew -<TAB>
--help  --help  -h  -h  -- help for krew
--v     -v              -- number for the log level verbosity
```

I've tested that this fix works for both `kubectl` and `tanzu` plugins.

jpmcb (approved):
LGTM! Nice one @marckhouzam !!

PR #2063, opened 2023-11-02
Improve API to get flag completion function

marckhouzam:
With #1943 we aimed to allow programs to have access to flag completion functions.
There were two issues with this change:
1. it was not handling persistent flags (I missed that in the review)
2. it was not similar to the existing `c.RegisterFlagCompletionFunc()` because it provided a global function, which is not as future proof.

This PR proposes a simpler API that matches the `c.RegisterFlagCompletionFunc()` API.  By removing the global function `GetFlagCompletion()` we are more future proof if we ever move from a global map of flag completion functions to something associated with the command.

The commit also makes this API work with persistent flags by using `c.Flag(flagName)` instead of `c.Flags().Lookup(flagName)`.

The commit also adds unit tests.

cc @jpmcb @avirtopeanu-ionos @maxlandon

### DIFF
--- a/completions.go
+++ b/completions.go
@@ -145,23 +145,18 @@ func (c *Command) RegisterFlagCompletionFunc(flagName string, f func(cmd *Comman
 	return nil
 }
 
-// GetFlagCompletion returns the completion function for the given flag, if available.
-func GetFlagCompletion(flag *pflag.Flag) (func(cmd *Command, args []string, toComplete string) ([]string, ShellCompDirective), bool) {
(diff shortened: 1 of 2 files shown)


marckhouzam:
@maxlandon do you see a problem with this API if we do a release with this PR but not #2012?

maxlandon:
Let me be wise and ask @jpmcb for this view on this.
I'd prefer to accept the judgement of a third mind here, whatever his decision/opinion is.

jpmcb (approved):
This looks good - I could go either way on getting this in with a release and not #2012. But I'm happy to also look at that and help get that through.

Issue #1825, opened 2022-10-12
Issue passing local flag to subcommand


IvanRibakov:
I have a root command with persistent flag and a subcommand with a local flag. When executing subcommand, I seem to be unable to pass the local flag. No matter how I do it, I get `Error: unknown flag:`.

Here's the minimal example that reproduces my situation:

```
package main

import (
	"fmt"
	"github.com/spf13/cobra"
)

var (
	testFlag string
	subFlag bool
	rootCmd = &cobra.Command{
		Use:   "root",
	}
)

func init() {
	rootCmd.PersistentFlags().StringVar(&testFlag, "rootFlag", "", "")

	subCmd := &cobra.Command{
		Use:   "subcommand",
	}
	subCmd.LocalFlags().BoolVar(&subFlag, "subFlag", false, "")
	subCmd.RunE = func(cmd *cobra.Command, args []string) error {
		fmt.Println(testFlag)
		fmt.Println(subFlag)
		return nil
	}

	rootCmd.AddCommand(subCmd)
}

func main() {
	rootCmd.Execute()
}
```

All of the above invocations produce an error `Error: unknown flag: --subFlag`:

```
$ go run . subcommand --rootFlag blah --subFlag
$ go run . subcommand --subFlag --rootFlag blah
$ go run . --rootFlag blah subcommand --subFlag
```

Can someone please explain me what am I missing here?

marckhouzam:
Hi @IvanRibakov. 
Try using `subCmd.Flags().BoolVar()` instead of `subCmd.LocalFlags()`.

I admit that I don't fully understand all these flag functions, but I use code I've seen work before.

IvanRibakov:
Hi @marckhouzam,

`subCmd.Flags()` indeed works, but it sounds like a workaround to me that limits using flags with the same short version character on different levels.

I'd still argue that original problem is a bug, or at a very minimum, a gap in documentation.

johnSchnake:
The docs use the `Flags()` function so they seem accurate; I am also not sure the fully history of the function origins but if you would like to add a note into the docs to disambiguate this I dont think there is any problem with that.

https://github.com/spf13/cobra/blob/main/user_guide.md#local-flags

niamster:
> The docs use the `Flags()` function so they seem accurate; I am also not sure the fully history of the function origins but if you would like to add a note into the docs to disambiguate this I dont think there is any problem with that.
> 
[...]

PR #2064, opened 2023-11-04
flags: clarify documentation that LocalFlags related function do not modify the state

niamster:
This addresses https://github.com/spf13/cobra/issues/1825

A user might confuse `LocalFlags` (or `InheritedFlags`) as a way to set local command flags in the same way as it is done via `Flags` (or `PersistentFlags`).
It appears that  the purpose of`LocalFlags` is to return a `FlagSet` for further examination.
Technically it's possible to update some flags returned by `LocalFlags` as `FlagSet` references original flags using pointers.

This PR attempts to clarify the documentation. 
Also this PR removes `lflags` and `iflags` from the `Command` structure to avoid confusion as they are not used beyond the scope of `LocalFlags` or `InheritedFlags` methods. Just recreate the instance on every call to relevant function.

NOTE: it's not possible to make `LocalFlags` return a value (rather than a pointer) as it breaks some APIs.

NOTE: local tests pass
```
>make all > /dev/null 2>&1 && echo OK
OK
```


command.go
```diff
@@ -1666,58 +1663,57 @@ func (c *Command) LocalNonPersistentFlags() *flag.FlagSet {
 }
 
 // LocalFlags returns the local FlagSet specifically set in the current command.
+// This function does not modify the flags of the current command, it's purpose is to return the current state.
 func (c *Command) LocalFlags() *flag.FlagSet {
 	c.mergePersistentFlags()
 
-	if c.lflags == nil {
```
marckhouzam, line 1672: The use of this member variable is for efficiency. We probably don’t want to remove it

niamster: Hi Marc, thank you for taking a look!
Does this optimization really matter? I can't imagine any code calling `LocalFlags` in a loop or something like that.
It seems to me that these fields are confusing as they appear to represent command state rather than to cache internal state of some functions.
If you prefer to keep this optimisation, do you agree to rename it `lflags` to something like `lflagsTmpState`?

niamster: The `lflags` field was introduced https://github.com/spf13/cobra/commit/5c9146990baf14f44c4c470befa56c7ee5b62130 and its initial purpose is not clear to me.

niamster: Hi @marckhouzam do you really want me to keep these fields? If so are you OK with renaming them as I suggested before? Thanks!

niamster: Hi @marckhouzam, will you have time to look at this PR? What would be the best way to move forward? Thanks!

niamster: Hello @marckhouzam is there any chance you can find some time to review this PR (clarify the documentation of the `LocalFlags` function? Thanks!

marckhouzam: @niamster I found this commit which introduced `iflags` and it seems to indicate there is an improvement in performance: https://github.com/spf13/cobra/commit/6202b5942b8f31501d14d8cb192790d8a10302f4

Although your code changes look safe, I prefer not to change things without a good reason;  cobra is being used by so many programs that we need to prioritize stability.

For example, what if a program stored the pointer returned by `InheritedFlags()` for some reason.  With this PR a second call to `InheritedFlags()` would return a different pointer.  Would it matter?  Probably not, but I cannot be sure.  So, let's not invite some unexpected problem 😄 

If you want to limit the PR to the comment changes, that would be fine.

niamster: Hi Marc, thank you for finding some time to look into this issue. OK, yes I understand your concern. I'll try to find some time next week to limit this PR to comment changes only 😄

niamster: Hi @marckhouzam I'm sorry for the delay. I've updated this PR to address the documentation only, can you please take a look? Thanks!

marckhouzam (approved):
Thanks!

PR #2080, opened 2023-11-26
fix: reset flags' value to default if flags have been parsed

### LINKED ISSUE #2079: [Bug] SetArgs doesn't reset flag value
See following test code. Call `SetArgs` won't reset flag value, `Execute()` also won't reset flag value too.
```go
func TestSetArgs(t *testing.T) {
	getCMD := func() *Command {
		cmd := &Command{
			Use: "testcmd",
				a, _ := cmd.Flags().GetBool("a")
				b, _ := cmd.Flags().GetBool("b")
				c, _ := cmd.Flags().GetBool("c")
				switch {
				case a && b, a && c, b && c:
					return fmt.Errorf("a,b,c only one could be true")
				}
				return nil
			},
		}
		f := cmd.Flags()
		f.BoolP("a", "a", false, "a,b,c only one could be true")
		f.BoolP("b", "b", false, "a,b,c only one could be true")
		f.Bool("c", false, "a,b,c only one could be true")
		return cmd
	}

	cmd := getCMD()

	// step 1
	cmd.SetArgs([]string{
		"--a=true",
	})
	assertNoErr(t, cmd.Execute()) // PASS

	// step 2
	cmd.SetArgs([]string{
		"--b=true",
	})
	t.Log(cmd.Flags().Changed("a")) // true
	assertNoErr(t, cmd.Execute()) // FAIL, output error msg is "a,b,c only one could be true"

	// step 3
	cmd.SetArgs([]string{
		"--c=true",
	})
	assertNoErr(t, cmd.Execute()) // FAIL, output error msg is "a,b,c only one could be true"

	// step 4
	cmd.SetArgs([]string{
		"--a=false",
		"--b=false",
		"--c=true",
	})
	assertNoErr(t, cmd.Execute()) // PASS
}
```

levisyin:
close #2079

### DIFF
--- a/command.go
+++ b/command.go
@@ -1833,7 +1833,9 @@ func (c *Command) ParseFlags(args []string) error {
 	// if flag has been parsed, we should reset flags' value to default
 	if c.Flags().Parsed() {
 		c.Flags().Visit(func(f *flag.Flag) {
-			f.Value.Set(f.DefValue)
+			if err := f.Value.Set(f.DefValue); err != nil {
+				c.PrintErrf("reset argument[%s] value error %v", f.Name, err)
+			}
 		})
 	}
 	err := c.Flags().Parse(args)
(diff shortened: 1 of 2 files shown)


marckhouzam:
To avoid any risk backwards-compatibility problems in existing projects, we will not fix this.
Users of Cobra can fix this themselves in their own code.
See https://github.com/spf13/cobra/issues/2079#issuecomment-1867991505

PR #2085, opened 2023-12-04
Add a Data map[string]any to store arbitrary command data

maxlandon:


Luap99:
Shouldn't this be just a `any`? Making this `map[string]any` is already somewhat limiting depending on the use case. If this is just `any` you can still put in a `map[string]any` with that or if you do not need it you may directly store a value without having to go through the string map lookup.

maxlandon:
Hello @Luap99, thanks for the input.

If several plugins/paths need different objects stored with the command, who decides who has precedence ? 

If plugin A wants a struct on monday, and then on tuesday plugin B wants another, then it can only delete Plugin A struct to replace it. 

Now plugin A can't work, because plugin B replaced the contents altogether, and it has no way of communicating it to plugin A.

So I think there is no way around a map here...

Luap99:
Yeah 

> Hello @Luap99, thanks for the input.
> 
> If several plugins/paths need different objects stored with the command, who decides who has precedence ?
> 
> If plugin A wants a struct on monday, and then on tuesday plugin B wants another, then it can only delete Plugin A struct to replace it.
> 
> Now plugin A can't work, because plugin B replaced the contents altogether, and it has no way of communicating it to plugin A.
> 
> So I think there is no way around a map here...

Ok yes that sounds like a valid reason for a map. I am not really sure how to common several plugins that need to store data attached to the command struct are but I guess it is better to be safe and use the map.

Maybe rethinking, is it really necessary to store the data inside the cobra command? Couldn't the "plugin" just define a wrapper type like
```
type PluginCommand struct {
    *cobra.Command
    data ...
}
```
and just accept this type in the plugin APIs?

command.go
```diff
@@ -111,6 +111,11 @@ type Command struct {
 	// command does not define one.
 	Version string
 
+	// Data are key/value pairs of arbitrary types that can be used by applications or "plugin" libraries
+	// that wish to store and use data associated with a given command. This data will thus share the same
+	// lifetime as the command itself.
+	Data map[string]any
```
marckhouzam, line 117: `any` was introduced in go 1.18.
Cobra still supports go 1.15.  I know these are very old versions, but I am the opinion that if we don't need to drop support, why do it.

So, can we change `any` to `interface{}`?

marckhouzam (commented):
The extra field seems reasonable to me.
Please see one comment which will fix CI.

maxlandon:
Hello,
Coming back to this small PR which is super important to all projects using cobra in a closed-loop console, with added completion engines. The possibility of such engines (or any other plugin using a command) to store data pertaining to each command will be very useful for correct garbage collection.

Thanks @marckhouzam .

maxlandon:
Hello @marckhouzam,

Sorry to ping you here and there since 3 days, but I'm trying to push those little enhancements related to plugins and completions being merged, so that cobra commands can be totally and cleanly used in closed-loop console applications.
It already works great, but if it can be perfect (for example, related to garbage-collecting), I would appreciate.

This precise PR is also actually just a line of code !

Thanks for any time you'll find for this !

PR #1944, opened 2023-04-04
Internationalization (I18N)

Goutte:
> **Work In Progress**
> Incomplete but good enough for my own use.

## How

- Detects language from `ENV`  (`LANGUAGE` > `LC_ALL` > `LANG`)
- Loads embedded `toml` translation files
- Translates

### Pluralization

Output in English of some flag errors has changed a little, since pluralization is dynamic.
Eg: `accepts at most 2 args, received 3` and `accepts at most 1 arg, received 3`
If that's not something you think we should have, and should stick with `arg(s)` to ease parsing for example (but… regex), I can revert this.

### `go 1.16`

`go 1.16` is required for `go-i18n v2`.  I enjoy `goi18n extract` and `merge` too much to bother downgrading to `v1`.

### Translations flow documentation

> Not documented yet.
> I'll start sketching it here…

The translation flow is as follows:

1. Create a new `Message` in a wrapper function in `localizer.go`
2. Call that function somewhere in cobra
3. Run `make i18n_extract`
4. Edit the newly generated `translations/translate.*.toml` files
5. Run `make i18n_merge`, it will write to the `translations/active.*.toml` files
6. Delete the now useless `translations/translate.*.toml` files

### Not all strings translated

I focused on user-facing strings because those were the most useful to me.
I might do some more but probably won't do them all.

## Bouerk

`go.sum` went bananas, and I'm not sure what happened.   Help appreciated.

Issue #1134, opened 2020-06-11
Multi-language support for CLIs?


thomasgloe:
I would like to support multiple languages for my CLI using cobra. Implementation for commands is no problem, but is it correct that there is currently no support for the text output generated by cobra itself (e.g., "Usage", "Flags", "Use "mycmd [command] --help" for more information about a command.")?

BunnyBrewery:
Are you talking about if there is multi-language support for default help message in Cobra?

thomasgloe:
Yes and I've already checked the source code, where strings are encapsulated in the UsageTemplate. The way to go seems to change the usage template with SetUsageTemplate.

If I have enough time, would it be of interest to include a small example in the docs?

jharshman:
@thomasgloe I'd be interested to see a PR for this if you wanted to take a shot at it.

hitzhangjie:
`SetsageTemplate` may only affects the template. If we want to support multiple languages, we may consider the description of commands and flags.

I use `go-i18n` to support multiple languages in my cobra cli.

Goutte:
I've reviewed cobra these past days (using it for [git spend](https://github.com/Goutte/git-spend)), and I've come to the conclusion that cobra itself should have some form of _i18n_ for default content.

I'm glad you're not against it for arcane reasons :) – it's just work, and this I understand.

There are some decisions that are best discussed beforehand, though.

## Choosing a translation file format

I'm partial to `toml` in our case, since we won't really need the tree structure of `yaml`.
The other formats are just not human-friendly enough, and even though there are really nice GUIs for translation, I prefer keeping the translation files as readable as possible.

## Embedding `toml` translation files

This appears to be the _easy_ way of handling i18n.

Embedding (`go:embed`) _ALL_ translations may add some kilobytes to cobra.
I'm okay with it, personally, but some of y'all may know ways _(I don't)_ to distribute "lightweight" versions of cobra (with only the _english_ file), along with the fully translated one, for people who desperately need lightweight.

Embedding also kind of slightly breaks the philosophy of package managers (`.deb`), since each cobra-based cli app will end up with its own translation files for the internals of cobra, whereas they could be shared, ideally.   This is tricky.

## `goi18n extract` or not?

Usage of `goi18n extract` requires writing the translation fetching code very verbosely,
[...]

PR #2090, opened 2023-12-14
Implement localization using gettext files — I18N — L10N

Goutte:
See #1134 

- Recipe to extract new translations from the Go code: `make i18n_extract`
- Embedded `MO` files
- Detect language from environment variables
- Some strings were pluralized

---

Replaces #1944


Goutte:
I know it's a lot of added lines, but more than half are from the `POT` and `PO` files.  I added the _french_ translation as an example, for good measure.

marckhouzam:
Thanks @Goutte!

A couple of things to help get this reviewed:
1- can resolve the conflicts with main
2- can you add in the description how to test this (including the best way to install `xgotext`)

marckhouzam:
@Goutte could you explain what the process would be, if we adopted this PR, for a cobra contributor to add a new string printout?  It is not clear to me when `make i18n_extract` should be run and which `*.po`/`default.pot`/`*.mo` files need to be modified by hand and when, etc.

At this point, clarifying these steps in the PR description is sufficient, but if the review goes well, we'll need those details in some cobra documentation.

Also, I'm also interested in understanding how a project using Cobra would proceed to add translations, or is that completely up to the project and this PR does not help?

marckhouzam:
> @Goutte could you explain what the process would be, if we adopted this PR, for a cobra contributor to add a new string printout? It is not clear to me when `make i18n_extract` should be run and which `*.po`/`default.pot`/`*.mo` files need to be modified by hand and when, etc.

I see this already in the docs.  Nice work.
Can you clarify what you mean by "Make sure your software has also updated the `MO` files"?
Do we absolutely need to use a special editor like poedit or can this be done from the command-line?

args.go
```diff
@@ -33,15 +34,15 @@ func legacyArgs(cmd *Command, args []string) error {
 
 	// root command with subcommands, do subcommand checking.
 	if !cmd.HasParent() && len(args) > 0 {
-		return fmt.Errorf("unknown command %q for %q%s", args[0], cmd.CommandPath(), cmd.findSuggestions(args[0]))
+		return fmt.Errorf(gotext.Get("LegacyArgsValidationError"), args[0], cmd.CommandPath(), cmd.findSuggestions(args[0]))
```
marckhouzam, line 37: How about leaving the original english string as the index for all the `gotext.Get()` calls?
It would make reading the code much easier and it would make working with the `*.po` files easier as the original text would be right there as the index.

If the `%` formatting gives a problem, maybe we can replace it with `%%` in the index?  It is not ideal, but it would be manageable.  So, for this line here we would instead use (unless there is a better way?)
```
return fmt.Errorf(gotext.Get("unknown command %%q for %%q%%s"), args[0], cmd.CommandPath(), cmd.findSuggestions(args[0]))

```

args.go
```diff
@@ -16,6 +16,7 @@ package cobra
 
 import (
 	"fmt"
+	"github.com/leonelquinteros/gotext"
```
marckhouzam, line 19: We could import this as `i18n`, so that instead of calling `gotext.Get()` we would call `i18n.Get()`.
This may be clearer when browsing the code?

marckhouzam (commented):
Just started looking at this but I'm staring to get it.
Seems good at first impression.

But, I feel we absolutely need a way to generate the `*.mo` files from the command-line, through the makefile in fact.

args.go
marckhouzam: I’ve looked I’d gotext in more detail, and the `%` form will work just fine as long as the parameters are in the `gotext.Get()` call instead of outside.

marckhouzam:
Something to discuss later, I just don’t want to forget. 

We will need a way for a program using cobra to disable these translations. For example, if a `mycli` project does not translate its own strings, it might not want the cobra strings to get translated as it would give an inconsistent user experience.

So we probably want the project to explicitly have turn on translations through a global setting.

Goutte:
Thanks for the comprehensive review, @marckhouzam !

> Also, I'm also interested in understanding how a project using Cobra would proceed to add translations, or is that completely up to the project and this PR does not help?

This PR does not help devs to implement their I18N.
They are free to choose whatever solution they like.

We could perhaps provide some doc/sugar on how to use the same translation toolkit as cobra, since the relevant libraries are already included ?

> Can you clarify what you mean by "Make sure your software has also updated the MO files"? Do we absolutely need to use a special editor like poedit or can this be done from the command-line?

Good question.  Since gettext is mature, I'm pretty confident there are CLI tools out there handling this. _(or we'll have to make one, preferably in Golang, using cobra of course)_

I'd like the MO to be generated by CI as well, if only to prevent contributors to inject nasty things in the binary files.

I'll look into this.

> So we probably want the project to explicitly have turn on translations through a global setting.

Very good point.  Partially translated apps are awkward, which is precisely why I wanted cobra to support i18n.  I'm not sure right now how to proceed to add a global setting to cobra, but I'll look into it and come back to ask for help if I don't find how.  Do give tips if you'd prefer it to be done in some specific way :)

args.go
Goutte: This is the crux of the matter.

I've done a whole bunch of implementations of translations in the past, and every time I try to use plain english instead of keys I end up either regretting it or refactoring heavily to keys.

Here's some food for thought when using raw english as keys:

- cosmetically changing the english string (typo, whitespace, etc.) invalidates all the existing translations (possibly hundreds of files need to be updated, for nothing)
- compels usage of `context` shenanigans for strings that are the same in english in different contexts but different in other languages
- too many gettext parsers out there choke on nasty translation ids like `the "quoted q'tara" <tag>`
- sometimes it creates a bias towards english in the structure of strings, that need to be fixed by translators, straight in the code, once again invalidating all existing translations

-----

I'm very aware that using keys makes the code harder to read and understand.  This is mitigated a little by a careful choice of the wording of the key.

I had to pick one way or the other ; it was not an easy choice, but it's one I made many times and I decided to go with hindsight from past experiences.

I'm not adamant on this, quite the contrary.
I listed some of the key points of my decision above (probably forgot some) ; I'll let y'all be the final judges.   Good luck !

marckhouzam:
To allow a project to enable translations I’m thinking of two approaches you can take:
1. A global setting like cobra has for other features:  https://github.com/spf13/cobra/blob/41227856cd731aeeeb5a2f7203b3a4c16b81fc67/cobra.go#L52-L66
2. A field in the root command like we have for tuning shell completions https://github.com/spf13/cobra/blob/41227856cd731aeeeb5a2f7203b3a4c16b81fc67/command.go#L203

if you think different settings might be useful eventually, I would recommend using a struct so we could grow it overtime

marckhouzam:
> This PR does not help devs to implement their I18N. They are free to choose whatever solution they like.
>
> We could perhaps provide some doc/sugar on how to use the same translation toolkit as cobra, since the relevant libraries are already included ?

Thanks for clarifying, it makes sense. Let’s leave this for a potential separate PR and focus on cobra’s strings .

Luap99:
I think cobra should have a built tag to disable/enable translations. Right now it looks like all translations are embedded in the final binary. That will lead to unnecessary bloat for users who do not want to use it so a build tag would be the better option IMO.

And these translations should be opt in not opt out IMO.

marckhouzam:
> I think cobra should have a built tag to disable/enable translations IMO.

Good idea +1

> And these translations should be opt in not opt out IMO.

Agreed 👍

Goutte:
> I think cobra should have a built tag to disable/enable translations. Right now it looks like all translations are embedded in the final binary. That will lead to unnecessary bloat for users who do not want to use it so a build tag would be the better option IMO.
> 
> And these translations should be opt in not opt out IMO.

That's a good idea.

I think I managed to do this using the "locales" build tag.

Now I need to tweak the test suite to use that build tag,
or perhaps even make two runs, with and without.

Anyhow, for now the tests are broken until I figure this out.

localizer.go
```diff
@@ -0,0 +1,138 @@
+package cobra
```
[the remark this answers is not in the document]
Goutte: I had noticed the `make lint` complaining about _something_ "not goimport'ed", I had no idea it was the copyright notice.  Thanks !

Goutte:
> Any update on this?

Nope ; I was hospitalized and totally forgot about this.

The tests pass, but the build without the locales bloat has not been tested, although it looks like embedding of the MO files has been removed if the `locales` tag is unset.

I'll do the rebase to resolve the conflicts, but sadly Time has caught up with me and I won't be able to spend a lot of it pursuing this.

Goutte:
## Summary post-rebase

- I've dropped the I18N of the default usage template string (I've localized the default usage function instead -- much easier), because the new signature of the (new?) template func won't accept my ~~hack~~ ad hoc `CommandUsageTemplateData` as `interface{}`.
- Fixed a small typo "topcis" → "topics" along the way
- I can't find any tests checking the french translation, and I'm not sure how to add one
- Existing tests should pass

args.go
decentral1se: Yes, AFAIU, `gotext` only supports matching on `gotext.Get` at this time.

args.go
decentral1se: I support the key approach! The problems listed here are sadly all too common without using keys.

decentral1se: I've been working a bit more with keys/no keys approach and I do also have some more sympathy for the non-key'd approach now that I see the level of inconvenience for the source developers. It's not much but is a barrier. Tbh if this is a controversial change, I'd rather drop it and proceed with getting the PR out than let in languish in debate.

decentral1se:
To be clear, I'm willing to pick this work up and try to get it over the line if someone from `cobra` core has cycles to map out how to finish it. I think @Goutte has made an amazing start and a great change here and it would be amazing to get it out into the world. I'm running into issues myself (https://git.coopcloud.tech/toolshed/abra/issues/629) with `cobra` i18n not being supported...

marckhouzam:
I would love to get internationalization added to cobra. If you want to take this over @decentral1se , I’ll do my best to carve out some time to help you along.

decentral1se:
@marckhouzam amazing, tysm! I guess the easiest thing to do is just this: you draw up a checklist of things you want to see resolved / done from this PR and then i'll try to get it done? There's a few open questions left hanging in the comments above, maybe you want to make some final decisions and call it? I don't mind following a check-list of things we can implement now which matches your capacity.

PR #2094, opened 2023-12-28
Add test for func in cobra.go

korovindenis:
Add test for:

- rpad(s string, padding int) string
- ld(s, t string, ignoreCase bool) int
- stringInSlice(a string, list []string) bool

in cobra.go

### DIFF
--- a/cobra_test.go
+++ b/cobra_test.go
@@ -133,8 +133,8 @@ func TestStringInSlice(t *testing.T) {
 			expected: false,
 		},
 		{
-			name:     "String not in slice (case-insensitive)",
-			a:        "PEAR",
+			name:     "String in slice (case-insensitive)",
+			a:        "APPLE",
 			list:     []string{"orange", "banana", "apple", "grape"},
 			expected: false,
 		},
--- a/cobra_test.go
+++ b/cobra_test.go
@@ -40,3 +40,185 @@ func TestAddTemplateFunctions(t *testing.T) {
 		t.Errorf("Expected UsageString: %v\nGot: %v", expected, got)
 	}
 }
+
+func TestLevenshteinDistance(t *testing.T) {
+	tests := []struct {
+		name       string
+		s          string
+		t          string
+		ignoreCase bool
+		expected   int
+	}{
+		{
+			name:       "Equal strings (case-sensitive)",
+			s:          "hello",
+			t:          "hello",
+			ignoreCase: false,
+			expected:   0,
+		},
+		{
+			name:       "Equal strings (case-insensitive)",
+			s:          "Hello",
+			t:          "hello",
+			ignoreCase: true,
+			expected:   0,
+		},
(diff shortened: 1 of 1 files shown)


cobra_test.go
```diff
@@ -40,3 +40,185 @@ func TestAddTemplateFunctions(t *testing.T) {
...
+		},
+		{
+			name:       "One empty string",
+			s:          "abc",
+			t:          "",
+			ignoreCase: false,
+			expected:   3,
+		},
+		{
+			name:       "Both empty strings",
```
marckhouzam, line 95: Is this kind of a duplicate of the test called `Empty strings`, or does the `ignoreCase` matter?

cobra_test.go
```diff
@@ -40,3 +40,185 @@ func TestAddTemplateFunctions(t *testing.T) {
...
+			expected: true,
+		},
+		{
+			name:     "String not in slice (case-sensitive)",
+			a:        "pear",
+			list:     []string{"orange", "banana", "apple", "grape"},
+			expected: false,
+		},
+		{
+			name:     "String in slice (case-insensitive)",
```
marckhouzam, line 136: I don't think we need case-insensitive tests since the function being testes is case-sensitive.
So, for this test, since `apple` does not match in the test above, do we really need to test `APPLE`?

cobra_test.go
```diff
@@ -40,3 +40,185 @@ func TestAddTemplateFunctions(t *testing.T) {
...
+			a:        "",
+			list:     []string{"orange", "banana", "apple", "grape"},
+			expected: false,
+		},
+		{
+			name:     "Empty string in empty slice",
+			a:        "",
+			list:     []string{},
+			expected: false,
+		},
```
marckhouzam, line 164: How about adding a test that empty strings can match?
```
		{
			name:     "Empty strings match",
			a:        "",
			list:     []string{"orange", ""},
			expected: true,
		},
```

marckhouzam (commented):
Thanks @korovindenis.
More tests is always better.
Some small nits if you agree with them.

cobra_test.go
korovindenis: These tests are similar, but ignoreCase matters.

cobra_test.go
korovindenis: Agreed, this test is unnecessary

cobra_test.go
korovindenis: Yes, that's a good scenario. Added the test.

cobra_test.go
marckhouzam: Sorry I got mixed up. `apple` matches in the above test so checking that `APPLE` does not match has value. It confirms the function is case-sensitive. 

It is the `PEAR` test that is not needed

marckhouzam (approved):
Thank you!

PR #2095, opened 2024-01-06
fish completions: fix double-evaluation of commandline

krobelus:
We capture the commandline tokens using fish's "commandline --tokenize" (-o).
That function turns

	echo 'some argument $(123)'

into two arguments while removing the quotes

	echo
	some argument $(123)

Later we pass "some argument $(123)" without quotes to the shell's
"eval". This is wrong and causes spurious evaluation of the parenthesis
as command substitution.

Fix this by escaping the arguments.

The downside of this change is that things like "$HOME" or "~" will
no longer be escaped. Changing this requires changes in fish, which
I'm working on.

Reproduce the issue by pasting the completion script at
https://github.com/fish-shell/fish-shell/issues/10194#issuecomment-1879563545
to a file "grafana-manager.fish" and running

	function grafana-manager; end
	source grafana-manager.fish

Then type (without pressing Enter)

	grafana-manager public-dashboards delete --organization-id 3 --dashboard-name "k8s (public)" <TAB>

Fixes https://github.com/fish-shell/fish-shell/issues/10194

### DIFF
--- a/fish_completions.go
+++ b/fish_completions.go
@@ -45,7 +45,7 @@ function __%[1]s_perform_completion
     __%[1]s_debug "Starting __%[1]s_perform_completion"
 
     # Extract all args except the last one
-    set -l args (commandline -opc)
+    set -l args (commandline -opc | string escape)
     # Extract the last arg and escape it in case it is a space
     set -l lastArg (string escape -- (commandline -ct))
 


fish_completions.go
```diff
@@ -45,7 +45,7 @@ function __%[1]s_perform_completion
     __%[1]s_debug "Starting __%[1]s_perform_completion"
 
     # Extract all args except the last one
-    set -l args (commandline -opc)
+    set -l args (commandline -opc | string escape)
```
krobelus, line 48: cc @marckhouzam 

FWIW the blame for this mistake is on me, I made the [suggestion to use it](https://github.com/spf13/cobra/pull/1249#discussion_r563337287)

stellarblaze (approved):
Thank you fixing this.

stellarblaze:
@krobelus, do you know what needs to be done for this PR to be merged and included in the next release?

stellarblaze:
+ @marckhouzam, could you please take a look?

krobelus:
> The downside of this change is that things like "$HOME" or "~" will no
> longer be escaped.

The API to avoid this downside while still fixing the issue at hand has been merged in https://github.com/fish-shell/fish-shell/pull/10212, which will be in the upcoming fish release.
It's fairly straightforward to upgrade to the new API, I can roll a patch by tomorrow.

ichekrygin:
Circling back on this issue - do you think this can be merged?

ccoVeille:
Thanks for resurrecting this old PR.

I'm a bit lost, not only because I'm new to the project.

I tried to read the issue, the discussion that happened on it. Then this PR and what was told here too. Then I looked at the fish issue.

From my understanding, and I could be totally wrong:
- there was a bug in fish
- the bug in fish was causing a bug with cobra when using fish
- this PR was opened to fix the bug in cobra when using fish.

For now, I think I'm good.

But then, from what I can see:
- the bug in fish is now fixed
- this PR is still not merged

Here is where I am.

So now, let me ask:
- is there still a bug with cobra when using fish, now the bug in fish is now fixed ?
- is there still a need for this PR ?

ichekrygin:
Hello @ccoVeille , thank you for looking into this. 

I understand that we still need this PR to address the issue of fish shell completions. The fix in Fish shell was necessary to avoid the undesirable side effects that this PR would introduce if merged as-is. As noted in the PR description:

"The downside of this change is that things like $HOME or ~ will no longer be escaped. Changing this requires changes in fish, which I'm working on."
Now that the upstream fix has been merged in Fish, this PR can be updated to take advantage of the new API—avoiding the side effects while still resolving the original problem.

+ @krobelus

ccoVeille:
OK thanks @ichekrygin.

It's way clearer now.

I might be wrong but based on your comment, I understand the PR shouldn't be merged like this, and requires more work.

krobelus:
I can update it but we'll also need someone with commit access

ccoVeille (approved):
LGTM 👍, even if my knowledge is very limited on this part of the code.

awsms:
Any chance to get this merged one day?

PR #2097, opened 2024-01-12
feat: Add the `WithUsage` function.

octo:
Add a wrapper for `PositionalArgs` that supplements errors with usage information.

The error messages provided by the `PositionalArgs` functions is quite terse, e.g.

> accepts *n* arg(s), received *m*

In many cases this will not be enough information for users to correct their mistake. The new `WithUsage` function includes the usage string in the error message to improve the user experience.

**Example:**

```go
cmd := &cobra.Command{
	Use:  "example <arg>",
	Args: cobra.WithUsage(cobra.ExactArgs(1)),
}
```

### DIFF
--- a/args.go
+++ b/args.go
@@ -129,3 +129,16 @@ func MatchAll(pargs ...PositionalArgs) PositionalArgs {
 func ExactValidArgs(n int) PositionalArgs {
 	return MatchAll(ExactArgs(n), OnlyValidArgs)
 }
+
+// WithUsage wraps another PositionalArgs function. If the function fails
+// (returns a non-nil error), the error is supplemented with the command's
+// usage message, providing the user with the information required to run the
+// command correctly.
+func WithUsage(wrapped PositionalArgs) PositionalArgs {
+	return func(cmd *Command, args []string) error {
+		if err := wrapped(cmd, args); err != nil {
+			return fmt.Errorf("%w\n\n%s", err, cmd.UsageString())
+		}
+		return nil
+	}
+}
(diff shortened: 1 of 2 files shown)


marckhouzam:
I like this. 
And I like the use of an example function in the tests, which we haven’t done before.
Nice job.

args.go
```diff
@@ -129,3 +129,16 @@ func MatchAll(pargs ...PositionalArgs) PositionalArgs {
...
 }
+
+// WithUsage wraps another PositionalArgs function. If the function fails
+// (returns a non-nil error), the error is supplemented with the command's
+// usage message, providing the user with the information required to run the
+// command correctly.
+func WithUsage(wrapped PositionalArgs) PositionalArgs {
+	return func(cmd *Command, args []string) error {
+		if err := wrapped(cmd, args); err != nil {
+			return fmt.Errorf("%w\n\n%s", err, cmd.UsageString())
```
nirs, line 140: This is better than other errors in flags since we separate the error from the usage message with a blank line. It would be nice to add the same for other usage errors.

One issue with showing usage is that long usage message cause the error to scroll out of the view, so the user need to scroll up to find what was the error. This new option allows using the option only for commands with short usage text, but having different behavior for commands does not feel the right solution.

I think that the current behavior not showing usage in for positional arguments errors in a design issue in cobra and inconsistent. It would be better to make this behavior the default behavior instead of adding a wrapper.

Also this seems to ignore the SilenceUsage option. In Execute() we have:

```
1132         // If root command has SilenceUsage flagged,
1133         // all subcommands should respect it
1134         if !cmd.SilenceUsage && !c.SilenceUsage {
1135             c.Println(cmd.UsageString())
1136         }
```

It would be best if we can propagate positional arguments errors to Execute and have all error handled in the same place.

PR #2098, opened 2024-01-15
Add Taikun CLI to list of projects

Smidra:
I'd like to add Taikun to the list o CLIs that use the Cobra framework.
The ClI repo is opensource and can be found [here](https://github.com/itera-io/taikun-cli)

Thanks.


marckhouzam (approved):
Thanks

PR #2099, opened 2024-01-15
More linting

scop:
Some old patches I've had around, refreshed.


bash_completions.go
```diff
@@ -597,19 +597,16 @@ func writeRequiredFlag(buf io.StringWriter, cmd *Command) {
 		if nonCompletableFlag(flag) {
 			return
 		}
-		for key := range flag.Annotations {
-			switch key {
-			case BashCompOneRequiredFlag:
```
jpmcb, line 602: cc @marckhouzam - was the idea to someday have other cases in the annotation? If so, we may want to keep around this switch / case

jpmcb (commented):
Nice - looks great. One small comment.

bash_completions.go
marckhouzam: Although we _may_ add new annotations, I don’t expect to evolve the bash completion v1 logic but only the v2 one anyway, so removing the switch seems fine

jpmcb (approved):

marckhouzam (approved):
Thanks!

PR #2126, opened 2024-03-21
fix(bash): Fix bash completion for suggestions that contain special characters.

### LINKED ISSUE #1740: Shell completions with whitespace seem to work differently on bash vs. other supported shells

JeffFaer:
Special characters include whitepace, so this is more general than
#1743. This should also fix #1740.

I added some test cases to cobra-completion-testing. This PR makes them
pass. It also doesn't trip any of the performance regression tests. I'm
happy to submit those tests as a PR as well.
  - https://github.com/JeffFaer/cobra-completion-testing/tree/special_characters
  - https://github.com/JeffFaer/cobra-completion-testing/commit/52254c1a25b9816cf31a617cf9667e61a1bd2259

### DIFF
--- a/bash_completionsV2.go
+++ b/bash_completionsV2.go
@@ -68,13 +68,6 @@ __%[1]s_get_completion_results() {
     lastChar=${lastParam:$((${#lastParam}-1)):1}
     __%[1]s_debug "lastParam ${lastParam}, lastChar ${lastChar}"
 
-    if [[ -z ${cur} && ${lastChar} != = ]]; then
-        # If the last parameter is complete (there is a space following it)
-        # We add an extra empty parameter so we can indicate this to the go method.
-        __%[1]s_debug "Adding extra empty parameter"
-        requestComp="${requestComp} ''"
-    fi
-
     # When completing a flag with an = (e.g., %[1]s -n=<TAB>)
     # bash focuses on the part after the =, so we need to remove
     # the flag part from $cur
--- a/bash_completionsV2.go
+++ b/bash_completionsV2.go
@@ -220,24 +220,15 @@ __%[1]s_handle_completion_types() {
         # completions at once on the command-line we must remove the descriptions.
         # https://github.com/spf13/cobra/issues/1508
         local tab=$'\t' comp
-        local matches=()
         for comp in "${completions[@]}"; do
             [[ -z $comp ]] && continue
             # Strip any description
             comp=${comp%%%%$tab*}
             # Only consider the completions that match
             if [[ $comp == "$cur"* ]]; then
-                # Strictly speaking we could append directly to COMPREPLY here.
-                # But there's a pretty big performance hit involved with
-                # creating one subshell to printf %%q for each completion that
-                # matches. Instead, batch all the matches up so we can quote
-                # them all at once in a single printf call.
-                matches+=( "$comp" )
+                COMPREPLY+=( "$comp" )
            fi
         done
-        while IFS='' read -r comp; do
-            COMPREPLY+=( "$comp" )
-        done < <(printf "%%q\n" "${matches[@]}")
         ;;
 
     *)
--- a/bash_completionsV2.go
+++ b/bash_completionsV2.go
@@ -59,7 +59,10 @@ __%[1]s_get_completion_results() {
     # Prepare the command to request completions for the program.
     # Calling ${words[0]} instead of directly %[1]s allows handling aliases
     args=("${words[@]:1}")
-    requestComp="${words[0]} %[2]s ${args[*]}"
+    requestComp="${words[0]} %[2]s"
+    if [[ "${#args[@]}" -gt 0 ]]; then
+        requestComp+="$(printf " %%q" "${args[@]}")"
+    fi
 
     lastParam=${words[$((${#words[@]}-1))]}
     lastChar=${lastParam:$((${#lastParam}-1)):1}
@@ -224,15 +227,24 @@ __%[1]s_handle_completion_types() {
         # completions at once on the command-line we must remove the descriptions.
         # https://github.com/spf13/cobra/issues/1508
         local tab=$'\t' comp
-        while IFS='' read -r comp; do
+        local matches=()
+        for comp in "${completions[@]}"; do
             [[ -z $comp ]] && continue
             # Strip any description
             comp=${comp%%%%$tab*}
             # Only consider the completions that match
             if [[ $comp == "$cur"* ]]; then
-                COMPREPLY+=("$comp")
-            fi
-        done < <(printf "%%s\n" "${completions[@]}")
+                # Strictly speaking we could append directly to COMPREPLY here.
+                # But there's a pretty big performance hit involved with
+                # creating one subshell to printf %%q for each completion that
+                # matches. Instead, batch all the matches up so we can quote
+                # them all at once in a single printf call.
+                matches+=( "$comp" )
+           fi
+        done
+        while IFS='' read -r comp; do
+            COMPREPLY+=( "$comp" )
+        done < <(printf "%%q\n" "${matches[@]}")
         ;;
 
     *)
@@ -247,7 +259,12 @@ __%[1]s_handle_standard_completion_case() {
 
     # Short circuit to optimize if we don't have descriptions
     if [[ "${completions[*]}" != *$tab* ]]; then
-        IFS=$'\n' read -ra COMPREPLY -d '' < <(compgen -W "${completions[*]}" -- "$cur")
+        # compgen's -W option respects shell quoting, so we need to escape.
+        local compgen_words="$(printf "%%q\n" "${completions[@]}")"
+        # compgen appears to respect shell quoting _after_ checking whether
+        # they have the right prefix, so we also need to quote cur.
+        local compgen_cur="$(printf "%%q" "${cur}")"
+        IFS=$'\n' read -ra COMPREPLY -d '' < <(IFS=$'\n'; compgen -W "${compgen_words}" -- "${compgen_cur}")
         return 0
     fi
 
@@ -271,7 +288,7 @@ __%[1]s_handle_standard_completion_case() {
         __%[1]s_debug "COMPREPLY[0]: ${COMPREPLY[0]}"
         comp="${COMPREPLY[0]%%%%$tab*}"
         __%[1]s_debug "Removed description from single completion, which is now: ${comp}"
-        COMPREPLY[0]=$comp
+        COMPREPLY[0]="$(printf "%%q" "${comp}")"
     else # Format the descriptions
         __%[1]s_format_comp_descriptions $longest
     fi
--- a/bash_completionsV2.go
+++ b/bash_completionsV2.go
@@ -59,17 +59,18 @@ __%[1]s_get_completion_results() {
     # Prepare the command to request completions for the program.
     # Calling ${words[0]} instead of directly %[1]s allows handling aliases
     args=("${words[@]:1}")
-    requestComp="${words[0]} %[2]s"
-    if [[ "${#args[@]}" -gt 0 ]]; then
-        # Previous args should already be escaped...
-        requestComp+=" ${args[*]::${#args[@]}-1}"
-        # ...but the current arg might not yet be escaped.
-        requestComp+=" $(printf "%%q" "${args[${#args[@]}-1]}")"
-    fi
+    requestComp="${words[0]} %[2]s ${args[*]}"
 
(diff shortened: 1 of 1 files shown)


ncw:
It would be nice to have this PR merged - many rclone users have noticed that we can't complete file names with spaces in :-(

marckhouzam:
Sorry for the delay.  I will be reviewing this next.

marckhouzam:
@JeffFaer 
> I'm happy to submit those tests as a PR as well.
> https://github.com/JeffFaer/cobra-completion-testing/tree/special_characters
> https://github.com/JeffFaer/cobra-completion-testing/commit/52254c1a25b9816cf31a617cf9667e61a1bd2259

That would be great!

JeffFaer:
https://github.com/marckhouzam/cobra-completion-testing/pull/25

marckhouzam:
I think there may be some confusion here.
When doing completion for special characters, the characters should be automatically escaped.
So if the completion choice is `a>b`, doing `prog [tab][tab]` should yield `prog a\>b`.
You can rebase #1743 and try it out as it behaves like this.
You can also try doing such completion with zsh and see how it should behave.

JeffFaer:
I've updated these changes to reflect the behavior you're looking for. It automatically escapes the last completion suggestion.

JeffFaer:
Hmm there's actually a bad interaction here with `COMP_WORDBREAKS`. The default `COMP_WORDBREAKS` seems to include many special characters by default.

The bash completion script is handling a couple wordbreaks (`=` and `:`) "correctly" by removing everything up to and including the wordbreak from proposed `COMPREPLY`s so that readline doesn't duplicate the same prefix again. That's problematic for special characters that need to be escaped like `>` because that means the completion script won't be able to escape them anymore...

JeffFaer:
I took a pass at a way to address the `COMP_WORDBREAKS` interaction (tl;dr: remove all wordbreaks that would need to be escaped)

bash_completionsV2.go
```diff
@@ -173,8 +169,9 @@ __%[1]s_process_completion_results() {
         __%[1]s_handle_completion_types
     fi
 
-    __%[1]s_handle_special_char "$cur" :
-    __%[1]s_handle_special_char "$cur" =
+    __%[1]s_handle_wordbreaks "$cur"
```
JeffFaer, line 172: At surface value, this line looked wrong to me because I didn't think it was correct to handle wordbreaks like this when the user requested menu-completion. But after some manual testing I'm pleasantly surprised to find that this is correct, and menu-completion does still expect the same behavior as "regular" completion.

marckhouzam:
@JeffFaer Thanks for the update!  It looks really good.  However, while testing with the updated program of https://github.com/marckhouzam/cobra-completion-testing/pull/25 I'm noticing a subtle behaviour, that I feel is incorrect.

```
bash-5.2$ ./testprog prefix special-chars bash4>[tab]
bash-5.2$ ./testprog prefix special-chars bash4\>redirect
```
In the above case, when the user presses `[tab]` after the redirect symbol `>` they should get some file completion and _not_ the `>` being escaped.  In the above case, since the user did not escape the `>` themselves, it means they want to use it as a redirect command, so Cobra should not escape it.

If the user where to escape the `>` themselves, that is when Cobra should continue the shell completion:
```
bash-5.2$ ./testprog prefix special-chars bash4\>[tab]
# Below is what I would expect, but this is not how the PR behaves
bash-5.2$ ./testprog prefix special-chars bash4\>redirect
```
Comparing it to `zsh` completion helps see the expected behaviour.
So, to summarize, this is what we need:
```
bash-5.2$ ./testprog prefix special-chars bash4[tab]
bash-5.2$ ./testprog prefix special-chars bash4\>redirect

bash-5.2$ ./testprog prefix special-chars bash4\>[tab]
bash-5.2$ ./testprog prefix special-chars bash4\>redirect

bash-5.2$ ./testprog prefix special-chars bash4>[tab]
# Normal file completion should happen here
```
The same approach should apply to the other special characters: ` `, `|`, `$`, `;`.
For `#`, zsh requires it to be escaped to continue completion (`bash5\#<TAB>`), but I'm not sure why, but let's do the same.

JeffFaer:
It sounds like you're looking to have special characters escaped as soon as they end up on the command line instead of after the full word has been completed. I've updated the PR to do that. PTAL. Escaping special characters as soon as they end up on the command line lets us simplify some of the wordbreak complexity, too

bash_completionsV2.go
```diff
@@ -224,20 +225,28 @@ __%[1]s_handle_completion_types() {
...
-        done < <(printf "%%s\n" "${completions[@]}")
+           fi
+        done
+
+        __%[1]s_escape_compreply
+        ;;
+
+    63)
+        # Type: Listing completions after successive tabs
+        __%[1]s_handle_standard_completion_case
```
JeffFaer, line 243: I'm not escaping here so that the list of results on multiple tabs doesn't include any extra escape characters...is that the behavior you're expecting?

marckhouzam: Could you give an example of the behaviour you are referring too?

JeffFaer: `testprog prefix special-chars bash<tab><tab>`

The list of options it reports is not escaped (i.e. `bash4>redirect`), but if you were to do `bash4<tab>` you'd get the escaped version (`bash4\>redirect`)

marckhouzam:
@JeffFaer I worked many hours trying to make sure this was safe to merge, and in the end I felt there were too many changes I didn't understand well enough so I tried to simplify the change as much as I could.  By the time I was done and had something working, it looked a lot like #1743.  So, I feel the fair thing to do is to go back to #1743 and make it work.

I was able to do sort all this out thanks to the great tests you wrote in https://github.com/marckhouzam/cobra-completion-testing/pull/25, so I want to merge that one also.

I'm going to pause working on this one and see if we can get #1743 working.
Thanks for your continue efforts.

bash_completionsV2.go
marckhouzam: I think we should escape it in the list.  I will do that in #1743

marckhouzam: Hmm, I notice that zsh and fish don't show the escaping in their list.  I'll think about it...

bash_completionsV2.go
```diff
@@ -247,7 +256,13 @@ __%[1]s_handle_standard_completion_case() {
 
     # Short circuit to optimize if we don't have descriptions
     if [[ "${completions[*]}" != *$tab* ]]; then
-        IFS=$'\n' read -ra COMPREPLY -d '' < <(compgen -W "${completions[*]}" -- "$cur")
+        # compgen's -W option respects shell quoting, so we need to escape.
+        local compgen_words="$(__%[1]s_escape_suggestions "${completions[@]}")"
+        if [[ "${BASH_VERSION}" = 3* ]]; then
+            # bash3 compgen doesn't handle escaped # symbols correctly.
+            compgen_words="${compgen_words//\\#/#}"
```
marckhouzam, line 263: In the end I decided not to escape the `#` since the shell passes to the go program without problems.

bash_completionsV2.go
marckhouzam: I didn't understand why this change was needed so for safety, I did not take it.

bash_completionsV2.go
```diff
@@ -224,20 +225,28 @@ __%[1]s_handle_completion_types() {
         # completions at once on the command-line we must remove the descriptions.
         # https://github.com/spf13/cobra/issues/1508
         local tab=$'\t' comp
-        while IFS='' read -r comp; do
+        for comp in "${completions[@]}"; do
```
marckhouzam, line 228: Not sure why this was needed so I didn't take it.

bash_completionsV2.go
```diff
@@ -224,20 +225,28 @@ __%[1]s_handle_completion_types() {
...
                 COMPREPLY+=("$comp")
-            fi
-        done < <(printf "%%s\n" "${completions[@]}")
+           fi
+        done
+
+        __%[1]s_escape_compreply
+        ;;
+
+    63)
```
marckhouzam, line 241: This was elegant.  However, because it was still required to escape characters for COMP_TYPE==9 so that we could filter completion on what the user typed, I went with a different approach.

To avoid showing escaped characters in the menu list, I didn't escape the completions when there were more than one.  We want to insert in the command-line the escape characters, and since we only insert in the command-line when there is a single completion (except for COMP_TYPE 37 or 42), that seemed like a good way to know when to escape or not.

bash_completionsV2.go
```diff
@@ -224,20 +225,28 @@ __%[1]s_handle_completion_types() {
...
+        # Type: Listing completions after successive tabs
+        __%[1]s_handle_standard_completion_case
         ;;
 
     *)
         # Type: complete (normal completion)
         __%[1]s_handle_standard_completion_case
+        __%[1]s_escape_compreply
```
marckhouzam, line 249: I liked the elegance of trying to escape the characters at the end.  But it turned out we also needed to escape them from within `__%[1]s_handle_standard_completion_case`, so I decided to do everything from within that function.

bash_completionsV2.go
```diff
@@ -247,7 +256,13 @@ __%[1]s_handle_standard_completion_case() {
...
     # Short circuit to optimize if we don't have descriptions
     if [[ "${completions[*]}" != *$tab* ]]; then
-        IFS=$'\n' read -ra COMPREPLY -d '' < <(compgen -W "${completions[*]}" -- "$cur")
+        # compgen's -W option respects shell quoting, so we need to escape.
+        local compgen_words="$(__%[1]s_escape_suggestions "${completions[@]}")"
+        if [[ "${BASH_VERSION}" = 3* ]]; then
+            # bash3 compgen doesn't handle escaped # symbols correctly.
+            compgen_words="${compgen_words//\\#/#}"
+        fi
+        IFS=$'\n' read -ra COMPREPLY -d '' < <(IFS=$'\n'; compgen -W "${compgen_words}" -- "${cur}")
```
marckhouzam, line 265: I ended-up using the logic you wrote here (except for the BASH 3 stuff).
I gave you credit in the commit and put you as a co-author.

marckhouzam:
@JeffFaer Thanks again for your hard work.  This was very tricky.  The version I plan to commit is now posted in #1743 but I made you a co-author because I was able to get things working thanks to different part of your work as well.

If you can give a spin to #1743 and see if you are happy with it, it would help me make sure I didn't forget something you might have noticed during your hard work.  Thanks!

I will close this PR in favor of #1743

PR #2105, opened 2024-02-01
Show the actual available flags in useline

lugray:
Use common useline notation for available flags:

- `[-f foo]` for (standard) optional flags
- `[-f foo | -b bar]` for mutually exclusive flags
- `-f foo` for required flags
- `{-f foo | -b bar}` for mutually exclusive flags where one is required
- `[-f foo -b bar]` for flags required as a group

For a boolean flag `f`, ` foo` is dropped from the above. The foo/bar comes from `flag.UnquoteUsage`.

PR #2140, opened 2024-05-03
Consistent annotation names

nirs:
Add `Annotation` suffix to the private annotations to allow nicer code using the constants.

For example one can use the current annotation names as a temporary variable instead of unclear shortcut. Instead of this:

    rag := flagsFromAnnotation(c, f, requiredAsGroup)
    me := flagsFromAnnotation(c, f, mutuallyExclusive)
    or := flagsFromAnnotation(c, f, oneRequired)

We can use now:

    requiredAsGrop := flagsFromAnnotation(c, f, requiredAsGroupAnnotation)
    mutuallyExclusive := flagsFromAnnotation(c, f, mutuallyExclusiveAnnotation)
    oneRequired := flagsFromAnnotation(c, f, oneRequiredAnnotation)

Example taken from #2105.

### DIFF
--- a/flag_groups.go
+++ b/flag_groups.go
@@ -23,9 +23,9 @@ import (
 )
 
 const (
-	requiredAsGroup   = "cobra_annotation_required_if_others_set"
-	oneRequired       = "cobra_annotation_one_required"
-	mutuallyExclusive = "cobra_annotation_mutually_exclusive"
+	requiredAsGroupAnnotation   = "cobra_annotation_required_if_others_set"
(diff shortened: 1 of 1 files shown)


marckhouzam:
Thanks @nirs .  This is fine but I will prevent #2105 from merging.  So I won't merge this one right away, review #2105 first and see where we stand.

marckhouzam (approved):
LGTM
Since #2105 needs a new revision, we can merge this and have #2105 rebase.

PR #2148, opened 2024-05-18
Remove fully inactivated linters

nirs:
"interfaces", "maligned", and "golint" do not produce any report and cause golangci-lint to fail as we can see in the CI.

- Remove "interfaces" and "maligned" linters that have no replacement.
- Replace golint with stylecheck

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -49,7 +49,7 @@ linters:
     #- scopelint
     #- staticcheck
     #- structcheck ! deprecated since v1.49.0; replaced by 'unused'
-    #- stylecheck
+    - stylecheck
     #- typecheck
     - unconvert
     #- unparam
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -33,16 +33,13 @@ linters:
     #- gocyclo
     #- gofmt
     - goimports
-    - golint
     #- gomnd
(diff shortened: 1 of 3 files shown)


marckhouzam:
Thanks for this @nirs and for being so quick to fix the problem.
I'm going to merge this right away to fix the build.

Below are the warnings we used to get and that have some recommendations (I can see those when using golangci-lint `1.55.0`):
```
WARN The linter 'golint' is deprecated (since v1.41.0) due to: The repository of the linter has been archived by the owner. Replaced by revive.
WARN The linter 'maligned' is deprecated (since v1.38.0) due to: The repository of the linter has been archived by the owner. Replaced by govet 'fieldalignment'.
WARN The linter 'interfacer' is deprecated (since v1.38.0) due to: The repository of the linter has been archived by the owner.
```
For `maligned` we do use `govet` already so we're good.
For `interfacer` as you point out, there is not recommended replacement.
For `golint`, the warning mentions using `revive`.

I quickly tried revive and it flags a bunch of unused parameters to functions.  Doesn't seem critical although it could prove useful in some situation.  Either way, it is not important for this PR.

Thanks again!

marckhouzam (approved):

PR #2153, opened 2024-06-07
feat: add flag prerequisite check

sam80180:
Add function for checking whether flags are effective only if some other flags are set


james-whitaker-arm:
Nice, I was looking for something like this (assuming I've understood this feature correctly).

It would be great if you could say:  flag X only makes sense in the presence of another flag Y. Such that if flag X is entered but flag Y isn't, an error is displayed. Allowing flag X to be provided by itself may be harmless, but it's misleading and not good CLI design.

PR #138, opened 2015-09-01
Allow SetHelpFunc to overwrite the --help flag instead of just the help subcommand

eparis:

PR #2158, opened 2024-06-10
Pass arguments to the help function

### LINKED ISSUE #2154: Arguments not passed to the help function
I can override the help function using:
- `cmd.SetHelpFunc(func(c *cobra.Command, args []string))`.

Notice that the new help function I will define takes a command and its arguments as parameters, just like we have in many cobra functions.

I may choose to print a different help text based on what args the user has typed.
However, my help function does not get passed the correct args.

There are two situations where the help function is called:
1. when using the `--help/-h` flag (e.g., `prog sub arg1 -h`)
2. when using the `help` command  (e.g., `prog help sub arg1`)

The below test shows that both cases are problematic (see program below).
Try it on playground: https://go.dev/play/p/UvaMTHq5Mqp
```
# Notice that the "args" should be "arg1 arg2" but are not
Calling: prog sub arg1 arg2 -h
parameter cmd: sub
param args: [sub arg1 arg2 -h]

# Notice that the "args" should be "arg1 arg2" but are not
Calling: prog help sub arg1 arg2
parameter cmd: sub
param args: []
```

<details><summary>Test program</summary>

```go
import (
	"fmt"
	"strings"

	"github.com/spf13/cobra"
)

func main() {
	var rootCmd = &cobra.Command{Use: "root"}
	rootCmd.AddCommand(&cobra.Command{Use: "sub"})

	rootCmd.SetHelpFunc(func(c *cobra.Command, args []string) {
		fmt.Printf("parameter cmd: %s\nparam args: %v\n", c.Name(), args)
	})

	args := []string{"sub", "arg1", "arg2", "-h"}
	fmt.Printf("Calling: prog %s\n", strings.Join(args, " "))
	rootCmd.SetArgs(args)
	rootCmd.Execute()

	args = []string{"help", "sub", "arg1", "arg2"}
	fmt.Printf("\nCalling: prog %s\n", strings.Join(args, " "))
	rootCmd.SetArgs(args)
	rootCmd.Execute()
}
```
</details>

marckhouzam:
Fixes #2154 

This PR makes sure that when Cobra calls `HelpFunc()(cmd, args)`, the proper arguments are passed to the function.
With this, when the help function is overridden, the new function can choose to print the help output based on what the args are.

For example, say I write my own `ls` program, I could override the help function using `SetHelpFunc()` so that
- `ls -h` would print `List the content of the current directory`
- `ls mydir -h` would instead print `List the content of the "mydir" directory`

This PR fixed the two cases where the help function is called:
1. when using the `--help/-h` flag (e.g., `prog sub arg1 -h`)
1. when using the` help` command (e.g., `prog help sub arg1`)

Another value of this fix is for programs that use plugins and want to ask the plugin what the help output should be.
For instance the `tanzu` CLI does this, where if I type `tanzu help cluster list` the help function will call the `cluster` plugin but needs to tell it that it needs the help for the `list` command.  This `list` argument needs this PR to get passed to the help function.  One can see how the `tanzu` code had to use `os.Args` to [work around this bug](https://github.com/vmware-tanzu/tanzu-cli/blob/6a11ce93cd4e811e213e8439e090e1d73a053fd3/pkg/cli/plugin_cmd.go#L192-L201).

Unit tests have been added which also illustrate the different cases that that PR fixes.
A test program is provided in #2154 to show the problem before this PR.

Please refer to a couple of comments I will add in the PR review for a concern about backwards-compatibility.

### DIFF
--- a/command.go
+++ b/command.go
@@ -470,8 +470,11 @@ func (c *Command) HelpFunc() func(*Command, []string) {
 }
 
 // Help puts out the help for the command.
-// Used when a user calls help [command].
-// Can be defined by user by overriding HelpFunc.
+// Kept for backwards compatibility.
+// No longer used because it does not allow
+// to pass arguments to the help command.
+// Can be a simple way to trigger the help
+// if arguments are not needed.
 func (c *Command) Help() error {
 	c.HelpFunc()(c, []string{})
 	return nil
@@ -1119,7 +1122,13 @@ func (c *Command) ExecuteC() (cmd *Command, err error) {
 		// Always show help if requested, even if SilenceErrors is in
 		// effect
 		if errors.Is(err, flag.ErrHelp) {
-			cmd.HelpFunc()(cmd, args)
+			// The call to execute() above has parsed the flags.
+			// We therefore only pass the remaining arguments to the help function.
+			argWoFlags := cmd.Flags().Args()
+			if cmd.DisableFlagParsing {
+				argWoFlags = flags
+			}
+			cmd.HelpFunc()(cmd, argWoFlags)
 			return cmd, nil
 		}
 
@@ -1260,14 +1269,14 @@ Simply type ` + c.displayName() + ` help [path to command] for full details.`,
 				return completions, ShellCompDirectiveNoFileComp
 			},
 			Run: func(c *Command, args []string) {
-				cmd, _, e := c.Root().Find(args)
+				cmd, remainingArgs, e := c.Root().Find(args)
 				if cmd == nil || e != nil {
 					c.Printf("Unknown help topic %#q\n", args)
 					CheckErr(c.Root().Usage())
 				} else {
 					cmd.InitDefaultHelpFlag()    // make possible 'help' flag to be shown
 					cmd.InitDefaultVersionFlag() // make possible 'version' flag to be shown
-					CheckErr(cmd.Help())
+					cmd.HelpFunc()(cmd, remainingArgs)
 				}
 			},
 			GroupID: c.helpCommandGroupID,
--- a/command_test.go
+++ b/command_test.go
@@ -12,6 +12,7 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+//nolint:goconst
 package cobra
 
 import (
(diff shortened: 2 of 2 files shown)


command.go
```diff
@@ -1119,7 +1122,13 @@ func (c *Command) ExecuteC() (cmd *Command, err error) {
 		// Always show help if requested, even if SilenceErrors is in
 		// effect
 		if errors.Is(err, flag.ErrHelp) {
-			cmd.HelpFunc()(cmd, args)
```
marckhouzam, line 1122: The use of `args` here was added in this [commit](https://github.com/spf13/cobra/pull/138/commits/c2d19e9c4338e36a2f2176537a4f67b3ee59f4e7) and it does seem that it was a mistake; I believe the `flags` variable should have been used instead.

command.go
```diff
@@ -1119,7 +1122,13 @@ func (c *Command) ExecuteC() (cmd *Command, err error) {
...
+			// The call to execute() above has parsed the flags.
+			// We therefore only pass the remaining arguments to the help function.
+			argWoFlags := cmd.Flags().Args()
+			if cmd.DisableFlagParsing {
+				argWoFlags = flags
+			}
+			cmd.HelpFunc()(cmd, argWoFlags)
```
marckhouzam, line 1131: I wanted to discuss backwards-compatibility for this change.
Before this PR cobra would pass all the arguments on the command-line (which was incorrect).
With this PR cobra will pass only the arguments after removing the sub-command and flags, as it does when calling the `*Run/RunE` functions.

It is possible that some programs worked around this bug and fixing it may affect them.

However, considering that before this PR the help function sometime received all the args (when using the `--help` flag) but no args at all when called from the `help` command, it makes it less likely that a workaround used the actual `args` parameter.

So, I feel it is ok to make this change but if we do maybe we should include a note in the release notes.

@jpmcb how do you feel about this?

command.go
```diff
@@ -1260,14 +1269,14 @@ Simply type ` + c.displayName() + ` help [path to command] for full details.`,
...
-				cmd, _, e := c.Root().Find(args)
+				cmd, remainingArgs, e := c.Root().Find(args)
 				if cmd == nil || e != nil {
 					c.Printf("Unknown help topic %#q\n", args)
 					CheckErr(c.Root().Usage())
 				} else {
 					cmd.InitDefaultHelpFlag()    // make possible 'help' flag to be shown
 					cmd.InitDefaultVersionFlag() // make possible 'version' flag to be shown
-					CheckErr(cmd.Help())
+					cmd.HelpFunc()(cmd, remainingArgs)
```
marckhouzam, line 1279: I don't feel this has any backwards-compatibility implications since the `args` value was always empty before this change for this case.

marckhouzam:
@jpmcb Could you have a look at this?  It is 7 lines of actual code with the rest being tests.

@nirs I've see recently how you are providing some great help to the Cobra project and your input here would be appreciated.

command_test.go
```diff
@@ -1079,6 +1080,192 @@ func TestHelpExecutedOnNonRunnableChild(t *testing.T) {
...
+
+	rootCmd.SetHelpFunc(func(c *Command, args []string) {
+		helpCalled = true
+		if c.Name() != "child" {
+			t.Errorf(`Expected command name: "child", got %q`, c.Name())
+		}
+		if len(args) != 2 || args[0] != "arg1" || args[1] != "arg2" {
+			t.Errorf("Expected args [args1 arg2], got %v", args)
+		}
+	})
```
nirs, line 1257: We repeat the code of the help function many times, maybe add a helper struct keeping the , so we can do:

     h := helper{t: t, command: "child", args: []string{"args1", "arg2"}}
     rootCmd.SetHelpFunc(h.helpFunc)
     _, err := executeCommand(rootCmd, "help", "child", "arg1", "arg2")

     if !h.helpCalled {
         ...
     }

nirs:
> This PR fixed the two cases where the help function is called:
> 
> 1. when using the `--help/-h` flag (e.g., `prog sub arg1 -h`)
> 2. when using the` help` command (e.g., `prog help sub arg1`)

I think this change is not needed and we can keep the current behavior, and fix the 
case of the more common and useful -h/--help.

> Another value of this fix is for programs that use plugins and want to ask the plugin what the help output should be. For instance the `tanzu` CLI does this, where if I type `tanzu help cluster list` the help function will call the `cluster` plugin but needs to tell it that it needs the help for the `list` command. This `list` argument needs this PR to get passed to the help function.

Why would someone type:

    tanzu help cluster list

and then type work hard to remove the help command to run the actual command, instead of adding -h to the actual command:

    tanzu cluster list -h

and finally remove the -h for running the actual command?

I guess tanszu smart help works both for help and -h/--help, so if we fix -h,--help, we can remove the workarounds?

It we spend time on the help command, we should make it easy to show different content for "help command". For example try `git log -h` and `git help log`. This is the biggest problems with Go tools - the online help is not good enough and we never have manual pages.

command.go
```diff
@@ -1119,7 +1122,13 @@ func (c *Command) ExecuteC() (cmd *Command, err error) {
 		// Always show help if requested, even if SilenceErrors is in
 		// effect
 		if errors.Is(err, flag.ErrHelp) {
-			cmd.HelpFunc()(cmd, args)
+			// The call to execute() above has parsed the flags.
+			// We therefore only pass the remaining arguments to the help function.
+			argWoFlags := cmd.Flags().Args()
```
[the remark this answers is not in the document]
marckhouzam: `argWoFlags` is used `execute()` but agree it is not as clear.  Might as well use the better `remainingArgs` since we are in a different function.

command.go
```diff
@@ -1119,7 +1122,13 @@ func (c *Command) ExecuteC() (cmd *Command, err error) {
...
+			if cmd.DisableFlagParsing {
+				argWoFlags = flags
+			}
```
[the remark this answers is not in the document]
marckhouzam: For commands that have `DisableFlagParsing == true`, we don't parse the arguments and `cmd.Flags().Args()` is not filled, so we need to compensate explicitly.  It's not super elegant I agree but this is how its done here too: https://github.com/spf13/cobra/blob/e94f6d0dd9a5e5738dca6bce03c4b1207ffbc0ec/command.go#L933-L936

I'll add a comment to clarify.

command_test.go
```diff
@@ -1079,6 +1080,192 @@ func TestHelpExecutedOnNonRunnableChild(t *testing.T) {
...
+func TestHelpOverrideOnRoot(t *testing.T) {
+	var helpCalled bool
+	rootCmd := &Command{Use: "root"}
+	rootCmd.SetHelpFunc(func(c *Command, args []string) {
+		helpCalled = true
+		if c.Name() != "root" {
+			t.Errorf(`Expected command name: "root", got %q`, c.Name())
+		}
+		if len(args) != 2 || args[0] != "arg1" || args[1] != "arg2" {
+			t.Errorf("Expected args [args1 arg2], got %v", args)
```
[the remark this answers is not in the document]
marckhouzam: That's much better.  I started making the change and realized that `slices.Equal` was added in Go 1.18.
Cobra uses go 1.15 to avoid forcing projects to upgrade.
The approach is debatable but is for another discussion 😄 
So I'm forced to leave this change as is.

command_test.go
marckhouzam: I like it!
I tried to do what you suggest.

command.go
marckhouzam: Your use case description is correct.  However, there is another (obscure) situation, which is actually the one I'm really trying to fix 😄.

The `tanzu` CLI uses plugins to provide many of its sub-commands.  So a user may do:
`tanzu help cluster list`
to get the help on the `cluster list` command.
However, the `tanzu` CLI is only has `cluster` as a sub-command and needs to call that "cluster" plugin to ask it for the help for the `list` command.  Therefore, the arguments passed to the original `tanzu help` command are required so which which sub-command of the plugin the user is asking about.

marckhouzam: Currently, the `tanzu` CLI code has to use `os.Args` to get that info: https://github.com/vmware-tanzu/tanzu-cli/blob/6a11ce93cd4e811e213e8439e090e1d73a053fd3/pkg/cli/plugin_cmd.go#L192-L201

marckhouzam:
> > This PR fixed the two cases where the help function is called:
> > 
> > 1. when using the `--help/-h` flag (e.g., `prog sub arg1 -h`)
> > 2. when using the` help` command (e.g., `prog help sub arg1`)
> 
> I think this change is not needed and we can keep the current behavior, and fix the case of the more common and useful -h/--help.

Answered here: https://github.com/spf13/cobra/pull/2158#discussion_r1635050426

> > Another value of this fix is for programs that use plugins and want to ask the plugin what the help output should be. For instance the `tanzu` CLI does this, where if I type `tanzu help cluster list` the help function will call the `cluster` plugin but needs to tell it that it needs the help for the `list` command. This `list` argument needs this PR to get passed to the help function.
> 
> Why would someone type:
> 
> ```
> tanzu help cluster list
> ```
> 
> and then type work hard to remove the help command [...]

I agree that the `help` command is not very useful and I personally never use it.
However, I still have to make it work for others who may use it.

> I guess tanzu smart help works both for help and -h/--help, so if we fix -h,--help, we can remove the workarounds?

The same override of `HelpFunc()` does work for both `help` and `-h/--help` but that is because Cobra uses the same `HelpFunc()` in both cases.  So we need to fix it for both cases.

> It we spend time on the help command, we should make it easy to show different content for "help command". For example try `git log -h` and `git help log`. This is the biggest problems with Go tools - the online help is not good enough and we never have manual pages.

This sounds like a good idea.  Programs using Cobra can do this already by overriding the `help` command completely, and provide something different between `help` and `-h`.  But maybe we can make it easier for them so that we can promote the idea of having different output in both cases.  I like the example of `git help log` vs `git log-h` that you provide.  It would be great if Cobra could guide projects to do something like that.

PR #2167, opened 2024-07-08
closes #2130 new command flag ErrorOnUnknownSubcommand

### LINKED ISSUE #2130: Need an error when the command is not found in the subcommand
hi

## Description
When I execute a job on our task center, I need to get the final result of the execution, for example:

`myapp job1 args1` 

When job1 doesn't exist, I can get the error:

```
Error: unknown command "job1" for ""
Run ' --help' for usage.
```

Then the task center can be accessed by `echo $?` to get the final execution result.
However, when I use a subcommand, I can't get the error. 

## How to reproduce

For example : 

```go
package main

import (
	"github.com/spf13/cobra"
)

func main() {
	var rootCmd = &cobra.Command{}

	configCmd := cobra.Command{
		Use: "config",
	}

	getCmd := cobra.Command{
		Use: "set",
			println("this is the config set")
		},
	}
	setCmd := cobra.Command{
		Use: "get",
			println("this is the config get")
		},
	}
	configCmd.AddCommand(&getCmd)
	configCmd.AddCommand(&setCmd)
	rootCmd.AddCommand(&configCmd)

		panic(err.Error())
	}
}
```

When I execute:

```bash
myapp config xxx
```

I can get the following prompt without an error.

```bash
Usage:
   config [command]

Available Commands:
  get         
  set         

Flags:
  -h, --help   help for config

Use " config [command] --help" for more information about a command.
```

## Expectations

Need to return error

------------------------------

 Is there some special consideration here?
https://github.com/spf13/cobra/blob/main/command.go#L1123

andremueller:
Proposed a backward compatible solution to https://github.com/spf13/cobra/issues/2130
Hereto a new flag `ErrorOnUnknownSubcommand` was added to the `Command` struct. Unit tests were added accordingly (TestSubcommandExecuteMissingSubcommand and TestSubcommandExecuteMissingSubcommandWithErrorOnUnknownSubcommand.

### DIFF
--- a/command.go
+++ b/command.go
@@ -44,6 +44,10 @@ type Group struct {
 	Title string
 }
 
+// ErrUnknownSubcommand is returned by Command.Execute() when the subcommand was not found.
+// Hereto, the ErrorOnUnknownSubcommand flag must be set true.
+var ErrUnknownSubcommand = errors.New("subcommand is unknown")
+
 // Command is just that, a command for your application.
 // E.g.  'go run ...' - 'run' is the command. Cobra requires
 // you to define the usage and description as part of your command
@@ -254,6 +258,17 @@ type Command struct {
 	// SuggestionsMinimumDistance defines minimum levenshtein distance to display suggestions.
 	// Must be > 0.
 	SuggestionsMinimumDistance int
+
+	// ErrorOnUnknownSubcommand controls the behavior of subcommand handling.
+	// When the flag is set true the behavior of Command.Execute() will change:
+	// If a sub-subcommand is not found ErrUnknownSubcommand will be returned on calling
+	// Command.Exec() instead of the old behavior where a nil error was sent.
+	// If the flag is false (default) the old behavior is performed.
+	// For this behavior the child subcommand must be nil.
+	// Example: in root/service/run - there would be an existing subcommand `run`
+	// in root/service/unknown - there would be an unknown subcommand `unknown` therefore returning an error
+	// `service` must have a nil Command.Run() function for this.
+	ErrorOnUnknownSubcommand bool
 }
 
 // Context returns underlying command context. If command was executed
@@ -923,9 +938,12 @@ func (c *Command) execute(a []string) (err error) {
 	}
 
 	if !c.Runnable() {
-		return flag.ErrHelp
+		if c.ErrorOnUnknownSubcommand {
+			return ErrUnknownSubcommand
+		} else {
+			return flag.ErrHelp
+		}
 	}
-
 	c.preRun()
 
 	defer c.postRun()
--- a/command_test.go
+++ b/command_test.go
@@ -17,6 +17,7 @@ package cobra
 import (
 	"bytes"
 	"context"
+	"errors"
 	"fmt"
 	"io"
 	"os"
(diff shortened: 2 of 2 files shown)


command.go
```diff
@@ -923,9 +938,12 @@ func (c *Command) execute(a []string) (err error) {
 	}
 
 	if !c.Runnable() {
-		return flag.ErrHelp
+		if c.ErrorOnUnknownSubcommand {
```
marckhouzam, line 941: `c` is the current command here.  Therefore I would need to set this new flag for every command throughout my program to get the new behaviour globally.

Do you feel this feature should be enabled per command or globally?

If globally, then you can add a global variable in `cobra.go` instead of in the command structure.  Then the program can enable this feature in a single location.

andremueller: Hey there! You are right. I think the global behavior is more what I (and probably others) want! I will change that!

andremueller: Hi @marckhouzam. Did you have a chance to review my changes? Would the behavior be so ok?

marckhouzam:
@andremueller Sorry for the delay but I had to find time to look into the many issues and PRs that tried to address this issue in the hopes of not missing anything.

Looking at #1056 and its PR #1068 there is a case we need to handle which is the additional help topics.
An additional help topic is a command that doesn't have a `Run()/RunE()` and no children, or children that also don't have a `Run()/RunE()`.  It is basically commands that don't do anything except print the help text.  Look for `IsAdditionalHelpTopicCommand()` to see the exact definition.

We need to handle this case properly when `EnableErrorOnUnknownSubcommand == true` so that programs can use your new option even if they also have those help topics.

After giving it some thought I think we can handle everything in the same fashion, as long as we don't return an error when there are no arguments.  I'll put comments in-line to clarify.

Here is a very small program that creates a program with commands and with help topics like so, to help you see what I mean:
```   
     __ child -> grandchild    
    /
root 
    \_ topic -> subtopic
```
```
package main

import (
	"fmt"
	"os"

	"github.com/spf13/cobra"
)

var rootCmd = &cobra.Command{
	Use:   "prog",
	Short: "desc for root",
}

var child = &cobra.Command{
	Use:   "child",
	Short: "desc for child",
}

var grandchild = &cobra.Command{
	Use:   "grandchild",
	Short: "desc for grandchild",
	// This has a Run so it is a real command (not a help topic)
		fmt.Println("grandchild called")
	},
}

var topic = &cobra.Command{
	Use:   "topic",
	Short: "desc for topic",
	// command without a run and for which all children don't have a run, so it is a help topic
}

var subtopic = &cobra.Command{
	Use:   "subtopic",
	Short: "desc for subtopic",
	// leaf command without a run, so it is a help topic
}

func main() {
	cobra.EnableErrorOnUnknownSubcommand = true
	rootCmd.AddCommand(child, topic)
	child.AddCommand(grandchild)
	topic.AddCommand(subtopic)
		os.Exit(1)
	}
}
```

command.go
```diff
@@ -44,6 +44,10 @@ type Group struct {
 	Title string
 }
 
+// ErrUnknownSubcommand is returned by Command.Execute() when the subcommand was not found.
+// Hereto, the ErrorOnUnknownSubcommand flag must be set true.
+var ErrUnknownSubcommand = errors.New("subcommand is unknown")
```
marckhouzam, line 49: Let's format this command message like the one in `legacyArgs()` to have a consistent behaviour.
`fmt.Errorf("unknown command %q for %q%s", args[0], cmd.CommandPath(), cmd.findSuggestions(args[0]))`

command.go
```diff
@@ -923,9 +927,12 @@ func (c *Command) execute(a []string) (err error) {
 	}
 
 	if !c.Runnable() {
-		return flag.ErrHelp
+		if EnableErrorOnUnknownSubcommand {
+			return ErrUnknownSubcommand
```
marckhouzam, line 931: Let's only return the error if there is at least one arg.
This will allow to get the usage printed without an error when running a valid command that doesn't have a Run.  This is something users do a lot.  For example running `helm repo` should print the usage without showing an error.

You should be able to move up the following lines to get the number of args:
```
argWoFlags := c.Flags().Args()
	if c.DisableFlagParsing {
		argWoFlags = a
	}
```

command_test.go
```diff
@@ -174,6 +175,102 @@ func TestSubcommandExecuteC(t *testing.T) {
...
+	const grandchildName = "grandchild"
+	EnableErrorOnUnknownSubcommand = true
+	defer func() { EnableErrorOnUnknownSubcommand = defaultErrorOnUnknownSubcommand }()
+	childCmd := &Command{Use: childName, Run: nil}
+	child2Cmd := &Command{Use: grandchildName, Run: emptyRun}
+	rootCmd.AddCommand(childCmd)
+	childCmd.AddCommand(child2Cmd)
+
+	// test existing command
+	c, output, err := executeCommandC(rootCmd, childName)
```
marckhouzam, line 238: This case should not show an error but should print the usage like before.

marckhouzam (commented):
Please run `make lint` to find the linter error.

leoluz:
Thank you @andremueller for working on this PR. We, from the Argo CD team, are also interested in this feature.

andremueller:
Hey @marckhouzam. I tried to include your proposals in the code. Actually one linting issue fixed which was not mine. However, I need to remove the common error "class" ErrUnknownSubcommand.

Would it be a good idea to refactor all 
`fmt.Errorf("unknown command xx %s", ...)`
with
`
fmt.Errorf("%w xx %s", ErrUnknownCommand, ....)
`
such that the main function could check with
`
if errors.Is(err, ErrUnknownCommand)
`
in favor of
`
if strings.HasPrefix(err.Error(), "unknown command" {...}
`

Hint: ErrUnknownCommand would be thrown in both cases: unknown command and unknown subcommand.

nirs:
> Would it be a good idea to refactor all `fmt.Errorf("unknown command xx %s", ...)` with `fmt.Errorf("%w xx %s", ErrUnknownCommand, ....)` such that the main function could check with `if errors.Is(err, ErrUnknownCommand)` in favor of `if strings.HasPrefix(err.Error(), "unknown command" {...}` ?

Yes, there is no sane way to handle `fmt.Errorf("unknown command xx %s", ...)`. It should be used only for fatal error.

Maybe cobra needs a mechanism for handling unknown commands? this can be used to implement plugins.

dcaldastp:
Any updates on this PR?

andremueller:
Yes would be interesting! What do you think @marckhouzam? Is there more work to be done on upstream Cobra before this feature can be added - regarding this error handling issue? I think this is the main show stopper to this time.

reneleonhardt:
Sad that it takes a year to fix this problem 😞

PR #2172, opened 2024-07-16
Fix deprecation comment for Command.SetOutput

thaJeztah:
Deprecation comments should be at the start of a paragraph [1], and because of that have a whitespace above them [2];

> To signal that an identifier should not be used, add a paragraph to its
> doc comment that begins with Deprecated: followed by some information
> about the deprecation (...)

With the whitespace missing, some tools, including pkg.go.dev [3] don't detect it to be deprecated.

[1]: https://go.dev/wiki/Deprecated
[2]: https://go.dev/doc/comment#paragraphs
[3]: https://pkg.go.dev/github.com/spf13/cobra@v1.8.1#Command.SetOutput

Before this patch:

Screenshot 2024-07-16 at 17 43 46
Screenshot 2024-07-16 at 17 43 37

With this patch:

Screenshot 2024-07-16 at 17 43 18
Screenshot 2024-07-16 at 17 43 27

### DIFF
--- a/command.go
+++ b/command.go
@@ -281,6 +281,7 @@ func (c *Command) SetArgs(a []string) {
 
 // SetOutput sets the destination for usage and error messages.
 // If output is nil, os.Stderr is used.
+//
 // Deprecated: Use SetOut and/or SetErr instead
 func (c *Command) SetOutput(output io.Writer) {
 	c.outWriter = output


marckhouzam (approved):
Thanks for figuring this out!

thaJeztah:
you're welcome! I actually came to the repo to open https://github.com/spf13/cobra/pull/2173, but then noticed the incorrect comment. I've had similar issues in various other repositories, and it's easy to overlook them because the work in _most_ cases (without the whitespace), but then.. don't work in other cases 😄

PR #2174, opened 2024-07-27
fix(completions): Complete map flags multiple times

gabe565:
Cobra currently allows Slice and Array flags to be completed multiple times, but it doesn't support multiple map flags like `stringToString` and `stringToInt` in the same way.

This PR adds support for map flags to be set multiple times.

### DIFF
--- a/completions.go
+++ b/completions.go
@@ -401,8 +401,9 @@ func (c *Command) getCompletions(args []string) (*Command, []string, ShellCompDi
 			doCompleteFlags := func(flag *pflag.Flag) {
 				if !flag.Changed ||
 					strings.Contains(flag.Value.Type(), "Slice") ||
-					strings.Contains(flag.Value.Type(), "Array") {
-					// If the flag is not already present, or if it can be specified multiple times (Array or Slice)
+					strings.Contains(flag.Value.Type(), "Array") ||
+					strings.HasPrefix(flag.Value.Type(), "stringTo") {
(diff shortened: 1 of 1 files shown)


completions.go
```diff
@@ -401,8 +401,9 @@ func (c *Command) getCompletions(args []string) (*Command, []string, ShellCompDi
 			doCompleteFlags := func(flag *pflag.Flag) {
 				if !flag.Changed ||
 					strings.Contains(flag.Value.Type(), "Slice") ||
-					strings.Contains(flag.Value.Type(), "Array") {
-					// If the flag is not already present, or if it can be specified multiple times (Array or Slice)
+					strings.Contains(flag.Value.Type(), "Array") ||
+					strings.Contains(flag.Value.Type(), "To") {
+					// If the flag is not already present, or if it can be specified multiple times (Array, Slice, or To)
```
marckhouzam, line 406: Instead of using “To” could we use “stringTo” to lower the risk of mis-detecting this case?
Specifically there is a `golangflag` type which possibly has a `Type()` we are not expecting (I’m not familiar with that typé of flash).

marckhouzam (commented):
Nice! I was not aware of these types.

completions.go
gabe565: Yeah, that's a better idea. Thanks for the feedback!

gabe565: I also went ahead and changed `strings.Contains()` to `strings.HasPrefix()`

marckhouzam (approved):
Thank you!

Issue #823, opened 2019-02-11
Default command


rmasci:
Why is it so difficult to have a default command?  So if I am making a program that does 4 things:

MyCommand Item1 filename
MyCommand Item2 -a 1 -b 2 filename
MyCommand Item3 -c 1 -d 2 -e 3 filename
MyCommand Item4 -a1 -e 3 filename

So if a user just types, 'MyCommand filename', it runs as if they typed 'MyCommand item1 filename' by default?  If command item1 errors out then throw an error, but try item1 if no command specified.

GustavoKatel:
I solved adding a `find` operation and changing the default args before executing the root command.
Something like this:

```go
cmd, _, err := rootCmd.Find(os.Args[1:])
if err != nil || cmd == nil {
    // Not found
    args := append([]string{"Item1"}, os.Args[1:]...)
    rootCmd.SetArgs(args)
}
```

GustavoKatel:
I solved adding a `find` operation and changing the default args before executing the root command.
Something like this:

```go
cmd, _, err := rootCmd.Find(os.Args[1:])
if err != nil || cmd == nil {
    // Not found
    args := append([]string{"Item1"}, os.Args[1:]...)
    rootCmd.SetArgs(args)
}
```

1138-4EB:
@rmasci , @GustavoKatel , you might find #725 useful.

rmasci:
Thank You!!

jharshman:
@rmasci it seems like your question has been answered by #725 .  If so, please close out this issue.

1138-4EB:
@jharshman, since there is no intuitive one-liner solution to this question, I think it might be worth keeping this issue open. Although the solution in #725 works, it is not clean and it is not properly integrated in the codebase. Therefore, I think this can be kept open until some equivalent to `checkRootAlias` is added to the codebase.

innovia:
> I solved adding a `find` operation and changing the default args before executing the root command.
> Something like this:
> 
> ```go
> cmd, _, err := rootCmd.Find(os.Args[1:])
> if err != nil || cmd == nil {
>     // Not found
>     args := append([]string{"Item1"}, os.Args[1:]...)
>     rootCmd.SetArgs(args)
> }
> ```

just a quick bug fix (should be checking if `cmd.args == nil`  b/c cmd is struct flags, help and more...) as well as context for where to place this

```go
func Execute() {
	cmd, _, err := rootCmd.Find(os.Args[1:])

[...]

PR #2178, opened 2024-08-16
args: add new `UnknownCommandError` type

mvo5:
This commit makes unknown commands a proper error type so that the caller can check for this specific error type without having to do string comparisons.

This is useful in the context of https://github.com/spf13/cobra/issues/823


mvo5:
Anything I can do here to help with this PR? If people are just busy that is fine of course (and I will be patient :)

ccoVeille:
As mentioned [here](https://github.com/spf13/cobra/pull/2274#issuecomment-2849666912) by @eth-p the following PR supersedes this one

- #2266

So I would like to focus on #2274 if you agree @mvo5 

So I would like to ask you @eth-p to check if what @mvo5 brought here is also present in #2274 and if not to add it

PR #2180, opened 2024-08-24
Fix --version help and output for plugins

nirs:
When writing a plugin and setting CommandDisplayNameAnnotation you want the display name to be used in help text. This did not work when setting command Version since the help text and version templates used the command Name().

This change makes command DisplayName() public, so it can be used in templates or other code that want to use this value, and fix --version help text and template to use the command DisplayName().

### DIFF
--- a/command.go
+++ b/command.go
@@ -1190,7 +1190,7 @@ func (c *Command) InitDefaultHelpFlag() {
 	c.mergePersistentFlags()
 	if c.Flags().Lookup("help") == nil {
 		usage := "help for "
-		name := c.displayName()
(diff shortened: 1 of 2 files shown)


nirs:
Lint errors are fixed in https://github.com/spf13/cobra/pull/2181

marckhouzam:
I’ll let you rebase this one first @nirs

nirs:
@marckhouzam can be merged now

nirs:
@marckhouzam can you check gain?

marckhouzam (approved):
Thanks for catching this!

PR #2183, opened 2024-08-24
Improve site formatting

nirs:
Improve site formatting to make it easier to maintain and have nicer output.

Preview:
- https://github.com/nirs/cobra/blob/site-formatting/site/content/user_guide.md
- https://github.com/nirs/cobra/blob/site-formatting/site/content/active_help.md


site/content/user_guide.md
```diff
@@ -4,13 +4,13 @@ While you are welcome to provide your own organization, typically a Cobra-based
...
-        commands.go
-        here.go
-      main.go
+▾ appName/
+  ▾ cmd/
+      add.go
+      your.go
+      commands.go
+      here.go
+  main.go
```
nirs, line 13: Not sure if main.go is indented correctly. We can replace it with tree output which is much more clear to users and also used later below.

```console
appName
├── cmd
│   ├── add.go
│   ├── your.go
│   ├── commands.go
│   └── here.go
└── main.go
```

marckhouzam (approved):
Nice!
Thanks!

site/content/user_guide.md
marckhouzam: Yeah, feel free to change this if you prefer.

marckhouzam: In a follow-up 😄

Issue #2188, opened 2024-09-11
Combining shorthand flags before command names does not work


nicula-stripe:
How to reproduce:

* Compile the following source code which defines some commands & flags (`go build ./mycli.go`):
```go
package main

import (
	"fmt"
	"os"

	"github.com/spf13/cobra"
)

var (
	aFlag bool
	bFlag bool
	cFlag string
)

func main() {
	// Define commands
	var rootCmd = &cobra.Command{
		Use: "mycli",
	}
	var firstCmd = &cobra.Command{
		Use: "first",
	}
	var secondCmd = &cobra.Command{
		Use: "second",
		Run: func(cmd *cobra.Command, args []string) {
			if aFlag {
				fmt.Println("flag A is set")
			}
			if bFlag {
				fmt.Println("flag B is set")
			}
			if cFlag != "" {
				fmt.Printf("flag C is set with value: %s\n", cFlag)
			}
		},
	}

	// Define flags
	rootCmd.PersistentFlags().BoolVarP(&aFlag, "set-A", "A", false, "Set the A flag")
	rootCmd.PersistentFlags().BoolVarP(&bFlag, "set-B", "B", false, "Set the B flag")
	rootCmd.PersistentFlags().StringVarP(&cFlag, "set-C", "C", "", "Set the C flag with a value")

	firstCmd.AddCommand(secondCmd)
	rootCmd.AddCommand(firstCmd)

	if err := rootCmd.Execute(); err != nil {
		fmt.Println(err)
		os.Exit(1)
	}
}
```
* Notice how this command does not work:
```
$ ./mycli -ABC elephant first second
Error: unknown command "elephant" for "mycli"
Run 'mycli --help' for usage.
unknown command "elephant" for "mycli"
```
* However, those commands work:
```
$ ./mycli -ABC=elephant first second
flag A is set
flag B is set
flag C is set with value: elephant
$ ./mycli first second -ABC elephant
flag A is set
flag B is set
flag C is set with value: elephant
```

So why does `./mycli -ABC elephant first second` not work? It seems like it should have no problem noticing that the last flag in the bundle, `-C`, takes an argument, and therefore not consider `elephant` as a command name, but as an argument to `-C`.

Is this a bug, a missing feature, or intended behavior?

Thanks!

marckhouzam:
This sounds familiar. You may want to look at past issues. 

Does `./mycli -C elephant first second` work?

nicula-stripe:
@marckhouzam 
[...]

PR #2189, opened 2024-09-14
Fix shorthand combination edge case

inicula:
Fixes: #2188


inicula:
cc @marckhouzam

command_test.go
```diff
@@ -2253,14 +2253,70 @@ func TestTraverseWithParentFlags(t *testing.T) {
 	if err != nil {
 		t.Errorf("Unexpected error: %v", err)
 	}
-	if len(args) != 1 && args[0] != "--add" {
+	if len(args) != 1 || args[0] != "--int" {
```
inicula, line 2256: The original condition seems wrong. `--add` wasn't mentioned anywhere else in the repo and from how I understand the test, `||` makes sense here, not `&&`.

command_test.go
```diff
@@ -2312,7 +2368,7 @@ func TestTraverseWithBadChildFlag(t *testing.T) {
 	if err != nil {
 		t.Errorf("Unexpected error: %v", err)
 	}
-	if len(args) != 1 && args[0] != "--str" {
+	if len(args) != 1 || args[0] != "--str" {
```
inicula, line 2371: See above.

inicula:
Could you please take another look? @marckhouzam 

Thanks! :)

PR #2194, opened 2024-10-01
Closes #2193

### LINKED ISSUE #2193: BUG: Incorrect copy of command context from parent to child
Found in cobra v1.8.1 (latest at time of writing)

Reproduce:
I execute the root command with a context ( ExecuteContext(ctx) ), my arguments resolve to a sub command and the context is passed down to the executing command- CORRECT
I execute this again with a different context.  This time the context is not passed and I get the previous context.  - INCORRECT

The context is changed in the root command so if my args indicated the root command, this command would execute with the different context as expected.  This means that the behaviour is inconsistent with sub commands compared to the root command

I think the problem lies in command.go at lines 1111-1115

```go
// We have to pass global context to children command
// if context is present on the parent command.
if cmd.ctx == nil {
    cmd.ctx = c.ctx
}
```
cmd is the exucte command and c is the root
This check only copies the context if the child has no context.  It seems that it should check if the root has a context and pass this.
The comment above these lines suggest this, the code should read

```go
// We have to pass global context to children command
// if context is present on the parent command.
if c.ctx == nil {
    cmd.ctx = c.ctx
}
```
Consider if there should be any check at all, whatever context the root has, it was passed with the execute command and is the context expected in the executing command.

AtricoSoftware:
Context is not copied from parent to child
Comment on code states copy if parent has context but code implements copy if child doesn't have context
Fix makes behaviour consistent between root and sub commands w.r.t. contexts

### DIFF
--- a/command.go
+++ b/command.go
@@ -1109,11 +1109,8 @@ func (c *Command) ExecuteC() (cmd *Command, err error) {
 		cmd.commandCalledAs.name = cmd.Name()
 	}
 
-	// We have to pass global context to children command
-	// if context is present on the parent command.
-	if cmd.ctx == nil {
-		cmd.ctx = c.ctx
-	}
+	// Copy context
+	cmd.ctx = c.ctx
 
 	err = cmd.execute(flags)
 	if err != nil {


AtricoSoftware:
I would suggest removing the check completely.
At the moment, if the user sets a context in the execute, it is used, otherwise any previous context is used
With no check, the context used is what the user sets or the background - no resue of a context unless the user specifies this.

lukaspj:
I concur.

This issue makes it very hard to use Cobra in tests so this would be a great fix.

PR #2200, opened 2024-10-23
add MarkIfFlagPresentThenOthersRequired

faizan-siddiqui:
This PR introduces a new flag group function, `MarkIfFlagPresentThenOthersRequired`, which enforces a "one-way required together" relationship among flags. This means that if a **primary flag** is set, then other **dependent flags** must also be set. This allows users to make certain flags conditionally required based on the presence of another flag, while maintaining flexibility when the primary flag is not provided.

### Example
Consider a command called `get-range` that uses the flags `--start` and `--end` to specify a date range. By default, the command can run without either flag, using default start and end values (e.g., `start=someDefaultStart` and `end=Now`). However, you may want to enforce that if the user specifies an `--end`, they must also specify a `--start`.

```go
cmd := &cobra.Command{
    Use: "get-range",
}

cmd.Flags().String("start", "", "Start date")
cmd.Flags().String("end", "", "End date")

// If --end is set, --start is required
cmd.MarkIfFlagPresentThenRequired("end", "start")
```

- If the command is called with only `--end=value`, it will trigger an error because `--start` is missing.
- If called without any flags or only with `--start`, it will run using the default values or the provided start date.

This change makes it easier to enforce dependencies between flags while allowing flexible defaults when the primary flag is not specified.

### DIFF
--- a/flag_groups.go
+++ b/flag_groups.go
@@ -23,9 +23,10 @@ import (
 )
 
 const (
-	requiredAsGroupAnnotation   = "cobra_annotation_required_if_others_set"
-	oneRequiredAnnotation       = "cobra_annotation_one_required"
-	mutuallyExclusiveAnnotation = "cobra_annotation_mutually_exclusive"
+	requiredAsGroup             = "cobra_annotation_required_if_others_set"
+	oneRequired                 = "cobra_annotation_one_required"
+	mutuallyExclusive           = "cobra_annotation_mutually_exclusive"
+	ifPresentThenOthersRequired = "cobra_annotation_if_present_then_others_required"
 )
 
 // MarkFlagsRequiredTogether marks the given flags with annotations so that Cobra errors
@@ -76,6 +77,25 @@ func (c *Command) MarkFlagsMutuallyExclusive(flagNames ...string) {
 	}
 }
 
+// MarkIfFlagPresentThenOthersRequired marks the given flags so that if the first flag is set,
+// all the other flags become required.
+func (c *Command) MarkIfFlagPresentThenOthersRequired(flagNames ...string) {
+	if len(flagNames) < 2 {
+		panic("MarkIfFlagPresentThenRequired requires at least two flags")
+	}
+	c.mergePersistentFlags()
+	for _, v := range flagNames {
+		f := c.Flags().Lookup(v)
+		if f == nil {
+			panic(fmt.Sprintf("Failed to find flag %q and mark it as being in an if present then others required flag group", v))
+		}
+		// Each time this is called is a single new entry; this allows it to be a member of multiple groups if needed.
+		if err := c.Flags().SetAnnotation(v, ifPresentThenOthersRequired, append(f.Annotations[ifPresentThenOthersRequired], strings.Join(flagNames, " "))); err != nil {
+			panic(err)
+		}
+	}
+}
+
 // ValidateFlagGroups validates the mutuallyExclusive/oneRequired/requiredAsGroup logic and returns the
 // first error encountered.
 func (c *Command) ValidateFlagGroups() error {
@@ -90,10 +110,12 @@ func (c *Command) ValidateFlagGroups() error {
 	groupStatus := map[string]map[string]bool{}
 	oneRequiredGroupStatus := map[string]map[string]bool{}
 	mutuallyExclusiveGroupStatus := map[string]map[string]bool{}
+	ifPresentThenOthersRequiredGroupStatus := map[string]map[string]bool{}
 	flags.VisitAll(func(pflag *flag.Flag) {
 		processFlagForGroupAnnotation(flags, pflag, requiredAsGroupAnnotation, groupStatus)
 		processFlagForGroupAnnotation(flags, pflag, oneRequiredAnnotation, oneRequiredGroupStatus)
 		processFlagForGroupAnnotation(flags, pflag, mutuallyExclusiveAnnotation, mutuallyExclusiveGroupStatus)
+		processFlagForGroupAnnotation(flags, pflag, ifPresentThenOthersRequired, ifPresentThenOthersRequiredGroupStatus)
 	})
 
 	if err := validateRequiredFlagGroups(groupStatus); err != nil {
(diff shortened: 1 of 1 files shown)


faizan-siddiqui:
@marckhouzam Would love to get this in thanks! Let me know if you have any questions :)

faizan-siddiqui:
@dpetersen @markbates @EdwardBetts Would love to get this in thanks! Let me know if you have any questions :)

danpalmer:
@marckhouzam is there anything we can do to get this moving? Alternatively we're happy to have discussions about it being out of scope, it just felt fairly obviously useful to us.

marckhouzam:
I like the idea. I’ll put it on my queue to review. Thank you.

flag_groups.go
```diff
@@ -206,6 +231,38 @@ func validateExclusiveFlagGroups(data map[string]map[string]bool) error {
...
+				}
+			}
+
+			// If any dependent flags are unset, trigger an error
+			if len(unset) > 0 {
+				return fmt.Errorf(
+					"if the first flag in the group [%v] is set, all other flags must be set; the following flags are not set: %v",
+					flagList, unset,
+				)
+			}
```
ccoVeille, line 260:
```suggestion
			if len(unset) > 0 {
				return fmt.Errorf(
					"%v is set, all other flags [%v] must be set; the following flags are not set: %v",
					flagList[0], flagList[1:], unset,
				)
			}
```

Maybe something like this would be more readable and understood

Or this

```suggestion
			if len(unset) > 0 {
				return fmt.Errorf(
					"%v is set, the following flags must be provided: %v",
					flagList[0], unset,
				)
			}
```

flag_groups_test.go
```diff
@@ -43,22 +43,25 @@ func TestValidateFlagGroups(t *testing.T) {
...
 		{
 			desc: "No flags no problem",
 		}, {
-			desc:                "No flags no problem even with conflicting groups",
-			flagGroupsRequired:  []string{"a b"},
-			flagGroupsExclusive: []string{"a b"},
+			desc:                            "No flags no problem even with conflicting groups",
+			flagGroupsRequired:              []string{"a b"},
+			flagGroupsExclusive:             []string{"a b"},
+			flagGroupsIfPresentThenRequired: []string{"a b"},
```
ccoVeille, line 64: Can you add a test about using


```suggestion
			flagGroupsIfPresentThenRequired: []string{"a b", "b a"},
```

This one is similar to the required group that also exists.

Then, it's up to you to decide if it should be discouraged or not

Please also consider adding a test about

```suggestion
			flagGroupsIfPresentThenRequired: []string{"a b", "a c"},
```

Sane here, you should decide if you want to discourage it or not

Same about a test about

```suggestion
			flagGroupsIfPresentThenRequired: []string{"a b c", "a c b"},
```


```suggestion
			flagGroupsIfPresentThenRequired: []string{"a b c", "b c"},
```

I'm asking for all these tests because I'm unsure how it would be expected to work

flag_groups.go
```diff
@@ -221,6 +278,7 @@ func sortedKeys(m map[string]map[string]bool) []string {
 // - when a flag in a group is present, other flags in the group will be marked required
 // - when none of the flags in a one-required group are present, all flags in the group will be marked required
 // - when a flag in a mutually exclusive group is present, other flags in the group will be marked as hidden
+// - when the first flag in an if-present-then-required group is present, the second flag will be marked as required
```
ccoVeille, line 281:
```suggestion
// - when the first flag in an if-present-then-required group is present, the other flags will be marked as required
```

faizan-siddiqui:
Hi @ccoVeille thank you for the review, I have committed your suggested changes

flag_groups.go
```diff
@@ -76,6 +77,25 @@ func (c *Command) MarkFlagsMutuallyExclusive(flagNames ...string) {
...
+	if len(flagNames) < 2 {
+		panic("MarkIfFlagPresentThenRequired requires at least two flags")
+	}
+	c.mergePersistentFlags()
+	for _, v := range flagNames {
+		f := c.Flags().Lookup(v)
+		if f == nil {
+			panic(fmt.Sprintf("Failed to find flag %q and mark it as being in an if present then others required flag group", v))
+		}
+		// Each time this is called is a single new entry; this allows it to be a member of multiple groups if needed.
```
ccoVeille, line 92: Is this tested? I mean covered by a test that will ensure it will stay like this if there is a future refactoring

flag_groups.go
```diff
@@ -23,9 +23,10 @@ import (
 )
 
 const (
-	requiredAsGroupAnnotation   = "cobra_annotation_required_if_others_set"
-	oneRequiredAnnotation       = "cobra_annotation_one_required"
-	mutuallyExclusiveAnnotation = "cobra_annotation_mutually_exclusive"
+	requiredAsGroupAnnotation             = "cobra_annotation_required_if_others_set"
+	oneRequiredAnnotation                 = "cobra_annotation_one_required"
+	mutuallyExclusiveAnnotation           = "cobra_annotation_mutually_exclusive"
+	ifPresentThenOthersRequiredAnnotation = "cobra_annotation_if_present_then_others_required"
```
ccoVeille, line 29: Could you rename them in a way they share a common prefix, not suffix

These constants are defined at the package level, so IDE will suggest them in any method when coding in that package

I would like to see something like this


```suggestion
	annotationGroupRequired             = "cobra_annotation_required_if_others_set"
	annotationRequiredOne                 = "cobra_annotation_one_required"
	annotationMutuallyExclusive           = "cobra_annotation_mutually_exclusive"
	annotationGroupDependent = "cobra_annotation_if_present_then_others_required"
```

The names of the constants are just suggestions to give you an idea

faizan-siddiqui:
Thanks @ccoVeille I have addressed your comment and added an explicit test case for ensuring the annotation is appended as expected and is guarded against future refactorings, please review if you can :)

ccoVeille (approved):
Thanks for working on the changes I suggested.

Let's @marckhouzam review everything now

faizan-siddiqui:
Thanks @ccoVeille for helping out with the review and suggested changes
@marckhouzam would love to get this in, please let me know if any issues

faizan-siddiqui:
Hi @marckhouzam , @ccoVeille , is there anything else you may need to have this merged in? Let me know if there are any questions.

PR #2210, opened 2024-12-28
completion: improve detection for flags that accept multiple values

thaJeztah:
The completion code attempts to detect whether a flag can be specified more than once, and therefore should provide completion even if already set.

Currently, this code depends on conventions used in the pflag package, which uses an "Array" or "Slice" suffix or for some types a "stringTo" prefix.

Cobra allows custom value types to be used, which may not use the same convention for naming, and therefore currently aren't detected to allow multiple values.

The pflag module defines a [SliceValue] interface, which is implemented by the Slice and Array value types it provides (unfortunately, it's not currently implemented by the "stringTo" values).

This patch adds a reduced interface based on the [SliceValue] interface mentioned above to allow detecting Value-types that accept multiple values. Custom types can implement this interface to make completion work for those values.

I deliberately used a reduced interface to keep the requirements for this detection as low as possible, without enforcing the other methods defined in the interface (Append, Replace) which may not apply to all custom types.

Future improvements can likely still be made, considering either implementing the SliceValue interface for the "stringTo" values or defining a separate "MapValue" interface for those types.

Possibly providing the reduced interface as part of the pflag module and to export it.

[SliceValue]: https://github.com/spf13/pflag/blob/d5e0c0615acee7028e1e2740a11102313be88de1/flag.go#L193-L203

### DIFF
--- a/completions.go
+++ b/completions.go
@@ -270,6 +270,14 @@ func (c *Command) initCompleteCmd(args []string) {
 	}
 }
 
+// sliceValue is a reduced version of [pflag.SliceValue]. It is used to detect
+// flags that accept multiple values and therefore can provide completion
+// multiple times.
+type sliceValue interface {
+	// GetSlice returns the flag value list as an array of strings.
+	GetSlice() []string
+}
+
 func (c *Command) getCompletions(args []string) (*Command, []string, ShellCompDirective, error) {
 	// The last argument, which is not completely typed by the user,
 	// should not be part of the list of arguments
@@ -399,7 +407,10 @@ func (c *Command) getCompletions(args []string) (*Command, []string, ShellCompDi
 		// If we have not found any required flags, only then can we show regular flags
 		if len(completions) == 0 {
 			doCompleteFlags := func(flag *pflag.Flag) {
+				_, acceptsMultiple := flag.Value.(sliceValue)
+
 				if !flag.Changed ||
+					acceptsMultiple ||
 					strings.Contains(flag.Value.Type(), "Slice") ||
 					strings.Contains(flag.Value.Type(), "Array") ||
 					strings.HasPrefix(flag.Value.Type(), "stringTo") {
--- a/completions.go
+++ b/completions.go
@@ -270,10 +270,10 @@ func (c *Command) initCompleteCmd(args []string) {
 	}
 }
 
-// sliceValue is a reduced version of [pflag.SliceValue]. It is used to detect
+// SliceValue is a reduced version of [pflag.SliceValue]. It is used to detect
 // flags that accept multiple values and therefore can provide completion
 // multiple times.
-type sliceValue interface {
+type SliceValue interface {
 	// GetSlice returns the flag value list as an array of strings.
 	GetSlice() []string
 }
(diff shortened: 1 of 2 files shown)


completions.go
```diff
@@ -270,6 +270,14 @@ func (c *Command) initCompleteCmd(args []string) {
 	}
 }
 
+// sliceValue is a reduced version of [pflag.SliceValue]. It is used to detect
+// flags that accept multiple values and therefore can provide completion
+// multiple times.
+type sliceValue interface {
```
marckhouzam: I think that if we want to allow programs to do
`var _ sliceValue = (*customMultiString)(nil)` like in the tests, we need to export the type.

marckhouzam:
I pushed a commit addressing the comment, and doing a slight refactoring.

marckhouzam (approved):
Very nice!
@thaJeztah If you have comments on the commit I pushed on top of yours, please let me know and we can address in a follow-up

thaJeztah:
Thanks! Changes  make sense;  I wasn't sure if we wanted a more solid / generic approach (possibly a dedicated interface or perhaps annotations for flags), hence I initially chose not to export the interface to keep some wiggle-room  😅

thaJeztah:
@marckhouzam are you a maintainer for `spf13/pflag` as well? I recently noticed that some  of our dependencies  started to force  an untagged version of the package (current master) through go modules; was wondering if a patch /  minor release is planned so that we can go back to using a tagged release in those codebases; https://github.com/spf13/pflag/compare/v1.0.5...d5e0c0615acee7028e1e2740a11102313be88de1

marckhouzam:
No, I’m not a maintainer for pflag

completions.go
```diff
@@ -270,6 +270,14 @@ func (c *Command) initCompleteCmd(args []string) {
 	}
 }
 
+// SliceValue is a reduced version of [pflag.SliceValue]. It is used to detect
+// flags that accept multiple values and therefore can provide completion
+// multiple times.
+type SliceValue interface {
+	// GetSlice returns the flag value list as an array of strings.
+	GetSlice() []string
+}
```
ccoVeille, line 279: Can you do something like this either here or in the tests?

(This is untested pseudocode written on a phone, please check and adapt)

```suggestion
// SliceValue is a reduced version of [pflag.SliceValue]. It is used to detect
// flags that accept multiple values and therefore can provide completion
// multiple times.
type SliceValue interface {
	// GetSlice returns the flag value list as an array of strings.
	GetSlice() []string
}

var _ SliceValue = (*pflag.SliceValue)(nil)
```

The idea is to validate pflag.SliceValue implements the interface you define

completions_test.go
```diff
@@ -671,6 +671,29 @@ func TestFlagNameCompletionInGoWithDesc(t *testing.T) {
...
 }
 
+// customMultiString is a custom Value type that accepts multiple values,
+// but does not include "Slice" or "Array" in its "Type" string.
+type customMultiString []string
+
+var _ SliceValue = (*customMultiString)(nil)
+
+func (s *customMultiString) String() string {
+	return fmt.Sprintf("%v", *s)
```
ccoVeille, line 681: What about this?

```suggestion
	return fmt.Sprint(*s)
```

completions.go
ccoVeille: @thaJeztah I'm unsure if you saw my comment last week

completions_test.go
ccoVeille: Samr here

PR #2223, opened 2025-01-22
Add aliases to doc markdown

divolgin:
Adds aliases to generated markdown docs, similar to how it's handled in the default help deplate:

```
Aliases:
  {{.NameAndAliases}}{{end}}{{if .HasExample}}
```


marckhouzam:
So this would change the existing behavior? I wonder if this is a backwards compatibility issue?

divolgin:
> So this would change the existing behavior? I wonder if this is a backwards compatibility issue?

It changes generated doc format by adding the list of aliases. I guess you could call it "changes to existing behavior".

marckhouzam:
I think this would be safe to add but I’m not sure. 
@jpmcb  do you think this could break existing programs?

PR #2224, opened 2025-01-29
fix: indentation level of examples section

johanfagerstroem:
Fix indentation level of the examples section in the usage template.

### DIFF
--- a/command.go
+++ b/command.go
@@ -1931,7 +1931,7 @@ Aliases:
   {{.NameAndAliases}}{{end}}{{if .HasExample}}
 
 Examples:
-{{.Example}}{{end}}{{if .HasAvailableSubCommands}}{{$cmds := .Commands}}{{if eq (len .Groups) 0}}
+  {{.Example}}{{end}}{{if .HasAvailableSubCommands}}{{$cmds := .Commands}}{{if eq (len .Groups) 0}}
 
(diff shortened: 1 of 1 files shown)


marckhouzam:
Thanks @johanfagerstroem .  Although the indentation you suggest is how the help template should always have been, it is too late to change it now as it would affect programs that have worked around it by inserting their own indentation in their example string. 

So we just have to leave things as is.

johanfagerstroem:
Thank you for responding @marckhouzam! Would you accept a strings.TrimPrefix()?

marckhouzam:
> Thank you for responding @marckhouzam! Would you accept a strings.TrimPrefix()?

I think that is still risky: maybe a program has invented their examples with four spaces, instead of the two cobra would be putting.

johanfagerstroem:
Doesn't matter - if we remove _all_ prefixing whitespace chars, and then prepend two, the net will always be two.

marckhouzam:
Not if the examples are multiline.

marckhouzam:
Here’s an example of a format I’m concerned about (note that in this case, the Tanzu CLI overrides the template completely, but it shows that the format could be multi line with more than two spaces leading each line).
https://github.com/vmware-tanzu/tanzu-cli/blob/b4a85781d178956dbb6497f98fb0bd9351f795bf/pkg/command/cert.go#L139-L150

johanfagerstroem:
Got it - I'll wrap it somehow. No big deal. Thank you for your time!

PR #2229, opened 2025-02-02
Allow to reset the templates to the default

marckhouzam:
Follow-up to #1956.

This commit allows a program to reset any of the three templates to their default behaviour, as it was possible to do before the change of #1956.

For example, one could imagine that a program could call `SetUsageTemplate("{{A short help}}")`, but if a particular set of arguments is passed in by the user, the program would want to reset to the default help by calling `SetUsageTemplate("")` (basically cancelling the previous call to `SetUsageTemplate()`).

This used to be possible before #1956, and this PR brings back this support.

Granted, this is a bit of a corner case, but with so many programs using Cobra, let's not make any assumptions and avoid unexpected changes. 

cc @aarzilli @jpmcb

### DIFF
--- a/command.go
+++ b/command.go
@@ -315,6 +315,10 @@ func (c *Command) SetUsageFunc(f func(*Command) error) {
 
 // SetUsageTemplate sets usage template. Can be defined by Application.
 func (c *Command) SetUsageTemplate(s string) {
+	if s == "" {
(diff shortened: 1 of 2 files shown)


aarzilli (approved):
LGTM -- good catch

jpmcb (approved):

PR #2230, opened 2025-02-06
doc: add Conduit

raulb:
Hi! 

We recently added a CLI to a our Conduit project, and we use Cobra to power it.

Here's a blog post mentioning it https://meroxa.com/blog/introducing-the-new-conduit-cli:-a-powerful-tool-for-managing-your-pipelines/

Thanks!

### DIFF
--- a/site/content/projects_using_cobra.md
+++ b/site/content/projects_using_cobra.md
@@ -8,6 +8,7 @@
 - [Cilium](https://cilium.io/)
 - [CloudQuery](https://github.com/cloudquery/cloudquery)
 - [CockroachDB](https://www.cockroachlabs.com/)
+- [Conduit](https://github.com/conduitio/conduit)
 - [Constellation](https://github.com/edgelesssys/constellation)
 - [Cosmos SDK](https://github.com/cosmos/cosmos-sdk)
 - [Datree](https://github.com/datreeio/datree)


marckhouzam (approved):
Nice!

I couldn't help notice that you have deactivated the `conduit completion` command.
(https://github.com/ConduitIO/conduit/blob/b145fc42f22ab26381af3e2c3d5c8883be28cdb5/cmd/conduit/main.go#L29)
If you don't like that conduit has a `completion` command, you can hide it instead of deactivating it (`cmd.CompletionOptions.HiddenDefaultCmd = true
`).  But then you can still tell your users that they can use shell completion with conduit.

I've just tried it with zsh (with a little cobra trick) and it seems to works fine.
```
# these two first lines use the cobra generic completion script
# from any cobra tool, in this case kubectl
$ source <(kubectl completion zsh)
$ compdef _kubectl conduit

$ conduit <TAB>
config             -- Shows the configuration to be used when running Conduit.
connector-plugins  -- Manage Connector Plugins
connectors         -- Manage Conduit Connectors
help               -- Help about any command
init               -- Initialize Conduit with a configuration file and directories.
open               -- Open in a web browser
pipelines          -- Initialize and manage pipelines
processor-plugins  -- Manage Processor Plugins
processors         -- Manage Processors
run                -- Run Conduit
version            -- Show the current version of Conduit.
```

PR #2233, opened 2025-02-15
chore: Upgrade dependencies for v1.9.0

jpmcb:
Quick dependency upgrades in preparation of v1.9.0 - cc @marckhouzam - used:

```
go get -u ./...
```

to pull the latest for us to consume: doesn't look like much, just `pflag`


marckhouzam (approved):
LGTM
Thanks

PR #2238, opened 2025-02-18
The default ShellCompDirective can be customized for a command and its subcommands

### LINKED ISSUE #2209: Feature request: Set a default completion function
In my experience, not many flags actually require completion of files.

Some have dynamic completions, but a great deal of them take free-form string parameters where no completion can be offered.
Unfortunately, the default completion for a flag with no completion function in Cobra is file completion.
This forces me to register a `cobra.NoFileCompletions` for each flag of that category.

As an example, completion for `docker run` has 56 flags where no meaningful FlagCompletionFunc exists.
In a first attempt, I registered 56 handlers with `cmd.RegisterFlagCompletionFunc("flag-name", cobra.NoFileCompletions)` just to silence their file completion.
Following a hint of @thaJeztah, I replaced them with a programmatically installed default handler, see [here](https://github.com/docker/cli/commit/06260e68f3686a0368284a8ce9b910e056bb523b#diff-50fe347eb4bdafebc190aa42b2caa34b089a538cd444b8a151a763372e93c0cf).

I would very much welcome a solution in Cobra, where I can set a default handler on command level, something like

```go
cmd.CompletionOptions.DefaultCompHandler = cobra.NoFileCompletions
```

With this feature, I'd only have to define the completion functions for flags that actually have meaningful completions.

/cc @thaJeztah

albers:
Closes #2209, #2221 

#2221 proposed a solution where the custom `DefaultShellCompDirective` only applied to a given command.

In https://github.com/spf13/cobra/pull/2221#discussion_r1923845125, alternative approaches were discussed.
We agreed on a recursive solution, where a change applies to a command and its subcommands.

To ease review, I open a new PR for this approach.

### DIFF
--- a/completions.go
+++ b/completions.go
@@ -115,6 +115,13 @@ type CompletionOptions struct {
 	DisableDescriptions bool
 	// HiddenDefaultCmd makes the default 'completion' command hidden
 	HiddenDefaultCmd bool
+	// DefaultShellCompDirective sets the ShellCompDirective that is returned
+	// if no special directive can be determined
+	DefaultShellCompDirective *ShellCompDirective
+}
+
+func (receiver *CompletionOptions) SetDefaultShellCompDirective(directive ShellCompDirective) {
+	receiver.DefaultShellCompDirective = &directive
 }
 
 // Completion is a string that can be used for completions
@@ -480,6 +487,14 @@ func (c *Command) getCompletions(args []string) (*Command, []Completion, ShellCo
 		}
 	} else {
 		directive = ShellCompDirectiveDefault
+		// check current and parent commands for a custom DefaultShellCompDirective
+		for cmd := finalCmd; cmd != nil; cmd = cmd.parent {
+			if cmd.CompletionOptions.DefaultShellCompDirective != nil {
+				directive = *cmd.CompletionOptions.DefaultShellCompDirective
+				break
+			}
+		}
+
 		if flag == nil {
 			foundLocalNonPersistentFlag := false
 			// If TraverseChildren is true on the root command we don't check for
--- a/completions_test.go
+++ b/completions_test.go
@@ -4016,3 +4016,60 @@ func TestInitDefaultCompletionCmd(t *testing.T) {
 		})
 	}
 }
+
+func TestCustomDefaultShellCompDirective(t *testing.T) {
+	rootCmd := &Command{Use: "root", Run: emptyRun}
+	rootCmd.PersistentFlags().String("string", "", "test string flag")
+	// use ShellCompDirectiveNoFileComp instead of the default, which is ShellCompDirectiveDefault.
+	rootCmd.CompletionOptions.SetDefaultShellCompDirective(ShellCompDirectiveNoFileComp)
+
+	// child1 inherits the custom DefaultShellCompDirective.
+	childCmd1 := &Command{Use: "child1", Run: emptyRun}
+	// child2 resets the custom DefaultShellCompDirective to the default value.
+	childCmd2 := &Command{Use: "child2", Run: emptyRun}
+	childCmd2.CompletionOptions.SetDefaultShellCompDirective(ShellCompDirectiveDefault)
+
+	rootCmd.AddCommand(childCmd1, childCmd2)
+
+	testCases := []struct {
+		desc              string
+		args              []string
+		expectedDirective string
+	}{
+		{
+			"flag completion on root command with custom DefaultShellCompDirective",
+			[]string{"--string", ""},
+			"ShellCompDirectiveNoFileComp",
+		},
+		{
+			"flag completion on subcommand with inherited custom DefaultShellCompDirective",
+			[]string{"child1", "--string", ""},
+			"ShellCompDirectiveNoFileComp",
+		},
+		{
+			"flag completion on subcommand with reset DefaultShellCompDirective",
+			[]string{"child2", "--string", ""},
+			"ShellCompDirectiveDefault",
+		},
+	}
+
+	for _, tc := range testCases {
+		t.Run(tc.desc, func(t *testing.T) {
+			args := []string{ShellCompNoDescRequestCmd}
+			args = append(args, tc.args...)
+
+			output, err := executeCommand(rootCmd, args...)
+
+			if err != nil {
+				t.Errorf("Unexpected error: %v", err)
+			}
+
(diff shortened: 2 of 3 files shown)


completions.go
```diff
@@ -115,6 +115,13 @@ type CompletionOptions struct {
 	DisableDescriptions bool
 	// HiddenDefaultCmd makes the default 'completion' command hidden
 	HiddenDefaultCmd bool
+	// DefaultShellCompDirective sets the ShellCompDirective that is returned
+	// if no special directive can be determined
+	DefaultShellCompDirective *ShellCompDirective
```
ccoVeille, line 120: I'm a bit annoyed by something here

We have an exported field with a pointer.

I remember the discussion that occurred in previous PR about that.

But here, the value the user may expect to be able to pass this field when creating an instance of the struct.

But there is a problem, the existing values they may want to pass are constants (with a iota btw)

But then... You cannot pass a reference to a constant.

I mean this cannot be written

```go
_ = cobra.Command{
	       DefaultShellCompDirective: &cobra.ShellCompDirectiveNoFileComp
}
```

So something else need to be found. i think

Either removing the pointer, using variables and not constants, adding a setter...

completions.go
```diff
@@ -115,6 +115,13 @@ type CompletionOptions struct {
...
+}
+
+func (receiver *CompletionOptions) SetDefaultShellCompDirective(directive ShellCompDirective) {
+	receiver.DefaultShellCompDirective = &directive
```
marckhouzam, line 124: if the above suggestion works, we can remove this

completions.go
marckhouzam: Ahhh, my suggestion doesn’t make sense because it’s the compiler that sets the value of `DefaultShellCompDirective` to 0 for every command. That’s why the pointer works because the compiler sets it to nil. 

I need to think about this

albers: In order to support the use case of a child command resetting the `DefaultShellCompDirective` to `ShellCompDirectiveDefault` while its parent has a custom `DefaultShellCompDirective` (see [test](https://github.com/spf13/cobra/pull/2238/files#diff-0a8ea1f7695bd9a687f50ecd2d3b7d0f0c04d535045d01d3209320ca8cff4597R4028-R4030)), we need a way to initialize the value of `DefaultShellCompDirective` to a new default value like `ShellCompDirectiveUndefined` or a pointer, while still preserving backwards compatibility.

My first approch may feel a bit clumsy, but it definitely fulfills the requirements. You can create the command with either

```go
	cmd := &Command{Use: "command", Run: emptyRun}
	cmd.CompletionOptions.SetDefaultShellCompDirective(ShellCompDirectiveNoFileComp)
```

or
```go
	directive := ShellCompDirectiveNoFileComp

	rootCmd := &Command{
		Use:               "root",
		Run:               emptyRun,
		CompletionOptions: CompletionOptions{DefaultShellCompDirective: &directive},
	}
```

We could improve (?) this approach by making `DefaultShellCompDirective` private and introduce a getter so that we have a symmetric access pattern that does not tempt users to directly set it to the constant.

The only alternative I see is to create a factory method `newCommand` that takes care of initializing `DefaultShellCompDirective`, something like

```go
    cmd := newCommand(&Command{Use: "command", Run: emptyRun})
```

But this is definitely not an improvement to the API.

marckhouzam: @albers your examples have convinced me that this is good enough. Let’s go with it as is, if you can just fix the docs

ccoVeille: I don't get where we are going.

If the argument was "yes, the setter approach was better", I would have say yes. But, here it sounds like something strange to me.

Expecting people to deal with pointer on constants. It looks a bit odd. 

here are some suggestions:

- do not use pointer, but add a completion option boolean that could be used for the parent setting

```go
	childCmd := &Command{
		Use:               "foo",
		CompletionOptions: CompletionOptions{UseParentDefaultShellCompDirective: true,},
	}
```

I'm unsure this one makes sense, but maybe it would help you guys to find something better around a boolean.

- keep the pointers, but change the type of ShellComp from constants to variables

people could use

```go
	rootCmd := &Command{
		Use:               "root",
		Run:               emptyRun,
		CompletionOptions: CompletionOptions{DefaultShellCompDirective: &ShellCompDirectiveNoFileComp},
	}
```

- add a function helper

```go
func ShellCompDirectiveHelper(so ShellCompDirective) *ShellCompDirective{
	return &so
}
```

```go
	rootCmd := &Command{
		Use:               "root",
		Run:               emptyRun,
		CompletionOptions: CompletionOptions{DefaultShellCompDirective: ShellCompDirectiveHelper(ShellCompDirectiveNoFileComp)},
	}
```

albers: > ```go
> 	childCmd := &Command{
> 		Use:               "foo",
> 		CompletionOptions: CompletionOptions{UseParentDefaultShellCompDirective: true,},
> 	}
> ```

This approach does not fulfill the requirements because it is not recursive.
The user would have to define the property on each subcommand that should inherit the setting.

> * keep the pointers, but change the type of ShellComp from constants to variables

The `ShellCompDirectives` are constants for a good reason.
The should not be modifyable.

> * add a function helper
> 
> ```go
> func ShellCompDirectiveHelper(so ShellCompDirective) *ShellCompDirective{
> 	return &so
> }
> ```
Well, we could add a `ToPointer` function for that purpose, WDYT @marckhouzam ?

completions.go
albers: We still need this.

albers:
@marckhouzam I updated the documentation.
Maybe we should drop a word about handling of constants in pointers and offer a helper as suggested by @ccoVeille so that the feature can be used either as
```go
	cmd := &Command{Use: "command", Run: emptyRun}
	cmd.CompletionOptions.SetDefaultShellCompDirective(ShellCompDirectiveNoFileComp)
```
or 
```go
	rootCmd := &Command{
		Use:               "root",
		Run:               emptyRun,
		CompletionOptions: CompletionOptions{DefaultShellCompDirective: ToPointer(ShellCompDirectiveNoFileComp)},
	}
```

albers:
ping @marckhouzam

albers:
@marckhouzam Is there anything I can do to move this PR forward?

marckhouzam (commented):
Sorry for the delay.
Thanks @albers !

marckhouzam (approved):

Issue #2240, opened 2025-02-20
Command passed to SetHelpFunc doesn't have a context


Frassle:
When using `SetHelpFunc` the function gets given a `Command` object and you'd think you could use `Context()` to get the right context off the command, but it doesn't actually get set. 

We hit this in pulumi: https://github.com/pulumi/pulumi/issues/18659

I can work around via just a context.Background call instead, but feels like this should just work as the parent "help" command does have a context.

Frassle:
Will see if I can raise a PR to fix this.

PR #2241, opened 2025-02-20
Flow context to command in SetHelpFunc

Frassle:
Fixes https://github.com/spf13/cobra/issues/2240

### DIFF
--- a/command_test.go
+++ b/command_test.go
@@ -2921,3 +2921,30 @@ func TestUnknownFlagShouldReturnSameErrorRegardlessOfArgPosition(t *testing.T) {
 		})
 	}
 }
+
+func TestHelpFuncExecuted(t *testing.T) {
+	helpText := "Long description"
+
+	child := &Command{Use: "child", Run: emptyRun}
+	child.SetHelpFunc(func(cmd *Command, args []string) {
+		_, err := cmd.OutOrStdout().Write([]byte(helpText))
+		if err != nil {
+			t.Error(err)
+		}
+
+		// Test for https://github.com/spf13/cobra/issues/2240
+		if cmd.Context() == nil {
+			t.Error("Context is nil")
+		}
+	})
+
+	rootCmd := &Command{Use: "root", Run: emptyRun}
+	rootCmd.AddCommand(child)
+
+	output, err := executeCommand(rootCmd, "help", "child")
+	if err != nil {
+		t.Errorf("Unexpected error: %v", err)
+	}
+
+	checkStringContains(t, output, helpText)
+}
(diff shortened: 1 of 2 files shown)


command_test.go
```diff
@@ -2921,3 +2921,29 @@ func TestUnknownFlagShouldReturnSameErrorRegardlessOfArgPosition(t *testing.T) {
...
+
+	rootCmd := &Command{Use: "root", Run: emptyRun}
+	rootCmd.AddCommand(child)
+
+	output, err := executeCommand(rootCmd, "help", "child")
+	if err != nil {
+		t.Errorf("Unexpected error: %v", err)
+	}
+
+	checkStringContains(t, output, rootCmd.Long)
```
marckhouzam, line 2945: I don't get this check here...
`rootCmd.Long` is empty; I assume you meant to use `child.Long`?
But then the `output` is also empty because the `HelpFunc` does not actually print any help.  I assume it should, so you can verify it?  Then you don't need to do the `helpExecuted` check.

marckhouzam (commented):
This seems reasonable.  Thanks @Frassle !

command_test.go
Frassle: Ah copy paste without full understanding :) I'll fix it up to do the print

command_test.go
```diff
@@ -2921,3 +2921,30 @@ func TestUnknownFlagShouldReturnSameErrorRegardlessOfArgPosition(t *testing.T) {
...
+
+	child := &Command{Use: "child", Run: emptyRun}
+	child.SetHelpFunc(func(cmd *Command, args []string) {
+		_, err := cmd.OutOrStdout().Write([]byte(helpText))
+		if err != nil {
+			t.Error(err)
+		}
+
+		// Test for https://github.com/spf13/cobra/issues/2240
+		if cmd.Context() == nil {
```
marckhouzam, line 2936: This looks much better, thanks

 I’ve been thinking about what we’re trying to achieve with this PR. I think we wanna make sure that the correct context can be retrieved by the help function. Right?

Therefore, I was thinking that in this test we should set a context on the root command and then instead of checking for nil here, we should check that we got the same context than the one we created. 

What do you think @Frassle ?

Frassle: Yes that seems reasonable to check in the test.

command_test.go
```diff
@@ -2921,3 +2921,34 @@ func TestUnknownFlagShouldReturnSameErrorRegardlessOfArgPosition(t *testing.T) {
 		})
 	}
 }
+
+func TestHelpFuncExecuted(t *testing.T) {
+	helpText := "Long description"
+
+	// Create a context that will be unique, not just the background context
+	executionCtx, cancel := context.WithCancel(context.Background())
```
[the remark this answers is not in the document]
Frassle: Ah that's cleaner, I'd just thrown a "defer cancel()" on

marckhouzam (approved):
Thanks!

PR #2246, opened 2025-03-04
Allow to override build date with SOURCE_DATE_EPOCH

bmwiedemann:
Allow to override build date with `SOURCE_DATE_EPOCH`
to make builds reproducible.
See https://reproducible-builds.org/ for why this is good and https://reproducible-builds.org/specs/source-date-epoch/ for the definition of this variable.

This patch was done while working on reproducible builds for openSUSE.


doc/rest_docs.go
```diff
@@ -123,7 +124,14 @@ func GenReSTCustom(cmd *cobra.Command, w io.Writer, linkHandler func(string, str
 		buf.WriteString("\n")
 	}
 	if !cmd.DisableAutoGenTag {
-		buf.WriteString("*Auto generated by spf13/cobra on " + time.Now().Format("2-Jan-2006") + "*\n")
+		now := time.Now()
+		if epoch := os.Getenv("SOURCE_DATE_EPOCH"); epoch != "" {
+			unixEpoch, err := strconv.ParseInt(epoch, 10, 64)
+			if err == nil {
```
marckhouzam, line 130: According to the spec, if the variable is poorly formatted, we should return an error.
Can you handle it like it is done here
https://github.com/spf13/cobra/blob/4f9ef8cdbbc88c5302be95e0e67fd78ebbfa9dd2/doc/man_docs.go#L129C4-L131C5

marckhouzam (commented):
Thanks for this.
Somehow Cobra already used this variable in man_docs.go but not in the other formats.
Would you mind also adding this check in md_docs.go?

It would also be nice to have a test to make sure the variable is respected.

doc/rest_docs.go
bmwiedemann: I wanted to keep the code simpler, because in practice, the variable is either unset or valid and we have used this simplification in other places.
But if you insist, we can copy the approach from man_docs. Is code-duplication okay or should it moved out somehow somewhere (I'm not so familiar with golang)

PR #2248, opened 2025-03-05
Update command.go

styee:
Fix spelling in the help output.

topcis -> topics

### DIFF
--- a/command.go
+++ b/command.go
@@ -2025,7 +2025,7 @@ func defaultUsageFunc(w io.Writer, in interface{}) error {
 		fmt.Fprint(w, trimRightSpace(c.InheritedFlags().FlagUsages()))
 	}
 	if c.HasHelpSubCommands() {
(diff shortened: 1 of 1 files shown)


marckhouzam (approved):
I can’t believe I missed that 😞
Thanks for fixing it. 
The additional help topics are very rarely used so I’m hoping this won’t require a patch release. How do you feel about it @styee ?

styee:
Thanks! Very likely nobody has noticed to this point, I don't think it needs a patch release.

marckhouzam:
Cutting a new release is not a big deal from the cobra side. It’s just that it might have so many projects feel like they have to look into the new release that I’m hesitant to do small ones like this. 
cc @jpmcb 

@caarlos0 would you be able to just use the latest main branch for the moment?

caarlos0:
yes, no worries :)

Issue #2257, opened 2025-03-12
Completions modify os.Args


jonathan-conder:
I'm trying to workaround https://github.com/spf13/cobra/issues/1877 by inspecting `os.Args` by hand. In some cases cobra accidentally inserts `--` into `os.Args`.

Minimal example:
```go
package main

import (
	"os"
	"strings"

	"github.com/spf13/cobra"
)

func main() {
	cmd := &cobra.Command{
		TraverseChildren: true,
		ValidArgsFunction: func(cmd *cobra.Command, args []string, toComplete string) ([]cobra.Completion, cobra.ShellCompDirective) {
			cobra.CompErrorln(strings.Join(os.Args[1:], ", "))
			return nil, cobra.ShellCompDirectiveDefault
		},
	}
	if err := cmd.Execute(); err != nil {
		panic(err)
	}
}

```

Expected behaviour:
```console
$ go run . __complete x
[Debug] [Error] __complete, x
:0
Completion ended with directive: ShellCompDirectiveDefault
```

Actual behaviour:
```console
$ go run . __complete x
[Debug] [Error] __complete, --
:0
Completion ended with directive: ShellCompDirectiveDefault
```

marckhouzam:
Apologies for missing this @jonathan-conder .  Clearly this is a bug. Thanks @veeceey for opening #2356  to address the problem

veeceey:
> Apologies for missing this [@jonathan-conder](https://github.com/jonathan-conder) . Clearly this is a bug. Thanks [@veeceey](https://github.com/veeceey) for opening [#2356](https://github.com/spf13/cobra/pull/2356) to address the problem

Thanks Marck my pleasure!!

Issue #1877, opened 2022-12-12
Custom autocompletion after double dash (`--`)


richard-kramer:
I'm currently trying out the cobra framework and stumbled upon an issue with dynamic autocompletion and `cmd.ArgsLenAtDash()`.

I'm trying to provide autocompletion for a predefined set of possible positional parameters until a `--` is added. Everything after that should be passed down to another command (or otherwise processed differently to all `args` before the double dash) and should provide different autocompletion results.

Unfortunately, because of the added `--` [here](https://github.com/spf13/cobra/blob/923592041e4ae38e2ae5c3cc66c3eb1e0c27261f/completions.go#L294), the result of `cmd.ArgsLenAtDash()` inside `ValidArgsFunction` is always at least `0`, even when there is no `--` in my command. This makes it impossible to detect, if a `--` **hasn't** been added to the arguments yet. As soon as I add another argument `behind` the `--`, `cmd.ArgsLenAtDash()` returns the correct result.

Now I'm not sure, if this is simply not a supported use case, if I'm missing something or if this is actually a bug.

### Example Code

This is an example `cmd/test.go` file inside an initialized cobra project, added to the root command.

```go
// cmd/test.go

// …

var validArgs = []string{"foo", "bar", "baz"}

// TestCmd represents the test command
var TestCmd = &cobra.Command{
  Use:   "test",
  ValidArgsFunction: func(cmd *cobra.Command, args []string, toComplete string) ([]string, cobra.ShellCompDirective) {
    cobra.CompDebugln(fmt.Sprint(cmd.ArgsLenAtDash()), true) // => always at least `0`, even when it should be `-1`

    possibleArgs := append(validArgs, "--")

    if cmd.ArgsLenAtDash() < 0 {
      return possibleArgs, cobra.ShellCompDirectiveNoFileComp
    }

    return nil, cobra.ShellCompDirectiveDefault
  },
  Run: func(cmd *cobra.Command, args []string) {
    cmdArgsEnd := len(args)
    if cmd.ArgsLenAtDash() > -1 {
      cmdArgsEnd = cmd.ArgsLenAtDash()
    }

    fmt.Println("test called")
    fmt.Println()
    fmt.Printf("args: %+q\n", args[:cmdArgsEnd])
    fmt.Printf("ArgsLenAtDash: %d\n", cmd.ArgsLenAtDash())
    fmt.Printf("args after --: %+q\n", args[cmdArgsEnd:])
  },
}

// …
```

[...]

PR #2258, opened 2025-03-12
fix: Prevent getCompletions from modifying os.Args

jonathan-conder:
Possible fix for https://github.com/spf13/cobra/issues/2257.

Tested using this test case:
```go
func TestModifiedArgs(t *testing.T) {
	cmd := &Command{TraverseChildren: true}
	args := []string{"__complete", ""}
	cmd.SetArgs(args)
		t.Errorf("Unexpected error: %v", err)
	}
	if len(args) != 2 || args[1] != "" {
		t.Error("Modified args")
	}
}
```


marckhouzam:
Before going this route, let's see if we can improve pflag with the required API as suggested in #1877

PR #2271, opened 2025-04-25
Add function to print command aliases

commit111:
This PR adds a function in `md_doc` to print the list of Aliases of a command in the markdown

### DIFF
--- a/doc/md_docs.go
+++ b/doc/md_docs.go
@@ -29,6 +29,16 @@ import (
 
 const markdownExtension = ".md"
 
+func printAliases(buf *bytes.Buffer, cmd *cobra.Command) {
+	if len(cmd.Aliases) > 0 {
+		buf.WriteString("### Aliases\n\n")
+		for _, alias := range cmd.Aliases {
+			buf.WriteString(fmt.Sprintf("* `%s`\n", alias))
+		}
+		buf.WriteString("\n")
+	}
+}
+
 func printOptions(buf *bytes.Buffer, cmd *cobra.Command, name string) error {
 	flags := cmd.NonInheritedFlags()
 	flags.SetOutput(buf)


ccoVeille:
Is this a WIP PR?

Unless I'm wrong, nothing calls this new method. So what is expected here?

doc/md_docs.go
```diff
@@ -72,6 +72,14 @@ func GenMarkdownCustom(cmd *cobra.Command, w io.Writer, linkHandler func(string)
 		buf.WriteString(fmt.Sprintf("```\n%s\n```\n\n", cmd.UseLine()))
 	}
 
+	if len(cmd.Aliases) > 0 {
+		buf.WriteString("### Aliases\n\n")
+		for _, alias := range cmd.Aliases {
+			buf.WriteString(fmt.Sprintf("* `%s`\n", alias))
+		}
+		buf.WriteString("\n")
+	}
```
ccoVeille, line 81: 🎨 this is a pure stylistic feedbacks.

It's open for debate.

Maybe the rendering could be different when there is only one alias.

Because this might seem empty as a section

```
### Aliases

* foobar
```

So, here maybe another rendering could be considered.

When multiple aliases are available

```
Aliases: foobar, fooquz
```

And when there is only one

```
Aliases: foobar
```

Please note, I didn't use a section on purpose

commit111:
Ok, I am done with changing it.

doc/md_docs.go
```diff
@@ -72,6 +72,17 @@ func GenMarkdownCustom(cmd *cobra.Command, w io.Writer, linkHandler func(string)
...
 	}
 
+	if len(cmd.Aliases) > 0 {
+		buf.WriteString("Aliases: ")
+		for _, alias := range cmd.Aliases {
+			buf.WriteString(fmt.Sprintf("`%s`", alias))
+			if alias != cmd.Aliases[len(cmd.Aliases)-1] {
+				buf.WriteString(", ")
+			}
+		}
```
ccoVeille, line 82: Your code is fine.

I would have used 

```go
"`"+strings.Join("`, `", cmd.Aliases)+"`"
```

But your code somehow respects the existing code. So why not. But then I would suggest you something like this

```suggestion
		for i, alias := range cmd.Aliases {
			if i > 0 {
				buf.WriteString(", ")
			}
			buf.WriteString(fmt.Sprintf("`%s`", alias))
		}
```

I way simpler to read and maintain

commit111: I agree

commit111: The reason I made it the way it was before was so that the last alias would not have a comma

doc/md_docs.go
```diff
@@ -72,6 +72,17 @@ func GenMarkdownCustom(cmd *cobra.Command, w io.Writer, linkHandler func(string)
 		buf.WriteString(fmt.Sprintf("```\n%s\n```\n\n", cmd.UseLine()))
 	}
 
+	if len(cmd.Aliases) > 0 {
+		buf.WriteString("Aliases: ")
+		for i, alias := range cmd.Aliases {
+			if i > 0 {
+				buf.WriteString(", ")
+			}
+			buf.WriteString(fmt.Sprintf("`%s`", alias))
```
ccoVeille, line 81: Here I'm surprised by this syntax, that was already used in the file I concede

I feel like this would be more natural.

> [!NOTE]
> This is more a note to myself to look deeper than something I expect you to change

This could have been used

```suggestion
			fmt.Fprintf(buf, "`%s`", alias)
```

@commit111 please look at this that seems a better option 

https://github.com/spf13/cobra/pull/2271#discussion_r2060904663

doc/md_docs.go
```diff
@@ -72,6 +72,17 @@ func GenMarkdownCustom(cmd *cobra.Command, w io.Writer, linkHandler func(string)
...
+		}
+		buf.WriteString("\n")
+	}
```
ccoVeille, line 84: Please note this could be simpler, and faster


```suggestion
		buf.WriteString("Aliases: `")
		for i, alias := range cmd.Aliases {
			if i > 0 {
				buf.WriteString("`, `")
			}
			buf.WriteString(alias)
		}
		buf.WriteString("`\n")
	}
```

ccoVeille (approved):
LGTM 👍

Thanks

Let's now wait for @marckhouzam to review it

lionello:
@marckhouzam did you get a chance to look at this?

PR #2272, opened 2025-04-25
Simplify aliases code

commit111:
Simplify the code for printing command aliases


commit111:
Closed because I unintentionally pushed this to the wrong upstream.

PR #2275, opened 2025-05-23
feat: Add ClearInitializers and ClearFinalizers method

Kirdock:
These methods let you clear any set initializer and finalizer.

## Why this change?
You may want to start tests cleanly. TestA could add finalizers, and another one, TestB, could also add some, but then TestB also has and executes TestA's finalizers.

### DIFF
--- a/cobra.go
+++ b/cobra.go
@@ -100,10 +100,20 @@ func OnInitialize(y ...func()) {
 	initializers = append(initializers, y...)
 }
 
+// ClearInitializers clears all set initializers of OnInitialize
+func ClearInitializers() {
+	initializers = initializers[:0]
+}
+
 // OnFinalize sets the passed functions to be run when each command's
 // Execute method is terminated.
 func OnFinalize(y ...func()) {
 	finalizers = append(finalizers, y...)
+}
+
+// ClearFinalizers clears all set finalizers of OnFinalize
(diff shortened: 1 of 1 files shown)


Kirdock:
@ccoVeille had the same thought, so does anything speak against having a `Finalize` function on a command that is executed right after [this line](https://github.com/spf13/cobra/blob/db3ddb5cf4968b320ea21f6ee5c2d2202a2e7b22/command.go#L961) or after [retrieving the flags](https://github.com/spf13/cobra/blob/db3ddb5cf4968b320ea21f6ee5c2d2202a2e7b22/command.go#L963)?

Kirdock:
Follow-up: https://github.com/spf13/cobra/pull/2276

ccoVeille:
> @ccoVeille had the same thought, so does anything speak against having a `Finalize` function on a command that is deferred right after [this line](https://github.com/spf13/cobra/blob/db3ddb5cf4968b320ea21f6ee5c2d2202a2e7b22/command.go#L961) or after [retrieving the flags](https://github.com/spf13/cobra/blob/db3ddb5cf4968b320ea21f6ee5c2d2202a2e7b22/command.go#L963)?

Nothing I know, maybe @marckhouzam will have some thoughts about it

PR #2279, opened 2025-05-28
Upgrade golangci-lint to v2, address findings

scop:
See individual commits for details.


ccoVeille (approved):
Thanks.

ccoVeille:
For anyone wondering

These 3 existed in v1
      - gosimple 
      - staticcheck 
      - stylecheck

And were merged into staticcheck in the v2

So it's normal that gosimple and style check are removed.

Also, the other linter removal are fine because they were deprecated.

Thanks for your PR @scop

marckhouzam (approved):
Thanks guys!

PR #2284, opened 2025-06-10
chore: add reuse toml

fabianburth:
#### What this PR does / why we need it
Preparation for contribution of OCM to NeoNephos.

This is just a first step, the copyright notices as well as the package supplier will have to be adjusted, too. This will be done once we know to what exactly it should be changed!


fabianburth:
Sorry, that was embarrassing!

PR #2285, opened 2025-06-11
fix: quote args in fish shell completion

### LINKED ISSUE #1933: command line args should be quoted in generated fish completion script
more specifically here in [fish_completions.go:56][1]:

```fish
set -l requestComp "%[10]s=0 $args[1] %[3]s $args[2..-1] $lastArg"
```

where the references `$args[1]`, `$args[2..-1]` and `$lastArg` should have been quoted.

[1]: https://github.com/spf13/cobra/blob/c8a20a16bafa2dc85d304b59b2e677c7f75a73a6/fish_completions.go#L56

Downstream issue: https://github.com/mikefarah/yq/issues/1502

n9v9:
The arguments used for dynamic completions in fish were missing quotes. This resulted in an error when one of the arguments had a `*` in it whose expansion matched no file. Once the completion got triggered, fish errored with the message `No matches for wildcard [...]`. Don't quote the first argument as that is the name of the binary.

Closes #1933


fish_completions.go
```diff
@@ -46,14 +46,14 @@ function __%[1]s_perform_completion
 
     # Extract all args except the last one
     set -l args (commandline -opc)
-    # Extract the last arg and escape it in case it is a space
+    # Extract the last arg and escape it in case it is a space or wilcard
```
ccoVeille, line 49:
```suggestion
    # Extract the last arg and escape it in case it is a space or wildcard
```

ccoVeille (approved):

marckhouzam:
This seemed reasonable but when I run the tests in https://github.com/marckhouzam/cobra-completion-testing I'm getting failures.  I haven't investigated yet.

marckhouzam:
I pushed a commit on top of the PR to properly handle when the arguments are flags.
The the problem was that when completing something like "prog --hello <TAB>" the new `string join` and `string escape` commands in the fish scripts would receive `--hello` and would treat it as a flag instead of an argument.  By using `string join --` and `string escape --` we tell fish that all arguments following the -- are not to be treated as flags.

marckhouzam (approved):
Thanks!

PR #2292, opened 2025-07-13
Add two more projects using cobra

TheKhanj:
How about these two? 🤷


marckhouzam:
Are these projects you are involved with @TheKhanj ?  I prefer not adding projects that we don’t have confirmation that the maintainers agree to be listed here…

PR #2296, opened 2025-07-23
command: move required flag check before PreRun(E)

timlamballaistessensohn:
The current implementation checks the presence of required flags after `PreRun` or `PreRunE` functions are run. This commit moves that validation to happen before those steps.

There are use cases where we want to use flags inside `PreRun`/`PreRunE` functions. Checking for presence of required flags in those cases involves manually iterating and checking the flags inside your prerun functions. If we move the validation up, we get the validation for free.

The only "counter" usecase would be when users use `PreRun`/`PreRunE` to set required flags. This, however seems like a poor usecase that defeats the purpose of having required flags and/or default values in the first place.

### DIFF
--- a/command.go
+++ b/command.go
@@ -981,6 +981,12 @@ func (c *Command) execute(a []string) (err error) {
 			parents = append(parents, p)
 		}
 	}
+	if err := c.ValidateRequiredFlags(); err != nil {
+		return err
+	}
+	if err := c.ValidateFlagGroups(); err != nil {
+		return err
+	}
 	for _, p := range parents {
 		if p.PersistentPreRunE != nil {
 			if err := p.PersistentPreRunE(c, argWoFlags); err != nil {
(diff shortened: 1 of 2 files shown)


command.go
```diff
@@ -1003,14 +1009,12 @@ func (c *Command) execute(a []string) (err error) {
 	} else if c.PreRun != nil {
 		c.PreRun(c, argWoFlags)
 	}
-
 	if err := c.ValidateRequiredFlags(); err != nil {
 		return err
 	}
 	if err := c.ValidateFlagGroups(); err != nil {
 		return err
 	}
```
ccoVeille, line 1017: So you are validating them twice?

timlamballaistessensohn: Ehh apologies, those were left there during testing so I could comment out the new lines and verify the tests. I accidentally pushed a dirty commit. Updated now, and thanks for reviewing @ccoVeille .

timlamballaistessensohn:
Understood that no breaking changes are desired, also found that this change has been proposed multiple times in the past, apologies for wasted time.

FTR, I am fixing my problem by adding `validateRequiredFlags` calls at the start of all my `PreRun`/`PreRunE` functions 🥇

PR #2303, opened 2025-08-31
Bump pflag to 1.0.8

tomasaschan:
In pflag, we recently made a breaking change in naming to be more inclusive. This change, while doing nothing to address the same uninclusive naming in cobra itself, adjusts to those changes to ensure automatic bumps (if you do them) can continue uninterrupted.

### DIFF
--- a/command.go
+++ b/command.go
@@ -39,7 +39,7 @@ const (
 )
 
 // FParseErrWhitelist configures Flag parse errors to be ignored
-type FParseErrWhitelist flag.ParseErrorsWhitelist
+type FParseErrWhitelist flag.ParseErrorsAllowlist
 
 // Group Structure to manage groups for commands
 type Group struct {
@@ -1877,7 +1877,7 @@ func (c *Command) ParseFlags(args []string) error {
 	c.mergePersistentFlags()
 
 	// do it here after merging all flags and just before parse
-	c.Flags().ParseErrorsWhitelist = flag.ParseErrorsWhitelist(c.FParseErrWhitelist)
+	c.Flags().ParseErrorsAllowlist = flag.ParseErrorsAllowlist(c.FParseErrWhitelist)
 
 	err := c.Flags().Parse(args)
 	// Print warnings if they occurred (e.g. deprecated flag messages).
--- a/go.mod
+++ b/go.mod
@@ -5,6 +5,6 @@ go 1.15
 require (
 	github.com/cpuguy83/go-md2man/v2 v2.0.6
 	github.com/inconshreveable/mousetrap v1.1.0
-	github.com/spf13/pflag v1.0.6
+	github.com/spf13/pflag v1.0.8
 	gopkg.in/yaml.v3 v3.0.1
 )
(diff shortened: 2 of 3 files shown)


nadeemc:
Until this is merged, those running into issues can use this go.mod workaround:

```go.mod
// Needed until PR fix is merged:
// https://github.com/spf13/cobra/pull/2303
replace github.com/spf13/cobra => github.com/tomasaschan/cobra v1.9.2-0.20250831200618-527c4d0e1829
```

marckhouzam:
@nadeemc curious why someone would need a workaround since I assume cobra will only break if pointing to the new pflag? Maybe for projects that directly include pflag as well as cobra?

marckhouzam (approved):
I like that such names are being updated.
Thanks for being proactive and opening this PR.

Although I would like to make a similar change for cobra, our policy is to be completely backwards compatible, so for the moment at least, I don’t think we should go beyond this current PR

marckhouzam:
Is this merge good enough or do we need a cobra release for this?  I still don’t quite get when things break…

nadeemc:
@marckhouzam It will need a release to bump spf13/cobra to v1.9.2 or v1.10.0. Otherwise, anyone using `go get -u ./...` within a directory with `go.mod` will see something like this:

```diff
require (
	github.com/spf13/cobra v1.9.1
)

require (
-	github.com/spf13/pflag v1.0.7 // indirect
+	github.com/spf13/pflag v1.0.8 // indirect
)
```

Which will break the build without spf13/cobra also having a version bump and being picked up by the `go get -u` command.

Airblader:
As the change has been reverted now, this one needs to be reverted as well, otherwise things will break yet again.

jpmcb:
> curious why someone would need a workaround since I assume cobra will only break if pointing to the new pflag? Maybe for projects that directly include pflag as well as cobra?

@marckhouzam - yes, this is exactly the scenario that would break downstream consumers of both `pflag` and `cobra`: anyone who upgrades to `pflag` v1.0.8 (via something like `go get -u ./...`) breaks their `cobra` builds. This is mostly a design choice by the original Go authors: there's not really the concept of shadow dependencies or multi-versioned dependencies unless you _explicitly_ put them in something like `github.com/user/package/v2` where the code path itself is the version.

The blast radius of this was quite large actually: I got pinged by some people that `kubernetes/kubernetes` builds were failing on their `main` branch which also ensures that AWS, Google cloud, Nvidia, etc. etc. are not able to build their distros on the tip of their CI/CDs:

https://prow.k8s.io/view/gs/kubernetes-ci-logs/logs/ci-kubernetes-unit-dependencies/1962249657098702848

In the past, cobra has probably been overly cautious about taking any kind of breaking change since there are just so many projects that use it: the APIs are stable and assumed to be non-breaking. Doing anything that breaks that semver contract erodes alot of trust in the library (and us as maintainers).

tomasaschan:
> The blast radius of this was quite large actually

As you can probably tell from my original PR message, I had _no idea_ the blast radius of that change in pflag would be this massive. Basically, I (incorrectly) assumed that Cobra users wouldn't notice unless they explicitly tried to upgrade pflag, but of course I realized Cobra itself would need a few small changes to be able to keep rolling forward - hence this PR.

The immediate issues have been resolved now with v1.0.9 which restores the removed names, so anything that got bitten by some autobump thingy should also be fixed by the same autobump thingy without further action required.

That said, it also seems that anything important that broke because of the pflag change used pflag indirectly through Cobra. If you'd like, I'd be happy to include (but not author... PRs welcome!) a set of integration tests on the pflag side to ensure that any change merged there does not have this type of effect downstream.

PR #2305, opened 2025-09-01
chore: upgrade pflags v1.0.9

jpmcb:
Edit: blast radius seems to primarily be from projects that use both cobra and pflag. The use of the re-exported type:

https://github.com/spf13/cobra/blob/main/command.go#L41-L42

shouldn't break projects now that we've cut a v1.10.0 with those changes. This now just upgrades pflag without the need to emergency roll back the original variable type changes.

---

This is an emergency roll forward to unblock `kubernetes/kubernetes` (see https://github.com/kubernetes/kubernetes/issues/133809 and [failing prow-bot](https://prow.k8s.io/view/gs/kubernetes-ci-logs/logs/ci-kubernetes-unit-dependencies/1962249657098702848)) and other large projects / cloud providers that directly consumed breaking changes off cobra related to downstream breaking changes in `spf13/pflags` v1.0.8:
* https://github.com/spf13/pflag/issues/445

that were merged into `spf13/cobra` in:
* https://github.com/spf13/cobra/pull/2303

We obviously want to use these more inclusive names so we'll plan to roll forward a change in the future for `ParseErrorsWhitelist` --> `ParseErrorsAllowlist` once the dust has settled and we understand what the blast radius for such a breaking change is.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -5,6 +5,6 @@ go 1.15
 require (
 	github.com/cpuguy83/go-md2man/v2 v2.0.6
 	github.com/inconshreveable/mousetrap v1.1.0
-	github.com/spf13/pflag v1.0.8
+	github.com/spf13/pflag v1.0.9
 	gopkg.in/yaml.v3 v3.0.1
 )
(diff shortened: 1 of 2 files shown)


marckhouzam (commented):
This also changed go-md2man, was that pulled in by pflag?

PR #2306, opened 2025-09-04
Move away from archived `gopkg.in/yaml.v3`

saschagrunert:


marckhouzam:
Apologies @saschagrunert but this PR fell through the cracks and was done again in #2336.
But thank you for your efforts, we will try to avoid such mistakes in the future.

PR #2313, opened 2025-09-30
feat: Command.SetVersionFunc

caarlos0:
Currently, we can use `Command.Version` and `Command.SetVersionTemplate`, which generally works fine.

I would like to propose a `Command.SetVersionFunc`, similar to `command.SetHelpFunc`.

The rationale behind it is that the caller can style the output conditionally, e.g.:

```
$ pgr -v
styled version output

$ pgr -v | cat
plain-text version output (output not a tty)
```

Right now, the way around it is to, before executing the root command, check if output is a terminal, detect color profile, build and set a styled template.

This works, but can interfere with content being piped in/out (e.g. if you want to detect terminal background color and other terminal features).

### DIFF
--- a/command.go
+++ b/command.go
@@ -649,7 +649,7 @@ func (c *Command) getVersionTemplateFunc() func(w io.Writer, data interface{}) e
 	return defaultVersionFunc
 }
 
-// getVersionFunc sreturns the version function for the command going up the
+// getVersionFunc returns the version function for the command going up the
 // command tree if necessary.
 func (c *Command) getVersionFunc() func(*Command) error {
 	if c.versionFunc != nil {


command_test.go
```diff
@@ -1389,8 +1389,8 @@ func TestShorthandVersionFlagOnlyAddedIfVersionNotDefined(t *testing.T) {
 }
 
 func TestUsageIsNotPrintedTwice(t *testing.T) {
-	var cmd = &Command{Use: "root"}
-	var sub = &Command{Use: "sub"}
+	cmd := &Command{Use: "root"}
+	sub := &Command{Use: "sub"}
 	cmd.AddCommand(sub)
```
ccoVeille, line 1394: Could you revert the changes introduced by what seems to be gofumpt

command_test.go
```diff
@@ -2952,3 +2952,18 @@ func TestHelpFuncExecuted(t *testing.T) {
...
+		return err
+	})
+
+	output, err := executeCommandWithContext(context.Background(), rootCmd, "-v")
+	if err != nil {
+		t.Errorf("Unexpected error: %v", err)
+	}
+
+	checkStringContains(t, output, "custom version function: v2.3.4")
+}
```
ccoVeille, line 2969: Could you test when versionFunc is only available on the parent?

Thanks

ccoVeille (commented):
Some minor remarks

Issue #2320, opened 2025-10-18
Feature Question: Do you want to have an interactive option for command execution?


OkieOth:
I am playing around with some CLI tools and from time to time I am annoyed by typing all the provided sub commands in the required chain and to add also the required parameters ... good for scripts, a bit to much copy/pasting for playing around.

I created for me some helper functions to use a fuzzy search approach to step from command to command, to make the experience in case more interactive. Since I am satisfied with the extension for myself, I wonder if you would like that I provide a PR to this code?

I see the number of open PRs and want with my question only avoid additional pressure :)

BR,
Eiko

OkieOth:
Thank you for the response ... and don't worry, I will prepare a PR ... if it doesn't fit with your general idea for the lib, I will put it in a separate repo. It seems nicely to work in linux with zellij, not sure if it works for instance with powershell.

OkieOth:
As mentioned in the PR, my changes could be a bit harsch. If you like and want to include the changes, but need changes in the PR feel free to nudge me

OkieOth:
I am not able to lower the dependencies ... so I closed my PR. Since I still like my idea, I will later share the link to my repo lib (with examples here) ... maybe you find it useful.

Thank you for looking into my stuff. Thanky you for your efforts and keeping the idea of open source alive :)

BR, Eiko

OkieOth:
@marckhouzam I implemented the mentioned feature here:https://github.com/OkieOth/ic0bra 
... and I still like it X-D

Implementing it in my own repo allowed also to include a history. 

At least the repo illustrates what I had in mind. 

Cheers Eiko

marckhouzam:
Thanks!  I’ll definitely have a look

PR #2321, opened 2025-10-20
Add interactive run

OkieOth:
This is the promised PR for issue #2320. Although I like the experience, are the changes in the project a bit harsch.

If you reject this PR, I will release the changes in a dedicated extension lib.

I am looking forward.

### DIFF
--- a/interactive.go
+++ b/interactive.go
@@ -102,7 +102,7 @@ func isFlagRequired(f *pflag.Flag) bool {
 
 func isRepeatableFlag(f *pflag.Flag) bool {
 	switch f.Value.Type() {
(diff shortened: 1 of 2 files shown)


go.mod
```diff
@@ -1,10 +1,37 @@
 module github.com/spf13/cobra
 
-go 1.15
+go 1.24
```
marckhouzam: Because cobra is used by so many projects, we cannot change the go version, or else we would force all our users to update their go version, which can be a big task.

OkieOth: I will change the go version and provide a proper example documentation

PR #2333, opened 2025-11-25
perf: Loop over the bash variable directly instead of starting subprocesses.

JeffFaer:
Today, the loop is printing the completions one value per line and then reading one line to a variable. We can just skip the whole IO rigamarole by looping over the completions directly.

I had added this optimization to https://github.com/spf13/cobra/pull/2126 but it doesn't look like it got picked up by https://github.com/spf13/cobra/pull/1743.

### DIFF
--- a/bash_completionsV2.go
+++ b/bash_completionsV2.go
@@ -326,7 +326,7 @@ __%[1]s_handle_standard_completion_case() {
     local longest=0
     local compline
     # Look for the longest completion so that we can format things nicely
-    while IFS='' read -r compline; do
+    for compline in "${completions[@]}"; do
         [[ -z $compline ]] && continue
 
         # Before checking if the completion matches what the user typed,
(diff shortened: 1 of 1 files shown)


JeffFaer:
It shouldn't. The quotes around the array variable should prevent word splitting like that.

```sh
$ arr=( "one" "two" "three four" "five")
$ for n in "${arr[@]}"; do
  echo "$n"
done
$ for n in ${arr[@]}; do
  echo "$n"
done
```

marckhouzam:
Ok.  Give me a bit of time to revive the completion tests that I haven't looked at in a while and run them.
Ref: https://github.com/marckhouzam/cobra-completion-testing

JeffFaer:
Anecdotally, this appears to shave off something like 10ms:

```bash
#!/usr/bin/env bash

arr=()
for ((i=0; i<1000; i++)); do
    arr+=($i)
done

trials=1

loop_arr_1() {
    sum=0
    for ((j=0; j<trials; j++)); do
        for i in "${arr[@]}"; do
            ((sum+=i)) || true
        done
    done
}

loop_arr_2() {
    sum=0
    for ((j=0; j<trials; j++)); do
        while IFS='' read -r i; do
            ((sum+=i)) || true
        done < <(printf '%s\n' "${arr[@]}")
    done
}

time loop_arr_1
time loop_arr_2
```

```
real	0m0.003s
user	0m0.003s
sys	0m0.000s

real	0m0.016s
user	0m0.013s
sys	0m0.004s
```

marckhouzam (approved):
The regressions tests pass.
LGTM

PR #2334, opened 2025-11-25
fix(bash): Escape common prefixes that are inserted by the first completion request.

JeffFaer:
On the first tab (`COMP_TYPE=9`), Bash will automatically insert the common prefix of the matching completion options. We need to make sure those completion options are escaped since some part of them might end up on the command line.

On subsequent tabs (`COMP_TYPE=63`), Bash will display the completion options to the user, so we don't want to escape them for that `COMP_TYPE`.

### DIFF
--- a/bash_completionsV2.go
+++ b/bash_completionsV2.go
@@ -321,6 +321,9 @@ __%[1]s_handle_completion_types() {
 }
 
 __%[1]s_escape_compreply() {
+  if (( ${#COMPREPLY[@]} == 0 )); then
+    return
+  fi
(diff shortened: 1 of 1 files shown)


JeffFaer:
Friendly ping :)

JeffFaer:
@marckhouzam Is there anything else you need from me before reviewing this PR?

JeffFaer:
@marckhouzam Hey Mark! Please let me know what else you'd like to see to get this PR merged! I've been using it on my own fork of Cobra for months without incident!

PR #2335, opened 2025-11-28
Added custom padding for help template

irapandey:
While working with cobra on a project - I found that the need to add custom padding 
This PR would enable user to add a custom value for padding for helper text.

### DIFF
--- a/command.go
+++ b/command.go
@@ -225,6 +225,9 @@ type Command struct {
 	commandsMaxUseLen         int
 	commandsMaxCommandPathLen int
 	commandsMaxNameLen        int
+	namePadding               int
+	usagePadding              int
+	commandPathPadding        int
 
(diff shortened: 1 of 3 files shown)


marckhouzam:
Thanks  @irapandey .  Can you please explain when you imagine such a feature being used?  I don't want to complicate the API without understanding what value it adds.  Thanks.

irapandey:
Hey @marckhouzam 

I understand your concern - I was working on a project that used helper template.

Screenshot 2025-12-18 at 11 14 24 PM

I ran into the need of custom padding when I couldn't adjust the spacing between `application` and `Deploy`.

This looks way too small for a detail. If I would have to change the spacing, I would have to write the custom template which is just more work on the user end. 

Having the option to add customization will be really helpful to the user. 

This way the user can make use of the existing template with their formatting. I can propose more formatting options that can be added.

PR #2344, opened 2026-01-05
doc: Put description in separate section

adombeck:
The `Long` description was prepended to the Synopsis section which it does not belong to - the synopsis should, by convention, only be the syntax of the command and its options.

It's now added to a separate Description section.

### DIFF
--- a/doc/md_docs.go
+++ b/doc/md_docs.go
@@ -63,13 +63,17 @@ func GenMarkdownCustom(cmd *cobra.Command, w io.Writer, linkHandler func(string)
 
 	buf.WriteString("## " + name + "\n\n")
 	buf.WriteString(cmd.Short + "\n\n")
-	if len(cmd.Long) > 0 {
-		buf.WriteString("### Synopsis\n\n")
-		buf.WriteString(cmd.Long + "\n\n")
-	}
 
 	if cmd.Runnable() {
+		if len(cmd.Long) > 0 {
+			buf.WriteString("### Synopsis\n\n")
+		}
 		fmt.Fprintf(buf, "```\n%s\n```\n\n", cmd.UseLine())
+	}
+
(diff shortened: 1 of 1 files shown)


adombeck:
I can add a test if you agree with the change.

doc/md_docs.go
```diff
@@ -63,15 +63,19 @@ func GenMarkdownCustom(cmd *cobra.Command, w io.Writer, linkHandler func(string)
 
 	buf.WriteString("## " + name + "\n\n")
 	buf.WriteString(cmd.Short + "\n\n")
-	if len(cmd.Long) > 0 {
-		buf.WriteString("### Synopsis\n\n")
-		buf.WriteString(cmd.Long + "\n\n")
-	}
 
 	if cmd.Runnable() {
+		if len(cmd.Long) > 0 {
```
marckhouzam, line 68: Why would this only be printed if `long > 0` now?

adombeck: That was already the case. The `### Synopsis` section header was previously printed in line 67, which was only printed if `len(cmd.Long) > 0`.

marckhouzam (commented):
Can you provide some kind of reference as to why we should use the new format you suggest?  I don’t want someone to come along later with a different opinion to challenge this change.

adombeck:
> Can you provide some kind of reference as to why we should use the new format you suggest? 

My reasoning is that the synopsis should only contain the formal description of how to run the command and what command line options it takes (e.g. `lxc launch [<remote>:]<image> [<remote>:][<name>] [flags]`). That's the convention for [man pages](https://manpages.ubuntu.com/manpages/trusty/man1/man.1.html#description) and [POSIX utility descriptions](https://pubs.opengroup.org/onlinepubs/9799919799/utilities/V3_chap01.html#tag_18_04).

PR #2346, opened 2026-01-06
doc: Don't sort subcommands in "See Also" section

adombeck:
The subcommands in the `Command.Commands()` slice are sorted alphabetically by default. That can be disabled by setting `cobra.EnableCommandSorting = false` in which case the order in which subcommands are added via `Command.AddCommand` is retained. That order is used in the "Available Commands:" section in the usage message. Let's also keep that order when listing subcommands in the "See Also" section of generated docs.

### DIFF
--- a/doc/man_docs.go
+++ b/doc/man_docs.go
@@ -20,7 +20,6 @@ import (
 	"io"
 	"os"
 	"path/filepath"
-	"sort"
 	"strconv"
 	"strings"
 	"time"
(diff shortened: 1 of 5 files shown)


ccoVeille:
I would prefer an option to do not sort them.

I feel like many people maybe surprised for things no longer being ordered

adombeck:
The commands are not sorted in the default usage function and default usage template, so why sort them by default in the generated docs? That seems unnecessarily inconsistent.

adombeck:
> The commands are not sorted in the default usage function and default usage template

Actually, the subcommands returned by [`Command.Commands`](https://github.com/spf13/cobra/blob/v1.10.2/command.go#L1332-L1339) *are* [sorted by default](https://github.com/spf13/cobra/blob/v1.10.2/cobra.go?tab=source#L47-L59). So this PR only changes the behavior if `cobra.EnableCommandSorting` was set to `false`, in which case the developers clearly want to retain the order in which commands were added via `Command.AddCommand`.

I edited the PR description to reflect that.

Issue #2338, opened 2025-12-08
[Feature Request] User provided Help exit code


vipcxj:
There is a project [argonaut](https://github.com/vipcxj/argonaut) I'm working on; this tool is used to simplify defining command-line arguments in the shell, similar to what cobra does in Go. Here is an example to help illustrate the issue.

This is a script the user wrote: cool.sh
```shell
#!/bin/bash
set -e
EVAL_CONTENT=`argonaut bind \
  --flag=flag1 \
  --flag-flag1-default=hello \\
  --flag-flag1-choices=hello,world \
  --flag=flags2 \
  --flag-flags2-multi=true \
  --flag-flags2-choices=one,two,three
  -- "$0" "$@"
`
eval "$EVAL_CONTENT"
```
Then run
```shell
./cool.sh --flags2 one --flags2 three
```
EVAL_CONTENT will be assigned
```
FLAG1='hello'
FLAG2='one,three'
```
So after executing eval, the user can directly use these two environment variables in their own scripts.
If you enter the wrong parameters, for example
```shell
./cool.sh --flags2 one --flags2 four
```
will output
```
Error: value four for flag flags2 is not in allowed choices [one two three]
Usage:
  cool.sh [flags]

Flags:
--flag1 string          (default "hello")
      --flags2 stringArray
  -h, --help                 help for cool.sh
```

So far, everything has worked perfectly, until I executed
```shell
./cool.sh --help
```
It's a complete disaster. Because EVAL_CONTENT will become the help message, and eval will obviously throw an error. The workaround I can think of right now is
1. Redirect the help message output to stderr
2. Return a nonzero exit code

As for redirecting the help output, I've already found the API, but when dealing with the latter I discovered that cobra doesn't support customizing the help's return value, because help never reaches the Run method and is handled by cobra beforehand, even if I define the help flag myself. I even think this is a bug: if the user defines help themselves, it indicates they want to handle help requests themselves, not have the framework do it.

PR #2348, opened 2026-01-08
feat: Add SetHelpExitCode to allow custom exit codes on help

maishivamhoo123:
### Description
Addresses issue #2338.

Currently, `cmd --help` returns exit code 0. This causes scripts using `eval $(cmd)` to crash because the shell tries to execute the help text as commands. This PR allows developers to set a custom exit code (e.g., 2) when help is invoked, preventing this behavior.

### Changes
- Added `helpExitCode` field to `Command` struct.
- Added `SetHelpExitCode(int)` method.
- Updated `ExecuteC` to return a specific error if `helpExitCode` is set and help is invoked.
- Added unit test `TestHelpExitCode`.

### Example Usage
```go
rootCmd.SetHelpExitCode(2)
// Running "app --help" will now print usage and exit with code 2

### DIFF
--- a/command.go
+++ b/command.go
@@ -257,6 +257,23 @@ type Command struct {
 	// SuggestionsMinimumDistance defines minimum levenshtein distance to display suggestions.
 	// Must be > 0.
 	SuggestionsMinimumDistance int
+
+	// helpExitCode is the exit code to return when the help flag is used.
+	// If 0 (default), the program returns nil error (exit code 0).
+	helpExitCode int
+}
+type ExitError struct {
+	Code int
+}
+
+func (e *ExitError) Error() string {
+	return fmt.Sprintf("command requested exit with code %d", e.Code)
+}
+
+// SetHelpExitCode sets the exit code to return when help is shown.
+// If set to a non-zero value, Execute() will return an ExitError wrapping this code.
+func (c *Command) SetHelpExitCode(code int) {
+	c.helpExitCode = code
 }
 
 // Context returns underlying command context. If command was executed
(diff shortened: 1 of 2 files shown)


ccoVeille (changes requested):
This won't work the way you think.

It won't return the expected exit code.

It will only return exit 1 and an error message.

Note: I'm not saying the feature is a bad idea (or a good one).

I'm just saying it doesn't do work

maishivamhoo123:
@ccoVeille Thank you for the guidance! I've updated the logic to use a custom ExitError struct. This ensures the specific integer code can be retrieved via errors.As instead of defaulting to a generic error 1.

Apologies for the oversight in the previous commit. I would appreciate your review when you have a moment.
Looking for guidance.
Thank you sir.

ccoVeille:
This won't work either... 

Take a look at CheckErr method and how code use it.

I feel like your PR requires more iterations. Adding a feature like this to such a project might not be easy.

 let's wait for @marckhouzam and his feedback

But if you have a need for this, maybe you could simply consider defining a custom helper command to your code using cobra (so without change in cobra) and then call os.Exit with the code you want.

hahahamid:
Hi! I also implemented this feature in another PR (#2382). Happy to share some observations:

1. My implementation uses a simpler approach - just using `fmt.Errorf` to return the exit code as an error message, which works correctly.
2. I added tests for multiple scenarios including child commands to verify the behavior.
3. The key thing to verify: make sure `os.Exit()` actually receives the correct code. You may need to check how the error propagates through the CLI runner.

Happy to help test or provide feedback on the implementation!

PR #2357, opened 2026-02-18
Fix #2330: Escape angle brackets in man page output

### LINKED ISSUE #2330: Angle brackets stripped from manpage output
The Incus project got a report that our automatically generated man pages are missing quite a bit of data.

That's because our Use line typically includes placeholders like this: `[<remote>:]<instance>`

Cobra seems to be generating some basic markdown and then feeding that to md2man.
Unfortunately, it doesn't mark those fields with a backtick or similar escaping so they just get lost in translation.

As a quick test, I've tried feeding cobra's generated markdown doc to md2man and the resulting manpages do contain the expected content.

https://github.com/lxc/incus/issues/2670 was the report on our end.

rodionsteshenko:
## Problem

Angle brackets in command `Use`, `Short`, `Long`, and flag `Usage` strings were stripped by md2man, which interpreted them as HTML tags. This caused placeholders like `[<remote>:]<instance>` to render as `[:]` in man pages.

Reported in #2330 by @stgraber (Incus project).

## Fix

Added `escapeAngleBrackets()` which replaces `<` with `\\<` and `>` with `\\>` before the markdown is passed to md2man. Applied to:
- `cmd.UseLine()` in SYNOPSIS
- `cmd.Short` in NAME
- Description (Long/Short) in DESCRIPTION
- `flag.Usage` in OPTIONS

## Test

Added `TestGenManAngleBracketsPreserved` that creates a command with angle-bracketed placeholders in Use, Short, and Long, generates a man page, and verifies the angle brackets survive in the rendered output.

- **Without the fix**: test fails — output contains `transfer [:] [:]` (all placeholder names stripped)
- **With the fix**: test passes — output preserves `<remote>`, `<source>`, `<dest>`

Tested locally on macOS ARM (Apple Silicon). Full test suite passes.

### DIFF
--- a/doc/man_docs.go
+++ b/doc/man_docs.go
@@ -140,6 +140,15 @@ func fillHeader(header *GenManHeader, name string, disableAutoGen bool) error {
 	return nil
 }
 
+// escapeAngleBrackets escapes angle brackets so they are not stripped by md2man
+// which interprets them as HTML tags. This is necessary for command usage strings
+// that contain placeholders like [<remote>:]<instance>.
+func escapeAngleBrackets(s string) string {
+	s = strings.ReplaceAll(s, "<", `\<`)
+	s = strings.ReplaceAll(s, ">", `\>`)
+	return s
+}
+
 func manPreamble(buf io.StringWriter, header *GenManHeader, cmd *cobra.Command, dashedName string) {
 	description := cmd.Long
 	if len(description) == 0 {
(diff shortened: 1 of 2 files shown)


truffle-dev (commented):
Reproduced the underlying issue and confirmed this fix resolves it. Rendering the raw usage string through the same `go-md2man/v2` path cobra uses:

```
=== RAW (unescaped) ===
\fBmyapp\fP copy [:]

=== ESCAPED (this PR) ===
\fBmyapp\fP copy [<remote>:]<instance> <dest>
```

So md2man (via blackfriday) parses `<remote>`, `<instance>`, `<dest>` as HTML tags and drops them, collapsing `[<remote>:]<instance> <dest>` to `[:]`. Escaping `<`/`>` to `\<`/`\>` before rendering preserves the placeholders verbatim, which is the correct fix at the right layer (pre-md2man, where the markdown is assembled).

`TestGenManAngleBracketsPreserved` passes on the branch, and `strings` is already imported, so the diff is self-contained.

One thing worth a sentence in a comment or the commit body: `escapeAngleBrackets` is applied to `description` (the Long text) as well as the usage line. That's correct for placeholders, but it means a Long description that legitimately contains a literal `<` or `>` (e.g. prose like "values < 10") will now render with a backslash. In roff output `\<` is shown as `<`, so it's cosmetically fine, but flagging it so a future reader doesn't think the escape leaked. Not a blocker.

Nice catch on the root cause.

Issue #2362, opened 2026-02-28
Extending a Cobra CLI into an MCP Server and AI Agent — architecture, patterns, and the flag/schema gap


OpenWaygate:
Hi there, I built [yutu](https://github.com/eat-pray-ai/yutu) — a YouTube CLI powered by cobra. Over time, I extended it into an [MCP](https://modelcontextprotocol.io/) server and an AI agent, all within the same binary. The `mcp` and `agent` modes are just cobra subcommands:

```shell
> yutu
Available Commands:
  agent       Start agent to automate YouTube workflows
  mcp         Start MCP server
  video       Manipulate YouTube videos
  playlist    Manipulate YouTube playlists
  ...
```

Three libraries make this work:

| Layer | Library |
|-------|---------|
| CLI   | [cobra](https://github.com/spf13/cobra) |
| MCP   | [modelcontextprotocol/go-sdk](https://github.com/modelcontextprotocol/go-sdk) |
| Agent | [adk-go](https://github.com/google/adk-go) |

The point of this post: **any cobra application can become an MCP server and an AI agent with minimal glue code** — and the main friction point is flag/schema duplication between cobra and MCP.

## Architecture

The key insight is that **all three interfaces share the same domain logic**. Only the "input layer" differs:

```
                    +--- CLI Flags ------ cobra ------+
main.go -> cmd/ ---+                                  +---> pkg/<resource>/
                    +--- MCP Schema ---- go-sdk ------+
                    |                                 |
                    +--- Agent --------- adk-go ------+
                          (reuses MCP tools via in-memory transport)
```

- **`pkg/<resource>/`**: Pure domain logic. Each resource (video, channel, playlist, ...) exposes methods like `List()`, `Insert()`, `Update()`, `Delete()` operating on a struct with functional options.
- **`cmd/<resource>/`**: Registers both cobra subcommands and MCP tools, calling the same `pkg/` methods.
- **`cmd/agent/`**: The agent connects to the MCP server via an in-memory transport and reuses all registered MCP tools - zero additional wiring per resource.

### Step 1: Domain logic in `pkg/`

Each resource is a self-contained package with a struct, functional options, and methods:

```go
// pkg/activity/activity.go
type Activity struct {
    ChannelId  string `json:"channel_id,omitempty"`
    MaxResults int64  `json:"max_results,omitempty"`
    // ...
}

func NewActivity(opts ...Option) IActivity[youtube.Activity] { /* ... */ }
[...]

PR #2363, opened 2026-02-28
Mark Flag as Enum

njayp:
Adds MarkFlagEnum to command object. Usage:

```go
cmd.MarkFlagEnum("format", []string{"json", "yaml", "table"})
```

issue #2362


ccoVeille:
It looks like what we have for completion func

I feel like it's what you should look at for ophis if you wwbtto be aware of possible values abd what they are used for

Issue #2359, opened 2026-02-19
InitDefaultHelpFlag is not i18n aware


decentral1se:
Follows https://github.com/spf13/cobra/issues/2301.

I have a bit of a situation where translation of long flags is desired but short flags is not.

If you have something like this in your CLI:

```go
	rootCmd.PersistentFlags().BoolVarP(
		&internal.Help,
		i18n.G("help"),
		"h",  // not translated
		false,
		i18n.G("help for myCli"),
	)
```

And you translate `"help"` to `"ayuda"`. Then in `InitDefaultHelpFlag` 👇 

https://github.com/spf13/cobra/blob/61968e893eee2f27696c2fbc8e34fa5c4afaf7c4/command.go#L1221

`helpFlagName` is no longer `"help"` and the code attempts to set `c.Flags().BoolP(helpFlagName, "h", false, usage)`. However, `-h` is already set and we see an explosion stacktrace:

```
panic: unable to redefine 'h' shorthand in "myCli" flagset: it's already used for "ayuda" flag
```

This is quite an edge-case but just another symptom of cobra not supporting i18n natively.

Just wanted to report that in.

PR #2367, opened 2026-03-10
fix: don't panic when -h shorthand is already in use

dimitar-trifonov:
InitDefaultHelpFlag previously always attempted to register the default help flag with shorthand "h". If a command (or its persistent flags) already used "h" for another flag, pflag would panic due to the duplicate shorthand.

Avoid the panic by checking whether shorthand "h" is already taken before adding the default help flag. When "h" is unavailable, register the help flag without a shorthand instead.

Refs: #2359

### DIFF
--- a/command.go
+++ b/command.go
@@ -1226,11 +1226,10 @@ func (c *Command) InitDefaultHelpFlag() {
 		} else {
 			usage += name
 		}
-		if c.Flags().ShorthandLookup("h") == nil {
-			c.Flags().BoolP(helpFlagName, "h", false, usage)
-		} else {
-			c.Flags().Bool(helpFlagName, false, usage)
+		if f := c.Flags().ShorthandLookup("h"); f != nil {
+			panic(fmt.Sprintf("shorthand '-h' is reserved for the help flag, but is already used by flag '%s'", f.Name))
 		}
+		c.Flags().BoolP(helpFlagName, "h", false, usage)
 		_ = c.Flags().SetAnnotation(helpFlagName, FlagSetByCobraAnnotation, []string{"true"})
 	}
 }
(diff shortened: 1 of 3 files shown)


ccoVeille (approved):

marckhouzam:
>  When "h" is unavailable, register the help flag without a shorthand instead.

This would be a new behaviour.  Currently, `-h` is reserved for the `help` flag and the panic warns a developer that there is a conflict with `-h`.  If we remove this, programs could at some time find that by mistake, the `-h` flag has been assigned to another command than `help`.

Could we adapt this PR to keep the panic but be smarter about it?  We need to be able to know what the real name of the help flag (after translation) so that we can check that name to know if the help flag has already been added or not.

command.go
```diff
@@ -1226,6 +1226,9 @@ func (c *Command) InitDefaultHelpFlag() {
 		} else {
 			usage += name
 		}
+		if f := c.Flags().ShorthandLookup("h"); f != nil {
+			panic(fmt.Sprintf("shorthand '-h' is reserved for the help flag, but is already used by flag '%s'", f.Name))
```
ccoVeille, line 1230: I would mention that InitDefaultFlag cannot be used because -h is already registered by f.Name

dimitar-trifonov: @ccoVeille you are correct: with the latest change we keep the "-h is reserved for help" behavior.
If a developer registers `-h` for another flag (e.g. `ayuda`), Cobra now panics early with a clearer message:
`panic: shorthand '-h' is reserved for the help flag, but is already used by flag 'ayuda'`

This is also covered by a unit test:
TestInitDefaultHelpFlagPanicsWhenHShorthandAlreadyUsed

cc: @marckhouzam

PR #2370, opened 2026-03-18
feat: add opt-in help-all command (ref #2362)

dpritchett:
## Summary

Adds `NewHelpAllCommand()` — an opt-in command that prints every visible, runnable command with args and description in one pass. Same wiring pattern as `completion`:

```go
rootCmd.AddCommand(cobra.NewHelpAllCommand())
```

- One line per command, columns aligned
- `--verbose` adds flag placeholders with type-aware formatting
- ~160 lines, no new dependencies

Context in #2362 ([see my comment](https://github.com/spf13/cobra/issues/2362#issuecomment-4085536594)). [Gist with `gh` output (~200 commands)](https://gist.github.com/dpritchett/aa0890e81a72e1f6524cf8faa28a2b56).

### Example (patched into a local clone of `gh`)

```sh-session
$ gh help-all
    gh api <endpoint>                                            # Make an authenticated GitHub API request
    gh issue create                                              # Create a new issue
    gh pr merge [<number> | <url> | <branch>]                    # Merge a pull request
    gh repo clone <repository> [<directory>] [-- <gitflags>...]  # Clone a repository locally
    gh run watch <run-id>                                        # Watch a run until it completes
    ...
```

[Full gist with ~200 commands](https://gist.github.com/dpritchett/aa0890e81a72e1f6524cf8faa28a2b56)

## What's here

- `helpall.go`: factory, tree walker, formatter, renderer
- `helpall_test.go`: 19 tests

## Roadmap

- [x] Implementation + tests
- [x] Proof of concept against `gh` CLI
- [ ] Maintainer feedback on design
- [ ] Doc page in `site/content/` if accepted
- [ ] Final review
- [ ] Land

### DIFF
--- a/helpall.go
+++ b/helpall.go
@@ -1,0 +1,173 @@
+// Copyright 2013-2023 The Cobra Authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package cobra
+
+import (
+	"fmt"
+	"strings"
+
+	"github.com/spf13/pflag"
+)
+
+// commandInfo is extracted data for one entry in the help-all reference output.
+type commandInfo struct {
+	Path  string   // full command path, e.g. "app serve"
+	Args  string   // argument spec from Use field, e.g. "[session_id]"
+	Flags []string // formatted flag placeholders, e.g. "[--all]"
+	Short string   // one-line description
+}
+
+// collectCommands walks the command tree and returns a commandInfo for each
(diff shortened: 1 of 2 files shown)


helpall.go
```diff
@@ -0,0 +1,173 @@
...
+}
+
+// NewHelpAllCommand returns a help-all command that prints a complete, aligned
+// command reference. Add it to root with root.AddCommand(). Pass --verbose to
+// include flag placeholders in the output.
+func NewHelpAllCommand() *Command {
+	var verbose bool
+
+	cmd := &Command{
+		Use:    "help-all",
```
ccoVeille, line 163: if we go in that direction I would use something like


```suggestion
func NewHelpAllCommand(argName s) *Command {
	var verbose bool

	cmd := &Command{
		Use:    argName,
```

There is no need to fix the positional argument to be "help-all", some users may want to use "help" or whatever they want

dpritchett: Good point, thanks. I've added a godoc example showing that the name is customizable via `.Use` on the returned `*Command` (see 8902815). This matches how cobra's own [completion command](https://github.com/spf13/cobra/blob/61968e893eee2f27696c2fbc8e34fa5c4afaf7c4/completions.go#L100) works internally (hardcoded default name, mutable after creation).

ccoVeille (approved):

PR #2372, opened 2026-03-20
feat(fish): implement ShellCompDirectiveFilterDirs

thaJeztah:
Fish supports directory-only completions through `__fish_complete_directories`, but Cobra's generated completion script treated this directive as unsupported.

Use Fish's native directory completion when `ShellCompDirectiveFilterDirs` is set.

### DIFF
--- a/site/content/completions/_index.md
+++ b/site/content/completions/_index.md
@@ -549,10 +549,7 @@ search  show  status
   * `BashCompSubdirsInDir` (filtering by directory)
 * The functions corresponding to the above annotations are consequently not supported and will be ignored for `fish`:
   * `MarkFlagFilename()` and `MarkPersistentFlagFilename()` (filtering by file extension)
-  * `MarkFlagDirname()` and `MarkPersistentFlagDirname()` (filtering by directory)
-* Similarly, the following completion directives are not supported and will be ignored for `fish`:
-  * `ShellCompDirectiveFilterFileExt` (filtering by file extension)
-  * `ShellCompDirectiveFilterDirs` (filtering by directory)
+* Similarly, the `ShellCompDirectiveFilterFileExt` (filtering by file extension) completion directive is not supported and will be ignored for `fish`.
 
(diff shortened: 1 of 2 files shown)


thaJeztah:
Failure is unrelated to my changes, but perhaps GolangCi-lint was updated;

```
  Error: completions.go:958:24: G703: Path traversal via taint analysis (gosec)
  		f, err := os.OpenFile(path,
  		                     ^
```

site/content/completions/_index.md
```diff
@@ -549,10 +549,7 @@ search  show  status
   * `BashCompSubdirsInDir` (filtering by directory)
 * The functions corresponding to the above annotations are consequently not supported and will be ignored for `fish`:
   * `MarkFlagFilename()` and `MarkPersistentFlagFilename()` (filtering by file extension)
-  * `MarkFlagDirname()` and `MarkPersistentFlagDirname()` (filtering by directory)
```
ccoVeille, line 552: I'm unsure why this line was removed

thaJeztah: I need to double check when I'm back at my computer, but I think the flow was; these options set a "bash" annotation on the flag, but the  code further down converts that to setting autocompletion for the flag, using the `ShellCompDirectiveFilterDirs` directive.

So they were ignored because that directive wasn't functional 😅

thaJeztah: Right, so `MarkFlagDirname` sets the `BashCompSubdirsInDir` annotation; https://github.com/spf13/cobra/blob/61968e893eee2f27696c2fbc8e34fa5c4afaf7c4/shell_completions.go#L96-L98

Which then gets translated to `ShellCompDirectiveFilterDirs` here; https://github.com/spf13/cobra/blob/61968e893eee2f27696c2fbc8e34fa5c4afaf7c4/completions.go#L416-L423


The `ShellCompDirectiveFilterDirs` previously didn't take effect on Fish, but with this change does

PR #2382, opened 2026-04-06
Add HelpExitCode feature to allow custom exit code when --help is triggered

### LINKED ISSUE #2338: [Feature Request] User provided Help exit code

hahahamid:
## Summary

Implements feature request #2338 - allows users to customize the exit code returned when `--help` is triggered.

## Changes

- Add `HelpExitCode` field to Command struct
- Modify `ExecuteC()` to return custom exit code when HelpExitCode is non-zero
- Add `SetHelpExitCode()` setter method for easy configuration
- Add tests covering default (0), custom (non-zero), and child command scenarios

## Why this matters
When embedding Cobra-based CLIs in shell scripts, users need a way to distinguish between help output and other errors. Currently, `--help` always returns exit code 0, making it impossible to handle specially in scripts.

## Usage

```go
rootCmd.SetHelpExitCode(1)
Now when users run myapp --help, the exit code will be 1 instead of 0, allowing scripts to handle it differently.

Fixes #2338

### DIFF
--- a/command.go
+++ b/command.go
@@ -238,6 +238,10 @@ type Command struct {
 	// SilenceUsage is an option to silence usage when an error occurs.
 	SilenceUsage bool
 
+	// HelpExitCode is the exit code returned when --help is triggered.
+	// If set to 0 (default), exit code is 0. If set to non-zero, that exit code is returned.
+	HelpExitCode int
+
 	// DisableFlagParsing disables the flag parsing.
 	// If this is true all flags will be passed to the command as arguments.
 	DisableFlagParsing bool
@@ -352,6 +356,11 @@ func (c *Command) SetHelpCommandGroupID(groupID string) {
 func (c *Command) SetCompletionCommandGroupID(groupID string) {
 	// completionCommandGroupID is used if no completion command is defined by the user
 	c.Root().completionCommandGroupID = groupID
+}
+
+// SetHelpExitCode sets the exit code returned when --help is triggered.
+func (c *Command) SetHelpExitCode(code int) {
+	c.HelpExitCode = code
 }
 
 // SetHelpTemplate sets help template to be used. Application can use it to set custom template.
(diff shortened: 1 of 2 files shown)


hahahamid:
Closing this PR as there's already an existing PR (#2348) addressing the same feature request (#2338). Thanks!

PR #2386, opened 2026-04-10
fix: pass correct args to HelpFunc (closes #2154)

### LINKED ISSUE #2154: Arguments not passed to the help function

scovl:
Two code paths in command.go were passing wrong arguments to the function registered via SetHelpFunc(func(*Command, []string)):

1. ExecuteC() — when --help/-h triggers flag.ErrHelp, the raw flag slice was passed instead of the post-parse positional arguments. Fixed by passing cmd.Flags().Args() (or the full slice when DisableFlagParsing is set).

2. InitDefaultHelpCmd() — the help <sub> built-in always passed an empty slice. Fixed by deriving the remaining args from cmd.Root().Find(args) so callers receive the extra tokens (e.g. plugin CLIs that inspect sub-command arguments).

Also extracts constant fmtSubCmdNameDesc in defaultUsageFunc (S1192) and centralises all duplicate test-string literals into a single const block in command_test.go (S1192 x43, S1186 x4).

Supersedes PR #2158 (open, merge conflicts).  @ccoVeille

### DIFF
--- a/command.go
+++ b/command.go
@@ -330,6 +330,9 @@ func (c *Command) SetFlagErrorFunc(f func(*Command, error) error) {
 }
 
 // SetHelpFunc sets help function. Can be defined by Application.
+// The help function receives the command and the remaining positional arguments
+// after flag parsing, consistent with Run and RunE.
+// If flag parsing is disabled (DisableFlagParsing), all arguments are passed.
 func (c *Command) SetHelpFunc(f func(*Command, []string)) {
 	c.helpFunc = f
 }
@@ -1150,7 +1153,17 @@ func (c *Command) ExecuteC() (cmd *Command, err error) {
 		// Always show help if requested, even if SilenceErrors is in
 		// effect
 		if errors.Is(err, flag.ErrHelp) {
-			cmd.HelpFunc()(cmd, args)
+			// execute() has already parsed the flags; pass only the remaining
+			// positional arguments to HelpFunc, consistent with Run/RunE.
+			// When DisableFlagParsing is set, flag parsing was not done so we
+			// fall back to the full flags slice which includes flags.
+			remainingArgs := cmd.Flags().Args()
+			if cmd.DisableFlagParsing {
+				remainingArgs = flags
+			}
+			// HelpFunc signature is func(*Command, []string) — it does not
+			// return an error. cmd.Help() also always returns nil.
+			cmd.HelpFunc()(cmd, remainingArgs)
 			return cmd, nil
 		}
 
@@ -1291,7 +1304,7 @@ Simply type ` + c.DisplayName() + ` help [path to command] for full details.`,
 				return completions, ShellCompDirectiveNoFileComp
 			},
 			Run: func(c *Command, args []string) {
-				cmd, _, e := c.Root().Find(args)
+				cmd, remainingArgs, e := c.Root().Find(args)
 				if cmd == nil || e != nil {
 					c.Printf("Unknown help topic %#q\n", args)
 					CheckErr(c.Root().Usage())
@@ -1303,7 +1316,7 @@ Simply type ` + c.DisplayName() + ` help [path to command] for full details.`,
 
 					cmd.InitDefaultHelpFlag()    // make possible 'help' flag to be shown
 					cmd.InitDefaultVersionFlag() // make possible 'version' flag to be shown
-					CheckErr(cmd.Help())
+					cmd.HelpFunc()(cmd, remainingArgs)
 				}
 			},
 			GroupID: c.helpCommandGroupID,
--- a/command_test.go
+++ b/command_test.go
@@ -2952,3 +2952,109 @@ func TestHelpFuncExecuted(t *testing.T) {
 
 	checkStringContains(t, output, helpText)
 }
+
+func TestHelpFuncArgs(t *testing.T) {
+	tests := []struct {
+		name               string
+		disableFlagParsing bool
+		addChildFlag       bool
+		rootRunnable       bool
+		argv               []string
+		wantCmd            string
+		wantArgs           []string
+	}{
+		{
+			name:         "help flag passes only positional args",
+			addChildFlag: true,
+			argv:         []string{"child", "arg1", "--myflag", "val", "arg2", "--help"},
+			wantCmd:      "child",
+			wantArgs:     []string{"arg1", "arg2"},
+		},
+		{
+			name:     "short -h passes only positional args",
+			argv:     []string{"child", "arg1", "arg2", "-h"},
+			wantCmd:  "child",
+			wantArgs: []string{"arg1", "arg2"},
+		},
+		{
+			name:     "help command passes args after subcommand path",
+			argv:     []string{"help", "child", "arg1", "arg2"},
+			wantCmd:  "child",
+			wantArgs: []string{"arg1", "arg2"},
+		},
+		{
+			name:     "no positional args with --help",
+			argv:     []string{"child", "--help"},
+			wantCmd:  "child",
+			wantArgs: []string{},
+		},
+		{
+			name:     "no positional args with help command",
+			argv:     []string{"help", "child"},
+			wantCmd:  "child",
+			wantArgs: []string{},
+		},
+		{
+			name:               "DisableFlagParsing keeps all args including flag-like tokens",
+			disableFlagParsing: true,
+			argv:               []string{"child", "arg1", "--flag-as-arg", "arg2"},
+			wantCmd:            "child",
+			wantArgs:           []string{"arg1", "--flag-as-arg", "arg2"},
+		},
+		{
+			name:         "root command --help",
+			rootRunnable: true,
+			argv:         []string{"--help"},
+			wantCmd:      "prog",
+			wantArgs:     []string{},
+		},
+	}
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			rootCmd := &Command{Use: "prog"}
+			if tt.rootRunnable {
+				rootCmd.Run = emptyRun
+			}
+
+			childCmd := &Command{Use: "child"}
+			if !tt.disableFlagParsing {
+				childCmd.Run = emptyRun
+			}
+			childCmd.DisableFlagParsing = tt.disableFlagParsing
+
+			if tt.addChildFlag {
+				childCmd.Flags().String("myflag", "", "a flag")
+			}
+
+			rootCmd.AddCommand(childCmd)
+
+			helpCalled := false
+			rootCmd.SetHelpFunc(func(c *Command, gotArgs []string) {
+				helpCalled = true
+				if c.Name() != tt.wantCmd {
+					t.Errorf("expected cmd %q, got %q", tt.wantCmd, c.Name())
+				}
+				if !reflect.DeepEqual(gotArgs, tt.wantArgs) {
+					t.Errorf("expected args %v, got %v", tt.wantArgs, gotArgs)
+				}
+			})
+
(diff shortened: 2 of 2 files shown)


command_test.go
ccoVeille, line 1: Could limit the PR to the feature you are testing and not adding changes that are unrelated.

Here you have tons of constants added for stylistic and opinionated reasons.

I'm not saying these stylistic changes are bad.

I'm saying it brings a lot of noises to the PR, file being unrelated to the changes you add, and possible conflicts with other pending PR

command.go
```diff
@@ -36,6 +36,10 @@ const (
 
 	helpFlagName    = "help"
 	helpCommandName = "help"
+
+	// fmtSubCmdNameDesc is the format used to print a sub-command name (padded)
+	// followed by its short description in the usage output.
+	fmtSubCmdNameDesc = "\n  %s %s"
```
ccoVeille: Please revert this changes

https://github.com/spf13/cobra/pull/2386#discussion_r3067811200

command_test.go
scovl: @ccoVeille 
You're right, and thanks for the feedback. My initial intent was to reduce pattern repetition by extracting constants, but I should have kept that in a separate PR to avoid noise and potential conflicts.

I've force-pushed with only the #2154 fix: the three command.go hunks (doc-comment, `ExecuteC()`, `InitDefaultHelpCmd()`) and the new tests appended at the end of command_test.go. No constants, no renames, no changes to existing tests.

Could you take another look?
Ty!

command_test.go
```diff
@@ -2952,3 +2952,196 @@ func TestHelpFuncExecuted(t *testing.T) {
...
+// overridingHelp is a helper for issue #2154 tests: it verifies that HelpFunc
+// receives the correct command name and positional arguments.
+type overridingHelp struct {
+	expectedCmd  string
+	expectedArgs []string
+	helpCalled   bool
+	err          error
+}
+
+func (o *overridingHelp) helpFunc(c *Command, args []string) {
```
ccoVeille: Here, it should be clearer that this satisfies the signature expected for SetHelpFunc, a comment maybe

command_test.go
ccoVeille: Thanks

command_test.go
```diff
@@ -2952,3 +2952,196 @@ func TestHelpFuncExecuted(t *testing.T) {
...
 }
+
+// overridingHelp is a helper for issue #2154 tests: it verifies that HelpFunc
+// receives the correct command name and positional arguments.
+type overridingHelp struct {
+	expectedCmd  string
+	expectedArgs []string
+	helpCalled   bool
+	err          error
+}
```
ccoVeille: I find this plus checkError a bit uncommon, also you are creating multiple TestHelpFunc functions

did you consider something like this? That would match existing code in tests ?

```go
func TestHelpFuncReceivesExpectedArgs_Table(t *testing.T) {
    type tc struct {
        name string

        // command setup
        disableFlagParsing bool
        addChildFlag       bool
        rootRunnable       bool // whether root has Run to avoid non-runnable behavior

        // args to execute
        argv []string

        // expectations
        wantCmd  string
        wantArgs []string
    }

    tests := []tc{
        {
            name:           "help flag passes only positional args (not flags)",
            addChildFlag:   true,
            argv:           []string{"child", "arg1", "--myflag", "val", "arg2", "--help"},
            wantCmd:        "child",
            wantArgs:       []string{"arg1", "arg2"},
        },
        {
            name:     "short -h passes only positional args",
            argv:     []string{"child", "arg1", "arg2", "-h"},
            wantCmd:  "child",
            wantArgs: []string{"arg1", "arg2"},
        },
        {
            name:     "`help child ...` passes args after subcommand path",
            argv:     []string{"help", "child", "arg1", "arg2"},
            wantCmd:  "child",
            wantArgs: []string{"arg1", "arg2"},
        },
        {
            name:     "no positional args with --help => empty slice",
            argv:     []string{"child", "--help"},
            wantCmd:  "child",
            wantArgs: []string{},
        },
        {
            name:              "DisableFlagParsing: help path receives full args (flag-like tokens kept)",
            disableFlagParsing: true,
            // Make child non-runnable (no Run/RunE) so Cobra returns flag.ErrHelp automatically
            argv:     []string{"child", "arg1", "--flag-as-arg", "arg2"},
            wantCmd:  "child",
            wantArgs: []string{"arg1", "--flag-as-arg", "arg2"},
        },
        {
            name:         "root command --help works",
            rootRunnable: true,
            argv:         []string{"--help"},
            wantCmd:      "prog",
            wantArgs:     []string{},
        },
    }

    for _, tt := range tests {
        t.Run(tt.name, func(t *testing.T) {
            rootCmd := &Command{Use: "prog"}
            if tt.rootRunnable {
                rootCmd.Run = emptyRun
            }

            childCmd := &Command{Use: "child"}
            if !tt.disableFlagParsing {
                // Only set Run when runnable cases want normal execution; for the DisableFlagParsing
                // case we want non-runnable behavior (no Run/RunE) as described in the PR.
                childCmd.Run = emptyRun
            }
            childCmd.DisableFlagParsing = tt.disableFlagParsing

            if tt.addChildFlag {
                childCmd.Flags().String("myflag", "", "a flag")
            }
            rootCmd.AddCommand(childCmd)

            helpCalled := false
            rootCmd.SetHelpFunc(func(c *Command, gotArgs []string) {
                helpCalled = true
                if c.Name() != tt.wantCmd {
                    t.Fatalf("expected cmd %q, got %q", tt.wantCmd, c.Name())
                }
                if len(gotArgs) != len(tt.wantArgs) {
                    t.Fatalf("expected args %v (len %d), got %v (len %d)",
                        tt.wantArgs, len(tt.wantArgs), gotArgs, len(gotArgs))
                }
                for i := range tt.wantArgs {
                    if gotArgs[i] != tt.wantArgs[i] {
                        t.Fatalf("expected args %v, got %v", tt.wantArgs, gotArgs)
                    }
                }
            })

            _, err := executeCommand(rootCmd, tt.argv...)
            if err != nil {
                t.Fatalf("unexpected error: %v", err)
            }
            if !helpCalled {
                t.Fatalf("help function was not called")
            }
        })
    }
}
```

This is pseudo code generated by Copilot when I asked him to tell me what would be the classic way to test.

I'm not expecting you to change it, but to tell me why you applied the pattern I can see in the PR

scovl: You're right, I overcomplicated it. Replaced the whole thing with a single table-driven `TestHelpFuncArgs` using `t.Run` subtests, same pattern as `TestStripFlags` and `TestFind`. 

Dropped the `overridingHelp` struct, assertions are now inline with `reflect.DeepEqual`. `TestDirectHelpCallNoPanic` stays separate since it calls `cmd.Help()` directly. Force-pushed, thanks for the review. 

I'll pay more attention next time.

command.go
```diff
@@ -1303,7 +1314,10 @@ Simply type ` + c.DisplayName() + ` help [path to command] for full details.`,
 
 					cmd.InitDefaultHelpFlag()    // make possible 'help' flag to be shown
 					cmd.InitDefaultVersionFlag() // make possible 'version' flag to be shown
-					CheckErr(cmd.Help())
+					// Pass the remaining args (those not part of the sub-command path)
+					// to HelpFunc so callers can inspect them (e.g. plugin CLIs).
+					_, remainingArgs, _ := cmd.Root().Find(args)
```
ccoVeille: The error possibly returned by find is discarded

command.go
```diff
@@ -1150,7 +1153,15 @@ func (c *Command) ExecuteC() (cmd *Command, err error) {
...
+			// execute() has already parsed the flags; pass only the remaining
+			// positional arguments to HelpFunc, consistent with Run/RunE.
+			// When DisableFlagParsing is set, flag parsing was not done so we
+			// fall back to the full flags slice which includes flags.
+			remainingArgs := cmd.Flags().Args()
+			if cmd.DisableFlagParsing {
+				remainingArgs = flags
+			}
+			cmd.HelpFunc()(cmd, remainingArgs)
 			return cmd, nil
```
ccoVeille, line 1167: I'm annoyed here

`CheckErr(cmd.Help())`

was here to exist on some condition, here you are not handling the error returned by cmd.HelpFunc, but here you always return nil.

Does it work as intended then?

command.go
scovl: @ccoVeille The concern is valid, but this behavior actually predates this PR. `CheckErr(cmd.Help())` was a no-op because `cmd.Help()` unconditionally returns `nil`:

```go
func (c *Command) Help() error {
    c.HelpFunc()(c, []string{})
    return nil // always nil
}
```

The original code used "false logic" it looked like error handling, but didn't actually guard anything. 

I used an explicit `_` on line 1319 as a **semantic marker**. It’s more honest to make the silence visible than to hide it behind a wrapper that implies handling. This makes it clear that we eventually need to either propagate the error properly or document why it's safe to ignore.

I'm happy to add a `//nolint` or an explanatory comment if you prefer!

ccoVeille: I agree with your analysis of existing checkErr calling something that always return nil

But here you use Find, find can return an error, in some edge cases

And here you are ignoring it.

https://github.com/spf13/cobra/blob/main/command.go#L775-L777

scovl: Hey @ccoVeille, thanks for the thorough review.

I studied your feedback more carefully and ran the full test suite to understand the issue. I've squashed everything into a single commit (`5240b43`) with the fix:

```go
cmd, remainingArgs, e := c.Root().Find(args)
```

One call, all three return values captured. The error `e` is checked immediately after (`if cmd == nil || e != nil`), so nothing is silently discarded anymore. I also added a comment in `ExecuteC` explaining why `return cmd, nil` is safe after calling `HelpFunc` (it doesn't return an error, and `cmd.Help()` always returns nil).

Could you take another look and let me know if this is good now?

ccoVeille (approved):

command.go
ccoVeille: Thanks this approach seems good enough

PR #2392, opened 2026-04-17
fix: permit empty string for ErrPrefix

### LINKED ISSUE #2226: permit empty string for ErrPrefix
The `SetErrPrefix` method introduced in https://github.com/spf13/cobra/pull/2023 is nice, but I wanted to set it to an empty string `""`, but discovered that this is treated the same as if you had not set it at all.

It would be preferable if this were a `*string` field, which would allow setting it to an empty string.

humanborch:
Currently, when a developer sets `cmd.SetErrPrefix("")`, the system incorrectly assumes the prefix was left unset due to `""` being the default zero-value for a string, and falls back to prepending `"Error: "` and a trailing space. This makes it impossible to output pure JSON errors because of the leading space injected by `c.PrintErrln`.

This PR introduces an unexported `errPrefixSet` boolean to cleanly demarcate when the user has explicitly `SetErrPrefix`.

* Modified `ErrPrefix()` getter to honor the marker without breaking the public `Command` struct API.
* Added a conditional downstream print block inside `ExecuteC` to prevent `fmt.Sprintln` from injecting a leading space when the prefix evaluates to empty.
* Added `TestRootErrPrefixEmpty` to `command_test.go` to explicitly prevent regressions.
* Existing unit tests pass.

Fixes #2226

### DIFF
--- a/command.go
+++ b/command.go
@@ -193,8 +193,8 @@ type Command struct {
 
 	// errPrefix is the error message prefix defined by user.
 	errPrefix string
-	// errPrefixSet tracks if the user explicitly set the errPrefix to allow for "" (empty string).
-	errPrefixSet bool
+	// errPrefixEmpty tracks if the user explicitly wants an empty errPrefix.
+	errPrefixEmpty bool
 
 	// inReader is a reader defined by the user that replaces stdin
 	inReader io.Reader
@@ -377,7 +377,12 @@ func (c *Command) SetVersionTemplate(s string) {
 // SetErrPrefix sets error message prefix to be used. Application can use it to set custom prefix.
 func (c *Command) SetErrPrefix(s string) {
 	c.errPrefix = s
-	c.errPrefixSet = true
+}
+
+// SetErrPrefixEmpty sets whether the error message prefix should be empty.
+// If set to true, this suppresses the default "Error: " prefix.
+func (c *Command) SetErrPrefixEmpty(empty bool) {
+	c.errPrefixEmpty = empty
 }
 
 // SetGlobalNormalizationFunc sets a normalization function to all flag sets and also to child commands.
@@ -644,8 +649,11 @@ func (c *Command) getVersionTemplateFunc() func(w io.Writer, data interface{}) e
 
 // ErrPrefix return error message prefix for the command
 func (c *Command) ErrPrefix() string {
-	if c.errPrefix != "" || c.errPrefixSet {
+	if c.errPrefix != "" {
 		return c.errPrefix
+	}
+	if c.errPrefixEmpty {
+		return ""
 	}
 
 	if c.HasParent() {
--- a/command.go
+++ b/command.go
@@ -375,14 +375,19 @@ func (c *Command) SetVersionTemplate(s string) {
 }
 
 // SetErrPrefix sets error message prefix to be used. Application can use it to set custom prefix.
+// To unset the error prefix, this function should be called with the empty string.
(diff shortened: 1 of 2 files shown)


marckhouzam:
Do we need a way to unset the error prefix now that using `“”` won’t unset?

marckhouzam:
This makes sense to me but I believe this has backwards-compatibility implications. Existing programs might be using `””` to unset the error prefix but this PR would change how that behaves.

It wouldn’t be pretty but we could have a function `SetErrorPrefixEmpty()` instead. And I’m sure there are other ways also.

humanborch:
Good catch on the backwards-compatibility issue! You're totally right that existing programs might be relying on `SetErrPrefix(""')` to unset the prefix back to the default.

I've pushed an update that reverts `SetErrPrefix(""')` back to its original behavior so we don't break anything. Following your suggestion, I added a new `SetErrPrefixEmpty(bool)` method instead. This gives developers a clean, explicit way to suppress the prefix without overloading the meaning of an empty string. I also added a test to ensure that `SetErrPrefix(""')` successfully restores the default prefix.

Let me know what you think of this approach!

command.go
```diff
@@ -377,6 +379,12 @@ func (c *Command) SetErrPrefix(s string) {
 	c.errPrefix = s
```
marckhouzam, line 381: If this function is called with an argument `""` it should unset the error prefix which should cause `ErrPrefix()` to check if the parent has an error prefix.

This won't happen if `SetErrPrefixEmpty(true)` has been called before.
I think we also need to call `c.errPrefixEmpty = false` in this function to truly unset the error prefix

command.go
```diff
@@ -377,6 +379,12 @@ func (c *Command) SetErrPrefix(s string) {
...
 }
 
+// SetErrPrefixEmpty sets whether the error message prefix should be empty.
+// If set to true, this suppresses the default "Error: " prefix.
+func (c *Command) SetErrPrefixEmpty(empty bool) {
+	c.errPrefixEmpty = empty
```
marckhouzam, line 389: Same idea that if `SetErrPrefix("hello")` was called before, this function should undo that, or else the error prefix will not actually show as empty.

So we should set `c.errPrefix = ""` here I think

command.go
marckhouzam, line 377: Can you add a comment saying that "to unset the error prefix, this function should be called with the empty string. To prevent an error prefix from being used, the function SetErrPrefixEmpty(true) should be used"

command.go
```diff
@@ -377,6 +379,12 @@ func (c *Command) SetErrPrefix(s string) {
 	c.errPrefix = s
 }
 
+// SetErrPrefixEmpty sets whether the error message prefix should be empty.
+// If set to true, this suppresses the default "Error: " prefix.
```
marckhouzam, line 386: Please a comment along the lines of `to unset the error prefix (so the parent chain can be checked), please use SetErrPrefix("")`

marckhouzam (commented):
I like this

humanborch:
@marckhouzam Thanks for the great review! I really appreciate the time you've taken to look into this and provide such helpful feedback. 

I've just pushed a commit that addresses your points:
- `SetErrPrefix` now explicitly unsets `errPrefixEmpty` (`c.errPrefixEmpty = false`).
- `SetErrPrefixEmpty(true)` now explicitly clears the prefix string (`c.errPrefix = ""`).
- Added the requested comments to both functions so the behavior is clear.

I'm thrilled to be able to contribute back to a project I use every day. Let me know if there's anything else needed!

PR #2398, opened 2026-04-28
feat: Add MatchAny positional args validator

walle250ai:
Adds MatchAny(pargs ...PositionalArgs) as the OR-semantics complement to the existing MatchAll. Returns nil if any validator passes; if all fail, returns a combined error listing each individual failure. Zero validators passed returns nil (vacuously true, consistent with MatchAll behaviour).

### DIFF
--- a/args.go
+++ b/args.go
@@ -135,6 +135,29 @@ func MatchAll(pargs ...PositionalArgs) PositionalArgs {
 	}
 }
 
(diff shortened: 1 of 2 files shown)


marckhouzam (approved):
Thanks. This seems reasonable  

You need to sign the CLA before this can be merged. Please see the instructions in the comment above.

marckhouzam:
Ah, and the linter failed.

PR #2423, opened 2026-06-26
test: fix completions test failure on Go 1.26

### LINKED ISSUE #2422: Test failed with golang 1.26
It complainted:
```
```

SudhanshuMatrix:
Under Go 1.26, `go vet` has tightened its format verb checks, flagging `%q` when used on the integer-based `ShellCompDirective` type:

`(*testing.common).Errorf format %q has arg directive of wrong type github.com/spf13/cobra.ShellCompDirective`

This fixes the issue by formatting the directive as a string using its unexported `string()` helper method, keeping the change isolated to the test file.

Fixes https://github.com/spf13/cobra/issues/2422

### DIFF
--- a/completions_test.go
+++ b/completions_test.go
@@ -3727,7 +3727,7 @@ func TestGetFlagCompletion(t *testing.T) {
 					t.Errorf("Unexpected completions %q", comps)
 				}
 				if tc.directive != directive {
-					t.Errorf("Unexpected directive %s", directive.string())
+					t.Errorf("Unexpected directive %q", directive.string())
 				}
 			}
 		})
--- a/completions_test.go
+++ b/completions_test.go
@@ -3727,7 +3727,7 @@ func TestGetFlagCompletion(t *testing.T) {
 					t.Errorf("Unexpected completions %q", comps)
 				}
 				if tc.directive != directive {
-					t.Errorf("Unexpected directive %q", directive)
+					t.Errorf("Unexpected directive %s", directive.string())
 				}
 			}
 		})


marckhouzam:
Thanks.  If you can check the CI errors, then we'll be good to go.

SudhanshuMatrix:
Hi Marc,

Regarding the CI errors, I have investigated them and plan to apply the following fixes:

1. **macOS Test Failure (`abort trap / missing LC_UUID load command`)**:
   Modern macOS runner environments strictly require compiled binaries to contain an `LC_UUID` load command. Some Go versions (like `1.21.x` in the test matrix) built using internal link mode omit this. I plan to modify the `Makefile` to accept `GO_TEST_FLAGS` and update the `.github/workflows/test.yml` workflow to dynamically pass `-ldflags=-linkmode=external` when running on macOS, which forces external linking and resolves this issue.

2. **`golangci-lint` Failures (`goconst`):**
   * **Test Files**: `goconst` is flagging repeated assertion strings in `_test.go` files. I plan to update `.golangci.yml` to exclude `_test.go` files from `goconst` checks.
   * **Production Files**: The linter is flagging common short/keyword strings (`"true"`, `"bash"`, `"zsh"`, `"fish"`, `"powershell"`). Rather than refactoring these into arbitrary package constants and adding unnecessary noise to the production code, I plan to add a text-matching exclusion rule under `linters.exclusions.rules` to ignore these specific false positives.

I will update the branch with these changes once you confirm this is the correct approach. Alternatively, if you think a different approach would be better, please let me know and I will apply those changes instead.

completions_test.go
```diff
@@ -3727,7 +3727,7 @@ func TestGetFlagCompletion(t *testing.T) {
 					t.Errorf("Unexpected completions %q", comps)
 				}
 				if tc.directive != directive {
-					t.Errorf("Unexpected directive %q", directive)
+					t.Errorf("Unexpected directive %s", directive.string())
```
ccoVeille: As you concert to string, you can keep the quotes, no?

```suggestion
					t.Errorf("Unexpected directive %q", directive.string())
```

SudhanshuMatrix: Ah, my bad. You are correct now that it is converted to a string, using `%q` is indeed much better to keep the quotes. I will change it now. Sorry for the oversight, and thanks for the suggestion and review!

ccoVeille (approved):

SudhanshuMatrix:
Hey @marckhouzam, thanks for taking the time to review this we really appreciate it

From what I can see, the CI failures aren't related to my changes; they also occur on the `main` branch.

If you're okay with it, I can open a separate PR to address the issue mentioned here: https://github.com/spf13/cobra/pull/2423#issuecomment-4815662356. I have included my proposed approach to the solution, and I'd really appreciate your feedback if you think there's a better way to handle it. Otherwise, if you prefer, you can assign it to someone else.

marckhouzam:
@SudhanshuMatrix I had written an answer asking for exactly what you have done but never pressed "send" 😞 .

thanks for taking the initiative.  I'll have a look.

marckhouzam:
So, yes, please open a separate PR configuring out the warning about constants please.

SudhanshuMatrix:
@marckhouzam  Done. I have opened a separate Pull Request here: https://github.com/spf13/cobra/pull/2429 with only the CI and linter configuration updates (the macOS linking workaround and the `goconst`/`nolintlint` rules), keeping the Go 1.26 formatting changes isolated.

ccoVeille (approved):

SudhanshuMatrix:
@marckhouzam Hi! Just a follow-up on this PR. The separate CI fix has now been merged, so this PR is now passing all checks. Whenever you have some time, I'd appreciate another review. Thanks!

SudhanshuMatrix:
Hey @marckhouzam, sorry to ping again  I know maintaining is volunteer work and I really appreciate the time you and the team put into this project. No pressure at all, just wanted to flag that this PR has been reviewed and approved by @ccoVeille , all CI checks are passing, and it's just a one-line, isolated change. Whenever you get a chance to take a look, I'd appreciate it. Thanks

PR #2426, opened 2026-06-30
fix: exclude hidden commands from usage padding calculation

### LINKED ISSUE #1272: Hidden Commands Should be Ignored When Calculating Use Indentation

jfjrh2014:
Hidden commands were included in the max name/use/commandPath length calculation in AddCommand, inflating usage padding for visible siblings. This excludes hidden commands from the calculation in both AddCommand and RemoveCommand.

Fixes #1272

### DIFF
--- a/command.go
+++ b/command.go
@@ -1345,18 +1345,21 @@ func (c *Command) AddCommand(cmds ...*Command) {
 			panic("Command can't be a child of itself")
 		}
 		cmds[i].parent = c
-		// update max lengths
-		usageLen := len(x.Use)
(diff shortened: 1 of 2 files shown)


jfjrh2014:
The golangci-lint failure is pre-existing: all 50 goconst violations are in test files (`*_test.go`) and exist on the main branch, not introduced by this PR. The two files this PR touches are `command.go` and `command_test.go` — neither introduces new goconst violations.

Verified: `command.go` has zero goconst errors in the lint output.

PR #2427, opened 2026-06-30
Fix directive formatting in completion test

HNO3Miracle:
Fix a completion test diagnostic that formats ShellCompDirective with %q. Current Go vet reports this as an invalid format string.

ShellCompDirective is a numeric bitmask type, so %v is the appropriate diagnostic format here.

Tested:
- go test ./...


marckhouzam:
Duplicate of #2423

PR #2429, opened 2026-07-01
fix: resolve macOS test link failure and update lint rules

SudhanshuMatrix:
### Description

This Pull Request addresses macOS test linking failures and linter configuration adjustments to ensure clean CI/CD runs and local builds.

As suggested in https://github.com/spf13/cobra/pull/2423#issuecomment-4846919718, this PR isolates these config and CI updates into a separate branch.

---

#### 1. macOS Missing Link Flags / Abort Trap
On macOS runner environments, test runs fail abruptly with `abort trap` due to a missing `LC_UUID` load command in the compiled test binaries:
```
******************** downloading dependencies ********************
go get -v ./...
******************** running tests with kyoh86/richgo ********************
richgo test -v ./...
     | dyld[6013]: missing LC_UUID load command in /private/var/folders/8j/sfr9qqcj73j4p6nhwcfpr0th0000gn/T/go-build959790180/b077/cobra.test
     | dyld[6013]: missing LC_UUID load command
     | signal: abort trap
FAIL | 	github.com/spf13/cobra	0.002s
     | dyld[6011]: missing LC_UUID load command in /private/var/folders/8j/sfr9qqcj73j4p6nhwcfpr0th0000gn/T/go-build959790180/b103/doc.test
     | dyld[6011]: missing LC_UUID load command
     | signal: abort trap
FAIL | 	github.com/spf13/cobra/doc	0.002s
make: *** [richtest] Error 1
Error: Process completed with exit code 2.
```

- **Failing CI Job Link**: [Link to failing macOS action run](https://github.com/SudhanshuMatrix/cobra/actions/runs/28704047114/job/85126780647)

- **Failing CI Job Link**: [Link to failing macOS action run](https://github.com/SudhanshuMatrix/cobra/actions/runs/28704047114/job/85126780639)
<details>
<summary><b>macOS Link Failure Screenshot</b></summary>
image

</details>

This is resolved by passing `-ldflags=-linkmode=external` to force external linking via the system linker on macOS. Since Go 1.24+ natively resolves this issue in its internal linker, this workaround is constrained to Go versions `< 1.24`.

#### 2. `golangci-lint` Failures (`goconst`)
The `goconst` linter reports multiple alerts on test files for repeating test assertions and commands:
```
Error: active_help_test.go:51:3: string `Completion ended with directive: ShellCompDirectiveDefault` has 48 occurrences, make it a constant (goconst)
		"Completion ended with directive: ShellCompDirectiveDefault", ""}, "\n")
		^
Error: active_help_test.go:62:10: string `The child command` has 7 occurrences, make it a constant (goconst)
		Short: "The child command",
		^
Error: active_help_test.go:198:3: string `Completion ended with directive: ShellCompDirectiveNoFileComp` has 100 occurrences, make it a constant (goconst)
		"Completion ended with directive: ShellCompDirectiveNoFileComp", ""}, "\n")
		^
Error: command_test.go:375:12: string `1.0.0` has 17 occurrences, make it a constant (goconst)
		Version: "1.0.0",
		         ^
```

- **Failing CI Job Link**: [Link to failing golangci-lint action run](https://github.com/SudhanshuMatrix/cobra/actions/runs/28704047114/job/85126780608)

<details>
<summary><b>Linter Failure Screenshot</b></summary>

image

</details>

We resolve this by excluding `_test.go` files from `goconst` checks.

#### 3. `nolintlint` Unused Directive Warnings
In Go/staticcheck versions where type compatibility checks do not trigger warnings on type assertions in tests, `nolintlint` fails the build reporting unused directives:
```
completions_test.go:2902:35: directive `//nolint:staticcheck` is unused (nolintlint)
```
However, removing the comment causes builds to fail on other matrix environments (e.g. older Go versions) where the warning *is* triggered. We resolve this by adding a targeted exclusion rule to allow unused `staticcheck` directives in `_test.go` files.

---

### Changes Proposed

1. **macOS Test Link Fix**:
   - Added `GO_TEST_FLAGS: -ldflags=-linkmode=external` environment variable for macOS platforms under Go `< 1.24` in the GitHub Actions `test-unix` job (with an explanatory comment).
   - Updated the `Makefile` (`test` and `richtest` targets) to reference `$(GO_TEST_FLAGS)`, keeping target execution platform-independent while resolving macOS-specific link issues.

2. **Linter Rule Exclusions & Hygiene (`.golangci.yml`)**:
   - Disabled `goconst` on all `_test.go` files to prevent noisy alerts on duplicated test strings.
   - Added `goconst` exclusions for common shell inputs (`true`, `bash`, `zsh`, `fish`, `powershell`).
   - Replaced the global `allow-unused: true` setting for `nolintlint` with a specific exclusion rule under `linters.exclusions.rules` that ignores unused `staticcheck` directives in `_test.go` files, keeping `nolintlint` fully active for all other linters and production files.

---

### Verification Results

* **Unit Tests**: All tests pass successfully locally (`go test ./...`).
* **Linter Validation**: Verified locally via `golangci-lint run` (0 issues). Both local and CI environments successfully validate with these changes.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -71,6 +71,8 @@ jobs:
         - 24
     name: '${{ matrix.platform }} | 1.${{ matrix.go }}.x'
     runs-on: ${{ matrix.platform }}-latest
+    # macOS runner environments require compiled binaries to contain an LC_UUID load command.
+    # We dynamically pass -ldflags=-linkmode=external on macOS to force external linking via the system linker.
     env:
       GO_TEST_FLAGS: ${{ matrix.platform == 'macOS' && '-ldflags=-linkmode=external' || '' }}
     steps:
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -72,9 +72,10 @@ jobs:
     name: '${{ matrix.platform }} | 1.${{ matrix.go }}.x'
     runs-on: ${{ matrix.platform }}-latest
     # macOS runner environments require compiled binaries to contain an LC_UUID load command.
-    # We dynamically pass -ldflags=-linkmode=external on macOS to force external linking via the system linker.
+    # Go 1.24+ natively resolves this by writing LC_UUID in Go's internal linker.
+    # For Go < 1.24, we pass -ldflags=-linkmode=external on macOS to force external linking via the system linker.
     env:
-      GO_TEST_FLAGS: ${{ matrix.platform == 'macOS' && '-ldflags=-linkmode=external' || '' }}
+      GO_TEST_FLAGS: ${{ matrix.platform == 'macOS' && matrix.go < 24 && '-ldflags=-linkmode=external' || '' }}
     steps:
 
     - uses: actions/checkout@v4
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -71,6 +71,8 @@ jobs:
         - 24
     name: '${{ matrix.platform }} | 1.${{ matrix.go }}.x'
     runs-on: ${{ matrix.platform }}-latest
+    env:
+      GO_TEST_FLAGS: ${{ matrix.platform == 'macOS' && '-ldflags=-linkmode=external' || '' }}
     steps:
 
     - uses: actions/checkout@v4
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -57,7 +57,17 @@ linters:
       - common-false-positives
       - legacy
       - std-error-handling
+    warn-unused: true
+    rules:
+      - path: '_test\.go'
+        linters:
+          - goconst
+      - linters:
+          - goconst
+        text: 'string `(true|bash|zsh|fish|powershell)`'
   settings:
+    nolintlint:
+      allow-unused: true
     govet:
       # Disable buildtag check to allow dual build tag syntax (both //go:build and // +build).
       # This is necessary for Go 1.15 compatibility since //go:build was introduced in Go 1.17.
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -65,9 +65,14 @@ linters:
       - linters:
           - goconst
         text: 'string `(true|bash|zsh|fish|powershell)`'
+      # Some Go/staticcheck versions in the test matrix don't trigger staticcheck warnings
+      # on type assertions/LHS variables in test files, while others do.
+      # To avoid build failures on both sides, we allow unused staticcheck nolints in test files.
+      - path: '_test\.go'
+        linters:
+          - nolintlint
+        text: 'directive `//nolint:staticcheck` is unused'
   settings:
-    nolintlint:
-      allow-unused: true
     govet:
       # Disable buildtag check to allow dual build tag syntax (both //go:build and // +build).
       # This is necessary for Go 1.15 compatibility since //go:build was introduced in Go 1.17.
--- a/Makefile
+++ b/Makefile
@@ -21,11 +21,11 @@ lint:
 
 test: install_deps
 	$(info ******************** running tests ********************)
-	go test -v ./...
+	go test -v $(GO_TEST_FLAGS) ./...
 
 richtest: install_deps
 	$(info ******************** running tests with kyoh86/richgo ********************)
-	richgo test -v ./...
+	richgo test -v $(GO_TEST_FLAGS) ./...
 
 install_deps:
 	$(info ******************** downloading dependencies ********************)


.github/workflows/test.yml
```diff
@@ -71,6 +71,8 @@ jobs:
         - 24
     name: '${{ matrix.platform }} | 1.${{ matrix.go }}.x'
     runs-on: ${{ matrix.platform }}-latest
+    env:
+      GO_TEST_FLAGS: ${{ matrix.platform == 'macOS' && '-ldflags=-linkmode=external' || '' }}
```
ccoVeille: Maybe a comment here explaining why this is needed

.golangci.yml
```diff
@@ -57,7 +57,17 @@ linters:
...
+    rules:
+      - path: '_test\.go'
+        linters:
+          - goconst
+      - linters:
+          - goconst
+        text: 'string `(true|bash|zsh|fish|powershell)`'
   settings:
+    nolintlint:
+      allow-unused: true
```
ccoVeille: I'm unsure about this.

It's an hammer for small problems.

I feel would have used a an exclusion in the files involved.

What is the exact errors that no longer report ?

I can see only 2 use cases for nolint in cobra

//nolint:gosec in code

Would //nolint:nolint works if added ?

//nolint:staticcheck in test files

I would use an esclusion rule about nolint reporting staticcheck in test files

Anyway whatever the solution a comment is required.

ccoVeille: Also, I don't replicate the nolint issue you are reporting

```console
$ go version                                                                                                                                                                           
go version go1.26.4 linux/amd64

$ golangci-lint version --short                                                                                                                                                        

$ golangci-lint run | tail -n 3
50 issues:
* goconst: 50
```

SudhanshuMatrix: Hi @ccoVeille ,
You're correct, `allow-unused: true` is too broad. I've removed the global setting and replaced it with a targeted exclusion rule under `rules` in `.golangci.yml` instead:
```yaml
      # Go/staticcheck versions behave differently on type assertions in test files.
      # We allow unused staticcheck nolints in test files to prevent failures on both sides.
      - path: '_test\.go'
        linters:
          - nolintlint
        text: 'directive `//nolint:staticcheck` is unused'

.github/workflows/test.yml
SudhanshuMatrix: Thanks for the review and suggestion.

On macOS, runner environments require compiled binaries to include an `LC_UUID` load command. To ensure this, we dynamically pass `-ldflags=-linkmode=external` on macOS, which forces external linking through the system linker and results in binaries containing the required `LC_UUID`.

.golangci.yml
SudhanshuMatrix: Why this happens & How to reproduce:
Different Go/staticcheck versions in the CI matrix behave differently on type assertions (e.g. `completions_test.go:2902`):

Some versions flag them (requiring `//nolint:staticcheck`).
Others do not, causing `nolintlint` to fail with `unused nolint directive`.

You can reproduce this by removing the exclusion rule and running `golangci-lint run` under a version that doesn't trigger the warning (like Go 1.26). You can also inspect the failed runs in my fork's Actions page at https://github.com/SudhanshuMatrix/cobra/actions before this fix was pushed.

.github/workflows/test.yml
ccoVeille: I'm surprised setup-go doesn't set it up

~~Maybe because it's an old setup-go that is used~~

Edit: I'm wrong setup-go is the latest

ccoVeille:
@SudhanshuMatrix 

Could you edit the description to provide a clear message if what was reported for the following use cases

- macOS missing flag

- nolint issue with staticcheck

I'm looking for copy paste for the error provided in the CI

If possible, please also with direct link to failing jobs.

I would like to be able to refer to them in conversation with other gophers to talk about it

SudhanshuMatrix:
@ccoVeille Done Let me know if there's anything else you'd like me to update.

ccoVeille (approved):

.github/workflows/test.yml
marckhouzam: Apparently this is happening because the MacOS runner was updated and now enforces that all executables must include an LC_UUID load command.



I asked Gemini why it is only failing for our 1.21 and 1.22 Go MacOS runners.  Further below is the answer, for posterity.  However, starting with Go 1.24, it works automatically.  @SudhanshuMatrix could you update the condition above to limit it to Go < 1.24?

**Why we see some success and some failures**

1. Why Go 1.17 to 1.20 pass
In older versions of Go (1.20 and below), the Go toolchain on macOS defaulted to external linking (CGO_ENABLED=1) for standard test execution when a C compiler (clang) was available in the environment—which it always is on GitHub Actions' macOS runners.

Because it used the system's external linker, clang and ld automatically stamped the LC_UUID onto the test binary. Go's internal linker was bypassed entirely for these tests, shielding them from the missing command issue.

2. Why Go 1.21 and 1.22 fail
Starting explicitly in Go 1.21, the Go team optimized the toolchain on macOS to use its internal linker by default more aggressively for pure Go code and basic tests to speed up compile times.

When your Cobra CI matrix hits Go 1.21 and 1.22 on a modern GitHub macOS runner (which forces strict dyld validation):

Go switches to internal linking.

Go's internal linker (at that time) forgets to write the LC_UUID load command.

dyld catches the omission and throws the abort trap.

3. Why Go 1.23 passes (The GitHub Actions Trap)
This is usually the most confusing part. Go 1.23's internal linker also lacked the LC_UUID command. So why does it pass?

It depends on the runner image provisioned by GitHub. GitHub Actions frequently pins older Go toolchains (like 1.21 and 1.22) to older or cross-compiled Intel/ARM runner configurations where setup-go downloads pre-compiled legacy binaries. Conversely, when setup-go sets up Go 1.23 on a brand new macOS GitHub runner runner, it often triggers default environment flags or utilizes fallback mechanism constraints (like specific Xcode SDK paths) that accidentally force Go 1.23 back into external linking mode.

4. Why Go 1.24+ passes
Go 1.24 natively resolved the root issue. The Go team updated cmd/link for the Darwin platform so that Go's internal linker now explicitly generates and writes a valid LC_UUID to the Mach-O header. No matter how it gets compiled, Go 1.24+ binaries comply perfectly with macOS's strict dynamic linker rules.

ccoVeille: Thanks for the investigation, it was not that obvious

SudhanshuMatrix: @marckhouzam Sorry for the delay I have updated the condition now

marckhouzam (approved):
LGTM
Thanks for this!

PR #2431, opened 2026-07-05
Fix hidden and deprecated commands changing padding in help text

### LINKED ISSUE #1272: Hidden Commands Should be Ignored When Calculating Use Indentation

natsu0529:
Hidden and deprecated commands are not shown in the help text, so they should not affect the padding used to align the visible commands. Currently, a subcommand with a long name that is hidden or deprecated makes the help text of its siblings wider than necessary.

Fixes #1272
Supersedes #1632 (now conflicting and inactive since 2022 — this PR re-applies the approach that was already approved there, rebased onto current main)

## Approach

As discussed in #1632, this checks `.Hidden` and `.Deprecated` explicitly rather than using `IsAvailableCommand()`, because the latter would incorrectly exclude doc-only commands (`AdditionalHelpTopic`, e.g. `go help go.mod`) which *are* shown in the help text (pointed out by @Luap99).

- `AddCommand`: skip hidden/deprecated commands when updating `commandsMaxUseLen`, `commandsMaxCommandPathLen` and `commandsMaxNameLen`
- `RemoveCommand`: skip them likewise when recomputing the max lengths (not covered by the original PR)
- Added `TestPadding` in `command_test.go` covering both cases

## Example

Before:

```
Available Commands:
  breakdown                                  Show full breakdown of costs
  comment                                    Post a comment
```

After:

```
Available Commands:
  breakdown   Show full breakdown of costs
  comment     Post a comment
```

(where a hidden `fig-autocomplete-very-long-hidden-command` subcommand exists)

## Regarding the open question in #1632

@jpmcb raised the concern that users who override the help template to display hidden/deprecated commands could be affected. For such users, this change only reduces the padding value returned by `NamePadding()`/`UsagePadding()`/`CommandPathPadding()`, so their custom templates keep working — long hidden entries would simply extend past the alignment column, exactly as any over-long visible entry already does today (the minimum paddings `minNamePadding` etc. are unchanged). Users who want the old behavior in a custom template can compute their own padding.

### DIFF
--- a/command.go
+++ b/command.go
@@ -1345,18 +1345,22 @@ func (c *Command) AddCommand(cmds ...*Command) {
 			panic("Command can't be a child of itself")
 		}
 		cmds[i].parent = c
-		// update max lengths
-		usageLen := len(x.Use)
-		if usageLen > c.commandsMaxUseLen {
-			c.commandsMaxUseLen = usageLen
-		}
-		commandPathLen := len(x.CommandPath())
-		if commandPathLen > c.commandsMaxCommandPathLen {
-			c.commandsMaxCommandPathLen = commandPathLen
-		}
-		nameLen := len(x.Name())
-		if nameLen > c.commandsMaxNameLen {
-			c.commandsMaxNameLen = nameLen
+		// hidden and deprecated commands are not shown in the help text
+		// so they should not affect the padding
+		if len(x.Deprecated) == 0 && !x.Hidden {
+			// update max lengths
+			usageLen := len(x.Use)
+			if usageLen > c.commandsMaxUseLen {
+				c.commandsMaxUseLen = usageLen
+			}
+			commandPathLen := len(x.CommandPath())
+			if commandPathLen > c.commandsMaxCommandPathLen {
+				c.commandsMaxCommandPathLen = commandPathLen
+			}
+			nameLen := len(x.Name())
+			if nameLen > c.commandsMaxNameLen {
+				c.commandsMaxNameLen = nameLen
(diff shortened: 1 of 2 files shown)


command.go
```diff
@@ -1345,18 +1345,22 @@ func (c *Command) AddCommand(cmds ...*Command) {
...
-		commandPathLen := len(x.CommandPath())
-		if commandPathLen > c.commandsMaxCommandPathLen {
-			c.commandsMaxCommandPathLen = commandPathLen
-		}
-		nameLen := len(x.Name())
-		if nameLen > c.commandsMaxNameLen {
-			c.commandsMaxNameLen = nameLen
+		// hidden and deprecated commands are not shown in the help text
+		// so they should not affect the padding
+		if len(x.Deprecated) == 0 && !x.Hidden {
```
natsu0529, line 1350: Note on semantics: this check happens at `AddCommand` time, so setting `Hidden`/`Deprecated` *after* the command has been added will not be reflected in the padding. This is consistent with the existing behavior — `Use` and `Name` are also snapshotted here, so mutating them after `AddCommand` has never updated the padding either. Visibility is expected to be set before `AddCommand`, which is also how cobra configures its own built-in commands (e.g. the completion command).

If we ever want to remove this whole class of staleness (including post-add `Use` changes), the paddings could be computed lazily in `UsagePadding()`/`NamePadding()`/`CommandPathPadding()` instead of cached — help rendering is not a hot path. I kept that out of scope here to stay aligned with the approach already approved in #1632.

PR #2433, opened 2026-07-07
feat: Add MarkFlagAppendEqual to suggest --flag= in shell completion

oscrx:
# Add `MarkFlagAppendEqual` to suggest `--flag=` in shell completion

## Summary

This adds an opt-in, per-flag mechanism to make shell completion suggest a
flag's long form with a trailing `=` and **no** trailing space
(e.g. `--flag=`), instead of `--flag ` (with a space).

It is intended for flags whose value is most naturally supplied with the
`--flag=value` form — in particular flags that carry an *optional* value via
`NoOptDefVal`, where the space-separated form (`--flag value`) does not bind the
value at all.

## Motivation

This came out of wanting to improve autocompletion for `kubectl`'s `--dry-run`
flag. `--dry-run` carries an optional value (`--dry-run=client` /
`--dry-run=server`, with a deprecated bare `--dry-run`), so it sets
`NoOptDefVal`. Today `kubectl <cmd> --dry-run<TAB>` completes to `--dry-run `
(with a space) and then offers file completion, even though the only working way
to pass a value is `--dry-run=client`. There is currently no way for a downstream
program like kubectl to make cobra suggest `--dry-run=` instead — hence this
opt-in flag.

For a flag with `NoOptDefVal` set, the only syntax that actually assigns a value
is `--flag=value`; `--flag value` treats `value` as a positional argument. Yet
today completion of the flag name produces `--flag` followed by a space, which:

- nudges the user toward the non-working space-separated form, and
- leaves them having to delete the space before typing `=value`.

Cobra used to suggest both `--flag` and `--flag=` for value flags, but that was
removed because suggesting *both* forms for *every* flag was noisy and
duplicative (see the long-standing comment in `getFlagNameCompletions`). This
change takes a different, narrower approach: it is **opt-in per flag**, so it
does not reintroduce duplicate suggestions for flags that don't want it.

## What changed

1. **New public API** `Command.MarkFlagAppendEqual(flagName string) error` and
   the `FlagAppendEqualAnnotation` key. Marking a flag makes its long-form
   completion be suggested as `--flag=`.

2. **`getFlagNameCompletions`** appends `=` to the suggested long form when the
   flag opted in.

3. **Fix to the existing `NoSpace` detection.** The code that adds
   `ShellCompDirectiveNoSpace` for a single `=`-suffixed completion compared the
   *entire* completion string against `"="`. When descriptions are enabled a
   completion is `"choice\tdescription"`, so it never ended in `=` and the
   branch was effectively dead. It now inspects only the choice (the text before
   the tab). It also OR-s the directive in (`|=`) instead of overwriting, so the
   pre-existing `ShellCompDirectiveNoFileComp` is preserved (the result is
   `NoSpace | NoFileComp` rather than `NoSpace` alone).

## Example

```go
cmd.Flags().String("dry-run", "none", `Must be "none", "server", or "client".`)
cmd.Flags().Lookup("dry-run").NoOptDefVal = "unchanged"
_ = cmd.MarkFlagAppendEqual("dry-run")
```

```
$ prog __complete cmd "--dry-run"
--dry-run=	Must be "none", "server", or "client".
:6            # ShellCompDirectiveNoSpace | ShellCompDirectiveNoFileComp
```

The shell inserts `--dry-run=` with the cursor right after `=`, ready for value
completion, and does not fall back to file completion.

## Backward compatibility

Fully backward compatible. The behavior is opt-in: flags that are not marked with
`MarkFlagAppendEqual` are completed exactly as before. The `NoSpace` detection
fix only affects completions that already end in `=`, which previously could not
be produced for flag names.

## Testing

- Added `TestFlagNameCompletionAppendEqual` covering: a marked flag
  (`--flag=` + `:6`), an unmarked flag (unchanged, `:4`), and the error returned
  for a non-existent flag.
- `go test ./...` passes with no regressions.

### DIFF
--- a/completions.go
+++ b/completions.go
@@ -182,6 +182,25 @@ func (c *Command) RegisterFlagCompletionFunc(flagName string, f CompletionFunc) 
 	return nil
 }
 
+// FlagAppendEqualAnnotation is the annotation key used to mark a flag so that
+// shell completion suggests its long form with a trailing "=" (e.g. "--flag=")
+// and without a trailing space. This is opt-in per flag, so it does not
+// reintroduce the duplicate "--flag"/"--flag=" suggestions that led to the
+// "=" form being removed for all flags.
+const FlagAppendEqualAnnotation = "cobra_annotation_flag_append_equal"
+
+// MarkFlagAppendEqual instructs shell completion to suggest the given flag's
+// long name with a trailing "=" and no trailing space. This is convenient for
+// flags whose value is most commonly provided with the "--flag=value" form,
+// such as flags that carry an optional value (NoOptDefVal).
+func (c *Command) MarkFlagAppendEqual(flagName string) error {
+	flag := c.Flag(flagName)
+	if flag == nil {
+		return fmt.Errorf("MarkFlagAppendEqual: flag '%s' does not exist", flagName)
+	}
+	return c.Flags().SetAnnotation(flagName, FlagAppendEqualAnnotation, []string{"true"})
+}
+
 // GetFlagCompletionFunc returns the completion function for the given flag of the command, if available.
 func (c *Command) GetFlagCompletionFunc(flagName string) (CompletionFunc, bool) {
 	flag := c.Flag(flagName)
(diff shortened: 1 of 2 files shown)


Luap99 (commented):
I am not a maintainer so just some comments.

Would it not be much simpler to do this by default with NoOptDefVal and not deal with another function/annotation at all for this. this makes so that applications actually benifit by default and not anybody must calls also MarkFlagAppendEqual for each NoOptDefVal they use

The completion code already handles NoOptDefVal in another path

oscrx:
Thanks for the review, I explicitly didn't touch the default behaviour.

The --flag= suggestion was actually the default once (#1048) and was deliberately disabled globally in #1070 because it always showed up for every value flag.
https://github.com/spf13/cobra/blame/ad460ea8f249db69c943a365fb84f3a59042d54e/completions.go#L610
Making it default again for NoOptDefVal flags would reintroduce that same noise — and since every boolean flag sets NoOptDefVal, it'd hit the most common flags hardest. MarkFlagAppendEqual is the opt-in way to bring = back only where it's actually correct, without regressing #1070.

oscrx:
Hey @marckhouzam do you have some time for a review?

oscrx:
Hi @marckhouzam @johnSchnake Following up since it's been a few weeks. This is currently blocking kubernetes/kubernetes#140299, so happy to make any changes needed to get it merge-ready. Let me know if there's anything I can do to help move review along.

oscrx:
Hey @jpmcb do you have some time by any change to look at this PR?
It's been a few weeks again and maybe tagging someone new will help :)

PR #2445, opened 2026-07-20
fix: required flags still show (default: ...) value

### LINKED ISSUE #1544: Required flags still show (default: ...) value
If you mark a flag as required using `MarkFlagRequired` then it doesn't have a default value, per se: it will always be present on the command line.

However, the help message still shows `(default: ...)` for that flag.

Is there any way to remove that part of the help message?

hazrid93:
Closes #1544.

Add a Command.FlagUsagesTrimRequired(fs) method that returns pflag's FlagUsages output for fs but with the trailing "(default ...)" suffix removed for flags marked required on this command (detected via the BashCompOneRequiredFlag annotation). It is usable in custom usage/help templates as {{.FlagUsagesTrimRequired .LocalFlags}}, so a tool can omit required flag defaults while leaving optional flags' defaults intact.

### DIFF
--- a/command.go
+++ b/command.go
@@ -24,6 +24,7 @@ import (
 	"io"
 	"os"
 	"path/filepath"
+	"regexp"
 	"sort"
 	"strings"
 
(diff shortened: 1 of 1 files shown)


command.go
```diff
@@ -1592,6 +1593,39 @@ func (c *Command) HasExample() bool {
...
+		if ok && len(ann) > 0 && ann[0] == "true" {
+			required[f.Name] = true
+		}
+	})
+	if len(required) == 0 {
+		return out
+	}
+	lines := strings.Split(out, "\n")
+	for i, line := range lines {
+		name := flagLineName(line)
```
Sanjays2402, line 1620: flagLineName stops at the first space/tab/=, but pflag renders the type on the same line ("--name string   usage"), so for a required string flag the trailing "(default \"x\")" you strip lives past the usage text and this matches fine. the one that bites: if a flag's usage text itself contains "(default ...)" you'll strip the wrong suffix since the regex is anchored to end of line only. probably rare enough to not matter, but worth a test with a multiline/awkward usage string.

PR #2447, opened 2026-07-20
feat: missing support for removing group

### LINKED ISSUE #1911: no support for removing group

hazrid93:
Closes #1911.

Add RemoveGroup(ids...) that removes the groups with matching IDs from the command, mirroring RemoveCommand's pattern. An ID with no matching group is a no op; removing a group does not remove the commands assigned to it.

### DIFF
--- a/command.go
+++ b/command.go
@@ -1397,6 +1397,23 @@ func (c *Command) AddGroup(groups ...*Group) {
 	c.commandgroups = append(c.commandgroups, groups...)
 }
 
+// RemoveGroup removes one or more command groups from a parent command by ID.
+// Groups whose ID matches one of the given ids are removed; an ID with no
+// matching group is a no-op. See issue #1911.
+func (c *Command) RemoveGroup(ids ...string) {
+	groups := []*Group{}
+main:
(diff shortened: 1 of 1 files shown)


command.go
```diff
@@ -1397,6 +1397,23 @@ func (c *Command) AddGroup(groups ...*Group) {
...
+	for _, group := range c.commandgroups {
+		for _, id := range ids {
+			if group.ID == id {
+				continue main
+			}
+		}
+		groups = append(groups, group)
+	}
+	c.commandgroups = groups
+}
```
Sanjays2402, line 1415: removing the group leaves any subcommand that still has GroupID set to it dangling — next Execute() runs checkCommandGroups() which panics with "group id '%s' is not defined for subcommand". should this also clear GroupID on children whose group just got removed, or at least doc that the caller has to reassign them first?

PR #2450, opened 2026-07-20
feat: ReparseFlags: force a clean re parse of flags

### LINKED ISSUE #1832: Function to reparse flags
It would be good for Cobra to have a function to  force reparsing for flags. eg. after the application has already started and the initial flag parsing has already been done.

```go
var reverse bool

cmd.Flags().BoolVarP(&reverse, "reverse", "r", false, "")

cmd.ParseFlags(args) // Neither work
cmd.Flags().Parse(args) // Neither work

```
My use case is a bit weird. I'm assuming most people's applications execute a command and then exit immediately.

In my case, the application continually runs. It executes commands, and then awaits a new command and loops.

Cobra seems to always remember the original flag values and never updates them when you run the command again with a different flag value.

hazrid93:
Closes #1832.

Add ReparseFlags(args): resets the command's local and persistent flags to their default values (each flag's Value.Set(DefValue) + Changed=false), then re parses via ParseFlags. Honors DisableFlagParsing (no op). Existing ParseFlags is unchanged.

### DIFF
--- a/command.go
+++ b/command.go
@@ -2070,3 +2070,25 @@ func defaultVersionFunc(w io.Writer, in interface{}) error {
 	_, err := fmt.Fprintf(w, "%s version %s\n", c.DisplayName(), c.Version)
 	return err
 }
+
+// ReparseFlags resets the command's local and persistent flags to their
+// default values and then re-parses the given args. This is useful when you
+// need to re-parse flags after the application has already started and the
+// initial flag parsing has already been done (e.g. to apply a new set of
+// args without leaving previously-parsed flags in place). See issue #1832.
+func (c *Command) ReparseFlags(args []string) error {
+	if c.DisableFlagParsing {
+		return nil
+	}
+
+	reset := func(fs *flag.FlagSet) {
+		fs.VisitAll(func(f *flag.Flag) {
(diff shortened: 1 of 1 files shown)


command.go
```diff
@@ -2070,3 +2070,25 @@ func defaultVersionFunc(w io.Writer, in interface{}) error {
...
+// need to re-parse flags after the application has already started and the
+// initial flag parsing has already been done (e.g. to apply a new set of
+// args without leaving previously-parsed flags in place). See issue #1832.
+func (c *Command) ReparseFlags(args []string) error {
+	if c.DisableFlagParsing {
+		return nil
+	}
+
+	reset := func(fs *flag.FlagSet) {
+		fs.VisitAll(func(f *flag.Flag) {
```
Sanjays2402, line 2085: does this actually round-trip for slice/array flags? pflag's stringArrayValue.Set appends after the first call and keeps its own internal `changed` bool that Set(f.DefValue) here never clears, and DefValue for a slice is the formatted form like "[a,b]" rather than a value Set can consume element-by-element. so reparsing a StringArray flag looks like it would append "[a,b]" onto the existing slice instead of resetting it. worth a test with a StringArray/StringSlice flag before/after.

PR #2455, opened 2026-07-22
fix: use DisplayName for default version flag usage

sonnemusk:
Align version flag usage with help flag DisplayName behavior; fix related comment typos.


sonnemusk:
Closing this PR. After review, this is too low-value / looks like bulk automated noise rather than a carefully chosen change. Sorry for the noise — will focus on real, high-signal fixes only.

PR #2459, opened 2026-07-25
fix: do not claim the -h shorthand if it is already used by another flag

### LINKED ISSUE #2359: InitDefaultHelpFlag is not i18n aware

Sanjays2402:
Closes #2359

### Problem

`InitDefaultHelpFlag` only checks whether a flag with the long name `help` exists before registering the default help flag with the `h` shorthand:

```go
if c.Flags().Lookup(helpFlagName) == nil {
    ...
    c.Flags().BoolP(helpFlagName, "h", false, usage)
}
```

If the user has already defined a flag that uses the `h` shorthand under a different long name (for example a translated `ayuda` flag, or any app-specific `-h` flag such as `--host`), the long-name lookup misses it and pflag panics:

```
panic: unable to redefine 'h' shorthand in "root" flagset: it's already used for "nothelp" flag
```

### Fix

Register the default help flag without a shorthand when `h` is already taken. This mirrors exactly how `InitDefaultVersionFlag` already handles the `v` shorthand, so the behaviour is consistent between the two built-in flags. `--help` still works; only the shorthand is yielded to the user flag that claimed it first.

### Test

Added `TestShorthandHelpFlagOnlyAddedIfShorthandNotDefined`, written in the same style as the existing `TestShorthandVersionFlagOnlyAddedIfShorthandNotDefined`.

Without the change to `command.go` the new test panics:

```
--- FAIL: TestShorthandHelpFlagOnlyAddedIfShorthandNotDefined (0.00s)
panic: unable to redefine 'h' shorthand in "root" flagset: it's already used for "nothelp" flag
```

With the change, the whole package passes:

```
$ go test .
ok  	github.com/spf13/cobra	3.881s
```


ccoVeille:
Sounds similar to 

- https://github.com/spf13/cobra/pull/2367

Sanjays2402:
Yes, same issue (#2359), but the two PRs resolve it in opposite directions.

#2367 keeps the panic and just replaces pflag's message with a clearer one saying `-h` is reserved for help. This one doesn't panic at all: if `h` is already taken, the default help flag is registered as `--help` with no shorthand, and the user's flag keeps `-h`.

The reason I went that way is that `InitDefaultVersionFlag` already does exactly this a few lines below:

```go
if c.Flags().ShorthandLookup("v") == nil {
    c.Flags().BoolP("version", "v", false, usage)
} else {
    c.Flags().Bool("version", false, usage)
}
```

So a user flag claiming `-v` silently yields the shorthand, while claiming `-h` blows up. Mirroring the version behaviour makes the two built-ins consistent and fixes the i18n case from #2359 without asking translators to avoid `-h`.

The counter-argument for #2367 is that `-h` is more load-bearing than `-v` and silently dropping it could surprise people who expect `-h` to always print help. If maintainers prefer that stance, #2367 is the right patch and I'll close this one. Happy either way, just flagging that the current asymmetry with the version flag is the part that seems unintended.

PR #2460, opened 2026-07-26
fix(args): use invalid arg instead of args[0] for OnlyValidArgs suggestions

sergioperezcheco:
OnlyValidArgs walks every positional arg and reports the first one that is not in ValidArgs, but it computed the "Did you mean?" suggestion from args[0] instead of the offending value. When a valid argument preceded the invalid one (e.g. cmd one tw where "one" is valid), the suggestion was based on "one" — an already-valid token — so no useful near-match was ever produced for the actual typo. The fix passes the invalid argument v to findSuggestions, matching the pattern already used by legacyArgs, so the suggestion reflects what the user actually mistyped.

Added a regression test that supplies a valid first arg, an invalid second arg that is a near-match for a subcommand, and asserts the suggestion points at that subcommand.


args.go
```diff
@@ -58,7 +58,7 @@ func OnlyValidArgs(cmd *Command, args []string) error {
 		}
 		for _, v := range args {
 			if !stringInSlice(v, validArgs) {
-				return fmt.Errorf("invalid argument %q for %q%s", v, cmd.CommandPath(), cmd.findSuggestions(args[0]))
+				return fmt.Errorf("invalid argument %q for %q%s", v, cmd.CommandPath(), cmd.findSuggestions(v))
```
Sanjays2402, line 61: passing v instead of args[0] is right, but findSuggestions -> SuggestionsFor only walks c.commands, it never looks at ValidArgs. so for the normal OnlyValidArgs case (valid args, no subcommands) you still get no "Did you mean" for either arg. your test only passes because it adds a twox subcommand — drop that and it goes quiet again. the suggestion source seems like the actual gap here.

sergioperezcheco:
Good catch. You're right — findSuggestions only walks c.commands so ValidArgs were never considered. I extended findSuggestions to also fuzzy-match ValidArgs via levenshtein distance (same threshold as subcommands), so the suggestion now covers both subcommands and valid arguments. Updated the test to drop the twox subcommand crutch and instead verify the suggestion comes directly from ValidArgs (e.g. 'twe' -> 'two').

PR #2461, opened 2026-07-27
test: fix printf vet error for ShellCompDirective on Go 1.26

### LINKED ISSUE #2422: Test failed with golang 1.26

sergioperezcheco:
Go 1.26 tightened the printf analyzer (golang/go#78486) so that %q is now rejected for non-string operands. The completion directive comparison in TestGetFlagCompletion uses %q to format directive, but directive is a ShellCompDirective (an int), so building the tests with Go 1.26 fails vet with:

  completions_test.go:3730:37: (*testing.common).Errorf format %q has arg directive of wrong type github.com/spf13/cobra.ShellCompDirective

This switches that one format verb from %q to %d. directive is the numeric completion directive bitmask, so %d is the accurate rendering and matches how ShellCompDirective is formatted elsewhere. I left the neighboring completions %q untouched since comps is []string, and verified there are no other %q-on-int sites in the file (the %q at line 4070 compares string fields). go vet ./... and the full test suite pass on Go 1.26.4.

Closes #2422.


ccoVeille:
Duplicate of 

sergioperezcheco:
You're right, thanks for pointing that out. Closing in favor of #2423 since it was first. My version used %d for the bitmask which is arguably more direct, but #2423's approach of using the string() helper is also fine — happy to help move #2423 forward if it's stuck.

PR #2467, opened 2026-07-28
fix(test): use %d instead of %q for ShellCompDirective in completion test

### LINKED ISSUE #2422: Test failed with golang 1.26

sergioperezcheco:
Go 1.26 tightens format-string checking (golang/go#78486) and now flags %q when the argument is an integer-backed type rather than a string. completions_test.go:3730 passes a ShellCompDirective (int) to %q, so `go vet` fails under the new toolchain.

This switches the verb to %d, which prints the numeric directive value — the same information the assertion was already trying to convey. The string comparisons on line 4070 stay on %q because those operands are real strings. go vet and the full completion test suite pass after the change.

Fixes #2422.


ccoVeille:
Why creating the PR 

- https://github.com/spf13/cobra/pull/2461

I thought you agreed it was a duplicate of

sergioperezcheco:
You're right, my apologies — I didn't realize #2423 was still open when I created this one. Closing as a duplicate since #2423 already addresses the same Go 1.26 vet issue.

PR #2473, opened 2026-08-01
fix: recompute command path lengths when a subtree is reparented

### LINKED ISSUE #2463: Help output depends on the order AddCommand is called (CommandPathPadding is snapshotted, never recomputed)
Two identical command trees render **different help text** depending only on the order in which `AddCommand` was called. Building bottom-up — construct a subcommand fully, then attach it to the root — produces misaligned columns.

Observed on cobra v1.10.2, go1.24, linux/amd64.

### Steps to reproduce

```go
package main

import (
	"fmt"

	"github.com/spf13/cobra"
)

func build(topFirst bool) {
	root := &cobra.Command{Use: "app", Run: func(*cobra.Command, []string) {}}
	sub := &cobra.Command{Use: "remote", Short: "remote ops", Run: func(*cobra.Command, []string) {}}
	// help topics (no Run) render under "Additional help topics", padded with CommandPathPadding
	t1 := &cobra.Command{Use: "authentication", Short: "how auth works"}
	t2 := &cobra.Command{Use: "tls", Short: "certificate setup"}

	if topFirst {
		root.AddCommand(sub)
		sub.AddCommand(t1)
		sub.AddCommand(t2)
	} else {
		sub.AddCommand(t1)
		sub.AddCommand(t2)
		root.AddCommand(sub)
	}
	_ = sub.Help()
}

func main() {
	fmt.Println("--- root.AddCommand(sub) first ---")
	build(true)
	fmt.Println("--- sub.AddCommand(topics) first ---")
	build(false)
}
```

### Current behaviour

```
--- root.AddCommand(sub) first ---
Additional help topics:
  app remote authentication how auth works
  app remote tls            certificate setup

--- sub.AddCommand(topics) first ---
Additional help topics:
  app remote authentication how auth works
  app remote tls        certificate setup      <- misaligned
```

### Expected behaviour

Both orders describe the same tree, so both should render identically (the first form).

### Cause

`AddCommand` computes the child's path length eagerly and caches it on the parent:

```go
commandPathLen := len(x.CommandPath())
if commandPathLen > c.commandsMaxCommandPathLen {
	c.commandsMaxCommandPathLen = commandPathLen
}
```

`CommandPath()` walks up through `parent`, so its value depends on how much of the tree is attached *at that moment*. In `sub.AddCommand(t1)` before `sub` has a parent, `t1.CommandPath()` is `"remote authentication"` (21) rather than `"app remote authentication"` (25). Attaching `sub` to `root` afterwards never invalidates the cached value, so `sub.commandsMaxCommandPathLen` stays 4 short forever.

```go
leaf.CommandPathPadding()   // 21 built bottom-up, 25 built top-down
leaf.NamePadding()          // 14 either way
leaf.UsagePadding()         // 25 either way
```

Only `CommandPathPadding` is affected, because `Name()` and `Use` are local to the command while `CommandPath()` is the only one of the three that depends on ancestry.

`RemoveCommand` already recomputes all three lengths from scratch, which suggests the cache is intended to stay accurate.

### Notes

- Bottom-up construction is the more natural pattern (build a subcommand in its own package, export it, attach it), so the broken order is the common one.
- This is separate from the hidden/deprecated-command padding reports in #2431 and #2426 — those are about *which* commands are counted, this is about *when* the length is computed.
- A fix could recompute on attach, or compute the max lazily at render time. Happy to open a PR once you've decided which shape you'd prefer.

r0h1tb:
Fixes #2463

## Problem

`AddCommand` caches `len(x.CommandPath())` on the parent at the moment a subcommand is attached. `CommandPath()` is derived by walking the parent chain, so attaching a subtree somewhere new lengthens the path of every command beneath it — invalidating anything cached earlier.

Two identical trees therefore render different help text depending only on construction order:

```go
// bottom-up: sub has no parent yet, so this caches "remote authentication" (21)
sub.AddCommand(t1); sub.AddCommand(t2); root.AddCommand(sub)

// top-down: caches "app remote authentication" (25)
root.AddCommand(sub); sub.AddCommand(t1); sub.AddCommand(t2)
```

`CommandPathPadding()` reads that cached value, so the "Additional help topics" column comes out misaligned in the bottom-up case.

## Fix

`AddCommand` now refreshes cached path lengths for the attached command and all of its descendants, via a small `recomputeCommandPathLengths` helper.

Two notes on scope:

- Only `commandsMaxCommandPathLen` is recomputed. `commandsMaxUseLen` and `commandsMaxNameLen` derive from `Use` and `Name`, which reparenting doesn't affect.
- The recursion is needed rather than a single level: a three-deep tree attached at the root has stale caches at every intermediate node, which the second test covers.

`RemoveCommand` already performs an equivalent recompute for its own children, so this brings `AddCommand` in line with it.

## Tests

`TestCommandPathPaddingIndependentOfAddCommandOrder` builds the issue's exact tree both ways and asserts the cached length matches, that it equals `len("app remote authentication")`, and that the rendered help output is byte-identical.

`TestCommandPathPaddingRecomputedForDeepSubtree` covers the multi-level case.

On `main` both fail:

```
max command path length = 21, want 25 (must not depend on AddCommand order)
nested max command path length = 11, want 19
```

With this change `go test ./...` is green across `cobra` and `cobra/doc`. `gofmt` and `go vet` are clean.


eugene-panin:
Thanks for the recursive refresh — I found the matching invalidation when a subtree is detached.

`main` passes this sequence, but this PR makes it fail because `AddCommand` refreshes the subtree with the longer path and `RemoveCommand` only clears `parent`:

```go
remote.AddCommand(authentication, tls)
root.AddCommand(remote)
root.RemoveCommand(remote)

// stale with this PR: "app remote authentication" length
// expected: len("remote authentication")
if got, want := tls.CommandPathPadding(), len(authentication.CommandPath()); got != want {
    t.Fatalf("padding = %d, want %d", got, want)
}
```

The minimal follow-up appears to be refreshing the detached subtree immediately after clearing its parent:

```go
command.parent = nil
command.recomputeCommandPathLengths()
```

I verified this in Docker against the PR branch: the reproduction fails before that line and passes after it, along with `go test ./...`, `go test -race ./...`, `go vet ./...`, and formatting. A regression test for the detach path would make the cache invalidation symmetric with the attach path.

r0h1tb:
Good catch — thanks @eugene-panin, this reproduces exactly as you describe and the suggested line is the right fix. Pushed in 4a5f267.

**Behavioural delta:** `RemoveCommand` now refreshes the detached subtree's cached path lengths, so `CommandPathPadding()` reflects the subtree's own root immediately after detach instead of the path it had while attached.

## One correction to the framing

The detach staleness isn't only a regression from this PR — it's already present on `main` when the tree is built top-down. `main` gets your bottom-up sequence right by accident, because without the attach-side refresh the cache was never updated to the longer value in the first place. Swap the two `AddCommand` calls and `main` fails the same assertion:

```go
root.AddCommand(remote)
remote.AddCommand(authentication, tls)
root.RemoveCommand(remote)

// main today: 25 (len("app remote authentication"))
// correct:    21 (len("remote authentication"))
```

I verified that against an unmodified `command.go`:

```
$ git checkout main -- command.go && go test -run TestCommandPathPaddingRecomputedOnRemoveCommand .
--- FAIL: .../top-down
    max command path length after detach = 25, want 21
    padding = 25, want 21
```

So this isn't papering over something the PR broke — `AddCommand` refreshing the subtree is what makes the missing detach-side invalidation observable in *both* construction orders, and the recompute makes the two paths symmetric. Worth stating plainly in case it affects how you'd want it reviewed.

## Tests

`TestCommandPathPaddingRecomputedOnRemoveCommand` runs your sequence in both orders as a table test; `TestCommandPathPaddingRecomputedOnRemoveCommandForDeepSubtree` covers the three-level case, since the detached subtree has stale caches at every depth just as the attached one does.

Copy-pasteable proof the tests pin the bug rather than restate the fix — revert the source hunk only and keep the tests:

```
$ git checkout HEAD~1 -- command.go
$ go test -run TestCommandPathPaddingRecomputedOnRemoveCommand .
--- FAIL: TestCommandPathPaddingRecomputedOnRemoveCommand/top-down
    max command path length after detach = 25, want 21
    padding = 25, want 21
--- FAIL: TestCommandPathPaddingRecomputedOnRemoveCommand/bottom-up
    max command path length after detach = 25, want 21
    padding = 25, want 21
--- FAIL: TestCommandPathPaddingRecomputedOnRemoveCommandForDeepSubtree
    nested max command path length after detach = 19, want 15
FAIL
```

`go test ./...`, `go test -race ./...`, `go vet ./...` and `gofmt -l .` are all clean with the change.

## Deliberately left out

`ResetCommands` has the same class of staleness and I did **not** touch it, to keep this PR to the reported issue. It clears `parent` and `commands` but leaves all three max-length caches populated, and `AddCommand` only ever grows them:

```go
root.AddCommand(&Command{Use: "authentication"})
root.ResetCommands()
root.AddCommand(&Command{Use: "ab"})
// maxPathLen=18 maxUseLen=14 maxNameLen=14; correct would be 6/2/2
```

That reproduces on `main` independently of this PR and also affects `commandsMaxUseLen`/`commandsMaxNameLen`, which the path-length recompute deliberately doesn't cover. Happy to fix it here if a maintainer prefers it bundled, or to file it separately — whichever is easier to review.

PR #2482, opened 2026-08-10
fix(powershell): handle empty and single-line completion output safely

### LINKED ISSUE #2481: PowerShell autocompletion script fails with runtime errors on empty or single-item completions
### Description

   ```text
   ```

   ```text
   ```

   ```powershell
   ```

### Root Cause Analysis

```powershell

    if ($Directive -eq "") {
        $Directive = 0
    }

```

### Proposed Fix

```go
	$Directive = 0
	if ($Out.Count -gt 0) {
		if ($DirectiveStr -ne "") {
		}
	}
```

VedantMadane:
**Description of the change**

Generated PowerShell completion scripts could throw at runtime when completion output was empty or had a single line, or when a completion value matched the directive string.

This change updates the template in `powershell_completions.go` to:

1. Force ` = @()` so a single line stays a one-element array (avoids `System.Char` from `[-1]`).
2. Only read the directive when `.Count -gt 0`.
3. Cast the last line to `[string]` before `TrimStart`.
4. Drop the directive with `Select-Object -First (.Count - 1)` instead of filtering equal values with `Where-Object`.

**Benefits**

- Empty completion output no longer crashes the completer.
- Single-line output (directive only) works.
- Completions that equal the directive string (e.g. `:0`) are kept.

**Possible drawbacks**

- Users must regenerate shell completion scripts to pick up the template change.

**Applicable issues**

- fixes #2481

**Additional information**

`go test . -run Pwsh` passes locally.


xhon-pelushi (approved):
Traced through the PowerShell semantics by hand (no pwsh available in this environment to execute it directly, so this is careful reading rather than a run):

**Root cause confirmed** — `$Out[-1]` on an auto-unwrapped single-element pipeline result indexes into the *string*, giving a `System.Char` with no `.TrimStart`. `$Out = @($Out)` forces the array shape before that ever happens. Correctly diagnosed.

**The truly-empty case also checks out, not just the single-line case** — when the completion function produces zero output objects, PowerShell's pipeline gives `$Out = $null`, not an empty array. `@($null)` is a *one*-element array containing `$null` (a PowerShell quirk worth being aware of), so `$Out.Count` is 1 and `$Out[-1]` is `$null`. `[string]$null` casts cleanly to `""`, `.TrimStart(':')` on that is safe, and the `-ne ""` check correctly leaves `$Directive` at its 0 default. So the "empty output" half of the title holds even for this trickier `$null`-vs-empty-array case, not just the one-line case the description walks through.

**Bonus fix not called out in the description**: the old code did `[int]$Directive = $Out[-1].TrimStart(':')` — casting straight to `[int]` inline. An empty string cast to `[int]` throws in PowerShell, so the very next line's `if ($Directive -eq "")` was already dead code, unreachable before your change. Extracting to `$DirectiveStr` and checking emptiness *before* casting fixes that too.

**One non-blocking note**: this repo's completion tests (this file included) only string-match against the generated template text — CI is Ubuntu-only, no pwsh runner. That's consistent with how the rest of `powershell_completions_test.go` already works, so not a gap introduced here, but since the original bug was a pure runtime failure this kind of test wouldn't have caught, it's worth actually running the generated script through a real PowerShell locally if you haven't already (sounds like you may have, given "possible drawbacks" mentions regeneration).

Logic holds up. LGTM.

PR #2483, opened 2026-08-11
feat: suggest matching flags on parse errors

### LINKED ISSUE #2207: Suggestions when "unknown flag" happens
Cobra will print automatic suggestions when "unknown command" errors happen.

It would be useful to have the same feature for flags.

Boulea7:
## Summary

- suggest nearby long flags when pflag reports an unknown flag
- include local and persistent flags, including with `TraverseChildren`
- preserve custom `FlagErrorFunc` behavior and exclude hidden flags and shorthand errors

Fixes #2207

## Testing

- `go test ./...`
- `go test -race ./...`
- `make all`
- `make lint`


otrumb (changes requested):
Requesting changes for a nil-command panic in the traversal path.

At `command.go:908`, `return nil, args, c.flagError(err)` can return both a nil command and nil error when a custom flag-error callback suppresses the error. With `TraverseChildren: true`, a registered `version` flag, arguments `--verion=true child`, and `SetFlagErrorFunc(func(*Command, error) error { return nil })`, `ExecuteC()` reaches the success path and panics when it dereferences `cmd` at line 1190.

I reproduced this on `a20bcaf54e27f56149745a3135e88307c924fea8`; the base commit returns the original unknown-flag error without panicking. Please ensure a suppressed traversal error cannot return a nil command as success, and add regression coverage.

Otherwise, focused flag-suggestion checks and the full tests, `go vet`, and `go build` passed on native Windows with CGO disabled.

PR #2487, opened 2026-08-17
Fix PowerShell autocompletion runtime exceptions

### LINKED ISSUE #2481: PowerShell autocompletion script fails with runtime errors on empty or single-item completions

TheRodzz:
Fixes #2481

This PR addresses the runtime exceptions that occur in PowerShell completions:
1. Null Pointer/MethodNotFound Exceptions on Empty Outputs by checking if $Out.Count is greater than 0.
2. System.Char vs. String Type Confusion by explicitly casting to [string] before calling TrimStart.
3. Filtering Collision on Identical String Values by cleanly removing the directive with Select-Object by index.


goransh-walia:
Clean and minimal — appreciate that it stays scoped to the actual fix.

TheRodzz:
> Clean and minimal — appreciate that it stays scoped to the actual fix.

Thanks @goransh-walia , I have signed the CLA, could you please approve the PR?

TheRodzz:
Hi @goransh-walia , Could you please review this, Thanks!

PR #2497, opened 2026-09-09
fix: handle empty and single-line output in PowerShell completion

### LINKED ISSUE #2481: PowerShell autocompletion script fails with runtime errors on empty or single-item completions

jabrailkhalil:
## Summary

Fixes #2481. The generated PowerShell completion script failed with runtime errors in three edge cases:

1. The completion command produced no output (e.g. the CLI crashed or returned nothing): `$Out[-1].TrimStart(':')` was called on a null value (`Cannot index into a null array`).
2. The completion command produced a single directive-only line: in newer PowerShell versions `-OutVariable` unwraps single-element results, so `$Out[-1]` returns a `System.Char` instead of a `System.String` and `TrimStart` is not available.
3. A completion value collided with the directive line: `$Out = $Out | Where-Object { $_ -ne $Out[-1] }` removed every line equal to the directive value, dropping legitimate completions (e.g. a completion whose value is `:0`).

The template in `powershell_completions.go` now:

- normalizes `$Out` into an array (`[Array]$Directives = $Out`) before reading the directive, so `$null`, a single string, or a multi-line result are all handled safely;
- parses the directive defensively and strips the last line by index with `Select-Object -First`, instead of filtering by value;
- returns early when there are no completions left after removing the directive, instead of indexing into `$Out`.

## Tests

- `go test ./...` — passes (cobra, cobra/doc)
- `go vet ./...` — passes
- `golangci-lint run --new-from-rev=origin/main --new ./...` — 0 issues

## Verification

Exercised the generated script in Windows PowerShell 5.1 against real binaries speaking the completion protocol (empty output, directive-only output, colliding directive/completion values, and a normal CLI built on cobra):

- empty output: old script crashed; new script completes with no errors
- directive-only output: old script surfaced a `PSArgumentNullException` on PS 5.1; new script completes with no errors
- colliding value: old script dropped the `:0` completion; new script returns it
- normal completions (subcommands, `ValidArgsFunction`): behavior is unchanged

Fixes #2481


jabrailkhalil:
The CI workflows for the current head `5499b76f9d` are waiting for contributor workflow approval (`action_required`): [Test](https://github.com/spf13/cobra/actions/runs/34506056609). Could a maintainer approve the pending runs when convenient? These runs have not executed their checks yet; I will address any failures once they run.

